PR #29, opened 2016-06-06
Specifying which page templates to enable elementor on.

jawngee:
This PR adds functionality to specify what page templates allow elementor to be used on, similar to how you can choose which post types.

The one problem with this is that WordPress's autosave functionality doesn't save template changes.  For example, you are editing a page and you select a new template.  WordPress won't then save that change unless you press the publish/update button yourself.  This is a problem because if you change the template and then click on "Edit with elementor" then it won't save the template when you exit to edit with elementor.


KingYes:
Hi @jawngee
I really appreciate you sending me this code. Unfortunately, we are working with a specific roadmap and this is not a feature we will add right now. I'll keep it in file in case we decide to use it in the future. Keep up the good work, I'd love to get more suggestions in the future.
I'd like us to continue to collaborate. Could you open an issue on GitHub so we can work together on other suggestions and ideas?

PR #46, opened 2016-06-08
carusel

yehudah:


assets/js/frontend.js
```diff
@@ -185,30 +185,67 @@
...
-				arrows: true
-			};
-
-		if ( data ) {
-			$.each( data, function( key, value ) {
-				options[ key ] = value;
-			} );
+			$carousel = $wrapper.children( '.elementor-carousel' );
+
+		var	slideToShow = $carousel.data( 'slidestoshow' ),
```
KingYes: Why you use extra spaces after the `var`?

PR #55, opened 2016-06-09
slideshow - img & imgbox title

yehudah:
new options
thumbnail navagation


includes/widgets/image-box.php
```diff
@@ -113,6 +113,16 @@ protected function _register_controls() {
 		);
 
 		$this->add_control(
+			'title_attr',
```
KingYes: The name is need to be: `image_title`.

includes/widgets/image.php
```diff
@@ -78,6 +78,16 @@ protected function _register_controls() {
 		);
 
 		$this->add_control(
+			'title',
```
KingYes: Again.. The name need to be: `image_title`.

KingYes:
Now it's okay for me. Thanks @yehudah

PR #125, opened 2016-06-22
Small typos

pedro-mendonca:
Typos in strings below:
'No **put** post_id'
'Testimonial Detail**e**s Position' *(if this was correctly found as a typo, there are a lot of coded variables with the same typo in `testemonial.php`and css files)


KingYes:
Thanks. The testimonial widget is under working, so it's okay for now. But thanks for you support ! :)

PR #148, opened 2016-06-26
added view and icon

yehudah:


includes/widgets/audio.php
```diff
@@ -28,13 +28,24 @@ protected function _register_controls() {
...
+				'default' => 'soundcloud',
+				'section' => 'section_title',
+			]
+		);
+
+		$this->add_control(
 			'link',
 			[
-				'label' => '',
+				'label' => 'Link',
```
KingYes: @yehudah: Add gettext here

PR #185, opened 2016-07-01
Typo in brand names.

pedro-mendonca:
Do these names need to be translatable?


arielk:
I think your right 👍

@KingYes what do you think?

KingYes:
@pedro-mendonca You are fast :)
Thanks!

PR #265, opened 2016-07-14
Fixes & Removed Translations

kobizz:


ramiy:
Why removing translation strings?

Issue #291, opened 2016-07-20
Adding more hooks & filters to settings and widget html output.


dtbaker:
Hey guys,

Great plugin! I'd love to use this in our next ThemeForest theme instead of Visual Composer.

A few more do_action() and apply_filters() through the code would be wonderful!

I need some extra hooks and filters on the generated HTML output (e.g. adding some more markup to the default image-carousel HTML output or adding some more "controls" in the _register_controls function).

If you're happy to accept pull requests I'd be happy to make a couple of suggestions that would make it easier for us to integrate more tightly with our themes. 

Cheers,
Dave

dtbaker:
For now I'm just doing this to load my own custom "Elementor Widget" which is a copy of your image-carousel widget.

```
<?php

if(!class_exists('dtbaker_elementor')) {
	class dtbaker_elementor {

		public function __construct() {
			add_action( 'elementor/widgets/widgets_registered', array( $this, 'widgets_registered' ) );
			add_action( 'wp_enqueue_scripts', [ $this, 'enqueue_scripts' ] );
			add_action( 'wp_enqueue_scripts', [ $this, 'enqueue_styles' ] );
		}

		public function widgets_registered() {

			if(defined('ELEMENTOR_PATH') && class_exists('Elementor\Widget_Base')){
				// get our own widgets up and running:
				// copied from widgets-manager.php

				$widget_file = 'plugins/elementor/widgets/image-slider.php';
				$template_file = locate_template($widget_file);
				if ( $template_file && is_readable( $template_file ) ) {
					require_once $template_file;
					Elementor\Plugin::instance()->widgets_manager->register_widget( 'Elementor\Widget_Image_Slider' );


				}
			}
		}

		public function enqueue_scripts(){

			if(class_exists('Elementor\Plugin')) {
				$widget_script_file = 'plugins/elementor/widgets/image-slider.js';
				wp_register_script(
					'elementor-image-slider',
					( file_exists( STYLESHEETPATH . '/' . $widget_script_file ) ? get_stylesheet_directory_uri() : get_template_directory_uri() ) . '/' . $widget_script_file,
					[
						'elementor-frontend',
					],
					Elementor\Plugin::instance()->get_version()
				);
				wp_enqueue_script( 'elementor-image-slider' );
			}
		}
		public function enqueue_styles(){

		}
	}

	$dtbaker_elementor = new dtbaker_elementor();
}
```

KingYes:
Hi @dtbaker 
[...]

PR #361, opened 2016-08-02
Filters for theme developers

dtbaker:
For issue #291


GreenRidingHood:
"topic: Adding a custom button class with theme filters: " from #291 , 
Hi! 
is there any documentation about that issue or example? 
I want to edit or customize theme button without touching plugin files
thanks..

PR #365, opened 2016-08-02
Updated correct style and titles

arielk:


includes/templates/types/local.php
```diff
@@ -253,7 +253,7 @@ public function admin_import_template_form() {
 					<input type="hidden" name="action" value="elementor_import_template">
 					<fieldset id="elementor-import-templates-form-inputs">
 						<input type="file" name="file" accept="application/json" required>
-						<input type="submit" class="button">
+						<input type="submit" class="button" value="Import">
```
KingYes: Please use gettext here

PR #436, opened 2016-08-19
Better HTML structure for Progress widget styling

daviedR:
Add new "elementor-progress-background" <div> and move inner text
element before the bar element for better styling possibilities.

Also add new variable: $progress-bar color for default progress bar color.


KingYes:
The last version of this I think we fix this, thanks!

PR #453, opened 2016-08-25
action to register new scripts and css

FrankM1:
This adds the ability to add more css and javascript in the editor
window since Elementor runs a new instance of  WP_Styles and WP_Scripts


KingYes:
We added actions for this, thanks! :)

PR #488, opened 2016-09-01
Made the plugin installable via composer.

jkrug:
With thoe little changes, you are able to install the plugin via composer and exclude the test packages for the live build. Thank you for the great work!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,6 @@
 {
 	"name": "pojome/elementor",
+        "type": "wordpress-plugin",
 	"description": "Elementor",
 	"minimum-stability": "dev",
 	"license": "gpl3",
@@ -9,7 +10,7 @@
 			"email": "yakir@pojo.me"
 		}
 	],
-	"require": {
+	"require-dev": {
 		"phpunit/phpunit": "4.8.8",
 		"satooshi/php-coveralls": "~1.0"
 	}


composer.json
```diff
@@ -1,5 +1,6 @@
 {
 	"name": "pojome/elementor",
+        "type": "wordpress-plugin",
```
cfoellmann: did you use spaces and not tabs here?

KingYes:
I don't think is needed for this project, but thank you :)

yangm97:
@KingYes it would be a nice addition for those of us managing our wordpress themes and plugins using composer (such as when you use the [bedrock](https://roots.io/bedrock/) boilerplate).

Not only the free version would benefit from this but more importantly the Pro version, since it's not available from wpackagist, as it's not pushed into wordpress.org, as requested on #4042

mindfullsilence:
@KingYes @yangm97 
I can assure you composer is a massive need. As @yangm97 pointed out, the bedrock boilerplate allows for installing plugins via composer, most of the time through wpackagist, and this isn't a fringe group of developers who use bedrock as boilerplate for WordPress. It allows for enterprise-level development and deployment of WordPress sites. 

Bedrock has about 130+ downloads per day, and according to ticket #4042 the ability to install the Pro plugin via composer has been requested by quite a few people for quite a long time. 

Even if it's not composer, a simple authenticated API endpoint that would allow us to programmatically download the plugin would be enough, the dev community can take it from there and figure out how to make it work with composer, our own bash scripts, or whatever else is needed for consistent deployments on an enterprise level. 

Right now we're pigeon-holed into a very laborious task for deployments and updates when using basic tools like Docker or load balancers for our WordPress sites because there's literally no other way.

PR #494, opened 2016-09-03
Image Carousel Filters.

dtbaker:
Two simple filters that will allow theme developers to customize the default Image Carousel slider output.

### DIFF
--- a/includes/widgets/image-carousel.php
+++ b/includes/widgets/image-carousel.php
@@ -493,7 +493,7 @@ class Widget_Image_Carousel extends Widget_Base {
 			}
 
 			// Allow themes to modify the individual slider HTML layout.
-			$image_html = apply_filters('elementor/widgets/image_carousel/image_html', $image_html, $instance, $attachment);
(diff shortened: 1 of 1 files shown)


KingYes:
In the last version of Elementor you can do that from our global hooks (filters and actions).

Thank you @dtbaker ! :)

dtbaker:
Thanks so much for adding those in!

iamntz:
> In the last version of Elementor you can do that from our global hooks (filters and actions).

@KingYes Can we find somewhere an example or more details on this?

KingYes:
For example filter to change the content:
https://github.com/pojome/elementor/blob/develop/includes/base/widget-base.php#L192

BTW, If you want to register new controls, you can use these hooks:
https://github.com/pojome/elementor/blob/develop/includes/base/element-base.php#L509-L541

We are working on "Skins" system, and when it's will be complete you will know about that :)

Thank you guys

iamntz:
Ah, but these are **old** hooks, hence the confusion :)

Thanks for the explanation.

PR #498, opened 2016-09-05
google maps filter for theme developers

dtbaker:
This will allow a theme developer to style the map output to better match their theme colors. Example:

image


dtbaker:
This can really make a theme look 1000% better:

image

KingYes:
In the last version of Elementor you can do that from our global hooks (filters and actions).

Thank you @dtbaker ! :)

PR #769, opened 2016-10-30
feature/global-widget

kobizz:


includes/base/element-base.php
```diff
@@ -24,17 +24,47 @@
...
+
+	public final static function add_edit_tool( $tool_name, $tool_data, $after = null ) {
+		if ( null === static::$_edit_tools ) {
+			self::_init_edit_tools();
+		}
+
+		if ( $after ) {
+			$after_index = array_search( $after, array_keys( static::$_edit_tools ) ) + 1;
+
+			static::$_edit_tools = array_slice( static::$_edit_tools, 0, $after_index, true ) +
```
KingYes: Please write comment what you doing here

PR #810, opened 2016-11-07
Added Background Overlay Settings to Column Background

WebTechStreet:
Section settings already allows to put background overlay along with
Background.

Same has been implemented for Column Background.

column-bg-overlay


KingYes:
Thanks :)
We merged your PR by #1153

PR #853, opened 2016-11-14
Hi！ I Find Problem

AquaHacker:
Hello, I found a problem;
In WordPress 4.6.1  》 I used the Betheme theme I found that the installation of the Ultimate Member 1.3.72   Elementor will become very slow but the actual load has been loaded for a long time to show;


KingYes:
Why you opened PR?

PR #889, opened 2016-11-19
Added hooks to allow customizing default elementor widgets

iamntz:
Hey guys, so here is the thing: I wanted to add some custom button style and i found out that there is no way. So I thought I should add a filter to allow just that. And instead of adding a filter only for button, I thought: what if... I add a filter for _all_ fields? So here it is :)

Usage:

```
add_filter('elementor/elements/add_control/button_type', function($args, $id)
{
  $args['options']['my-button'] = __('My Button Style!');
  return $args;
}, 10, 2);
```

### DIFF
--- a/includes/base/element-base.php
+++ b/includes/base/element-base.php
@@ -115,6 +115,9 @@ abstract class Element_Base {
 	}
 
 	public function add_control( $id, $args ) {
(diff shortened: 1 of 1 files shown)


dasigna:
interesting solution... where exactly did you put this in?

mbaierl:
This is not included in the latest version of Elementor?

Issue #939, opened 2016-11-24
Feature Request: Add and ID to an element


nickdgreen:
The ability to add an ID to an element so that it can be individually targeted with #styleStyle from add on CSS

arielk:
Why not using by CSS Classes?

peepshow:
ID's are sometimes a required if you're trying to combine with another plugin, JS or WP. Also ID's are helpful if you want to add anchor tags. ID's are generally considered better for performance when referenced in JS too as they are unique selectors.

arielk:
We know it. But we're really trying to find a way to do it in a way that will not create conflicts with duplicate ID.

Thanks for your feedback!

pingram3541:
I know this is closed but...

Duplicate ID's are likely not going to hurt anything (outside of js plugins or absolute positioning) as most browsers ignore that technically there is supposed to only be one by spec.

Many other builders already provide the ability to add id's without all hell breaking loose and it's not a complex thing to add in so maybe reconsider this request again.  ID's have one very important thing for me, hierarchical superiority which means writing less of these:

`!important`

The spec of ID attributes being unique is not really that big of a problem is it?

drawcard:
I just discovered how badly I need this feature, trying to implement [vivus.js](https://maxwellito.github.io/vivus/) on some SVG's in my layout. Vivus explicitly uses ID's to identify SVG elements, in order to trigger the animations / callbacks.

Perhaps a simple validator could be written that continuously detects all custom ID's on the page, and warns the user if there's a conflict. Or maybe the ID customisation feature is toggle for advanced users on the settings page.

I would also appreciate any reconsideration on this feature request :)

pingram3541:
A simple js object could be used with validation to ensure no duplicates are ever entered.

Would love an example on how to add a custom control to a section.  Inspecting the code it looks like it was designed to only allow for customizing or adding our own widgets but sections and columns are private and protected by design.  No?

nickdgreen:
I don't see the need for validation or duplicate checking the website designer/builder has to take some responsibility for the control of IDs and most designers/builders will use recognisable IDs pageOneSectionOne, pageOneSectionTwo etc. to me it's a gap in the product that can be filled without over thinking it.

pingram3541:
I agree, duplicate id's are likely not going to cause any problems for anyone w/ exception of an occasional script error or broken anchor but the suggestion was based off the data already being present and a very low scope to push all id's to an array to check against.

[...]

PR #940, opened 2016-11-26
Added is_anchor switch for heading

iamntz:
Sometimes I need to add anchors to some headings, but adding anchor elements sometimes may be an overkill.

### DIFF
--- a/includes/widgets/heading.php
+++ b/includes/widgets/heading.php
@@ -36,6 +36,18 @@ class Widget_Heading extends Widget_Base {
 		);
 
 		$this->add_control(
+			'is_anchor',
+			[
+				'label' => __( 'Is Anchor', 'elementor' ),
+				'type' => Controls_Manager::SWITCHER,
+				'default' => '',
+				'label_on' => __( 'Yes', 'elementor' ),
+				'label_off' => __( 'No', 'elementor' ),
+				'return_value' => 1,
+			]
+		);
+
+		$this->add_control(
 			'link',
 			[
 				'label' => __( 'Link', 'elementor' ),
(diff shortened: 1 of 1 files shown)


arielk:
We plan to add an option ID to any element in Elementor. You can follow it on this issue: #939 

Thanks for your PR!

iamntz:
Right, but this feature purpose would be to have a simple, predictible way of adding a html anchor for titles. Not for styling (which you would made for the _container_) but solely for anchoring.

arielk:
You can use bay Menu Anchor widget for this..

iamntz:
I know, but say you want to have a title like `My Awesome Title` anchored. You will manually add a menu anchor called `my-awesome-title`. Then if you want to change the title to `Puppies and feathers`, you'll also need to change the anchor, to match the title, othewise will not be too semantically. :)

In case you wonder about the usecase: I'm having some kind of documentation on a page, with various sections.

arielk:
It's a good idea but it's problematic. because it works well only with titles in English, but if you write with different languages (like: Hebrew, Arabic..) it doesn't work because the letters are different.

iamntz:
Hmmm, you're right. Maybe those users should fallback on menu anchor field?

arielk:
So I think the best option right now is to use the Anchor widget.

pingram3541:
My issue is scrollTop always scrolls past the anchor no matter what I try to do for vertical margin/padding on that anchor because it is an inline element (without custom css) so I have to remember to adjust that on the containing element which is often a trial and error process of testing a dozen page loads until I get it just right for all views.

Still don't understand what's so taboo about adding ID attributes to sections and columns when they are indeed a useful part of html spec and are likely not going to break anything as most browsers ignore that "there can only be one!" anyway.

iamntz:
@pingram3541 you can do like i did: fork the project, merge your changes, create PRs. Either the PR are accepted or not, you're able to do your job. The only downside is that you need to `git pull` every now and then to bee up to date.

Most needed changes are a matter of 1-2 lines of code, so there are very small chances of conflicts.

pingram3541:
Great suggestion.  I'll get cracking =)

PR #971, opened 2016-12-02
Custom control hooks

iamntz:
This is a tweaked version of one of my previous PR. #889 

This will allow you to do this:

1. tweak/edit existing fields via `elementor/elements/add_control` filter;
1. tweak existing fields output (e.g. add a custom `add_render_attribute`) via `elementor/widget/before_render_content`
1. Allow registering custom sections on existing fields, either before or after via `elementor/element/after_register_controls` hook;
1. Tweak the image field class.

This last point is added because i noticed there is no way of adding a custom class on the preview image.

### DIFF
--- a/includes/base/element-base.php
+++ b/includes/base/element-base.php
@@ -622,7 +622,9 @@ abstract class Element_Base {
 	private function _init_controls() {
 		Plugin::instance()->controls_manager->open_stack( $this );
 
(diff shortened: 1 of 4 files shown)


KingYes:
I'll come back here after the release `1.0.0` get publish. Thank you

PR #991, opened 2016-12-05
Allow to overwrite render attributes

kobizz:


includes/base/element-base.php
```diff
@@ -417,18 +417,18 @@ public function is_control_visible( $control, $values = null ) {
 		return true;
 	}
 
-	public function add_render_attribute( $element, $key = null, $value = null ) {
+	public function add_render_attribute( $element, $key = null, $value = null, $overwrite = false ) {
```
matipojo: mabye "overide"?

PR #992, opened 2016-12-05
extracted column wrap class names and added filter for it

iamntz:
Usecase:

```
add_filter('elementor/element/column/class_names', function ($classNames, $instance, $isInline) {

	if (!$isInline) {
		$settings = $instance->get_settings();

		if ($settings['background_is_dark']) {
			$classNames[] = 'has-dark-bg';
		}
	} else {
		$classNames[] = '{{ settings.background_is_dark ? " has-dark-bg " : "" }}';
	}

	return $classNames;
}, 10, 3);
```

### DIFF
--- a/includes/elements/column.php
+++ b/includes/elements/column.php
@@ -358,9 +358,22 @@ class Element_Column extends Element_Base {
 		<?php
 	}
 
+	private function get_column_wrap_class_names( $is_inline = false)
+	{
+		$class_names[] = 'elementor-column-wrap';
+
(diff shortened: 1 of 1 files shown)


KingYes:
And what's about JS template?

iamntz:
It's there in both cases :)

```
$classNames[] = '{{ settings.background_is_dark ? " has-dark-bg " : "" }}';
```

This is the only way i can think of in order to make it play nice and allow evaluation JS.

PR #1150, opened 2017-01-11
Proof of concept: Use unified templates in both JS and PHP world

### LINKED ISSUE #1147: API Request: Re-Use existing widgets
hy,

i would like to see functionality in the API to re-use existing widgets. in my special case i would like to use the image carousel functionality as wrapper for dynamic content which i retrieve from external sources.

it would be good for the frontend to provide the same user experience and easier for the developer.

so in this example, only the image source and image size controls would not be shown, instead other options for content sources. all other controls should stay the same. 

do you have any plans on building such possibilities in elementor (pro)?

matt

iamntz:
This would partially close #1147 and will *greatly* improve the extensibility of the current widgets, will reduce repeated markup and in general will produce nice code :)

I added this only on accordion widget, but hopefully it will be extended to all widgets and it will remove `echo`es among widgets, with `render` method just _returning_ the html, havnig the `widget-base.php` `render_content` method to be the single point of echo (this will also simplify tests if you decide at some point to add tests as well).

Right now, the solution is rather hacky¹, and requires html `render` to echoes things passing the settings as `settings` key:

```php
echo tpl::get('widgets/accordion', [
	'settings' => $settings
]);
```

I would rather have the whole thing as a oneliner:

```
echo tpl::get('widgets/accordion', $settings);
```

### How to set up
You just need to run `composer install --no-dev` to install php's handlebars.

This means that *ALL* js logic should be moved where it belongs: in the JS files, not in the template (e.g. all conditionals should be placed into models, not into the html). So this:

```js
<# if ( '' !== settings.image.url ) {
			var image = {
				id: settings.image.id,
				url: settings.image.url,
				size: settings.image_size,
				dimension: settings.image_custom_dimension,
				model: editModel
			};
```

Should NOT be present in the html but into JS. Basically we should have the templates as dumb as possible.

---

This will allow users to override templates by simply placing a `views/widgets/accordion.hbs` in *their* theme (or their plugin). Combined with some of my other PRs that adds various hooks (that allow editing _existing_ widget controls) we (users) could endlessly extend base widgets.

---

¹ Right now, it's just a proof of concept and I'll leave it like this. Time doesn't allow me to dig deeper into the code, and the current feedback (or rather the lack of) over existing PRs doesn't really encourage me either :)


iamntz:
Message received. Thanks :)

KingYes:
Sorry, I don't close your PR. I just release new version of Elementor and I remove the current branch, so GitHub close all assigned PR's :|

iamntz:
Three days later I would expect some feedback that is not silence. **Any** kind of feedback. :)`

Which, to be honest, makes me very biased on using/recommending elementor: on one hand, the plugin is great (overall), but, on the other hand, the team behind is very „closed” on suggestion and/or feedback (although I'm not 100% sure that _closed_ it's the right term), making developing around it somehow complicated.

WPDevHQ:
@iamntz - is your Master branch up to date with all of your proposals i.e. all merged?

This "will allow users to override templates by simply placing a views/widgets/accordion.hbs in their theme" sounds very interesting and would like to test it out in terms of usage as well as the extensibilty possibilities.

iamntz:
> is your Master branch up to date with all of your proposals i.e. all merged?

No, this was branched from 1.1.0, but this isn't a problem. This can be cherry picked into any branch... It's just a proof of concept though.

orditeck:
@iamntz your PRs look really useful. I found them because I was looking to do the same as you - expend the default styles on the button component to add my own. Couldn't believe it didn't exist and I was happy to see you proposed it. But the PR is closed... And all your PRs are closed & not merged. That's exactly what Elementor should be about IMHO : extensibility. So now Elementor is just not that and your very useful PRs got no attention from the contributor.  

However @iamntz I'm really grateful to you for your PRs. It allowed me to see that I should stay away from Elementor. Took me a good while to choose the plugin I'd use for a big proejct. I'm just gonna have to start searching again.

PR #1238, opened 2017-01-29
Add missing default values on examples

luizeof:
Add missing default values on examples


KingYes:
Hi @luizbills, This issue is already resolved. Thank you for your support.

Issue #822, opened 2016-11-08
[Enhancement] Line up text items in icon list widget


knireis:
When the text of a list item does not fit on 1 line, the text is displayed on a line below it. This is good, but the text starts full left side of the widget below the icon of the list item.
It looks better if there is an ident and the text on line 2 continues same ident as the first line. See image

![schermafdruk-van-2016-11-06-17-00-46](https://cloud.githubusercontent.com/assets/2702064/20095702/dcebc7f4-a5a6-11e6-9c58-851ee9f0147e.png)

WPDevHQ:
We resolved this with a bit CSS trick here: https://www.facebook.com/groups/Elementors/permalink/976594989132954/

But this should be built in to the plugin so I'll +1 it

knireis:
Thanks, but i can't get it to align properly

WPDevHQ:
Do you have a link we can take a look at?

You may need to adjust the values and target the correct element. For the solution linked to we are targeting the `.elementor-element.elementor-element-jzyvoci` element which will be different on your site.

knireis:
http://digna.computerhulpnieuw-vennep.nl/   it is now without the css



I did change the class for the target element and also tried some different
values. But it basically only got worse.... Probably my lack of knowledge

Regards,  Jasper

2016-11-14 14:23 GMT+01:00 Zulfikar Nore <notifications@github.com>:

WPDevHQ:
Try this....had to do some adjustment due to the inline styling.

`.elementor-element.elementor-element-5h3n5vn .elementor-icon-list-icon {
  width: 8%;
  margin-right: 2%;
  float: left;
  clear: left;
}
.elementor-element.elementor-element-5h3n5vn .elementor-icon-list-text {
  width: 90%;
  float: left;
  padding-left: 0;
}
.elementor-element.elementor-element-5h3n5vn .elementor-icon-list-text {
    text-indent:  0;
}
.elementor-element.elementor-element-sewis5j .elementor-widget-container {
    margin-top: 140px;
}`

For some strange reason the applied :not(:last-child) margin is lost so I had to apply one to the top of the button.
![icontextalign](https://cloud.githubusercontent.com/assets/16872407/20269453/04705258-aa7b-11e6-8451-5236a6051098.png)

knireis:
thanks, that is pretty close, only the button jumps to the side, see
attachment

Regards

2016-11-14 16:01 GMT+01:00 Zulfikar Nore <notifications@github.com>:

[...]

PR #1323, opened 2017-02-21
Icon List

kobizz:


arielk:
i think that this PR fix #822 - need test

haniobaid:
> I think that this PR fix #822 - need test

As of 27-feb-2019 the issue in 822 is still not working with the latest release of elementor pro / price table widget / wordpress. i.e. if the feature (bullet points) in the list body have icons and the text wraps to the second line, the second line incorrectly aligns with the icon instead of the first line of text which looks absolutely horrible. See screenshot

PR #1326, opened 2017-02-21
Accessibility: add `role` tags to the toggle widget

ramiy:


ramiy:
The heading should also have an `id="xxx"` tag and the region should have an `aria-labelledby="xxx"`.

I was thinking we can use the widget unique id + the toggle count. But I don't know how to return the widget unique id.

PR #1365, opened 2017-02-26
load un minified color-picker js if SCRIPT_DEBUG is defined

### LINKED ISSUE #1364: Respect SCRIPT_DEBUG constant when loading the `color-picker.min.js`
The plugin loads the minified version[ of color-picker](https://github.com/pojome/elementor/blob/master/includes/controls/color.php#L25) despite the SCRIPT_DEBUG constant is defined in `wp-config.php`

This is more of a problem when using trunk version of WordPress where the minified file is not present.

Nikschavan:
Fixes #1364


KingYes:
Thanks :)

PR #1421, opened 2017-03-06
switched from textarea to WYSIWYG for icon-box

artifactdev:
In the need of adding links to the iconbox description it should be WYSIWYG instead of textarea


KingYes:
Hey @artifactdev 

In the this we don't have any plain to add this. Maybe in the future.

Thank you for your contribution, but I deny this PR for now.

PR #1445, opened 2017-03-10
Correction in  js hook's example in Docs

anandau:
Made a correction examples of  following js hook
frontend/element_ready/{elementType.skinName}


KingYes:
Thanks!

PR #1460, opened 2017-03-14
Feature/under construction

kobizz:


includes/plugin.php
```diff
@@ -44,6 +44,12 @@ class Plugin {
 	public $revisions_manager;
 
 	/**
+	/**
```
matipojo: duplicate comment start `/**`

matipojo (changes requested):

PR #1570, opened 2017-04-01
Update php-hooks.md

Zarioza:
Element_Base class in method print_element() for static::get_type() return string 'element'.


KingYes:
Thanks. We merge this PR by this: https://github.com/pojome/elementor/pull/1574

PR #1641, opened 2017-04-25
If Image Carousel caption is set to none, don't print the markup for the figcaption

carasmo:
If Image Carousel caption is set to none, don't print the markup for the figcaption


KingYes:
Thanks!

carasmo:
You're welcome. I see that I made a spelling error. I wrote `carosel_caption` in 3 locations -- it should be `carousel_caption`. Will this be a pain to fix?

KingYes:
You can see my commit about this too :)

carasmo:
Much better! Thanks!

PR #1660, opened 2017-04-30
Animations

kobizz:


includes/elements/column.php
```diff
@@ -134,6 +134,103 @@ protected function _register_controls() {
 
 		$this->end_controls_section();
 
+		// Section Hover
+		$this->start_controls_section(
+			'_section_hover',
+			[
+				'label' => __( 'Elementor Hover', 'elementor' ),
```
arielk: Change the label to "**Element Hover**" @kobizz

includes/elements/column.php
```diff
@@ -134,6 +134,103 @@ protected function _register_controls() {
...
+		$this->add_group_control(
+			Group_Control_Background::get_type(),
+			[
+				'name' => 'background_hover',
+				'types' => [ 'none', 'classic', 'gradient' ],
+				'selector' => '{{WRAPPER}}:hover > .elementor-element-populated',
+			]
+		);
+
+		$this->add_control(
```
arielk: This is also need conditions ?

PR #1667, opened 2017-05-01
Increasing z-index for Elementor controls.

AustinGil:
In response to this issue: https://github.com/pojome/elementor/issues/1209


KingYes:
Thanks!

PR #1771, opened 2017-05-28
Feature/section handler

kobizz:


assets/dev/js/utils/module.js
```diff
@@ -143,10 +143,16 @@ var Module = function() {
 
 Module.prototype.__construct = function() {};
 
+Module.prototype.getConstructorID = function() {
```
KingYes: Remove this

assets/dev/js/utils/module.js
```diff
@@ -161,6 +167,12 @@ Module.extend = function( properties ) {
 
 	child.prototype.constructor = child;
 
+	var constructorID = ++Module.extendsCount;
```
KingYes: Please write comment about this. Why, how, and when..

includes/elements/section.php
```diff
@@ -1134,7 +1154,10 @@ public function before_render() {
 				<?php endif;
 			endif;
 
-			if ( in_array( $settings['background_overlay_background'], [ 'classic', 'gradient' ] ) ) : ?>
+			if (
```
KingYes: Please use variable for this condition

KingYes (changes requested):

PR #1881, opened 2017-06-22
register_controls method should be private

nextend:
register_controls method should be private as it can be called only once as the second time it throws class already defined errors. Proper access to this method is through: get_controls()


KingYes:
I like this, thanks :)

PR #1976, opened 2017-07-08
Merge your saved sections as a template!! 

elemounter:
Now you can only save sections separately, but it's not possible yet to merge the saved sections. Benefits: 
- With this option you will save time. 
- If you want to copy half of your page only, you don'y have to save all the sections separately.


KingYes:
What is it?

PR #2045, opened 2017-07-20
Elementor in wordpress categories

JuanFcoRomero:
Hello,

I tried to edit wordpress categories with Elementor but I can't. I think it will be a very demand new feature in Elementor for wordpress users. I will love see it!

(Posts > Categories and editing de Categori with Elementor)


KingYes:
Please open new issue about this.. Not Pull Request.

PR #2047, opened 2017-07-20
Fixed array iterator for old php versions

kobizz:


includes/controls/groups/base.php
```diff
@@ -15,6 +15,8 @@
 
 		$filtered_fields = $this->prepare_fields( $filtered_fields );
 
+		reset( $filtered_fields ); // For php < 7
```
KingYes: Write the comment above..

Issue #2136, opened 2017-08-04
Cannot get rid of plugin data collection message


nextend:
Elementor 1.6.2 free

`Love using Elementor? Become a super contributor by opting in to our anonymous plugin data collection and to our updates. We guarantee no sensitive data is collected. Learn more.`

The message does not goes away when I hit "No thanks".

https://www.youtube.com/watch?v=sBpl0Zx9foU&feature=youtu.be

arielk:
It looks like you have a JS error on the page caused by conflict with other plugins, try turning off plugins and check if it works out for you.

arielk:
@nextend It looks like you have a JS error on the page caused by conflict with other plugins, try turning off plugins and check if it works out for you.

nextend:
@arielk Why a JavaScript error can affect this button? It simply points to an url which loads fine: /***/wordpress3/wp-admin/admin.php?page=elementor&elementor_tracker=opt_out&_wpnonce=da1c857b9f

The "Sure! I'd love to help" button is not working either.

I made the debug for you:
The Tracker class init static method called after the following actions happen in WordPress:
```
auth_cookie_valid
auth_redirect
_admin_menu
admin_menu
admin_init
admin_bar_init
add_admin_bar_menus
elementor/admin/after_create_settings/elementor
```
Then the init method adds adds to the `admin_init` action the following:
`add_action( 'admin_init', [ __CLASS__, 'handle_tracker_actions' ] );`

So the problem is that the admin_init action happened earlier than the Tracker::init called. 

Possible solution:
```
	public static function init() {
		add_action( 'elementor/tracker/send_event', [ __CLASS__, 'send_tracking_data' ] );
		self::handle_tracker_actions();
		add_action( 'admin_notices', [ __CLASS__, 'admin_notices' ] );
	}
```

cavalierlife:
I was just coming here to submit this bug, so obviously I have the problem as well.

KingYes:
Thanks.

PR #2139, opened 2017-08-04
handle_tracker_actions method never called

nextend:
Fix for #2136


KingYes:
It's bug from our new autoloader. Fixed by https://github.com/pojome/elementor/commit/0766777c578f0aeaa61b1144f6737dabc2c0d2dc

Thanks..

PR #2147, opened 2017-08-07
Possibility to override easily default widget

matthieumota:
I don't know if you agree with that. But I purpose because it's overkill to create a complete button widget if I only want add/edit a simple control. It is enough for the developer to create a file in a specific location of his theme.
Example : I've created elementor/widgets/button_override.php in my template folder
"_override" must be respected to avoid mistakes with other classes.

```php
<?php

namespace Elementor;

class Widget_Button_Override extends Widget_Button
{
    // Write override here...
}
```


KingYes:
Thank you for your PR, but this feature is not part of our roadmap.

PR #2166, opened 2017-08-09
CSS select only elementor related iframes

norgeous:
Hello,
I would like to use iframes like this:
```
<div style="text-align:center;">
    <iframe src="localhost" width="800"></iframe>
</div>
```
iframes are indeed inline elements. so the above method should work fine...
However, Elementor is over-zealously selecting all iframes with the width 100% css rule.

Would be great if Elementor would only select the iframes that it inserts.

In this PR, I had a go at changing the `frontend.css` to do the above, but I can't be sure that I didn't break something that requires the iframe rules - may have accidentally ommited a few selectors because I don't know the project well enough yet.

Please, let me know your thoughts!

### DIFF
--- a/assets/css/frontend.css
+++ b/assets/css/frontend.css
@@ -1,4 +1,4 @@
-@charset "UTF-8";
+﻿@charset "UTF-8";
 .dialog-widget {
   position: fixed;
   height: 100%;
@@ -130,7 +130,9 @@ body.admin-bar .dialog-close-button {
   .elementor .elementor-widget:not(.elementor-widget-text-editor) figure {
     margin: 0; }
   .elementor embed,
-  .elementor iframe,
+  .elementor-custom-embed iframe,
+  .elementor-soundcloud-wrapper iframe,
+  .elementor-video-wrapper iframe,
   .elementor object,
   .elementor video {
     max-width: 100%;


joshmarom:
Hi @norgeous and thank you for your insight.

The CSS rule you are referring to is a part of our a group of reset rules that are meant to keep elements inside Elementor in a uniform basic style and prevent unexpected conflicts.

This specific rule is meant to prevent iframes (among other elements) from overflowing and exceeding the width of their parent containers (ie widget/column/section), by setting the **`max-width`** to 100% (this does not have any effect on the width unless it's greater than its parent).

We believe this behavior is most the commonly desired behavior and should, therefore, be set as default.

If for whatever reason you would like to have your iframe exceed its parent width, the appropriate way to handle that would be by selectively overriding the `max-width` value to `none`.

Thank you.

norgeous:
Hi @joshmarom, thanks for such a fast response!

Setting `max-width:none;` doesn't work for my purpose. My issue is with the fact that once `width:100%;` has been set on an iframe it will never respect the attribute `width="800"`. The CSS style cannot be overridden by any value of `width` such as `auto`, `unset`, `initial`, etc.

Here are some steps to repro my issue;
1. Use Elementor to put a 'html element' into a page with the following code:
```
<div style="text-align:center;">
    <iframe src="localhost" width="800" style="width:auto; max-width:none;"><iframe>
</div>
```
2. Save the page and view it on the front end.
3. The iframe has a width of `300px`, the default for iframes and is not respecting `width="800"`.
3. Use element inspector to turn on/off the inline styles on the element. Try setting the inline style to `width:auto;`, `width:unset;` or `width:initial;`.
4. Observe that the iframe either has a width of `100%` or `300px`, but never `800`.

I would still like to make a modification that will satisfy my requirements for not having the CSS width setting on all iframes, whilst preserving your requirements for responsive elements. I will attempt another separate PR after you respond to this question;

What do you think you would be the best way to prevent the `width:100%` getting to my iframe? I think there are few ways we can resolve this... I will attempt another shortly.

Thanks,
norgeous

PR #2177, opened 2017-08-11
WP CLI command to regenerate the elementor CSS.

### LINKED ISSUE #2161: [FEATURE REQUEST] WP CLI Command to regenerate the CSS
This command can be chained in the automatic site migration tools to clear the cache after the site is migrated to live from staging or vice-versa.

Right now this can be quickly run from WP CLI like this - `wp eval "Elementor\Plugin::$instance->posts_css_manager->clear_cache();" ` Although it would be nice to have this as a dedicated command.

Nikschavan:
First pass at adding the new WP CLI command to elementor.

Fixes #2161

### DIFF
--- a/includes/plugin.php
+++ b/includes/plugin.php
@@ -226,6 +226,10 @@ class Plugin {
 		if ( defined( 'DOING_AJAX' ) && DOING_AJAX ) {
 			new Images_Manager();
 		}
+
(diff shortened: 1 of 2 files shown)


KingYes:
https://github.com/pojome/elementor/issues/4861

PR #2217, opened 2017-08-20
Set font size inherit (fix #2098) for Heading widget

### LINKED ISSUE #2098: Feature Request - Apply Style to Anchor - Toggle
So many times I've seen someone on the forums struggle to overcome font and color changes when an element such as heading or text, etc includes anchors and the underlying theme css overrides the desired look.

Those of us w/ the know how can use custom css but its certainly not intuitive for future edits or clients making adjustments, they'll just think it's broken.

screen shot 2017-07-27 at 11 19 05 pm

But I had a thought...why not just place a toggle that switches the underlying style selector generated by Elementor to target an anchor as well?

` .elementor-element h2, .elementor-element h2 a { .... }`

` .elementor-element p, .elementor-element p a { .... }`

Of course these are simple examples but the proper selectors would be a tad more authoritative to get the job done (i.e. elementor body class + .. + unique element selector)

The toggle could simply be a check box when an anchor exists, right below the other choices for open in a new window and no follow.

screen shot 2017-07-27 at 11 20 41 pm

Or maybe just make it intuitive enough to detect a value in the url/link field and then provide it's own style color/font settings for the "Anchor" in the Style tab?

arielk:


KingYes:
Fix #2098

PR #2251, opened 2017-08-24
Lightbox UI color

joshmarom:


core/settings/general/model.php
```diff
@@ -104,7 +104,14 @@ public static function get_controls_list() {
 								'.elementor-lightbox' => 'background-color: {{VALUE}}',
 							],
 						],
-
+						'elementor_lightbox_ui_color' => [
+							'label' => __( 'UI Color', 'elementor' ),
+							'type' => Controls_Manager::COLOR,
+							'selectors' => [
+								'.elementor-lightbox .dialog-lightbox-close-button' => 'color: {{VALUE}}',
+								'.elementor-lightbox .elementor-swiper-button' => 'color: {{VALUE}}',
```
kobizz: You can just put both selectors in one key separated by comma

PR #2300, opened 2017-09-01
Fix problem with decimal comma produced in CSS files

milsorm:
There is decimal commas instead of decimal points on some locale settings (CZ, SK, ES).
But to CSS files there is necessary to produce only decimal points.

This solution fixed output to CSS files of that numeric values.


milsorm:
:-( but it was failing before too

PR #2437, opened 2017-09-19
Fix problem with decimal comma produced in CSS files in all CSS files

milsorm:
There is decimal commas instead of decimal points on some locale settings (CZ, SK, ES).
But to CSS files there is necessary to produce only decimal points.

This solution fixed output to CSS files of that numeric values.
In comparison with #2300 also solve e.g. column widths.


milsorm:
Sry for that but working with ugly git is exhaustive for me. Milan Sorm and milsorm are same accounts. I have no problem with CLA. I want to include such small bug correction into mainline. But I cannot spend several hours to creating pull requests which confirm all these conditions :(

Just use it or leave it. I have patch for myself.

milsorm:
Still an issue and it is pitty to apply this patch over new and new version of Elementor :-(

Djimusic:
@milsorm Would you like me to recommit this change since my account has no issues with the CLA?

milsorm:
Bad bad bad git :-( I have to patch it manually every update. Ugly!

edemir206:
I have exactly the same problem, but my language is Portuguese.

bainternet:
@milsorm 	

I'm going to close this for now, mainly because the suggested solution is a major performance hit.

But we will think of adding a filter hook to allow this when needed.

thanks, and sorry for the late reply.

Issue #852, opened 2016-11-14
Use a custom font in texts


mrprunen:
## Description
Elementor is really great. 
However sometimes you may need to utilize some of custom fonts (that not part in the google font family) for text elements.

temporary i've  found workaround using child theme with font-face and inject code into text content; 
es <p style="font-family: 'klinic_slabbook';">

Works for me, but 
It would be nice to find a solution to load webfonts in "fonts folder" of the theme and make sure that in addition to system fonts and google fonts,  "uploaded fonts" are selectable.

what do you think?
<img src="http://dev.eticru.org/tmpdocs/Immagine1.png" alt="Image1" />

<img src="http://dev.eticru.org/tmpdocs/Immagine2.png" alt="Image2" />

## Environment
					
== Server Environment ==
	Operating System: Linux
	Software: Apache
	MySQL version: 5.5.52
	PHP Version: 5.4.45
	PHP Max Input Vars: 1000
	PHP Max Post Size: 128M
	GD Installed: Yes
	Elementor Library: connected

== WordPress Environment ==
	Version: 4.6.1
	Site URL: http://dev.eticru.org
	Home URL: http://dev.eticru.org
	WP Multisite: No
	Max Upload Size: 128 MB
	Memory limit: 40M
	Permalink Structure: /%category%/%postname%/
	Language: it-IT
	Timezone: 0
	Debug Mode: Inactive

== Theme ==
	Name: GeneratePress Child ETICRU
	Version: 0.1
	Author: Marco Bruno
	Child Theme: Yes
	Parent Theme Name: GeneratePress
	Parent Theme Version: 1.3.40
	Parent Theme Author: Tom Usborne

== User ==
	Role: administrator
	WP Profile lang: it_IT
	User Agent: Mozilla/5.0 (Windows NT 10.0; WOW64; rv:49.0) Gecko/20100101 Firefox/49.0

== Active Plugins ==
	Clef
		Name: Clef
		Version: 2.6.1
		Author: Clef

	Elementor
		Name: Elementor
		Version: 0.10.7
		Author: Elementor.com

	Elementor Templater: ElemenTemplator
		Name: Elementor Templater: ElemenTemplator
		Version: 1.0.7
		Author: WPDevHQ

	GP Premium
		Name: GP Premium
		Version: 1.2.91
		Author: Tom Usborne

	iThemes Security
		Name: iThemes Security
		Version: 5.7.0
		Author: iThemes

	Limit Login Attempts
[...]

PR #2585, opened 2017-10-13
Font register filter

patryk-uszynski:
Hi,

I've added filter, which provides ability to register own font names. There isn't any good way to do this yet, so i think this might be helpful ;) Here - #852 for example.


patryk-uszynski:
Example usage:
```php
add_action('elementor/fonts_registered', function($fonts) {
	$fonts['My new font'] = 'local';

	return $fonts;
});
```

ramiy:
Nice one!

KingYes:
This code has a memory leak issue. We used `return` the data this way when the function executed all fonts data remove from the memory. In your code it's always keep in the memory inside the new property you created.

patryk-uszynski:
I wanted to avoid multiple `apply_filters` calls. What if I modify that, make property only for new fonts and `return` the data old way with `array_merge`?

patryk-uszynski:
@KingYes any suggestions?

KingYes:
Fixed by #3336

Thanks

PR #2604, opened 2017-10-18
Inline editing

kobizz:


includes/widgets/alert.php
```diff
@@ -287,24 +293,13 @@ protected function render() {
...
-			if ( 'show' === settings.show_dismiss ) {
-				html += '<button type="button" class="elementor-alert-dismiss">X</button>';
-			}
-
-			html += '</div>';
-		
-			print( html );
-		}
-		#>
+		<div class="elementor-alert elementor-alert-{{ settings.alert_type }}" role="alert">
```
KingYes: Return the old code

includes/widgets/button.php
```diff
@@ -430,7 +430,7 @@ protected function _content_template() {
 						<i class="{{ settings.icon }}"></i>
 					</span>
```
KingYes: Move this method to below

KingYes (changes requested):

PR #2629, opened 2017-10-25
php-doc: add @since and @access tags

matipojo:


matipojo:
@ramiy auto generated...

PR #2663, opened 2017-10-30
Image Carousel Links

nesoor:
Hey,
I am not sure if this is the right place to post my request ? 

It would be cool if I can put a link in each photo of the image carousel. 
To make it even more advanced it would be nice if you can place text under it and a button.


KingYes:
There: https://github.com/pojome/elementor/issues

PR #2697, opened 2017-11-06
Include your own font icons (linea-Icons)

WAM-Meerbusch:
A possibility to integrate own font icons like Linea Icons
So this z.b. in the FlipBox, Iconbox or can be selected.
Would really help if that would work


KingYes:
Please create new issue there: https://github.com/pojome/elementor/issues

PR #2703, opened 2017-11-06
Added Meetup to the Social Icons widget

rsiddall:
The new Meetup swarm icon was added in Font Awesome 4.7. The official background colour was eyedropped from an image in an article announcing the Meetup redesign.


KingYes:
Thanks! 👍

PR #2835, opened 2017-11-27
Accessibility: accessible widget search

ramiy:
The widget search is not accessible. It should either use `role="search"` or `type="search"`.

In addition, the `input` field should have a `label` for screen readers. We hide the `label` for regular users using `class="screen-reader-text"`, this way only screen readers "see" it.

Ref: https://www.w3.org/TR/html-aria/

elementor-widget-search


ramiy:
@arielk Please review the PR.

ramiy:
@kobizz, @matipojo, @joshmarom

on the same principle, when we define **Controls** with `'show_label' => false` argument, the Label should not be removed, it should have `class="screen-reader-text"`. This way screen readers "see" it but normal users don't.

Issue #1579, opened 2017-04-03
Enhancing the "Page Setting" panel


ramiy:
## Description

The page setting panel added in v1.4.0 is awesome! I like the direction where it's going.

The next step should be to allow users to update all the page content from Elementor - titles, excerpts, authors, publish date, etc.... Also, the content should be separated from the settings.

This is my suggestion:

![page-settings](https://cloud.githubusercontent.com/assets/576623/24593314/d7ae5da0-182a-11e7-849e-3e36b350d285.png)

WPDevHQ:
+1
WYSIWYG for the excerpt input. The featured image field would be an added bonus!
Maybe split it into 3 tabs instead of 2? i.e. Settings - Content - Style

kokers:
What would be great also, is to move this as a tab next to elements|global http://prntscr.com/hgvx45 I find myself going there quite a lot, especially when there are some custom settings added. This could be optional, but either way, Elements/Global in sidebar are related to current page. All options hidden under hamburger menu are somewhat global settings. Adding a link to page settings upfront is more fitting imo. It could be as a simple gear icon.

Issue #1571, opened 2017-04-02
Enhancment: Make controls "readonly" or "disabled"


ramiy:
## Description

Currently, plugin developers can't set controls as readonly or disabled. For example I can't display a [Text Field](https://github.com/pojome/elementor/blob/master/docs/content/controls/_text.md) as a read-only field, instead I need to hack the code or display HTML using the [Raw-HTML](https://github.com/pojome/elementor/blob/master/docs/content/controls/_raw-html.md) control.

We can add two new argument to define a field as 'readonly' and 'disabled' fields.

### Usage

```php
$this->add_control(
  'widget_title',
  [
     'label'       => __( 'Title', 'text-domain' ),
     'type'        => Controls_Manager::TEXT,
     'readonly'    => true,
     'disabled'    => true,
  ]
);
```

### Arguments

Argument       | Required   | Type         | Default                      | Description
------------   | :--------: | :------:     | :--------------------------: | ---------------------------------------------
`readonly`      | no         | *`bool`*   |  `false`                     | Display the input field as read-only
`disabled`  | no         | *`bool`*     | `false`                       | Display the input field as disabled

### Preview

![elementor](https://cloud.githubusercontent.com/assets/576623/24584130/72fa3cc2-176b-11e7-99de-834702cd3ae6.png)

ramiy:
**Update**: we can't use ```disabled``` because it removes the value. When saving data with a disabled field, it doesn't send the field value. This can cause data loss.

bainternet:
What is a typical use case for this?

ramiy:
I can think of two use case:

1. **DEVELOPMENT:** Elementor Plugins developers (and Elementor Pro) can use it to show disabled/readonly the field in free versions, and to show editable in Pro versions.

2. **USER EXPERIENCE:** With conditional logic in controls Elementor totally hide the field if the condition fails. With disabled/readonly fields, the user can see the extra field, but not change it.

LightXServices:
1+

preyes24973:
+1

djoo:
+1

black-eye:
+1

peteratdjeeni:
+1000

LightXServices:
A use chase could be to show data on the input using javascript.

[Logo]<https://www.lightx.co.il>



Neriya Rosner
CEO - LightX
P:  +13237403901<tel:+972545428866>
P:  +972545428866<tel:+972545428866>
[...]

PR #2837, opened 2017-11-27
Add excerpt field to Elementor page settings

ramiy:
Related: #1579 (Enhancing the "Page Setting" panel)

page-settings-excerpt

### DIFF
--- a/core/settings/page/manager.php
+++ b/core/settings/page/manager.php
@@ -90,6 +90,8 @@ class Manager extends BaseManager {
 
 		$post->post_title = $data['post_title'];
 
+		$post->post_excerpt = $data['post_excerpt'];
+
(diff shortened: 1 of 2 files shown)


arielk:
Thanks @ramiy 

But you need to check if the current CPT is support `excerpt` parameter

ramiy:
@arielk Done.

Tested on **Pages** that don't support excerpts. And then add excerpt support with [Page Excerpt Plugin](https://wordpress.org/plugins/page-excerpt/).

ramiy:
The PR hides the `excerpt` field when the post type has no excerpt support. But I think that it would be better to display the excerpt field and mark it with `disabled` attribute (see #1571).

KingYes:
@arielk Nice catch!

Thanks @ramiy

PR #2854, opened 2017-11-28
Fixes Image Box heading to have a working link even without an image

Jaska:


KingYes:
It does not make sense to display links using the Image Box widget without the image itself.

KingYes:
@Jaska We have further looked into this and apparently you were right.
We will release a fix for it in the upcoming update.
Thanks for your input.

Jaska:
Thanks!

PR #2861, opened 2017-11-29
a11y: Accessibility love

ramiy:
This PR fixed several accessibility issues. Separate commit for each issue.


ramiy:
## Replace "title" attributes with "aria-label" in <a> tags

The first commit removes `title` attributes from `<a>` tags and replaces them with `aria-label`.

From accessibility point of view, the `title` attribute doesn't help the screen reader. Instead we should use `aria-label`. But it should be used only to describe the text. For example: `<a href="">View</a>` view what? We should use: `<a href="" aria-label="View post">View</a>`.

Blind people can't see the screen, this is why we use `aria-label` to add more info for screen readers.

On the other hand, this code: `<a href="" aria-label="View">View</a>` it has no meaning. It does not add more information.

Reference from WordPress core: https://core.trac.wordpress.org/search?changeset=on&q=Accessibility+Remove+title+attributes&page=2&noquickjump=1

ramiy:
In the other 5 commits I'm removing the `title` attribute from elements like `<div>`, `<label>` and `<li>`.

In some places we need to **remove** the `title` attribute, in some places **move** the `title` attribute to inner elements, and in other places **replace** it with screen-reader-only text.

### Preview switcher:
elementor-preview

### Template library actions:
elementor-library

### Edit tools for widgets/columns/sections:
elementor-edit-tools

ramiy:
The progress widget commit above is based on https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_progressbar_role  and several other similar resources.

Issue #2450, opened 2017-09-20
Accordion widget looses its content while editing.


ryanlabelle:
## Description

When adding and editing the Accordion widget, the content disappears and it's not retrievable.

Related to: https://github.com/pojome/elementor/issues/2324

## Steps to reproduce

1. Fresh install of WordPress 4.8.2
2. Fresh install of Elementor  1.7.4
3. Activate Twenty Seventeen Theme
3. Edit Sample Page with Elementor
4. Drag Accordion Widget onto screen.
5. Start adding content to content area, adjusting sort order of accordions while editing and expanding / contracting boxes. Just play around with it. 

It's intermittent but there is something going on for sure. I took a video: https://cl.ly/1X0M1d2B2A1F

I also didn't see any javascript errors in the console. 

## Environment

					
== Server Environment ==
	Operating System: Darwin
	Software: Apache/2.4.23 (Unix) PHP/5.6.28 LibreSSL/2.2.7
	MySQL version: 5.7.16
	PHP Version: 5.6.28
	PHP Max Input Vars: 1000
	PHP Max Post Size: 8M
	GD Installed: Yes
	Elementor Library: Connected

== WordPress Environment ==
	Version: 4.8.2
	Site URL: http://local/elementor
	Home URL: http://local/elementor
	WP Multisite: No
	Max Upload Size: 8 MB
	Memory limit: 40M
	Permalink Structure: /%postname%/
	Language: en-US
	Timezone: 0
	Debug Mode: Active

== Theme ==
	Name: Twenty Seventeen
	Version: 1.3
	Author: the WordPress team
	Child Theme: No

== User ==
	Role: administrator
	WP Profile lang: en_US
	User Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/60.0.3112.113 Safari/537.36

== Active Plugins ==
	Elementor
		Version: 1.7.4
		Author: Elementor.com

== Debug ==
	Errors: There are no errors to display

ryanlabelle:
I've noticed that when switching the text control from WYSIWYG to TEXTAREA the content is not lost. Any idea when you'll be looking into this? Thank you for your awesome support.

bainternet:
If issue persists please contact our support at support@elementor.com

Issue #2324, opened 2017-09-05
Accordion widget looses some of its content while editing.


lukasalperowitz:
## Description

Accordion widget looses some of its content while editing.

## Steps to reproduce

* Add an accordion widget
* Add 8-10 entries
* Fill in header and content
* Save the page

Exspected behavior:
* All entries with header and contents are stored and visible on page

Actual behavior:
* Headlines of all entries are stored and visibile
* Content of some (2-3) elements is empty, content entered before is gone.

## Environment
		
== Server Environment ==
	Operating System: Linux
	Software: Apache/2.4.27 (Unix)
	MySQL version: 5.5.5
	PHP Version: 5.6.31
	PHP Max Input Vars: 1000
	PHP Max Post Size: 64M
	GD Installed: Yes
	Elementor Library: Connected

== WordPress Environment ==
	Version: 4.8.1
	Site URL: http://new.nightoftheprogfestival.com
	Home URL: http://new.nightoftheprogfestival.com
	WP Multisite: No
	Max Upload Size: 64 MB
	Memory limit: 128M
	Permalink Structure: /%year%/%monthnum%/%day%/%postname%/
	Language: de-DE
	Timezone: Europe/Berlin
	Debug Mode: Active

== Theme ==
	Name: Twenty Seventeen Child-Theme
	Version: 1.0
	Author: Lukas Alperowitz
	Child Theme: Yes
	Parent Theme Name: Twenty Seventeen
	Parent Theme Version: 1.3
	Parent Theme Author: the WordPress team

== User ==
	Role: administrator
	WP Profile lang: de_DE
	User Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_0) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/60.0.3112.113 Safari/537.36

== Active Plugins ==
	Custom Facebook Feed
		Version: 2.4.6
		Author: Smash Balloon

	Custom Twitter Feeds
		Version: 1.2.6
		Author: Smash Balloon

	Easy Google Fonts
		Version: 1.4.3
		Author: Titanium Themes

	Elementor
		Version: 1.7.1
		Author: Elementor.com

	Instagram Feed
		Version: 1.4.9
		Author: Smash Balloon

	Jetpack by WordPress.com
		Version: 5.2.1
		Author: Automattic

	Polylang
		Version: 2.2.1
		Author: Frédéric Demarle

	YouTube
		Version: 11.8.2
[...]

PR #2899, opened 2017-12-06
Fixed tinyMCE inside repeater

### LINKED ISSUE #2897: Tab content lost when dragging but releasing on same position
hy,
i see this issue when working in chrome and firefox on mac (latest versions). i use the tab-widget with a few tabs. when i start dragging one tab to another position but then release it to early, so that it jumps back to the original position, the content of this tab is lost.
this happens pretty often as sometimes when you click on it, it makes a small movement...

i have also seen it in various pages where we use elementor with different configuration of other themes/plugins/...

hope you can reproduce and fix it :)
matt

kobizz:
Fixes #2897, #2450, #2324


dahutt:
bildschirmfoto 2017-12-13 um 18 36 08

I still get sometimes only the text view

PR #2909, opened 2017-12-07
a11y: Use `aria-hidden=true` in screen-readers to hide icons used purely for decoration

ramiy:
Finishing the work started in #2861


ramiy:
This PR also adds screen-reader text to some icons.

**REPEATER:**

repeater

**PANEL FOOTER:**

panel-footer


And other places....

PR #2926, opened 2017-12-11
 Make Social Icon widget accessible

joshmarom:


ramiy:
@joshmarom why `sr-only`? I used `elementor-screen-only` in #2861. We need to set a standard.

Issue #2088, opened 2017-07-27
web font lazy load


luizbills:
An option to lazy load web fonts (google fonts and icon fonts) will increase performance and prevent from text blinking on page loading.

### Useful links

- https://www.zachleat.com/web/comprehensive-webfonts/
- https://github.com/typekit/webfontloader

bainternet:
I merged this issue with  #2987  , which is on the same subject so we could deal with it more efficiently. 
Please feel free to continue the conversation there.

Thanks for your feedback!

maxwellhibbert:
Better font loading is really needed. Currently ALL weights of any fonts used are loaded!

PR #2987, opened 2017-12-19
Change fonts to use Web Font Loader 

mnwalker:
Have changed font loading from CSS link to asynchronous JS using Type kit webfont loader as requested in #2088
See https://github.com/typekit/webfontloader for full info

**Benefits:**
Faster page loading and improved google pagespeed score by preventing rendering delays
Supports moving request to footer instead of head
Helps display fallback font rather than blank while waiting for font to load

**Downsides:**
Depends on Javascript

If the maintainers/community think this is beneficial but should be optional instead of mandatory let me know and will work at making it a configurable option.

### DIFF
--- a/includes/frontend.php
+++ b/includes/frontend.php
@@ -43,7 +43,6 @@ class Frontend {
 			add_action( 'wp_enqueue_scripts', [ $this, 'enqueue_styles' ] );
 		}
 
-		add_action( 'wp_head', [ $this, 'print_google_fonts' ] );
 		add_action( 'wp_footer', [ $this, 'wp_footer' ] );
 
 		// Add Edit with the Elementor in Admin Bar.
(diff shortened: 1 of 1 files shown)


bainternet:
@mnwalker 

Thank you, we are thinking about it but no as the only method to load fonts, meaning you will be optional.

benqzq:
Transferred from [here](https://github.com/pojome/elementor/issues/3045):

## Fonts aren't loaded in font-face

Whenever I use Google fonts from Elementor I get a notable decrease in Google Pagespeed Insights (GPI) score (`>=30 points`).

If I disable Elementor font-load, and load fonts manually via `@font-face`, I don't have that problem.

**Why it's important to fix it**

I had a case when a site had 100/100 (and that notably gave an SEO boost) and it downgraded its GPI score from 100 to about 67/100.

I think it's the only task left to make Elementor totally SEO friendly.

**Why @font-face is the only way to go**

From my experience, loading fonts in any way which isn't `@font-face` is doomed to fail. Google only desires manual font load as with `@font-face`.

**A possible handling algorithm**

1. If the user selects a font that doesn't exist, download the font-files from Google to a dedicated directory.
2. Add a `@font-face` rule in a dedicated typography CSS file.
3. Create a `font-family` command, for the selected font.

yotamselementor:
Hi @mnwalker ,
This PR was unintentionally closed - I removed the obsolete master branch which caused a chain effect of auto-closing all PRs directed at that branch.
I am looking for a way to reopen these PRs against main.

Thanks for your understanding!

PR #3109, opened 2018-01-08
Added tooltip to 'reset to default' popover-toggle button

kobizz:


ramiy:
@kobizz Please don't forget the a11y perspective. Ref: 98e83f99b07e89305737812aa01102d08804c223

PR #3149, opened 2018-01-11
Docs: add phpDocs to classes

ramiy:


includes/template-library/sources/local.php
```diff
@@ -14,36 +14,92 @@
...
+	 */
 	public static function get_template_type( $template_id ) {
 		return get_post_meta( $template_id, self::TYPE_META_KEY, true );
 	}
 
 	/**
-	 * @static
+	 * Is base templates screen.
+	 *
+	 * Whether the current screen is a base template.
```
bainternet: current screen base is edit and post type is template

bainternet:
@KingYes looks good.

PR #3199, opened 2018-01-16
Flatpickr script should be enqueued in the footer like the others

nicomollet:
Is there a reason for not enqueing flatpickr.js in the footer?
It is render blocking. 

Thanks.


KingYes:
Thanks

PR #3250, opened 2018-01-20
Update _media.md

stratboy:
Added an important comment for Utils usage

### DIFF
--- a/docs/content/controls/_media.md
+++ b/docs/content/controls/_media.md
@@ -18,7 +18,7 @@ $this->add_control(
      'label' => __( 'Choose Image', 'your-plugin' ),
      'type' => Controls_Manager::MEDIA,
      'default' => [
-        'url' => Utils::get_placeholder_image_src(),
+        'url' => Utils::get_placeholder_image_src(),  // Rember to add 'use Elementor\Utils' at the top of the document;
      ],
   ]
 );


Sheldon-Monk:
Have never done GitHub before, so hoping I am posting in the right  area. Please correct me if I've mis-posted.

When  you consider the next rev of Elementor, one feature you may want to  consider adding is one in the plugin Microthemer: the Font Family menu.  (You can see it at the time of about 3:10 in the YouTube video https://www.youtube.com/watch?v=NZFbc-KuMPk .  The drop-down menu they use is extremely helpful in that one can see  the actual typefaces, making selection far, far, far faster.  Just a thought....

ramiy:
@Sheldon-Monk This kind of change should be either applied to all the docs or to non of them.

As the use of namespaces is a common practice in PHP, your comments is redundant. It's a common knowledge.

stratboy:
@ramiy it's me I made the edit to the md file.

I think the comment should be left there because the specific Utils class it's in fact not present in all the examples. In fact, it's the only place I saw it for now. So the thing "apply to all or none", doesn't apply to this specific case. 

Also, even if I'm an experienced developer, I copied the code and (obviously) got errors. I then fixed it of course, but we should consider that not all the developers out there really are used to oop oriented programming and thus could not know the problem or easily fix it. Docs should be for all, not only for some.

Sheldon-Monk:
@stratboy  Thank you!
I just wanted to add to the improvement of the product, ans was directed here.  I can see that communication could be greatly improved if the instructions of what I should have done once here would have been included.  It's not intuitive for the average Elementor user.  (Maybe there's a better forum or repository or method for contributing potential upgrades.  You guys would know that better than I.)
Again, thank you.

Issue #3098, opened 2018-01-06
Feature Request: Enhance Image Box Widget


ramiy:
## Description

The **Image** widget allows the user to select the **Image Size**:

![elementorimage](https://user-images.githubusercontent.com/576623/34644642-da02aff4-f342-11e7-9378-f7f70b5177ee.png)

The same control should be added to the **Image Box** widget:

![elementorimagesize](https://user-images.githubusercontent.com/576623/34644655-2e9603ae-f343-11e7-9bc3-b583bd91cbda.png)

I should mention that The **Image Carousel** and the **Image Gallery** widgets also allows the user to select the **Image Size**.

bainternet:
@ramiy 

Seems fair, send PR and will have a look

amindiary:
Can you add border radius for image too? 
to be able make this:
<a href="https://hizliresim.com/azqlbz"><img src="https://i.hizliresim.com/azqlbz.png"></a>

ramiy:
@bainternet PR added!

ramiy:
Closing the issue as the PR was merged.

Issue #928, opened 2016-11-23
Add image size control for Image Box widget


matthiaswagnerwy:
hy,
when using the image box it should be possible to choose the desired size like it is done when placing a single image.

thanks :)

m9n:
Agreed! If the original image is really big, we're currently forced to use massively oversized images in the page.

For a temporary workaround, an editor element can be used, but @matthiaswagnerwy's suggestion is needed in the longer term.

m9n:
In fact, it would also be good to see a max-width control alongside image size.

DavidCaum:
Size control should present for all image used. In all widgets. Even background image.

ionlantern:
I'd love that feature! Make it happen soon ;)

grafx77:
Same here....need this feature.

matthiaswagnerwy:
while it is great to see how you develop elementor and also react to many feature requests, i am still wondering why image size is still not available in all widgets. i just stumbled upon the testimonial widget in a project and feel forced to upload duplicate images to my library in order to optimize loading times... when can we expect to see the selection in other widgets and also for background images?

PR #3251, opened 2018-01-20
Add "Image Size" control to the "Image Box" widget

ramiy:
Fixing #3098

image-box

----

This way I can set **Custom** sizes and use less bandwidth using smaller images!

image-box-custom-sizes

### DIFF
--- a/includes/widgets/image-box.php
+++ b/includes/widgets/image-box.php
@@ -99,6 +99,14 @@ class Widget_Image_Box extends Widget_Base {
 			]
 		);
 
+		$this->add_group_control(
+			Group_Control_Image_Size::get_type(),
+			[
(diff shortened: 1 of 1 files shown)


arielk:
Need more review by UI team

ramiy:
@KingYes

kobizz:
Looks good

ramiy:
@KingYes Ok, so we have a problem with this PR, the new control uses the `image_size` name, the same name Elementor uses in all widgets.

But this specific widget already uses the `image_size` name in a different control.

This is why Travis failed. I see a PHP notice in my `debug.log` file.

ramiy:
Also fixing #928

PR #3254, opened 2018-01-20
Escaping output data

ramiy:
## Performance vs. Security

This PR remove all the output data escaping functions for **better performance**.

#3113 PR adds more escaping functions to all output data for **better security**.


includes/admin.php
```diff
@@ -588,12 +588,12 @@ public function elementor_dashboard_overview_widget() {
 					?>
 				</div>
 				<div class="e-overview__create">
-					<a href="<?php echo esc_attr( Utils::get_create_new_post_url() ); ?>" class="button"><span aria-hidden="true" class="dashicons dashicons-plus"></span> <?php esc_html_e( 'Create New Page', 'elementor' ); ?></a>
+					<a href="<?php echo esc_attr( Utils::get_create_new_post_url() ); ?>" class="button"><span aria-hidden="true" class="dashicons dashicons-plus"></span> <?php _e( 'Create New Page', 'elementor' ); ?></a>
```
norewp: Should this not be esc_url instead of esc_attr?

PR #3262, opened 2018-01-21
 Video Widget - Play Icon: Replace the select control with a switcher control

### LINKED ISSUE #3241: Video Widget - Image Overlay Play Icon
## Description

The **Video Widget** allows the user to set a custom **Image overlay** with a **Play Icon**.

The play icon is based on the **Select Control** but it has only two options (Yes / No).

We should use **Switcher Control** for better UX, this way the user can change the selection with one button click instead of two.

video-play-icon

ramiy:
Fix #3241

## Current

video-play-icon

## New

video-play-icon2

### DIFF
--- a/includes/widgets/video.php
+++ b/includes/widgets/video.php
@@ -335,12 +335,11 @@ class Widget_Video extends Widget_Base {
 			'show_play_icon',
 			[
 				'label' => __( 'Play Icon', 'elementor' ),
-				'type' => Controls_Manager::SELECT,
+				'type' => Controls_Manager::SWITCHER,
 				'default' => 'yes',
-				'options' => [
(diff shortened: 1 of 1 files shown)


ramiy:
I can also update the text to **Show** / **Hide** just like text in the image-overlay control and some other controls in the video options section.

ramiy:
Also, the commit is backwards compatible as the `render()` function checks only for active state:

```php
<?php if ( 'yes' === $settings['show_play_icon'] ) : ?>
	<div class="elementor-custom-embed-play">
		<i class="eicon-play" aria-hidden="true"></i>
	</div>
<?php endif; ?>
```

This is why I added `'return_value' => 'yes',` to the control.

PR #3270, opened 2018-01-21
a11y: Video widget accessibility

ramiy:
## Description

The video widget need some accessibility love!

1. **Name** - The iframe should have an accessibility name (https://fae.disability.illinois.edu/rulesets/FRAME_2/), either `title` or `aria-label` tags.

2. **Action** - The play icon should have some text saying "Play video" or some thing like that. Currently, regular visitors **see** the play button and understand that clicking the icon actives the video, but blind people **don't see** the icon. We need to add an assistive text for them.

3. **Markup** - The play icon wrapper is a `<div>`, it should be a `<button>`. The second approach is to add `role="button"` to the `<div>` tag.

video-a11y


ashybaye:
Hello! Trying to make sure that the video components are indeed passing the accessibility checks. Was the `title` attribute added to the iframes? I still can't see it in the code when adding videos. Thank you!

Issue #3278, opened 2018-01-22
[Feature Request] Provide filter for droppable selector on column


jegbagus:
## Description

I need to create column to be following when scroll ( act as sticky sidebar). it will need to modify column wrapper structure. This change will cause column can't be added with another element anymore. its because strict selector on html5Droppable item.

so i propose to provide Filter on items of html5Droppable (column).

```javascript
onRender: function() {
	var self = this;

	var itemSelector = elementor.hooks.applyFilters( 'elements/column/render/droppable-item-selector', ' > .elementor-column-wrap > .elementor-widget-wrap > .elementor-element, >.elementor-column-wrap > .elementor-widget-wrap > .elementor-empty-view > .elementor-first-add', this );

	BaseElementView.prototype.onRender.apply( self, arguments );

	self.changeChildContainerClasses();

	self.changeSizeUI();

	self.$el.html5Droppable( {
		items: itemSelector,
		axis: [ 'vertical' ],
		groups: [ 'elementor-element' ],
		isDroppingAllowed: self.isDroppingAllowed.bind( self ),
		currentElementClass: 'elementor-html5dnd-current-element',
		placeholderClass: 'elementor-sortable-placeholder elementor-widget-placeholder',
		hasDraggingOnChildClass: 'elementor-dragging-on-child',
		onDropping: function( side, event ) {
			event.stopPropagation();

			var newIndex = jQuery( this ).index();

			if ( 'bottom' === side ) {
				newIndex++;
			}

			self.addElementFromPanel( { at: newIndex } );
		}
	} );
},
```

bainternet:
Currently, this feature is not supported, but we will look into it in the future if we see a large demand for it.

Thanks for the feedback.

PR #3279, opened 2018-01-22
Provide filter for droppable selector on column

jegbagus:
explained on #3278


bainternet:
@jegbagus 

Thanks for the PR, Currently we don't feel comfortable having a filter there.

thanks.

PR #3383, opened 2018-02-05
Bug fix: Image size control

ramiy:
When when using different key names for the media control and the image size control its not working. This PR fixes the issue.


ramiy:
Fixed the issue introduced in #3251

PR #3401, opened 2018-02-06
module: debug-mode

matipojo:
Open Elementor Editor and add `mode=debug`. for example: 
http://localhost/elementor/wp-admin/post.php?post=7737&action=elementor&mode=debug


ramiy:
But why in Elementor Core? why not a separate plugin?

PR #3429, opened 2018-02-08
Document Settings: add featured image support

ramiy:
@matipojo What do you think about that?


ramiy:
Implemented in #3456 in commit bd6a45df83f9703b1826c5a2b8412c5cbfe9bb4b

PR #3452, opened 2018-02-12
Add "caption_space" control to the Image Widget

ramiy:
The **Image Widget** allows us to add a caption. But the user can't control the space between the image and the caption.

This PR adds new slider control that adds top margin to the caption.

Note that the control was copied from the Image Box control.

image-spacing


ramiy:
image-caption-spacing

PR #3476, opened 2018-02-14
i18n: merge similar translation strings and update others

ramiy:
This PR removes two translation strings from the [Elementor plugin](https://translate.wordpress.org/projects/wp-plugins/elementor) by merging the string with similar translation strings.

Strings removed:
* `+ Add Images`
* `Add Image`

<s>Strings changed:
* `Choose Image` => `Choose an Image`</s>

----

i18n-a

i18n-b


ramiy:
The "`Choose Image`" to "`Choose an Image`" change was reverted due to @arielk request.

PR #3479, opened 2018-02-14
footer-saver: activate the `publish` button also if editor isn't chan…

matipojo:
…ged [fixes]


matipojo:
http://i.imgur.com/7VEpVAL.png

PR #3487, opened 2018-02-15
Allow responsive control class suffixes

alexej-d:
This PR would enable devs to set a cluss suffix for responsive controls like this:
```
'prefix_class' => 'align-',
'suffix_class' => '%s',
```

This way one can create responsive classes like ``.align-center-tablet`` which gives the framework more flexibility.

### DIFF
--- a/includes/base/controls-stack.php
+++ b/includes/base/controls-stack.php
@@ -738,7 +738,7 @@ abstract class Controls_Stack {
 	final public function get_class_controls() {
 		return array_filter(
 			$this->get_active_controls(), function( $control ) {
-				return ( isset( $control['prefix_class'] ) );
+				return ( isset( $control['prefix_class'] || isset( $control['suffix_class'] ) );
 			}
 		);
 	}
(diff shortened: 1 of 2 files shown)


kobizz:
In case you are using `add_responsive_control` method, you can achieve this by including a `%s` placeholder within the `prefix_class`.

For example:

```php
$this->add_responsive_control(
	'align',
	[
		'label' => __( 'Alignment', 'elementor' ),
		'options' => [
			'left'    => [
				'title' => __( 'Left', 'elementor' ),
				'icon' => 'fa fa-align-left',
			],
			'center' => [
				'title' => __( 'Center', 'elementor' ),
				'icon' => 'fa fa-align-center',
			],
			'right' => [
				'title' => __( 'Right', 'elementor' ),
				'icon' => 'fa fa-align-right',
			],
		],
		'prefix_class' => 'elementor%s-align-',
	]
);
```

In this case, every `%s` will be replaced with a `-` + the matching device.

Here is an example for such as possible selector output:

```css
.elementor-tablet-align-left
```

bainternet:
Thanks for the PR, Currently we don't feel comfortable with it and we haven't found the need for it.

thanks.

PR #3562, opened 2018-02-23
Feature/micro elements

MTempleS:
Elementor video widget doesn't support Dailymotion. Can somebody create this support (not only for Dalymotion, but also for others youtube alternatives host videos?

Thanks a lot.

Marcus


KingYes:
Create your issue there: https://github.com/pojome/elementor/issues

PR #3603, opened 2018-02-27
Dynamic Tags Support

matipojo:


includes/widgets/button.php
```diff
@@ -1,6 +1,8 @@
 <?php
 namespace Elementor;
 
+use Elementor\Modules\DynamicTags\Module;
```
kobizz: It's most recommended to use an alias for this `Module` class, since it's used outside of its context.

Maybe something like `DynamicTagsModule`.

KingYes: +1

KingYes (changes requested):

PR #3684, opened 2018-03-07
Fix issue with overriding media queries 

konradsajdak:
Throwing queries with "max-width" and "min-width" values lower in CSS than queries with only "min-width".

When Elementor generate CSS, puts media query with max and min value before query with only min value, so ex. "width" value couldn't be overridden. This generate problem with the columns sizes in responsive. In my case column size on tablet could override column size on desktop.

After this fix, when media queries are sorting and first element have max value but second doesn't, puts first element lower in CSS file.

### DIFF
--- a/includes/stylesheet.php
+++ b/includes/stylesheet.php
@@ -384,12 +384,12 @@ class Stylesheet {
 
 				$b_query = $this->hash_to_query( $b );
 
-				if ( isset( $a_query['max'] ) && !isset($b_query['max']) ) {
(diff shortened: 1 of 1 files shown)


kobizz:
Hey @KonradSajdak,

This issue has already been fixed by [this commit](https://github.com/pojome/elementor/commit/ebfeff83c8661c511ca6b587e98dd9647e1aff55#diff-f077073577d9c7f7c76a3cec37b3302e), but we really appreciate the time that you spent.

Thanks.

PR #3773, opened 2018-03-15
Feature/role manager JSCS and strings

bainternet:


includes/editor-templates/panel.php
```diff
@@ -56,12 +56,10 @@
 </script>
 
 <script type="text/template" id="tmpl-elementor-panel-footer-content">
-	<?php if ( Plugin::$instance->role_manager->user_can( 'design' ) ) : ?>
-	<div id="elementor-panel-footer-settings" class="elementor-panel-footer-tool elementor-leave-open tooltip-target" data-tooltip="<?php echo __( 'Settings', 'elementor' ); ?>">
+	<div id="elementor-panel-footer-settings" class="elementor-panel-footer-tool elementor-leave-open tooltip-target" data-tooltip="<?php esc_html_e( 'Settings', 'elementor' ); ?>">
```
ramiy: Wrong esc function. It should be `esc_arrt_e()` instead of `esc_html_e()`.

PR #3837, opened 2018-03-21
Elementor Feature Request

krann:
It would be great if the textbox widget had a checkbox which, when checked, took the 'Featured Image' and plonked it nicely into the text area so that the text wrapped around it. It would be like magic!
I like to 'justify-right' images. That way the text flows around it correctly and I like images to be about 50% of the width of the main text block. Those could be the Factory Settings for such a feature.
--


KingYes:
Open in new issue: https://github.com/pojome/elementor/issues

PR #3910, opened 2018-03-27
Fix issue with conditional repeater fields.

### LINKED ISSUE #3886: "Invalid supplied argument" warning for hidden repeater fields
Hi,

Please consider adding an additional check before this loop:

https://github.com/pojome/elementor/blob/59850ebed85c2e1c57cb6f914576677851e7c67d/includes/base/controls-stack.php#L1055

to make sure that $settings[ $control_name ] is not null. This is the case when the repeater field is displayed conditionally in the editor panel. Currently it fires off an "invalid argument supplied" warning in such cases.

Thanks!

albionselimaj:
Fixes #3886

### DIFF
--- a/includes/base/controls-stack.php
+++ b/includes/base/controls-stack.php
@@ -1717,7 +1717,7 @@ abstract class Controls_Stack {
 
 		foreach ( $controls as $control ) {
 			if ( 'repeater' === $control['type'] ) {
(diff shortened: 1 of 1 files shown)


KingYes:
Merge by #3916 
Thanks

PR #3955, opened 2018-03-29
Extracted button styles in another method

iamntz:
The reasoning is simple: while you can get the default sizing, you can't also get the default styling, which is pretty bad when you want to extend stuff a bit.

### DIFF
--- a/includes/widgets/button.php
+++ b/includes/widgets/button.php
@@ -94,6 +94,16 @@ class Widget_Button extends Widget_Base {
 		];
 	}
 
+	public static function get_button_types() {
+		return [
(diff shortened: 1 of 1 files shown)


iamntz:
One month later...

bainternet:
Thanks for the PR, Currently we don't see a real need for it since the current API allows you to update the control.

for example https://github.com/pojome/elementor/issues/3497#issuecomment-370736478

thanks again.

iamntz:
@bainternet thanks for the super quick reply. The reasoning is not _extending_ but _reusing_. There are some places where a button needs to have styles, but they don't, so....

PR #3999, opened 2018-04-04
Widgets advance tab

ramiy:
### Old View:

advanced-label-block

----

### New View:

advanced-label-inline

### DIFF
--- a/includes/elements/column.php
+++ b/includes/elements/column.php
@@ -614,7 +614,7 @@ class Element_Column extends Element_Base {
 		$this->start_controls_section(
 			'section_advanced',
 			[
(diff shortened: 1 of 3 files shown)


ramiy:
Also the section label is updated.

### Current Label:

advanced-sections-label-old

----

### New Label:

advanced-sections-label-new

iamntz:
Just an thought: depending on language, thelabels can be larger than that, leaving the input field smaller than it should.

Although the zindex field is OK to be smaller, the other ones (especially the ID and class) should be a bit bigger, in order to be readable.

ramiy:
@iamntz the Advanced tab is under design review, it will be enhanced in the coming versions of Elementor.

PR #4053, opened 2018-04-10
added CSS id to button widget

bainternet:


iamntz:
You already have this kind of control (which is present on all widgets):

Issue #4143, opened 2018-04-17
Widget_Base->parse_text_editor is not properly formatting texts


nextend:
Original code:
```PHP
protected function parse_text_editor( $content ) {
    /** This filter is documented in wp-includes/widgets/class-wp-widget-text.php */
    $content = apply_filters( 'widget_text', $content, $this->get_settings() );

    $content = shortcode_unautop( $content );
    $content = do_shortcode( $content );
    $content = wptexturize( $content );

    if ( $GLOBALS['wp_embed'] instanceof \WP_Embed ) {
        $content = $GLOBALS['wp_embed']->autoembed( $content );
    }

    return $content;
}
```

Probably in this method, you want to mimic WordPress `the_content` behavior, but this is not the right order. Also this method should do something similar that the core WordPress Text widget does, so you can use that as an example.

**Issue 1**

`wptexturize` should be sooner than `do_shortcode`

Explanation: `wptexturize` is parsing the HTML code with regular expression. It means that it can fail in special circumstances. Probably these special circumstances will come from shortcodes as they can insert inline scripts and such to the content. For example I have a ticket at WordPress where it fails: [https://core.trac.wordpress.org/ticket/43785#ticket](https://core.trac.wordpress.org/ticket/43785#ticket)

based on `wp-includes/default-filters.php`
```PHP
add_filter( 'widget_text_content', 'capital_P_dangit', 11 );
add_filter( 'widget_text_content', 'wptexturize'          );
add_filter( 'widget_text_content', 'convert_smilies',  20 );
add_filter( 'widget_text_content', 'wpautop'              );
add_filter( 'widget_text_content', 'shortcode_unautop'    );
add_filter( 'widget_text_content', 'do_shortcode',     11 );
```

**Issue 2**

Old themes used to add the `do_shortcode` to the `widget_text` filter. It also resulted that the shortcodes evaluated before the `wptexturize` and might result the same issue. This is why WordPress developers introduced a workaround to the text widget:
Check `/wp-includes/widgets/class-wp-widget-text.php` ->`widget` method

Before they run `widget_text` filter, they remove the `do_shortcode` from the filters and they place it back after the `widget_text` ran.

```PHP
/*
 * Suspend legacy plugin-supplied do_shortcode() for 'widget_text' filter for the visual Text widget to prevent
 * shortcodes being processed twice. Now do_shortcode() is added to the 'widget_text_content' filter in core itself
[...]

PR #4144, opened 2018-04-17
Clean up parse_text_editor filters

nextend:
Fix for issue #4143

### DIFF
--- a/includes/base/widget-base.php
+++ b/includes/base/widget-base.php
@@ -408,12 +408,22 @@ abstract class Widget_Base extends Element_Base {
 	 * @return string Parsed content.
 	 */
 	protected function parse_text_editor( $content ) {
+		$widget_text_do_shortcode_priority = has_filter( 'widget_text', 'do_shortcode' );
+		if ( false !== $widget_text_do_shortcode_priority ) {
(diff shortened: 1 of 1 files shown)


bainternet:
@nextend 

Thanks for the PR, Currently we don't feel comfortable having filters there, nor do we see a real need for it.

And also, in case that is is really needed it can be done via the current filters.

thanks again.

nextend:
@bainternet It is still an issue. Elementor does not use `do_shortcode` on the content in the Text editor widget. You should follow WordPress!

Switch the `wptexturize` and `do_shortcode` call and this issue will be solved:
```
	protected function parse_text_editor( $content ) {
		/** This filter is documented in wp-includes/widgets/class-wp-widget-text.php */
		$content = apply_filters( 'widget_text', $content, $this->get_settings() );

		$content = shortcode_unautop( $content );
		$content = wptexturize( $content );
		$content = do_shortcode( $content );

		if ( $GLOBALS['wp_embed'] instanceof \WP_Embed ) {
			$content = $GLOBALS['wp_embed']->autoembed( $content );
		}

		return $content;
	}
```

PR #4311, opened 2018-05-01
Release/2.1.0 Post Blog Widget - post nav widget

NAROLLES:
Hi
I'm sorry but I'm a beginner here
thank you for telling me if there is better place here to post my suggestion
I come on board of elementor support

Would it be possible to have an additional option to that of "skipping" one or more articles (which may be convenient in some cases). The addition of the option "do not display an article already displayed on the same page" (this exists at Divi and very practical,)

post nav widget - If possible, provide the following options on this widget: "on all posts", "by category" or by "tags", oceanwp allows it in the single item setting but not Elementor for this same feature

thanks


KingYes:
Please open new issue about this, no Pull request.

PR #4584, opened 2018-05-29
Better integration of custom fields

danielepol:
Hi, I'd like to see a better integration of dynamic content and custom fields: this is necessary if you really want Elementor can be used to design a website theme. Just a couple of examples: 

1. **Featured Image/Custom field as a background image.** When I create a post template, I can't use the featured image as a background for an Elementor section. This is too bad. 
2. **Get colour from a custom field.** VERY useful when you want to customize elements colors based on category. 
3. **A better Custom fields integration on blocks.** i.e. I can't use a custom field to define the countdown date or the link for a social icon. 
4. **Management of repeatable custom fields** inside blocks as carousel, accordion, etc... I'd like to create a repeatable custom field and put each value as a list block, or an accordion block (everything with repeatable elements). 
5. **Conditional view**. Show or hide elements based on logical conditions. I.e. I'd like to display a social icon only if the link field is not empty (obviously is useful only if I can use custom fields inside the block).

Thank you for considering my suggestions and keep up with the good job!


bainternet:
@danielepol 

Thanks,
Please follow guidelines for opening a new feature request, so we can help you better.

* Prerequisites.

and also each feature should be on its own request.

anyway, closing as this shouldn't be a pull request but feel free to open correctly again.

thanks again.

PR #4590, opened 2018-05-29
Fix Background with Dynamic Tag/CSS

### LINKED ISSUE #4577: Background featured image not positioned properly
Background dynamic featured image doesn't seem to position itself according to settings (Position, Attachment, Repeat, Size). Instead, it positions itself to default settings.
Thank you.

matipojo:
Fix #4577

### DIFF
--- a/includes/base/controls-stack.php
+++ b/includes/base/controls-stack.php
@@ -57,6 +57,18 @@ abstract class Controls_Stack {
 	private $_settings;
 
 	/**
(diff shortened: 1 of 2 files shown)


tiagoefreitas:
This bug has returned for me, may be related to the new global fonts.

It was working then I just changed a global font used in that section and it messed up the section background settings, it was on cover no-repeat and now on live it shows as defaults (repeat). It shows ok on the editor, so no idea what to do to fix it.

PR #4858, opened 2018-06-25
Posts widget layout

kkri:
Hello. 
Thank you for all Elementor Pro big features but I will be very happy if your posts widget will be empowered with new different layouts to allow us to create a gorgeous page like a magazine. Of course it will be used for other sites like events, or to show custom posts types or products and many other uses. 
As showing posts is common and very important to almost all kind of sites, hope it will be available in next versions, as a big part of it is already there (queries for listing posts). See this plugin to explain better what I mean https://www.templatemonster.com/it/plugin-per-wordpress-tipo-69506.html 
Thank you so much


bainternet:
Thanks for the feedback,
Please follow guidelines for opening a new feature request, so we can help you better.

* Prerequisites.
* Description and examples.

kkri:
Sorry if I made mistakes, but it's my first or second time here and it's not easy to understand how to write correctly, Could you point me exactly to what I miss? I'll be happy to do the right thing. Thank you.

Issue #4498, opened 2018-05-21
Import Elementor Page Template resets Wordpress Page Template


pingram3541:
- [ x ] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [ x ] The issue still exists against the latest stable version of Elementor.

## Description

When importing a page template from the library into a page, and choosing the option to also import the page settings (custom css, classes, id's, advanced styles...) the WordPress page template is reset to default.

## Steps to reproduce

1) Create a new page and set the template as Elementor Canvas
2) Edit page with Elementor (notice it loads w/ Canvas as expected)
3) Import a saved page template from the Elementor library, preferably one also that uses Canvas
4) Choose Yes to the option of also importing and overwriting the current page settings (needed to get any customizations of the page's settings stored in the template, mainly page-wide custom css from Elementor Pro)
5) Watch as template is imported but after a moment the entire preview reloads w/ what was imported but back to the theme's default template. One must manually go back and set the template to Canvas.

## Isolating the problem

- [ x ] This bug happens with only Elementor plugin active (and Elementor Pro).
- [ x ] This bug happens with a default WordPress theme active.
- [ x ] I can reproduce this bug consistently using the steps above.

## Environment

-mixed/all

mariovalney:
I'm falling on a problem like yours when I'm using `Elementor:instance()->document->save()` to create pages in a updater script.

For my case I debugged and get the problem:
https://github.com/pojome/elementor/blob/master/core/settings/page/manager.php#L181

Template is being replaced by `default` if no `$settings['template']` value is there.

Maybe it's causing your problem too.

PR #5054, opened 2018-07-13
Fix: Elementor reset Page Template

mariovalney:
Elementor saving process is reseting page templates ( `_wp_page_template` ) when we uses `Elementor:instance()->document->save()` to create pages (like in a updater script).

I found template is being replaced by default if no `$settings['template']` value is there:
https://github.com/pojome/elementor/blob/master/core/settings/page/manager.php#L181

Some points:

1. Maybe it's related to #4498 
2. A review is very needed as I was not able to check all the impact of this PR.


KingYes:
Merge by: #5067

Thanks!

PR #5250, opened 2018-07-30
search and replace PLEASE....

michelleist:
As an addition to copy pasting styles, search and replace anything would be awesome and helping a lot!


KingYes:
Please open new issue there: https://github.com/pojome/elementor/issues

Issue #5421, opened 2018-08-15
mobile column width percentage doesn't work with reversed columns


moseskerub:
## Prerequisites

- [X] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [X] The issue still exists against the latest stable version of Elementor.

## Steps to reproduce

1. add a new section with 3 columns
2. switch to mobile editing
3. set the first column width to 100%, and the other two 50%
4.watch the 50% columns next to each other
5. turn on reverse mobile columns in section settings
6. watch the two 50% columns one above the other instead of next to each other.

## Isolating the problem

- [X] This bug happens with only Elementor plugin active (and Elementor Pro).
- [X] This bug happens with a default WordPress theme active.
- [X] I can reproduce this bug consistently using the steps above.

moseskerub:
Sadly I'm not a developer of any sort, just an happy end user. But this bug didn't exist in earlier versions If I'm not mistaken

I can't add any details, but maybe the solution is to revamp the reverse columns mechanism?
maybe use an ordering system (flex allows that I think) so you could select the precedence for each column - and this could solve tablet reverse as well - which doesn't exist today

Hope it can help

arielk:
@moseskerub Thanks for your idea, we just fix this issue for next version and we added an option to reverse columns for tablet also: https://github.com/pojome/elementor/issues/2070

Issue #3393, opened 2018-02-05
Can you add the feature to have Reverse Tablet the same as you do mobile?


ljgarnes:

bainternet:
I merged this issue with  #2070, which is on the same subject so we could deal with it more efficiently. 
Please feel free to continue the conversation there.

Thanks for your feedback!

Issue #2070, opened 2017-07-24
[Feature Request]  Reverse column on tablet


dam62500:
Hello,

I think it's a good features to allow reverse columns for tablets like mobile feature.

mmjaeger:
+1

pingram3541:
Is it shameful for me to put a bump on this one?  It seems like a simple thing to implement no?

vongoh:
+1

dam62500:
Any news about that? I need to invert the columns even on tablet and I'm not the only about that. I think that'll be a good features :)

breizhwave:
just perfect solution   @dam62500

black-eye:
+1

denishacquin:
+1

PR #5472, opened 2018-08-19
Fix reverse-column bug (#5421) and add tablet breakpoint (#2070 , #3393)

joshmarom:


ramiy:
@joshmarom you should have used a `sprintf()` in this case.

PR #5974, opened 2018-10-23
Tweak: added css filename filter hook

### LINKED ISSUE #5968: Add filter to rename external css files
HI!
it would be really useful to have a filter to rename the external css files. This is because I often work with multisite with a single uploads folder.
Is this feature in your future projects for elementor?
Thank's

bainternet:
`elementor/files/file_name` closes #5968

### DIFF
--- a/core/files/base.php
+++ b/core/files/base.php
@@ -55,6 +55,16 @@ abstract class Base {
 	}
 
 	public function __construct( $file_name ) {
+		/**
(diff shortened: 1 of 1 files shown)


jawabuu:
Hey @bainternet , this is much appreciated. My question is if and how I can utilize this so that the css filename uses the postname rather than the id e.g. post-home.css vs post-33.css

PR #6005, opened 2018-10-28
Image caption source (needs update script) closes #5770

### LINKED ISSUE #5770: Single Post Featured Image Widget Caption should not be included
- [ x] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [ x] The issue still exists against the latest stable version of Elementor.

## Description

When using the Single Post Featured Image Widget to create a single post template, the Caption field doesn't make sense there. It expects text to be input such as "This is my caption". But the whole point of this widget is to be dynamic, and entering anything into the caption field results in the same caption being displayed under every post's Featured Image. I think the caption field should be removed from the widget.

bainternet:


includes/widgets/image.php
```diff
@@ -604,6 +658,33 @@ protected function _content_template() {
 				return;
 			}
 
+			var hasCaption = function() {
```
matipojo: Why each render needs to re-define the functions?

matipojo (changes requested):

includes/widgets/image.php
```diff
@@ -604,6 +658,33 @@ protected function _content_template() {
...
+				if( ! settings.caption_source || 'off' === settings.caption_source ) {
+					return false;
+				}
+				return true;
+			}
+
+			var ensureAttachmentData = function( id ) {
+				if ( 'undefined' === typeof wp.media.attachment( id ).get( 'caption' ) ) {
```
KingYes: In remove images?

KingYes (changes requested):

PR #6045, opened 2018-11-01
Plugin: New - Abstract element `Container` with a behavior like inner…

matipojo:


stevecove:
I know this isnt a discussion forum, but do you have any more info on this, as it looks to be extremely relevant for an upcoming project, and we are currently deciding on the layout system to use.

matipojo:
Hi @stevecove, Currently it's in the infrastructure level only. so it's not recommended to take it in account for your projects.

PR #6060, opened 2018-11-04
2.3.0 fixes

matipojo:


arielk:
@matipojo @kobizz What's the problem with the current query?

PR #6127, opened 2018-11-12
Aweber Integration to Elementor

rainx10:
I have no idea if I'm in the right place...
but is there a way we can integrate Aweber email auto responder service to Elementor? 

Thank be great.


KingYes:
Please post there: https://github.com/pojome/elementor/issues

PR #6183, opened 2018-11-15
Convert repeater data array to Backbone.Collection

evgeniyshay:
Repeater field in page/section settings tab causes an error on page/section revision change, because type of repeater data is Array instead of Collection. So this fix checks for data type in repeater fields and converts it from Array to Backbone Collection if needed.


bainternet:
@evgeniyshay 

We are aware of this issue, but the fix you are suggesting is not addressing the root cause of the issue. So, for now, we are not going to merge this, and we'll try to correct this in the right place.

thanks.

PR #6274, opened 2018-11-26
1. Read More: Fix + Widget. 2. Compatibility: Exclude library post type from Yoast SEO

matipojo:


includes/frontend.php
```diff
@@ -1116,6 +1116,10 @@ private function process_more_tag( $content ) {
 			return $parts['main'] . '<div id="more-' . $post->ID . '"></div>' . $parts['extended'];
 		}
 
+		if ( empty( $parts['more_text'] ) ) {
+			$parts['more_text'] = __( '(more&hellip;)' );
```
KingYes: elementor

KingYes (changes requested):

PR #6283, opened 2018-11-27
pass a reference to init hook

i30:
## Description
By passing a reference to the init hook, other developers can use that reference directly without having to instantiate the class again while adding actions to the hook. For example:

Instead of this:
```php
add_action( 'elementor/init', function() {
   \Elementor\Plugin::$instance->elements_manager->add_category( 
   	'theme-elements',
   	[
            'title' => __( 'Theme Elements', 'theme-domain' ),
   	    'icon' => 'fa fa-plug', //default icon
   	]
   );
} );
```

We can do this:
```php
add_action( 'elementor/init', function($elementor) {
  $elementor->elements_manager->add_category( 
   	'theme-elements',
   	[
   	    'title' => __( 'Theme Elements', 'theme-domain' ),
   	    'icon' => 'fa fa-plug', //default icon
   	]
   );
} );
```
It's more handy.

## Quality assurance

### DIFF
--- a/includes/plugin.php
+++ b/includes/plugin.php
@@ -480,7 +480,7 @@ class Plugin {
 		 *
 		 * @since 1.0.0
 		 */
-		do_action( 'elementor/init' );
+		do_action( 'elementor/init', $this );
 	}
 
 	public function on_rest_api_init() {


bainternet:
@i30 

`\Elementor\Plugin` is a singleton template class so you don't actually need (or even can, the constructor is private) to instantiate the class again.

You can see in your own snippet that you are accessing a static instance:
```PHP
\Elementor\Plugin::$instance
```

and as for not passing it around, `elementor/init` is a very general hook used for many cases, and there is no need to pass a reference to a static publicly accessible instance if it's only needed in some use cases.


but anyway, thanks for the PR.

Issue #1903, opened 2017-06-28
Feature request: Vimeo as a video background 


upekshapriya:
Be good to have Vimeo as an option for a video background as well as YouTube and MP4 file. See https://codepen.io/mattgrosswork/pen/jrdwK

HammerOz:
I thought I would try and use the vimeo source mp4 file link that you have access to as a pro user. The video is actually visible while editing, yet fails when viewing the live page. Any ideas? 

example file link:

 https://player.vimeo.com/external/226866110.hd.mp4?s=4c4c1ed74fbd38d861c83bff708ed17167957252&profile_id=174

i2g-lweiss:
It seems the Vimeo link generates a request for a token for access when selecting "use with your own player" (Settings/Video file in Vimeo).  If you paste that link into your browser it changes to one with a token.  You can use this generated link in elementor and it works everywhere but the token expires the next day.

bainternet:
Currently, Vimeo allows removing controls and background embeds just for a Plus, PRO, or Business member.

i2g-lweiss:
We are a pro member, have not found this option.  We are using the direct links to "use with your own player".  It's not about removing the controls, we just want to use a vimeo video link as a background.

HammerOz:
It's all a bit so so. I'm self hosting now

zimventures:
Throwing my +1 in as a feature request.

dasignlady:
I would also like to see this feature added. Gotta upload same video to youtube just to use it as background, when I can use vimeo in video widget?

mskks64:
"YouTube link or video file (mp4 is recommended)."

Mp4 from vimeo does not work.

dasignlady:
So, I just purchased a subscription to Publit.io and uploaded the section background video from there instead. It seems to be faster than the other services so I am happy with that. NOW, I need the option of using _that same source_ for my standalone videos but I can only choose from Vimeo, YT, Daily Motion or self-hosted. (The latter is not an option!) I do not understand why there is such a disparity between what sources we are able to use for one or the other? If adding a custom link works for background video, please add same option for the video widget. OR, make it so background video works on mobile as well, eliminating the need for a mobile-only separate video. Thanks in Advance!

danishiqbal4:
+1 for this feature

Deralict:
+1 for me too. Vimeo is such better quality so is more suitable for a background than YouTube.

Matherone:
+1.

[...]

PR #6326, opened 2018-12-03
play button removal for background videos on vimeo

thephilm:
Added an extra parameter for video module specifically for Vimeo for removing the play button that shows up in the middle (desktop) lower left (mobile) on autoplayed background videos. This is useful when using Vimeo as a background video in headers for example. ref: https://help.vimeo.com/hc/en-us/articles/115011183028-Embedding-background-videos

## PR Checklist

- [ ] The commit message follows our guidelines:  https://github.com/pojome/elementor/blob/master/.github/CONTRIBUTING.md

## PR Type

- [ ] Bugfix
- [x ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

Added an extra parameter for video module specifically for Vimeo for removing the play button that shows up in the middle (desktop) lower left (mobile) on autoplayed background videos.

*

## Description

*

## Test instructions
Embed a vimeo video with a  Plus, PRO, Business, or Premium vimeo account.
The play button that normally appears in the middle of the video will not be visible when selection is set to true.

*

## Quality assurance

- [x ] I have tested this code to the best of my abilities

Fixes #

### DIFF
--- a/includes/widgets/video.php
+++ b/includes/widgets/video.php
@@ -399,6 +399,20 @@ class Widget_Video extends Widget_Base {
 		);
 
 		$this->add_control(
+			'vimeo_background',
+			[
+				'label' => __( 'Show Play Button', 'elementor' ),
+				'type' => Controls_Manager::SWITCHER,
+				'label_off' => __( 'Yes', 'elementor' ),
+				'label_on' => __( 'No', 'elementor' ),
+				'default' => 'no',
+				'condition' => [
+					'video_type' => 'vimeo',
+				],
+			]
+		);
+		
+		$this->add_control(
 			'download_button',
 			[
 				'label' => __( 'Download Button', 'elementor' ),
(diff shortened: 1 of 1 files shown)


includes/widgets/video.php
```diff
@@ -896,6 +910,7 @@ public function get_embed_params() {
 				'vimeo_title' => 'title',
 				'vimeo_portrait' => 'portrait',
 				'vimeo_byline' => 'byline',
+				'vimeo_background' => 'background',
```
kobizz: Have you tested this? 
Why are you adding a control but not using it's value?
Additionally, keep in mind that this feature is available only if you are a Plus, PRO, Business, or Premium Vimeo member.

kobizz (changes requested):

includes/widgets/video.php
thephilm: You are correct, I also realize I have it so the default is to not display the play button which would not be ideal for most people. I guess this was the extent of my expertise, but I think it could be valuable as Vimeo is a very useful service for doing background videos especially with the amount of branding YouTube has recently added to their videos no matter what parameters you add.

bainternet:
I merged this issue with  #1903
, which is on the same subject so we could deal with it more efficiently. 
Please feel free to continue the conversation there.

Thanks for your feedback!

PR #6366, opened 2018-12-06
Fix the wp_the_query global when doing "Maintenance" redirection

adrianosferreira:
## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Description
Elementor is using `query_posts` function in order to "fix" the `$wp_query` with the Coming Soon page and it is working pretty well. However, it must also fix the global `$wp_the_query`, because if you later call the function `wp_reset_query` anywhere, it will reset the main query, basically replacing `$wp_query` with `$wp_the_query`. Then, instead of showing the coming soon page, it will display the original page that you have when you don't have the maintenance mode active.

This fix simply adjusts the `$wp_the_query` with the same value of `$wp_query`.

## Test instructions

- Enable Coming Soon feature
- Open the single.php file of your current theme and add this:

```php
get_header();

wp_reset_query();
```

- Access the site as a visitor and notice that the Coming Soon page is not displayed

## Quality assurance

- [ x ] I have tested this code to the best of my abilities

## Issue

https://github.com/pojome/elementor/issues/6365

### DIFF
--- a/includes/maintenance-mode.php
+++ b/includes/maintenance-mode.php
@@ -146,6 +146,8 @@ class Maintenance_Mode {
 			'p' => self::get( 'template_id' ),
 			'post_type' => Source_Local::CPT,
 		] );
+
+		$GLOBALS['wp_the_query'] = $GLOBALS['wp_query'];
 	}
 
 	/**


kagg-design (approved):
+1

bainternet:
@adrianosferreira 

Thanks for the PR, But the change you are suggesting is almost never recommended ( to overwrite  `$GLOBALS['wp_the_query']`). Also the case you are describing ( having `wp_reset_query` without an actuall query) is not something that should happen.

Now if you do however have a case like that and you do find yourself in the need to overwrite `$GLOBALS['wp_the_query']` then you can use the native `template_redirect` action hook at a priority lower than 11 ( like 12+ ) to do it, ex:

```PHP
add_action( 'template_redirect', function() {
	if ( \Elementor\Maintenance_Mode::get( 'template_id' ) === $GLOBALS['post']->ID ) {
		$GLOBALS['wp_the_query'] = $GLOBALS['wp_query'];
	}
}, 12 );
```

Thanks again.

adrianosferreira:
@bainternet 

>But the change you are suggesting is almost never recommended ( to overwrite $GLOBALS['wp_the_query'])

FYI the `query_posts` shouldn't be also used by plugins, per WP documentation:

>This function will completely override the main query and isn’t intended for use by plugins or themes. Its overly-simplistic approach to modifying the main query can be problematic and should be avoided wherever possible. In most cases, there are better, more performant options for modifying the main query such as via the ‘pre_get_posts’ action within WP_Query.

Using `query_posts` you are basically doing what you suggested to not to do, which is to overwrite the global directly:

```
function query_posts($query) {
    $GLOBALS['wp_query'] = new WP_Query();
    return $GLOBALS['wp_query']->query($query);
}
```

Anyway, my example in test instructions was quite simple, indeed, I can't see anyone using it in that way. However, what I wanted to explain with it is that if somebody will manipulates the global post data (exactly like Elementor is doing) in order to do something and then end up restoring it using `wp_reset_query` it will automatically break this Elementor's functionality.

PR #6367, opened 2018-12-06
Move message notice to flex

arielk:


ramiy:
25 more:
https://github.com/pojome/elementor/search?q=display%3A+table%3B&unscoped_q=display%3A+table%3B

PR #6432, opened 2018-12-14
Feature Request - Divs

mauritzbotha92:
Hi all,

I would like to request this feature.

Currently the only way to get more divs into your design would be to create a new section, select the 2 column section, and then delete one column. This allows you to have an additional div in your design which can be styled.

It would be better to have the "innersection" element be replaced with a div element, that allows you to keep nesting divs inside of divs (as complex as you want to make it). This allows you have a lot more styling functionality as I often find I am restricted with not enough divs on the webpage that I can work with. 

Many thanks!


mauritzbotha92:
Or maybe not be replaced, but add an additional element in elementor that allows for adding a lot more divs inside a section.

PR #6433, opened 2018-12-14
Added filter for content in Element_Base

thenovacreator:
## PR Type

- [x] Feature

## Summary

* Added render content filter for childs of Base_Element using print_element (e.x. Section / Column)

## Description

* Added render content filter for childs of Base_Element using print_element

## Test instructions

* Use the filter "elementor/{$element_type}/render_content" for e.x. a section to manipulate its content

## Quality assurance

- [ ] Docs have been added / updated (for bug fixes / features) (Could not)
-> Documentation should be extended according to this https://code.elementor.com/php-hooks/#elementorwidgetrender_content.

### DIFF
--- a/includes/base/element-base.php
+++ b/includes/base/element-base.php
@@ -660,6 +660,20 @@ abstract class Element_Base extends Controls_Stack {
 		ob_start();
 		$this->_print_content();
 		$content = ob_get_clean();
+		
+		/**
+		 * Render element content.
+		 *
+		 * Fires when Elementor renders the element content.
+		 *
+		 * The dynamic portion of the hook name, `$element_type`, refers to the element type.
+		 *
+		 * @since 1.0.0
+		 *
+		 * @param string $content The content of the element.
+		 * @param Element_Base $this The element.
+		 */
+		$content = apply_filters( "elementor/{$element_type}/render_content", $content, $this );
 
 		$should_render = ( ! empty( $content ) || $this->should_print_empty() );
 
@@ -679,7 +693,7 @@ abstract class Element_Base extends Controls_Stack {
 			$this->_add_render_attributes();
 
 			$this->before_render();
-			echo ( apply_filters( "elementor/{$element_type}/render_content", $content, $this ) );
+			echo $content;
 			$this->after_render();
 
 			$this->enqueue_scripts();
--- a/includes/base/element-base.php
+++ b/includes/base/element-base.php
@@ -679,7 +679,7 @@ abstract class Element_Base extends Controls_Stack {
 			$this->_add_render_attributes();
 
 			$this->before_render();
-			echo $content;
+			echo ( apply_filters( "elementor/{$element_type}/render_content", $content, $this ) );
 			$this->after_render();
 
 			$this->enqueue_scripts();
--- a/includes/base/element-base.php
+++ b/includes/base/element-base.php
@@ -660,7 +660,7 @@ abstract class Element_Base extends Controls_Stack {
 		ob_start();
 		$this->_print_content();
 		$content = ob_get_clean();
-		
+
 		/**
 		 * Render element content.
 		 *


thenovacreator:
We need this to implement a cache buster for our "splittest for elementor" plugin.

ramiy:
The correct syntax should be:

```
/**
 * Render element content.
 *
 * Fires when Elementor renders the element content.
 *
 * The dynamic portion of the hook name, `$element_type`, refers to the element type.
 *
 * @since 1.0.0
 *
 * @param string $content The content of the element.
 * @param Element_Base $this The element.
 */
$content = apply_filters( "elementor/{$element_type}/render_content", $content, $this );
echo $content;
```

You should apply filter on the variable you are filtering.

Also, it would be better to move this filter to line 663, where you define the `$content`.

thenovacreator:
Do recommend it like this?

thenovacreator:
What about this to be merged?

thenovacreator:
Is there something new going on here?

TerenceMilbourn:
+1 -- without this is not possible to exclude test content from the site's caching

hosting-io:
+1 - without this, the plugin is almost useless for most sites that are interested in performance as well as marketing. Almost all users that are looking at A/B testing will have some sort of caching in place i would guess

hsk-it:
Would love this, too!
+1 here!

WebZenStudio:
+1 Have been waiting forever for there to be a split testing option for Elementor. I have been using Thrive Themes specifically because of their A / B split testing functionality. I was thrilled to see a plugin for split testing with Elementor... and then disappointed that it won't work with WP Rocket or other caching plugins enabled. Disabling caching is not an option. Looking forward to being able to split test with Elementor.

backpackerstoday:
+1 Please Elementor Team, make our lifes even easier..

thenovacreator:
I can not understand how things went here. Except of the syntax correction there were almost no comments of the elementor team hear. In contrast to the other users posting here that they would love to have this. And then without any comment or hint for a solution (how ever this solution could look like) and any word about why this request is closed it is closed. 

That is really sad because without this we have really big issues to develop our plugin further. We may stop developing it since **really** usefull features would be depend on this and without that ... well what should we develop. And do not think that we did not care about this. But instead of bugging we think that the users voices in here would be enougth.

TerenceMilbourn:
It could be that just nobody noticed or, more likely, they want to save the solution for themselves, and rather than help you, end up as your competitor.  I don't know this is true, its just an idea, so just saying what I think might be what's happening.

RichardUUU:
I'm also interested in seeing this implemented.

hsk-it:
We really need this to be implemented, too. 
Closing this topic is not the right way to work on this feature. 
Really disappointed at this moment. 
Please rethink this point.

bainternet:
[**@hsk-it**](https://github.com/hsk-it) commented on [Jul 19, 2019, 10:30 PM GMT+3](https://github.com/elementor/elementor/pull/6433/#issuecomment-513349083 "2019-07-19T19:30:00Z - Replied by Github Reply Comments"):
> We really need this to be implemented, too. 
> Closing this topic is not the right way to work on this feature. 
> Really disappointed at this moment. 
> Please rethink this point.


[**@RichardUUU**](https://github.com/RichardUUU) commented on [Jul 19, 2019, 10:21 PM GMT+3](https://github.com/elementor/elementor/pull/6433/#issuecomment-513346549 "2019-07-19T19:21:14Z - Replied by Github Reply Comments"):
> I'm also interested in seeing this implemented.


[**@TerenceMilbourn**](https://github.com/TerenceMilbourn) commented on [Jul 19, 2019, 4:17 PM GMT+3](https://github.com/elementor/elementor/pull/6433/#issuecomment-513224943 "2019-07-19T13:17:51Z - Replied by Github Reply Comments"):
> It could be that just nobody noticed or, more likely, they want to save the solution for themselves, and rather than help you, end up as your competitor.  I don't know this is true, its just an idea, so just saying what I think might be what's happening.


[**@thenovacreator**](https://github.com/thenovacreator) commented on [Jul 19, 2019, 12:56 AM GMT+3](https://github.com/elementor/elementor/pull/6433/#issuecomment-513004037 "2019-07-18T21:56:00Z - Replied by Github Reply Comments"):
> I can not understand how things went here. Except of the syntax correction there were almost no comments of the elementor team here. In contrast to the other users posting here that they would love to have this. And then without any comment or hint for a solution (how ever this solution could look like) and any word about why this request is closed it is closed. 
> 
> That is really sad because without this we have really big issues to develop our plugin further. We may stop developing it since **really** useful features would be depend on this and without that ... well what should we develop. And do not think that we did not care about this. But instead of bugging we think that the users voices in here would be enough.


[**@backpackerstoday**](https://github.com/backpackerstoday) commented on [Jul 6, 2019, 2:12 PM GMT+3](https://github.com/elementor/elementor/pull/6433/#issuecomment-508917666 "2019-07-06T11:12:53Z - Replied by Github Reply Comments"):
> +1 Please Elementor Team, make our lifes even easier..


[**@WebZenStudio**](https://github.com/WebZenStudio) commented on [Jun 28, 2019, 8:59 PM GMT+3](https://github.com/elementor/elementor/pull/6433/#issuecomment-506825162 "2019-06-28T17:59:33Z - Replied by Github Reply Comments"):
> +1 Have been waiting forever for there to be a split testing option for Elementor. I have been using Thrive Themes specifically because of their A / B split testing functionality. I was thrilled to see a plugin for split testing with Elementor... and then disappointed that it won't work with WP Rocket or other caching plugins enabled. Disabling caching is not an option. Looking forward to being able to split test with Elementor.

Hi everyone,

This was closed via External Software, hence the lack of explanation, but not by accident. We closed this Mainly for the performance and resources hit it would cause passing each and every *element* ( section, column, widget) output via a filter.

Also, there are already hooks in place that allow manipulating the output:

https://github.com/elementor/elementor/blob/8bfc36ffc3c2031c5953f31096c3a0f2c68c7ad0/includes/base/element-base.php#L519-L529

https://github.com/elementor/elementor/blob/8bfc36ffc3c2031c5953f31096c3a0f2c68c7ad0/includes/base/element-base.php#L530-L542

https://github.com/elementor/elementor/blob/8bfc36ffc3c2031c5953f31096c3a0f2c68c7ad0/includes/base/element-base.php#L572-L584

https://github.com/elementor/elementor/blob/8bfc36ffc3c2031c5953f31096c3a0f2c68c7ad0/includes/base/element-base.php#L585-L594

thenovacreator:
@bainternet 

Thanks for the comment. The inital Problem is that all the hooks you are presenting here are not working for this case. I need to manipulate the html content directly.

So now i have to use the "elementor/frontend/section/should_render" event and this code

`global $globalRenderingSection;

if ($element->get_name() != "section") {
	return true;
}

if ($globalRenderingSection) {
	return true;
}

$globalRenderingSection = true;
ob_start();
$element->print_element();
$content = ob_get_contents();
ob_end_clean();

echo $this->cacheBuster->renderContent($content, $element);

$globalRenderingSection = false;

return false;`

To get things working

PR #6459, opened 2018-12-16
Feature/background upgrades

matipojo:


core/base/background-task.php
```diff
@@ -174,6 +174,28 @@ protected function handle() {
 		}
 	}
 
+	public function is_process_locked() {
```
KingYes: doc

modules/wp-cli/update.php
```diff
@@ -0,0 +1,108 @@
...
+			/** @var \WP_Site[] $sites */
+			$sites = get_sites();
+
+			foreach ( $sites as $keys => $blog ) {
+				// Cast $blog as an array instead of  object
+				$blog_id = $blog->blog_id;
+
+				switch_to_blog( $blog_id );
+
+				\WP_CLI::line( 'Site #' . $blog_id );
```
KingYes: blogname

KingYes (changes requested):

PR #6588, opened 2019-01-01
js/admin and gutenberg: new - add 'back to wp editor' confirm popup

nuritsha:


kobizz (approved):

portrayaloflife:
Oh snap is this legit happening? Have the same issue with clients, this feature would be amazing

PR #6609, opened 2019-01-03
Tweak: run merge only once

bainternet:


includes/controls/animation.php
```diff
@@ -130,9 +130,10 @@ public static function get_animations() {
 			 * @param array $additional_animations Additional Animations array.
 			 */
 			self::$_additional_animations = apply_filters( 'elementor/controls/animations/additional_animations', $additional_animations );
```
KingYes: Use local variable

includes/controls/hover-animation.php
```diff
@@ -105,9 +105,10 @@ public static function get_animations() {
 			 * @param array $additional_animations Additional Animations array.
 			 */
 			self::$_additional_animations = apply_filters( 'elementor/controls/hover_animations/additional_animations', $additional_animations );
```
KingYes: Use local variable

KingYes (changes requested):

Issue #6652, opened 2019-01-08
get_render_attribute_string ret empty params


AntonRzevskiy:
## Prerequisites

- [x] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [x] The issue still exists against the latest stable version of Elementor.

### Does not comply with the documentation!
https://github.com/pojome/elementor/blob/0da1fc6b8ee8f2cc968f9bac0fdbb40d0f7ce104/includes/base/element-base.php#L600-L601

It returns like `<div data-test>`
If `add_render_attribute( 'data-test', '' );`

Coz:
https://github.com/pojome/elementor/blob/0da1fc6b8ee8f2cc968f9bac0fdbb40d0f7ce104/includes/utils.php#L619
Don't care about)

## Isolating the problem

- [ ] This bug happens with only Elementor plugin active (and Elementor Pro).
- [x] This bug happens with a default WordPress theme active.
- [x] I can reproduce this bug consistently using the steps above.

bainternet:
@AntonRzevskiy 

`get_render_attribute_string` looks at the entire string, not if a specific attribute is empty. So simply by calling `add_render_attribute()` the string will never be empty. Which seems to comply with the documentation.

thanks for the feedback.

PR #6680, opened 2019-01-10
update(utils.php): filter atts. #6652

AntonRzevskiy:
Add is_printable method for render_html_attributes method. Now such functions as get_render_attribute_string will not return empty parameters.

## PR Checklist

- [x] The commit message follows our guidelines:  https://github.com/pojome/elementor/blob/master/.github/CONTRIBUTING.md

## PR Type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

* Improved functions `get_render_attribute_string` and `print_render_attribute_string`. Now they do not return empty HTML attributes, your code will become cleaner.

## Description

* Added new universal method `is_printable`. Now we can check the value before it is displayed.
* Fixed method `render_html_attributes`. Now we can not worry about the attributes that are passed to this function. In the string that will be returned will contain only the filled parameters.

## Test instructions

* Before PR:
```php
// in Widget::render
$this->add_render_attribute( '_test', 'data-test', [

	'0', // print " 0"
	0, // print " 0"
	null, // print " "
	false, // print " "
	true, // print " 1"
	array(), // print " Array"
	[], // print " Array" too
	'', // print " "
	new stdClass(), // Fatal error: Object of class stdClass could not be converted to string

] );
$this->print_render_attribute_string( '_test' );
```

* After PR:
```php
// in Widget::render
$this->add_render_attribute( '_test', 'data-test', [

	'0', // print " 0"
	0, // print " 0"
	null, // no print
	false, // no print
	true, // no print
	array(), // no print
	[], // no print
	'', // no print
	new stdClass(), // no print, no error

] );
$this->print_render_attribute_string( '_test' );
```

## Quality assurance

Fixes #

### DIFF
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -608,25 +608,58 @@ class Utils {
 	}
 
 	/**
-	 * Render html attributes
+	 * Render html attributes.
 	 *
 	 * @access public
-	 * @static
-	 * @param array $attributes
 	 *
-	 * @return string
+	 * @static
+	 *
+	 * @param    array    $attributes
+	 *
+	 * @return   string
 	 */
 	public static function render_html_attributes( array $attributes ) {
 		$rendered_attributes = [];
 
 		foreach ( $attributes as $attribute_key => $attribute_values ) {
+
 			if ( is_array( $attribute_values ) ) {
+
+				$attribute_values = array_filter( $attribute_values, array( 'Utils', 'is_printable' ) );
+
+				if ( ! $attribute_values ) continue; // if empty
+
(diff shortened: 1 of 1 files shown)


bainternet:
@AntonRzevskiy 

You are missing the point that HTML attributes can and need to be empty and as for value passed to it, its up to the developer.

thanks anyway.

AntonRzevskiy:
Yes indeed, perhaps in very rare situations, empty attributes may be useful. :+1:

PR #6773, opened 2019-01-17
i18n: merge similar translation strings

ramiy:
Merge two similar translation strings. And hard-code the `360px` value, this way translators won't  be able enter wrong value by mistake.

elementor

The original translation string located in the same file, 6 lines above.


ramiy:
@arielk @KingYes Please review this PR.

PR #6775, opened 2019-01-17
i18n: split translation strings

ramiy:
Make it easier to translate the Translation String, by splitting it into two smaller strings. The second string already exist.

elementor-i18n-3


ramiy:
**Note:** this change won't add new translation strings.

PR #6776, opened 2019-01-17
i18n: avoid using HTML tags in translation strings

ramiy:
Not all the translators know HTML, this is why you should avoid using unneeded HTML tags where they are not needed.

elementor-i18n-4

Replace: `<strong>%s</strong>` with simple `%s`, moving the HTML tags out of the string.


KingYes:
Thanks !

PR #6791, opened 2019-01-19
Bug Fix: styling image widget captions

ramiy:
This PR fixes the wrong condition used when displaying the caption styling section.

Currently the "**Caption**" section under the "**Style**" tab displayed only when `"caption_source" = "custom"` and when `"caption_source" != ""`.

The user can't really style the caption when he choose to use "Attachment Caption" because of this wrong condition set to display the section.


ramiy:
**Note:** The same issue needs to be fixed in the "featured image" widget in Elementor Pro (@matipojo).

PR #6835, opened 2019-01-22
Release/2.4.4

simplysianne:
Hi, I was told to put my request for a feature here, so here it is: 

I've been testing out ShowIt and love their multiple canvas view functionality that allows the site to be more interactive and allows for multiple content to appear on a page without having to jump over to another page and without scrolling. I think adding this type of functionality to Elementor would be very beneficial to up-leveling site design. 

Here's an example: https://simply-sianne-2.showitpreview.com/

The third section is the one that has multiple canvas views.

Thanks!!


KingYes:
Please open new issue, not Pull Request.

PR #6994, opened 2019-02-06
Sticky Scrolling Effect stop point

benstone326:
Hello there,

I was wonder that will it be available in the near future to set a stop point for the Sticky Scrolling Effects?

For example: I started scrolling my header for the section I named #form.
So in the Sticky Scrolling Effects there will be a drop down menu, where you can choose what section do you want to stop the effect or simply write the ID of the section.

I hope you understand what I ask :)

Best regards,
Benjamin


KingYes:
Open new issue about this. Not PR.

PR #7214, opened 2019-02-26
Release/2.5.0

mauritzbotha92:
Feature request

Hi all, I would like to request 2 minor things be added to elementor, not necessarily relevant to the latest release.

Ability to add overlay for images with normal and hover states which will allow you to have a colour overlay or gradient overlay for example directly on the image.

The ability to add padding for borders. I don't always want a border to be only on the outside of an image. Sometimes I would like a border to have padding on the element it applies to, so the border can be on the inside of the element as well with a space around it. And maybe add more border patterns to choose from as well.

Many thanks,
Mauritz


KingYes:
Please open new Issue, not PR.

PR #7274, opened 2019-03-02
POC: Transform Control

ramiy:
@joshmarom 

Proof of concept implementing `Transform Control` in Elementor.

transform

This is a group control. More advanced fields can be added to the popup from:
https://developer.mozilla.org/en-US/docs/Web/CSS/transform#Syntax


KingYes:
Hey @ramiy 

We doing this by other way (Custom Position).

Thank you for your PR.

Issue #7311, opened 2019-03-05
Elementor 2.5.1 Video Widget Bug


nerdtech:
## Prerequisites

- [x] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [x] The issue still exists against the latest stable version of Elementor.

## Description

The video widget disappears when you apply "fixed" or "absolute" positioning.

## Steps to reproduce

1. Drag elementors video widget into a section.
2. Edit the widget and go to the "advanced" tab.
3. Select "custom positioning".
4. Under positioning apply either "fixed" or "absolute".
5. The widget disappears.

## Isolating the problem

- [x ] This bug happens with only Elementor plugin active (and Elementor Pro).
- [x ] This bug happens with a default WordPress theme active.
- [ x] I can reproduce this bug consistently using the steps above.

## Environment

<details>
<summary>
					
== Server Environment ==
	Operating System: WINNT
	Software: Apache
	MySQL version: MySQL Community Server - GPL v8.0.15
	PHP Version: 7.2.15
	PHP Max Input Vars: 1000
	PHP Max Post Size: 256M
	GD Installed: Yes
	ZIP Installed: Yes
	Write Permissions: All right
	Elementor Library: Connected

== WordPress Environment ==
	Version: 5.1
	Site URL: http://localhost/wordpress
	Home URL: http://localhost/wordpress
	WP Multisite: No
	Max Upload Size: 256 MB
	Memory limit: 256M
	Permalink Structure: /%postname%/
	Language: en-US
	Timezone: 0
	Debug Mode: Inactive

== Theme ==
	Name: Elementor Hello Theme
	Version: 1.1.0
	Author: Elementor Team
	Child Theme: No

== User ==
	Role: administrator
	WP Profile lang: en_US
	User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/72.0.3626.119 Safari/537.36

== Active Plugins ==
	Advanced Custom Fields PRO
		Version: 5.7.12
		Author: Elliot Condon

	All-in-One WP Migration
		Version: 6.77
		Author: ServMask

	Classic Editor
		Version: 1.4
		Author: WordPress Contributors

	Columns Alignment Fix for Elementor
		Version: 1.1.0
		Author: SukiWP

	Custom Post Type UI
		Version: 1.6.1
		Author: WebDevStudios

	Elegant Tabs for Elementor
		Version: 1.0
[...]

Issue #7309, opened 2019-03-05
Global Widgets Spacing option is buggy


reardestani:
## Prerequisites
- [ ] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [x] The issue still exists against the latest stable version of Elementor.

## Description
Global **Widgets Spacing** is acting differently in the latest version (2.5.1). In previous stable version it was fine. I am not sure it's a bug or new intended behavior but it does not look correct.

Video: http://take.ms/jwB7h

## Steps to reproduce
What above video.

## Isolating the problem
- [x] This bug happens with only Elementor plugin active (and Elementor Pro).
- [x] This bug happens with a default WordPress theme active.
- [x] I can reproduce this bug consistently using the steps above.

gulhangulez:
Reza +1. Yes, I also faced with the same issue. https://monosnap.com/file/273jfu4SpRyEMlQkzLWa8Db5J1p3bz

shilo-ey:
Hi @reardestani 

Thanks for reporting. we will look into this and update here.

Thanks.

arielk:
Fix for v2.5.2

Thanks for your feedback

mdrideout:
Using Version 2.5.9, just saved one of my pages, and then randomly all of my widgets across the site got a margin-bottom of 20px. Can't seem to get rid of it. 
```
.elementor-widget:not(:last-child) {
    margin-bottom: 20px;
}
```

PR #7318, opened 2019-03-05
Absolute & Fixed position default & fix for widget space

arielk:


arielk:
Fix: https://github.com/elementor/elementor/issues/7311
Fix: https://github.com/elementor/elementor/issues/7309

PR #7325, opened 2019-03-05
Text Editor Widget: Add a separator before columns control

ramiy:
Before:

columns-before

After:

columns-after


ramiy:
@KingYes

PR #7510, opened 2019-03-17
include/utils: fix - do html entity decode before json encode

nuritsha:
Make sure bool parameters are printed as strings.

### DIFF
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -650,7 +650,7 @@ class Utils {
 	 */
 	public static function print_js_config( $handle, $js_var, $config ) {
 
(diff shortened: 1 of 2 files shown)


includes/utils.php
```diff
@@ -649,6 +649,15 @@ public static function get_meta_viewport( $context = '' ) {
 	 * @param mixed $config
 	 */
 	public static function print_js_config( $handle, $js_var, $config ) {
+
+		// decode the data the same way wp_localize_script would have done
```
matipojo: Capital D

nuritsha: Done

PR #7734, opened 2019-04-04
Upgrade DB: Fix - If cannot upgrade show add an admin notice. (Fix #7725)

### LINKED ISSUE #7725: Database upgrade does not run in v2.5, all form ID's gone causing back blank forms be sent
Found problem on bug reported in #7691 Still a open issue.

I have the problem here. Over 150 websites all with this same problem. All form IDs are gone, which results in blank emails / submits. And error messages when connected with ActiveCampaign / MailChimp.

By downgrading to Elementor Pro to 2.4.8 the issue is also resolved here (temp solution, because v2.5.4 does not working).

**Found the problem**: the database Upgrade script in the background is not running. This happens when these lines are in wp-config.php
`define('DISALLOW_FILE_EDIT', true); `
`define('DISALLOW_FILE_MODS',true);`

When commetting out this lines, the upgrade script can run in the background and all IDs are back again.

Question now is: Why does Elementor Pro v2.4.8 does not have this issue while v2.5.x does? And is there a way to force the database upgrade with the DISALLOW_FILE_EDIT and DISALLOW_FILE_MODS enabled? Or at least warn the user that the database upgrade is not been made so they / we know that action needs to be taken?

Manually changing 150+ wp-config.php files is a lot of work. 
So how to solve this problem in bulk other than reverting all the sites back to v2.4.8?

matipojo:


matipojo:
Fix #7725

PR #7769, opened 2019-04-07
Image Box Widget: update position icons

ramiy:
Replace confusing "align" icons with "position" icons.

position-icons

Because this control positions the image on the Top/Left/Right side of the text.


ramiy:
image-box

ramiy:
@arielk @KingYes

PR #7770, opened 2019-04-07
Icon Box Widget: update position icons

ramiy:
Replace confusing "align" icons with "position" icons.

position-icons

Because this control positions the icon on the Top/Left/Right side of the text.


ramiy:
icon-box

ramiy:
@arielk @KingYes

PR #7911, opened 2019-04-18
Feature/commands

MellowMinded:
I think it's appropriate to post here. Although I see it as an issue, you may not.

It has come to my dreadful surprise that the page builder has not a disable responsive mode. I mean, come on. Whyy? I am crying and going batshit crazy over some little mobile adjustment that makes the site uncomfortable to look at. For my page it would be much sexier with the desktop version on mobile.
Is there any way I still can disable responsive?


KingYes:
Open new issue, no PR.

PR #7919, opened 2019-04-20
Typography Control > Text Decoration > Add "Underline & Overline" option

ramiy:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

* Add "Underline & Overline" option to "Text Decoration" in Typography Control

## Description

Add "Underline & Overline" option to "Text Decoration" in Typography Control

```css
text-decoration: underline overline;
```

## Quality assurance

### DIFF
--- a/includes/controls/groups/typography.php
+++ b/includes/controls/groups/typography.php
@@ -169,6 +169,7 @@ class Group_Control_Typography extends Group_Control_Base {
 				'underline' => _x( 'Underline', 'Typography Control', 'elementor' ),
 				'overline' => _x( 'Overline', 'Typography Control', 'elementor' ),
 				'line-through' => _x( 'Line Through', 'Typography Control', 'elementor' ),
+				'underline overline' => _x( 'Underline & Overline', 'Typography Control', 'elementor' ),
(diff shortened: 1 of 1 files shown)


arielk:
@ramiy You can also add `text-decoration: underline overline line-through;`

ramiy:
@arielk Yes, I know.

I also think the "**Text Decoration**" should use the **Select2** control instead of the **Select** control, this way the user can combine `underline`, `overline` and `line-through`.

In addition, we can add conditional controls for [text-decoration-color](https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-color) and [text-decoration-style](https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-style). Only if the text decoration is not empty.

But first, you should check out #7944 (See the two screenshots) and read the logic behind the change.

arielk:
@ramiy 
As a matter of fact, all these options (including the original) are very niche, so I'm not sure why we should add it to the core.

Additionally, `text-decoration-color` and `text-decoration-style` are not supported in all browsers (IE & Edge of course) so we may need to take a step back with this Pull Request.

PR #8250, opened 2019-06-03
Send get url to get values in popup fields from the get method

sbc640964:
It is simple, like the subject, that it is possible to send parameters in GET to the forms on the popup


KingYes:
Open new issue about this, not PR.

PR #8281, opened 2019-06-10
Maintenance Mode: New - Add action hook on change mode.

matipojo:


includes/maintenance-mode.php
```diff
@@ -296,6 +296,12 @@ public function print_style() {
 		<?php
 	}
 
+	public function on_update_mode( $old_value, $value ) {
```
KingYes: explian

KingYes (changes requested):

Issue #5045, opened 2018-07-13
Quotes manipulation by Elementor


tufloc:
## Prerequisites
- [x] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [x] The issue still exists against the latest stable version of Elementor.

## Description
When using Elementor with another third-party plugin, Elementor seems to be changing the quotes in the plugin's code and as a result disrupting its operations.

Bellow is a screenshot of the source code of the affected page when Eelementor is running.

![bug](https://user-images.githubusercontent.com/1139134/42678349-49d6fa1a-8687-11e8-9f45-9cebd07c274e.png)

And here is a screenshot of the source code from the same affected page when Elementor is disabled.

![bug1](https://user-images.githubusercontent.com/1139134/42678423-84a8fbac-8687-11e8-968f-f962003cf2e5.png)

As you can see Elementor seems to be changing all the quotes that the plugin uses for its json code to 
a different kind of quotes, which causes disruption in the plugin code.

## Steps to reproduce

1. Use Elementor.
2. Use [Super Forms's](http://f4d.nl/super-forms/) conditional logic feature.

## Isolating the problem

- [x] This bug happens with only Elementor plugin active (and Elementor Pro).
- [x] This bug happens with a default WordPress theme active.
- [x] I can reproduce this bug consistently using the steps above.

## Environment

<details>
<summary>System Info</summary>
```
					
== Server Environment ==
	Operating System: Linux
	Software: Apache/2.4.18 (Ubuntu)
	MySQL version: 5.7.22
	PHP Version: 7.0.30-0ubuntu0.16.04.1
	PHP Max Input Vars: 1800
	PHP Max Post Size: 500M
	GD Installed: Yes
	ZIP Installed: Yes
	Write Permissions: There are some writing permissions issues with the following directories/files:
		 - WordPress root directory
	Elementor Library: Connected

== WordPress Environment ==
	Version: 4.9.6
	Site URL: https://trekworlds.com
	Home URL: https://trekworlds.com
	WP Multisite: No
	Max Upload Size: 500 MB
	Memory limit: 40M
	Permalink Structure: /%postname%/
	Language: en-US
	Timezone: 0
	Debug Mode: Inactive

== Theme ==
	Name: Bulmapress
	Version: 0.0.1
	Author: Scops
	Child Theme: No

== User ==
	Role: administrator
	WP Profile lang: en_US
[...]

PR #8300, opened 2019-06-11
Fix post title potentially breaking editor

VesterDe:
## PR Checklist

- [x ] The commit message follows our guidelines:  https://github.com/elementor/elementor/blob/master/.github/CONTRIBUTING.md

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

Fix post title potentially breaking editor, if it containing a single quote, for example.

## Description

Using esc_attr on the post_title as it is enqueued for frontend settings.

## Test instructions

Create a new Page/Post/<other post_type>, and put an single quote - ' -  in the title (French users are particularly likely to do so organically). Before this fix, you would get a "Preview could not be loaded" error, because the JSON generated from the frontend settings would have been corrupted by the stray quote.

By using esc_attr, it gets encoded to &#039; which fixes the error.

## Quality assurance

- [ x] I have tested this code to the best of my abilities

Fixes #

P.S. Should I have added some tests/docs for this?

### DIFF
--- a/includes/frontend.php
+++ b/includes/frontend.php
@@ -1048,7 +1048,7 @@ class Frontend extends App {
 			$post = get_post();
 			$settings['post'] = [
 				'id' => $post->ID,
-				'title' => $post->post_title,
+				'title' => esc_attr($post->post_title),
 				'excerpt' => $post->post_excerpt,
 			];
 		} else {


VesterDe:
Ok apparently this isn't a problem for a fresh install and no other plugins, so on second review this bug is a relative of

https://github.com/elementor/elementor/issues/5045

and a consequence of some plugin changing json encoding in some way. My pull request still fixes it, but the direct culprit is not clear.

bainternet:
@VesterDe 

Thanks for the PR but without being able to replicate this, which we can't, I don't think it's needed.

Thanks again.

PR #8770, opened 2019-08-04
Tweak: Canvas - add support of `wp_body_open`.

nuritsha:


modules/page-templates/templates/canvas.php
```diff
@@ -23,6 +23,7 @@
 </head>
 <body <?php body_class(); ?>>
 	<?php
+	Elementor\Modules\PageTemplates\Module::body_open();
```
matipojo: Why not add here `do_action( 'wp_body_open' );`?

KingYes: When we drop all `WP < 5.2`  support, we just remove that method.

matipojo: With `do_action( 'wp_body_open' )` we shouldn't do anything.

Issue #8784, opened 2019-08-06
PHP Notice:  Undefined offset: 0 in elementor/core/documents-manager.php on line 411


dbjpanda:
$caps[0] should be checked first.

```
if ( empty( $allcaps[ $caps[0] ] ) ) {
                return $allcaps;
            }
```
should be       
 ```
if ( isset( $caps[0] )){
            if ( empty( $allcaps[ $caps[0] ] ) ) {
                return $allcaps;
            }
        }
```

shilo-ey:
Thanks @dbjpanda !

A patch has already been submitted (#8734)

PR #8785, opened 2019-08-06
Update documents-manager.php #8784

dbjpanda:
Fix for #8784


KingYes:
Thank you, but it's fixed by #8734

PR #8904, opened 2019-08-26
 Make the divider MUCH better

joshmarom:
Now using Background-image instead of svg pattern
Now includes animation feature


ramiy:
@joshmarom you added elementor-pro textdomain.

PR #9045, opened 2019-09-10
Added missing contains and !contains conditions in PHP closes #9043

### LINKED ISSUE #9043: Conditions class does not match its JS counterpart
JS Conditions function has contains 

https://github.com/elementor/elementor/blob/5d257d0455d17cd4dae7b1d0d21c73d11ecfafb2/assets/dev/js/editor/utils/conditions.js#L20-L23

which works right in the editor to check if the string is in the control that has array as value

```
				'conditions' => array(
					'terms' => array(
						array(
							'name' => 'meta_elements',
							'operator' => 'contains',
							'value' => 'author',
						),
					)
				)
```

this option does not exists in the php version of the Conditions class

https://github.com/elementor/elementor/blob/5d257d0455d17cd4dae7b1d0d21c73d11ecfafb2/includes/conditions.php#L33-L56

Although this works  in the editor to show/hide conditional controls,  the value is always null in the frontend because the missing cases.

A simple `in_array `would match both switches

```
			case 'contains':
				return false !== in_array($right_value,$left_value);
			case '!contains':
				return false === in_array( $right_value, $left_value );
```

bainternet:


includes/conditions.php
```diff
@@ -42,6 +42,10 @@ public static function compare( $left_value, $right_value, $operator ) {
 				return false !== array_search( $left_value, $right_value );
 			case '!in':
 				return false === array_search( $left_value, $right_value );
+			case 'contains':
```
matipojo: It's same to 'in' '!in', just add the 'contain' cases:
```
case 'in':
case 'contains':
```

matipojo (changes requested):

includes/conditions.php
bainternet: it checks the other way around

includes/conditions.php
```diff
@@ -39,9 +39,13 @@ public static function compare( $left_value, $right_value, $operator ) {
 			case '!==':
 				return $left_value !== $right_value;
 			case 'in':
-				return false !== array_search( $left_value, $right_value );
+				return in_array( $left_value, $right_value );
```
matipojo: Use a `strict` compare in all `in_array`:
```suggestion
				return in_array( $left_value, $right_value, true );
```

matipojo (changes requested):

matipojo (approved):

PR #9482, opened 2019-11-05
Icons Control Inline Skin - Fixes and Tweaks

udidol:
- Fixed: Bugs in the Icons Control's Inline Skin
- Added the ability to exclude icon types (icon/svg) from the inline control skin.
- Changed the `ICONS` controls to `inline` skin and added recommended icons - in Accordion and Toggle widgets

## PR Type

- [X] Bugfix
- [X] Feature

Fixes #
Missing class prefix in the `ui` object building.

### DIFF
--- a/assets/dev/js/editor/controls/icons.js
+++ b/assets/dev/js/editor/controls/icons.js
@@ -307,14 +307,17 @@ class ControlIconsView extends ControlMultipleBaseItemView {
 			} );
 		}
 
-		const previewHTML = '<i class="' + iconValue + '"></i>';
+		if ( 'media' === skin || 'svg' !== iconType ) {
+			const previewHTML = '<i class="' + iconValue + '"></i>';
 
-		iconContainer.html( previewHTML );
+			iconContainer.html( previewHTML );
+		}
 
 		this.enqueueIconFonts( iconType );
 	}
 
 	setDefaultIconLibraryLabel( defaultIcon, iconContainer ) {
+		console.log( defaultIcon );
 		// Check if the control has a default icon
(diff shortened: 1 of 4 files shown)


assets/dev/js/editor/controls/icons.js
```diff
@@ -307,14 +307,17 @@ class ControlIconsView extends ControlMultipleBaseItemView {
...
 
-		iconContainer.html( previewHTML );
+			iconContainer.html( previewHTML );
+		}
 
 		this.enqueueIconFonts( iconType );
 	}
 
 	setDefaultIconLibraryLabel( defaultIcon, iconContainer ) {
+		console.log( defaultIcon );
```
kobizz: Remove it

includes/controls/icons.php
```diff
@@ -155,6 +159,9 @@ protected function get_default_settings() {
 			'recommended' => false,
 			'is_svg_enabled' => Svg_Handler::is_enabled(),
 			'skin' => 'media',
+			'inline_options' => [
+				'exclude' => [],
+			],
```
kobizz: Do it like the following example:
```
'inline_options' => [
    'icon',
    'svg',
],
```

kobizz (changes requested):

includes/controls/icons.php
udidol: It doesn't work. because the override settings don't fully _overwrite_ the default settings - it happens with `array_merge` (`\includes\controls\base.php` in the constructor).
So if you only include one of the two options (`icon` or `svg`) in the widget's control instance - the `array_merge` will add the other one. 

It has to be a nested array.

kobizz (approved):

PR #9594, opened 2019-11-17
History/Revisions: Remove - Delete revision is no longer supported. (…

matipojo:


KingYes:
See more: https://core.trac.wordpress.org/ticket/43709

PR #9724, opened 2019-11-29
Source image replace with placeholder image due to server timeout

maheshwaghmare:
The images are downloaded with `wp_safe_remote_get()`. Some times due to large image or server timeout, the function returns the empty file content.

And it returns `false`. It should be `$attachment` which holds the source image url.

When image not downloaded, then for blank images, the `placeholder.png` is loaded on the front end.

---

With filter `http_request_timeout`, we can increase the timeout, but, somehow the image is not downloaded, then it should show the source image.


KingYes:
Hi @maheshwaghmare

Thanks for your contribution.

Unfortunately, we won’t add this option due to the reason that it will allow potential hotlinking and issues with CORS policy.

Thanks again.

PR #9795, opened 2019-12-07
fix gruntjs svg icon src

akkspros:
fix gruntjs svg icon src in README.md file


KingYes:
This file is auto generated by grunt task. Thanks.

PR #9832, opened 2019-12-11
Fix: Set padding on icon span instead of text span (#9831) [ED-9757]

### LINKED ISSUE #9831: 🔗 🚀 Feature Request: Icon List Widget - Set text-indent spacing on the icon span (Related to #2167) - [ED-9757]
## Prerequisites

- [x] I have searched for similar features requests in both open and closed tickets and cannot find a duplicate.
- [x] The feature is still missing in the latest stable version of Elementor.

### **What problem is your feature request going to solve? Please describe.**
When using the **Icon List** widget in inline layout mode, utilizing the _Text Indent_ option adds left padding to the `<span>` containing the list item's text. This is an issue when you're using the **Icon List** widget without an icon in list item. Using the _Text Indent_ control on an item without an icon will cause the clickable area/hitbox of the item to have the added padding. 

This can cause undesired spacing when mixing list items that have icons and list items that don't have icons. It's especially noticeable when using the divider functionality, or in the case of underlining the `<a>` tag using custom CSS (which is how I found the root of this issue).

_**Note:** This also happens when the widget is set to default layout mode, but the use-case of mixing icon'ed and non-icon'ed list items is far less common._

### **Describe the solution you'd like**
I would like to propose switching the application of the _Text Indent_ option to adding the padding to the right of the `<span>` that contains the icon, rather than the current implementation of adding left padding to the `<span> that contains the text.

Since the `<span>` that contains an icon does not output if there isn't an icon selected, the text-indent isn't applicable when 

### **Describe alternatives you've considered**
Alternatively, a class could be added to list items that have an icon or don't have an icon and can be used to selectively output the text-indent. The CSS that is output for the text indent could target only list items that have an icon.

### **Possible Related Issue**
#2167

cirkut:
## PR Checklist

## PR Type

- [x] Bugfix
- [x] Feature
- [x] Code style update (formatting, local variables)
- [ ] Build related changes
- [ ] CI related changes

## Summary

* Text-Indent in Icon List widget items applies padding to icon element instead of text element

## Description

* When a list item doesn't have an icon, the anchor (if the list item has one) will have blank space padded on the left because the text-indent control would add padding to the left of the `<span>` that contains the text. This fixes that because the icon element does not output unless there is an icon selected, which means there will be no padding applied to the anchor tag in the event of a list item not having an icon selected.

## Test instructions

* Creating an Icon List widget and setting the layout to `inline`
* Remove an icon from any list item
* Add a number to the _Text Indent_ control
* Preview the page
* Verify that the number added to the _Text Indent_ control is added as CSS to the `.elementor-icon-list-icon` class as `padding-right` (in LTR languages, left in RTL languages) instead of the previous implementation of adding `padding-left` to the `.elementor-icon-list-text` element.

## Quality assurance

- [ ] I have added unit tests to verify the code works as intended

Fixes #9831


joshmarom (commented):
Seems okay

cirkut:
Wanted to follow up on this to see if there's any additional information I can provide or if there's anything else I can do to help get this change pushed live. This is specifically for the Icon List widget, but I've noticed this kind of issue appearing in a couple of other place (like the Button widget with icon spacing, though I think that only applies on the backend when editing the widget, whereas this issue relates to the final stored CSS).

cirkut:
Commenting on this again, would love to have this pushed live. It's purely a simple shift of what element adds the text-indent. Text-indent should be right padding on the icon, not the left of the text.

cirkut:
Commenting once again. Seriously guys, this is a freebie, it's already ready, and it makes the most sense. It's almost like you don't actually want people contributing to Elementor at all. What's the point in having this on GitHub if you're not going to merge even the most simple of changes?

rami-elementor (approved):

PR #9976, opened 2019-12-23
Fixed add_render_attribute output issue

### LINKED ISSUE #9975: add_render_attribute generate wrong output
## Description

When we use the "add_render_attribute" function without any '$key' and '$value' argument then the system will generate the wrong HTML like below.

    <div ="">This is Test</div>

Check the screenshot for more information.

Error

## Steps to reproduce

Please check below widget code to reproduce the bug.

    <?php

	protected function render() {
		
		$this->add_render_attribute( 'text_block' );

		$this->add_inline_editing_attributes( 'text_block' );

		echo '<div '.$this->get_render_attribute_string( 'text_block' ).'>';
			echo $settings['text_block'];
		echo '</div>';

	}

<details>
<summary>System Info</summary>
https://pastebin.com/qzLuV896

mukeshpanchal27:
Fix #9975

### DIFF
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -574,6 +574,10 @@ class Utils {
 		$rendered_attributes = [];
 
 		foreach ( $attributes as $attribute_key => $attribute_values ) {
+			if ( empty( $attribute_key ) ) {
+				return '';
+			}
+			
 			if ( is_array( $attribute_values ) ) {
 				$attribute_values = implode( ' ', $attribute_values );
 			}


matipojo:
Hi @mukeshpanchal27 

Thanks for the fix. 

I think the solution is to add some validations on `add_render_attribute`, to ensure that the value is not empty.

kobizz:
Hi @mukeshpanchal27 

Thanks for your contribution. We decided not to include your PR at this moment.

When using the `add_render_attributes` method, you should always specify a key and a value, either by using the method arguments or passing an array of keys/values.

Thank you!

PR #10072, opened 2020-01-06
Tweak: Added dynamic capabilities to Title control in Toggle widget closes #10070

### LINKED ISSUE #10070: Accordion Widget dynamic Nature
I don't know whether it's a bug or feature. However, I am adding it in bug section.

If you will check your accordion widget, it has got two options:

1. Title
2. Description

There is also a dynamic option for this widget.

Now, the issue is a dynamic option is only present for the title. There is no dynamic option to fetch text for the description.

I don't know if there is a way to fetch two vales in two different options using one dynamic field. Do let me know if there is one.

Or

Kindly add a dynamic option for the description too. It would be of great help.

Thanks

shilo-ey:


includes/widgets/accordion.php
```diff
@@ -110,6 +110,9 @@ protected function _register_controls() {
 				'type' => Controls_Manager::WYSIWYG,
 				'default' => __( 'Accordion Content', 'elementor' ),
 				'show_label' => false,
+				'dynamic' => [
```
matipojo: Maybe it should be a default for all text/wysiwyg controls?

matipojo (changes requested):

includes/widgets/accordion.php
shilo-ey: We can add it just in WYSIWYG

kobizz (dismissed):

includes/widgets/accordion.php
shilo-ey: @kobizz @matipojo Updated - removed dynamic registration from this PR. handled by udi in different PR

matipojo (approved):

kobizz (approved):

PR #10422, opened 2020-02-05
Background tasks: Fix - init common app. (Fix upgrade fail with a PHP fatal error)

matipojo:


KingYes:
@matipojo Fix code owner on this file.

PR #10432, opened 2020-02-06
Editor 2.9.0: Fix - Set default route, for all elements & add global widget…

iNewLegend:


assets/dev/js/editor/regions/panel/pages/editor/component.js
```diff
@@ -24,11 +24,17 @@ export default class Component extends ComponentBase {
 	defaultCommands() {
 		return {
 			open: ( args ) => {
-				this.openEditor( args.model, args.view );
+				if ( ! this.setDefaultTab( args ) ) {
+					// TODO: Remove - Backwards compatibility.
```
iNewLegend:
```suggestion
```

PR #10459, opened 2020-02-09
Frotnend: Fix - Stretch Element error when change to canvas.

matipojo:


kobizz (changes requested):
The stretch tool must not run at all when it's not represent any element. It needs a more profound investigation.

matipojo:
It's not fixing the problem.

PR #10615, opened 2020-02-20
Editor/CSS-Parser: Change - Filter style also in page_settings.

iNewLegend:


kobizz (changes requested):
We've already agreed on another approach for this issue

matipojo (approved):

kobizz (approved):

PR #10618, opened 2020-02-20
Fix: `<a>` Links appear with `underline` text decoration

shilo-ey:


assets/dev/scss/frontend/_global.scss
```diff
@@ -13,6 +13,11 @@
 	// Reset Style for some themes and browsers
 	hyphens: manual;
 
+	a {
+		box-shadow: none;
+		text-decoration: none;
```
matipojo: How can the Theme Style override this rule?

matipojo (changes requested):

assets/dev/scss/frontend/_global.scss
shilo-ey: It's stronger

kobizz (approved):

matipojo (approved):

PR #10691, opened 2020-02-27
 Feature/use panel open and close commands

iNewLegend:


matipojo (dismissed):

assets/dev/js/editor/regions/panel/component.js
```diff
@@ -24,8 +24,12 @@ export default class Component extends ComponentBase {
 
 	defaultCommands() {
 		return {
-			open: () => elementor.getPanelView().modeSwitcher.currentView.setMode( 'edit' ),
-			close: () => elementor.getPanelView().modeSwitcher.currentView.setMode( 'preview' ),
+			open: () => new class Open extends $e.modules.CommandBase {
```
kobizz: A class as a command??

iNewLegend: Most of the commands are like that. ( non using class is optional ) take look at `document/elements` for exmaple `wp-content/plugins/elementor/assets/dev/js/editor/document/elements/commands`.

kobizz: You should export it to a new file

kobizz (changes requested):

PR #10750, opened 2020-03-05
Fix: Whatsapp share link does not escape html entities (PT 553)

udidol:
## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

* Fix: Whatsapp share link does not escape html entities

## Description

* Added HTML entity decoding to all page titles, and added URL encoding to make sure WhatsApp handles special characters properly

## Test instructions

- Create a new page, and give the page a title with special characters like quotes and ampersands (`&`).
- Edit the page with Elementor
- Add a Share Buttons widget
- Add Share Buttons for all platforms, share the page and see that the page title and page URL are displayed correctly in the shared post


kobizz (commented):
Why do you always need to do the encoding?

udidol:
Because the post title can always have special characters in it, such as quotes and ampersands. That was the bug in the first place

kobizz (dismissed):

matipojo (approved):

includes/frontend.php
```diff
@@ -1152,10 +1152,9 @@ protected function get_init_settings() {
 			$post = get_post();
 			$title = wp_get_document_title();
 
-			if ( false !== strpos( $title, '&#8211' ) ) {
-				// Convert WP's core N-dash separator's HTMLEntity code to a regular dash character
-				$title = html_entity_decode( $title, ENT_NOQUOTES, 'UTF-8' );
-			}
+			// Convert WP's core N-dash separator's HTMLEntity code to a regular dash character,
+			// then encode the title to escape problematic characters
+			$title = rawurlencode( html_entity_decode( $title, ENT_QUOTES, 'UTF-8' ) );
```
matipojo: Can you add a PHP Unit test?

matipojo (changes requested):

tests/phpunit/elementor/includes/test-frontend.php
```diff
@@ -0,0 +1,24 @@
+<?php
```
matipojo: Move to utils test.

tests/phpunit/elementor/includes/test-frontend.php
```diff
@@ -0,0 +1,24 @@
...
+namespace Elementor\Testing\Includes;
+
+use Elementor\Testing\Elementor_Test_Base;
+use Elementor\Utils;
+
+class Elementor_Test_Frontend extends Elementor_Test_Base {
+	// This tests the urlencode_htmlentities utility method, to see that it properly decodes HTML Entities and then
+	// URL-encodes them to be used in requests
+	public function test_should_return_urlencoded_string() {
+		add_filter( 'document_title_parts', function () {
```
matipojo: Explain the filter hack.

matipojo (changes requested):

matipojo (dismissed):

includes/utils.php
```diff
@@ -656,4 +656,31 @@ public static function is_empty( $source, $key = false ) {
...
 	}
+
+	/**
+	 * Convert HTMLEntities to UTF-8 characters,
+	 * then encode the title to escape problematic characters in URLs
+	 *
+	 * @param $string
+	 * @return string
+	 */
+	public static function urlencode_htmlentities( $string ) {
```
kobizz: Do only `html_entity_decode` functionality

tests/phpunit/elementor/includes/test-utils.php
```diff
@@ -142,6 +142,28 @@ public function test_getYoutubeId() {
 		$this->assertNull( \Elementor\Embed::get_video_properties( 'https://www.youtube.com/' ) );
 	}
 
+	/**
+	 * Test Should Return a URL-Encoded String
+	 *
+	 * This tests the urlencode_htmlentities() utility method, to see that it properly decodes HTML Entities and then correctly URL-encodes them to be used in URLs, such as social media sharing
+	 */
+	public function test_should_return_urlencoded_string() {
```
kobizz: Fix the test according to the changes requested above

kobizz (changes requested):

includes/utils.php
```diff
@@ -656,4 +656,30 @@ public static function is_empty( $source, $key = false ) {
...
 		return defined( 'ELEMENTOR_PRO_VERSION' );
 	}
+
+	/**
+	 * Convert HTMLEntities to UTF-8 characters
+	 *
+	 * @param $string
+	 * @return string
+	 */
+	public static function decode_htmlentities( $string ) {
```
kobizz:
```suggestion
	public static function decode_html_entities( $string ) {
```

kobizz (changes requested):

includes/utils.php
matipojo: Why? now some titles will throw error at sharing.

kobizz: In order to keep the generic of this method, it will be used like this:

`rawurlencode( Utils::decode_htmlentities( YOUR_TEXT ) )`

kobizz (dismissed):

includes/utils.php
matipojo: But the only use case for this util is for encode URL's properly, so the user shouldn't think about `rawurlencode`.

The method name may be `Utils::encode_url`;

udidol: I agree with you on this. Done.

matipojo (approved):

kobizz (approved):

KingYes (approved):

PR #10784, opened 2020-03-08
Editor/Views/Widget: Fix - If new widget appear and it has dynamics b…

iNewLegend:


assets/dev/js/editor/elements/views/widget.js
```diff
@@ -147,7 +147,12 @@ WidgetView = BaseElementView.extend( {
 			var editModel = this.getEditModel(),
 				$template = jQuery( '#tmpl-elementor-' + editModel.get( 'widgetType' ) + '-content' );
 
-			this._templateType = $template.length ? 'js' : 'remote';
+			// In other words: If not not initial preview and it has dynamics then request from remote.
+			if ( elementor.previewAttachedOnce && editModel.get( 'settings' ).get( '__dynamic__' ) ) {
+				this._templateType = 'remote';
```
iNewLegend: Not good enough, will be called also from reload preview. and i assume may come from others, best way to optimize if can be done only when element created for the first time.

PR #10884, opened 2020-03-18
Fix: Theme style overrides the page Background Color - Page Settings …

udidol:
…> Style > Background (PT 567)

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

* Fix: Theme style overrides the page Background Color

## Quality assurance

### DIFF
--- a/includes/controls/groups/background.php
+++ b/includes/controls/groups/background.php
@@ -132,7 +132,7 @@ class Group_Control_Background extends Group_Control_Base {
 			'default' => '',
 			'title' => _x( 'Background Color', 'Background Control', 'elementor' ),
 			'selectors' => [
-				'{{SELECTOR}}' => 'background-color: {{VALUE}};',
+				'{{SELECTOR}}' => 'background: {{VALUE}};',
 			],
 			'condition' => [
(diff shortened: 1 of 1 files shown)


includes/controls/groups/background.php
```diff
@@ -132,7 +132,7 @@ public function init_fields() {
 			'default' => '',
 			'title' => _x( 'Background Color', 'Background Control', 'elementor' ),
 			'selectors' => [
-				'{{SELECTOR}}' => 'background-color: {{VALUE}};',
+				'{{SELECTOR}}' => 'background: {{VALUE}};',
```
matipojo: Background color should work together with background image (for example if a PNG includes a transparent area)

matipojo (changes requested):

includes/controls/groups/background.php
udidol: It works perfectly with a regular background color. 
I don't think there is a way to make it work with a gradient background without using JS.
What do you think? @matipojo

PR #10889, opened 2020-03-18
Frontend/Admin Bar: Fix - "Edit with Elementor" disappeared on WP >= 5.4. (Fix #10745)

### LINKED ISSUE #10745: "No Templates For Condition" after updating to WP 5.4
We are experiencing an issue where after updating the WP core to 5.4rc the templates of a page are no longer recognized, like so:

DeepinScreenshot_select-area_20200304212548

Reverting WP core back to 5.3 seems to solve the problem and the template along with the "Edit with Elementor" button re-appear.
DeepinScreenshot_select-area_20200304212616

This test was run on a clean db with only Elementor & Elementor Pro (latest) plugins activated and Hello Elementor theme.

matipojo:


KingYes (approved):

includes/frontend.php
```diff
@@ -216,6 +216,9 @@ public function init() {
 
 		// Add Edit with the Elementor in Admin Bar.
 		add_action( 'admin_bar_menu', [ $this, 'add_menu_in_admin_bar' ], 200 );
+
+		// Detect Elementor documents via their css that printed before the Admin Bar.
+		add_action( 'elementor/css-file/post/enqueue', [ $this, 'add_document_to_admin_bar' ] );
```
kobizz: Why don't you call `add_document_to_admin_bar` directly from where you enqueue the CSS?

kobizz (changes requested):

includes/frontend.php
matipojo: Because they are not related. frontend.php is the controller.

kobizz: I still think this is not the right way to deal with that, but I won't delay the release for that

kobizz (approved):

PR #10956, opened 2020-03-24
Fix: Pinterest social sharing cannot share posts with large featured images (PT 240)

udidol:
## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

* Fix: Pinterest social sharing returns error in posts with large featured images

## Description

* Since Pinterest has a problem sharing large images, I limited the featured image used to 1600px width, and use the closest thumbnail under that width.

## Quality assurance

### DIFF
--- a/includes/frontend.php
+++ b/includes/frontend.php
@@ -1163,11 +1163,32 @@ class Frontend extends App {
 
 			$title = Utils::urlencode_html_entities( wp_get_document_title() );
 
+			$thumb_id = get_post_thumbnail_id();
+			$thumb_size = 'full';
+			$full_thumb_meta = wp_get_attachment_image_src( $thumb_id, $thumb_size );
+
+			// If thumb is wider than 1600px, get the closest thumbnail size to it
+			// This is done because Pinterest has a problem sharing large images
+			if ( $thumb_id && $full_thumb_meta && ( 1600 < $full_thumb_meta[1] ) ) {
+				$thumbnail_sizes = Group_Control_Image_Size::get_all_image_sizes();
+
+				uasort( $thumbnail_sizes, function( $a, $b ) {
+					return $b['width'] - $a['width'];
+				} );
+
+				foreach ( $thumbnail_sizes as $size_slug => $size_meta ) {
+					if ( 1600 > $size_meta['width'] ) {
+						$thumb_size = $size_slug;
+						break;
+					}
+				}
+			}
+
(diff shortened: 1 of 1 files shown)


includes/frontend.php
```diff
@@ -1163,11 +1163,32 @@ protected function get_init_settings() {
...
+					}
+				}
+			}
+
 			$settings['post'] = [
 				'id' => $post->ID,
 				'title' => $title,
 				'excerpt' => $post->post_excerpt,
-				'featuredImage' => get_the_post_thumbnail_url(),
+				'featuredImage' => get_the_post_thumbnail_url( null, $thumb_size ),
```
matipojo: It should be used only for Pinterest, add it as an additional setting. and extract the logic to a separated method.

matipojo (changes requested):

includes/frontend.php
udidol: I think I will just try to pull the 'large' size thumbnail, and have the 'full' as a fallback - this is more economical, and less logic

includes/frontend.php
```diff
@@ -1163,11 +1163,19 @@ protected function get_init_settings() {
 
 			$title = Utils::urlencode_html_entities( wp_get_document_title() );
 
+			// Try to get the 'large' size featured image thumbnail
```
matipojo: Please explain why use the `large`.

matipojo (changes requested):

includes/frontend.php
udidol: Because it is the largest default WP size - the chances someone would change the dimensions of it are very small. It is the safest choice if we want to avoid using an expensive operation to find the largest image that would pass through Pinterest.

Anyway, I am also e-mailing with Pinterest showing them the issue, hopefully we will find the culprit.

matipojo: I mean explain in code (or maybe close the PR until it's ready for merge?)

udidol: Ok will do

udidol: Would rather not close it to not forget about it.
Marked it as WIP, this should prevent merge

udidol: Done

matipojo (approved):

kobizz (approved):

KingYes (approved):

Issue #10095, opened 2020-01-08
Icons Font Display Swap


jonathansoltero:
## Prerequisites

- [x] I have searched for similar features requests in both open and closed tickets and cannot find a duplicate.
- [x] The feature is still missing in the latest stable version of Elementor ( Elementor Pro. )

**What problem is your feature request going to solve? Please describe.**
PageSpeed Insights "Ensure text remains visible during webfont load" Fonts being loaded by Elementor such as the icons are causing a 1-second delay in load time.

I am not referring to custom fonts from Elementer Pro, I am referring to the bundled icons/fonts with Elementor, paths such as:

```
/elem/assets/lib/font-awesome/webfonts/fa-solid-900.woff2
/elem/assets/lib/eicons/fonts/eicons.woff2?5.5.0
```

**Describe the solution you'd like**
Adding `font-display: swap;` 

Update: Font Awesome did this themselves, just update to the latest version 5.15

**Describe alternatives you've considered**
I have added the code in myself but they were removed after the update. I don't want to keep manually adding the font display.

**Additional context**

EdwardBlack:
+1 for that Issue.

Even though you don't reference the custom uploaded fonts, there should also be an option to select what font-display-attribute you want to use in the interface.

stannvanheusden:
I second this. Not only for internal fonts: also for Google Fonts, TypeKit etc.

jdub850000:
Thirded!

dineshmistry:
Fourth!

draconnia:
On the one hand  **Lighthouse**  tells us to use  `font-display: swap` and delays the load
On the other hand, we are told that  `font-display: swap` is not ideal for an icon font, because there's no benefit in showing the character in a fallback typeface.
What can we do?

visrey:
I think this issue is fixed in the latest release of font awesome refer the ticket
 
https://github.com/FortAwesome/Font-Awesome/issues/16077

Upgrading the font-awesome version should fix the issue.

jonathansoltero:
> I think this issue is fixed in the latest release of font awesome refer the ticket
> 
> [FortAwesome/Font-Awesome#16077](https://github.com/FortAwesome/Font-Awesome/issues/16077)
> 
> Upgrading the font-awesome version to 5.13 should fix the issue.

What version is Elementor on now?

jonathansoltero:
Can we also consider adding <link rel=preload>

visrey:
[...]

PR #11037, opened 2020-04-02
Add font-display: swap to Font Awesome for better Google Page Speed Insights score

jankohoener:
## PR Type
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Description

Adds the property "font-display: swap" to the CSS code of Font Awesome. This ensures that text remains visible during page load and avoids a flash of invisible text (see also https://web.dev/font-display/)
In addition, this improves the Google Page Speed Insights score and is a fix for issue #10095.


joshmarom:
Hi @jankohoener. `font-display: swap` is not ideal for an icon font. There's no benefit in showing a fallback typeface as it will not communicate the meaning of the font, and will only be a distraction.

joshmarom (commented):
Hi @jankohoener. `font-display: swap` is not ideal for an icon font. There's no benefit in showing the character in a fallback typeface, as it will not communicate the meaning of the icon, that would only cause a distraction.

PR #11229, opened 2020-04-26
Feature/global panel

matipojo:


assets/dev/js/frontend/handlers/section/stretched-section.js
```diff
@@ -7,10 +7,19 @@ export default class StretchedSection extends elementorModules.frontend.handlers
 		elementorFrontend.addListenerOnce( handlerID, 'sticky:stick', this.stretch, this.$element );
 
 		elementorFrontend.addListenerOnce( handlerID, 'sticky:unstick', this.stretch, this.$element );
+
+		if ( elementorFrontend.isEditMode() ) {
+			this.onKitChangeStretchContainerChange = this.onKitChangeStretchContainerChange.bind( this );
+			elementor.channels.editor.on( 'kit:change:stretchContainer', this.onKitChangeStretchContainerChange );
```
kobizz, line 13: We should supply a general alternative for the `onGeneralSettingsChange` method

core/kits/assets/js/hooks/ui/document/elements/settings/update-stretch-container.js
```diff
@@ -0,0 +1,33 @@
+/**
```
kobizz, line 1: As I mentioned above, it should be handled generally

core/kits/documents/tabs/tab-base.php
```diff
@@ -0,0 +1,76 @@
+<?php
```
kobizz, line 1: Instead of duplicating a bunch of methods, why not call `$this->document->add_control`?

kobizz (changes requested):

assets/dev/js/frontend/handlers/section/stretched-section.js
matipojo: The frontend handlers shouldn't know about the editor at all. it's a temporary handling, until we will create a Preview Handler solution.

core/kits/documents/tabs/tab-base.php
matipojo: Because our standard is to call only `$this->add_control` in order to easier move controls between classes. this class will be a base also for Skin_Base.

assets/dev/js/frontend/handlers/section/stretched-section.js
kobizz: But you still calling the editor and listening to its events. Therefore it should be handled.

assets/dev/js/frontend/handlers/section/stretched-section.js
```diff
@@ -41,6 +50,10 @@ export default class StretchedSection extends elementorModules.frontend.handlers
 
 		super.onInit( ...args );
 
+		if ( elementorFrontend.isEditMode() ) {
+			elementor.channels.editor.on( 'kit:change:stretchContainer', this.onKitChangeStretchContainerChange.bind( this ) );
```
kobizz, line 54: Why is there a duplicate binding?

kobizz (changes requested):

core/kits/documents/tabs/tab-base.php
kobizz: If so, this class has to be located in another place

includes/autoloader.php
```diff
@@ -92,6 +92,7 @@ private static function init_classes_map() {
 			'Conditions' => 'includes/conditions.php',
 			'Controls_Manager' => 'includes/managers/controls.php',
 			'Controls_Stack' => 'includes/base/controls-stack.php',
+			'Sub_Controls_Stack' => 'includes/base/sub-controls-stack.php',
```
kobizz, line 95: Our goal is to get rid of this list and automatically load everything by NS. Therefore, this class shouldn't be on this list and has to be loaded from 'core/base' or something like this

kobizz (changes requested):

includes/autoloader.php
matipojo: It should be near the control stack itself.

kobizz (approved):

KingYes (approved):

PR #11281, opened 2020-05-03
WIP: JS Data API: $e.data + globals endpoint.

iNewLegend:


data/editor/document/processors/add-default-globals.php
```diff
@@ -0,0 +1,46 @@
...
+	private function add_globals_recursive( &$elements ) {
+		foreach ( $elements as $element_key => &$element_model ) {
+			$element_instance = Plugin::$instance->elements_manager->create_element_instance( $element_model );
+			$element_controls = $element_instance->get_controls();
+
+			foreach ( $element_controls as $control_name => $control ) {
+				if ( ! empty( $control['globals'] ) ) {
+					$element_settings = & $elements[ $element_key ]['settings'];
+
+					// Add default globals only when __globals__ does not exist.
```
matipojo, line 32: Explain that if the `__globals__`  is exist but empty, it's means that the globals is disabled even though the default is global.

matipojo (changes requested):

core/common/assets/js/api/core/data/cache.js
```diff
@@ -0,0 +1,187 @@
...
+	 * @return {boolean}
+	 */
+	delete( endpoint ) {
+		let result = false;
+
+		Object.keys( this.storage.getAll() ).forEach( ( endpointKey ) => {
+			if ( endpointKey === endpoint ) {
+				this.storage.removeItem( endpoint );
+				result = true;
+			} else if ( endpointKey.includes( endpoint + '/' ) ) {
```
iNewLegend, line 178: Use '/*' to determine that next parts need to be deleted.

data/base/controller.php
```diff
@@ -0,0 +1,266 @@
...
+		if ( $command ) {
+			$format = $command . '/' . $format;
+		} else {
+			$format = $format . $command;
+		}
+
+		$this->register_endpoint_format( $command, $format );
+	}
+
+	public function register_endpoint_format( $command, $format ) {
```
iNewLegend, line 219: Add PHPDOC.

PR #11445, opened 2020-05-24
ElementBase: Fix - Validate URL before render it.

matipojo:


core/base/document.php
```diff
@@ -899,6 +899,20 @@ public function delete() {
...
+				foreach ( $element_controls as $name => $config ) {
+					if ( Controls_Manager::URL === $config['type'] && isset( $element['settings'][ $name ] ) ) {
+						$element['settings'][ $name ]['url'] = (string) wp_http_validate_url( $element['settings'][ $name ]['url'] );
+					}
+				}
+
+				return $element;
+			} );
+		}
+
```
kobizz, line 915: This might cause a huge performance leak, and of course, if you do this check you don't need the second check inside `add_link_attributes` method

kobizz (changes requested):

matipojo:
On `add_link_attributes` it still needs a check for old pages that have
already been saved.

kobizz (approved):

KingYes (approved):

PR #11489, opened 2020-05-27
Convert schemes to Globals + Temp support for Globals in Schemes mechanism

udidol:


kobizz (commented):
-  Check globals inside groups

PR #11505, opened 2020-05-31
ElementBase: Fix - Validate URL and custom attributes before render.

matipojo:


includes/base/element-base.php
```diff
@@ -450,7 +450,7 @@ public function add_link_attributes( $element, array $url_control, $overwrite =
 		$attributes = [];
 
 		if ( ! empty( $url_control['url'] ) ) {
-			$attributes['href'] = str_replace( 'javascript:', '', $url_control['url'] );
+			$attributes['href'] = (string) wp_kses_bad_protocol( $url_control['url'], [ 'http', 'https' ] );
```
kobizz, line 453: The result is already a string. Why do you convert it?

kobizz (changes requested):

KingYes (dismissed):

includes/base/element-base.php
matipojo: Fixed.

kobizz (approved):

KingYes (approved):

PR #11512, opened 2020-06-01
Tweak: Add "justify" text alignment to columns & sections

### LINKED ISSUE #11480: Add " justify" to the text alignment in the column typography
## Prerequisites

- [X] I have searched for similar features requests in both open and closed tickets and cannot find a duplicate.
- [X] The feature is still missing in the latest stable version of Elementor ( Elementor Pro. )

**What problem is your feature request going to solve? Please describe.**

I often have websites with multiple text widget in a column and I'd like to justify the text. 

**Describe the solution you'd like**

In the Section -> Column -> Style  tab -> Typography, there is an option for the text alignment. Only 3 options are offered : left | center | right
It should also include a "Justify" option.

That kind option is already present in the Text widget style tab, so it should also be offered in other places where text alignment can be defined.

**Describe alternatives you've considered**

I can add CSS in the advance panel of the Column, or I can set the Justified option for each Text widget in the Style panel.

Those option works, but the CSS isn't user-friendly, and when we have a lot of text widget, it's long to set all widget individually. And especially painful if in the future the client decides to change the alignment.

**Additional context**

ramiy:
Fixes #11480

### DIFF
--- a/includes/elements/column.php
+++ b/includes/elements/column.php
@@ -683,6 +683,10 @@ class Element_Column extends Element_Base {
 						'title' => __( 'Right', 'elementor' ),
 						'icon' => 'eicon-text-align-right',
 					],
+					'justify' => [
+						'title' => __( 'Justified', 'elementor' ),
+						'icon' => 'eicon-text-align-justify',
+					],
 				],
 				'selectors' => [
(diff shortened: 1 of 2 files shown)


rodolphebertozzo:
I guess you have to change the branch to 3.0.

ramiy:
@shilo-ey

PR #11577, opened 2020-06-09
Widgets Manager: Fix - Don't mix kit controls with widgets controls.

matipojo:


includes/managers/widgets.php
```diff
@@ -265,6 +265,10 @@ public function get_widget_types_config() {
 	public function ajax_get_widget_types_controls_config( array $data ) {
 		$config = [];
 
+		// Init kit controls in order to avoid init kit group control from within another group control.
+		// Todo: find a better solution.
+		Plugin::$instance->kits_manager->get_active_kit_for_frontend()->get_controls();
```
kobizz, line 270: Can I have a better explanation of this solution?

kobizz (changes requested):

includes/managers/widgets.php
matipojo: After moving the `default_font` to the kit, the Typography group control initialized the kit controls at: https://github.com/elementor/elementor/blob/release/3.0.0/includes/controls/groups/typography.php#L91

and because the group control is a singleton, it's args are changed to the last kit group control.

PR #11660, opened 2020-06-17
JS API/Core/Data: Fix - commandExtractArgs.

iNewLegend:


core/common/assets/js/api/core/data.js
```diff
@@ -478,9 +487,9 @@ export default class Data extends Commands {
 	run( type, command, args ) {
 		args.options.type = type;
 
-		command = this.commandExtractArgs( command, args );
+		const extractedCommand = this.commandExtractArgs( command );
```
matipojo, line 490:
```suggestion
		const { command, args } = this.commandExtractArgs( command, args );
```

core/common/assets/js/api/core/data.js
```diff
@@ -478,9 +487,9 @@ export default class Data extends Commands {
...
 
-		return super.run( command, args );
+		return super.run( extractedCommand.command, { ... args, ... extractedCommand.args } );
```
matipojo, line 492:
```suggestion
		return super.run( command, args );
```

matipojo (changes requested):

matipojo (dismissed):

core/common/assets/js/api/core/data.js
```diff
@@ -189,27 +195,30 @@ export default class Data extends Commands {
...
 	 *
 	 * @param {string} command
-	 * @param {object} argsTarget
+	 * @param {object} args
 	 *
-	 * @returns {string} command
+	 * @returns {ExtractedCommand} command
 	 */
-	commandExtractArgs( command, argsTarget ) {
+	commandExtractArgs( command, args = {} ) {
```
matipojo, line 202: Change also the PHP method.

matipojo (changes requested):

core/common/assets/js/api/core/data.js
```diff
@@ -189,27 +195,30 @@ export default class Data extends Commands {
 	 * If the command have query convert it to args.
 	 *
 	 * @param {string} command
-	 * @param {object} argsTarget
+	 * @param {object} args
 	 *
-	 * @returns {string} command
+	 * @returns {ExtractedCommand} command
```
kobizz, line 200: Is `ExtractedCommand` a type? It seems like plain object

core/common/assets/js/api/core/data.js
kobizz, line 202: Why do you need the args parameter at all? It's not related to this method purpose

kobizz (changes requested):

core/common/assets/js/api/core/data.js
iNewLegend: Known object.

core/common/assets/js/api/core/data.js
iNewLegend: It used in cases, you want merge, args.

iNewLegend: handled.

matipojo (approved):

core/common/assets/js/api/core/data.js
kobizz: As far as I know, only instances object should use this kind of typing. Could you provide me with a reference?

matipojo: @kobizz see the typedef above:
https://github.com/elementor/elementor/pull/11660/files#diff-d0a28344df7ce0852c5a7c797db34c66R21

iNewLegend: this is how jsdoc works its create 'fake' type.

kobizz (approved):

KingYes (approved):

PR #11790, opened 2020-07-05
Fix - JSCS-ESlint

iNewLegend:


matipojo (approved):

core/common/assets/js/api/modules/command-data.js
```diff
@@ -142,15 +142,15 @@ export default class CommandData extends CommandBase {
 	 * @param [args={}]
 	 * @returns {{}} filtered result
 	 */
-	applyAfterDelete( data, args = {} ) {
+	applyAfterDelete( data, args = {} ) {// eslint-disable-line no-unused-vars
```
kobizz, line 145: We should find a better solution for that, instead of `eslint-disable-line no-unused-vars` everywhere

tests/qunit/mock/ajax/setup-ajax.js
```diff
@@ -1,6 +1,6 @@
-/* global jQuery */
 import * as Ajax from './index';
 
+// eslint-disable-next-line no-unused-vars
```
kobizz, line 3: Same as above

kobizz (changes requested):

matipojo (dismissed):

core/common/assets/js/api/modules/command-data.js
iNewLegend: Maybe we can use this:
````
/* eslint-disable no-unused-vars */

// ... your code here with unused vars...

/* eslint-enable no-unused-vars */
````

when we sure what we do?

iNewLegend: By the way: its not everywhere its in base classes + tests

matipojo (approved):

PR #11854, opened 2020-07-13
Add responsive approach for slider control

arielk:


assets/dev/scss/editor/panel/controls/_slider.scss
```diff
@@ -14,6 +14,10 @@
 			display: block;
 		}
 	}
+	
+	.elementor-control-unit-2 {
```
kobizz, line 18: Instead of overwriting the `.elementor-control-unit-2` you can do either:

1. Deleting this class from slider control and applying the width on the second class (`.elementor-slider-input`).
2. Changing the global value of `.elementor-control-unit-2` if you want to apply this behavior everywhere.

kobizz (changes requested):

PR #11876, opened 2020-07-15
Data/Manager: New - Simple cache, data API can called  multiple times.

iNewLegend:


matipojo (dismissed):

data/manager.php
```diff
@@ -52,6 +57,20 @@ public function get_controllers() {
...
 	}
 
+	private function get_cache( $key ) {
+		$result = null;
+
+		if ( isset( $this->cache[ $key ] ) ) {
+			$result = $this->cache[ $key ];
+		}
+
+		return $result;
```
kobizz, line 67:
```suggestion
		return self::get_items( $this->cache, $key );
```

kobizz (changes requested):

matipojo (approved):

kobizz (approved):

PR #11880, opened 2020-07-15
Globals: Added backwards compatibility for schemes in the globals mechanism

udidol:
## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

* Globals: Added backwards compatibility for schemes in the globals mechanism

## Quality assurance

### DIFF
--- a/core/kits/manager.php
+++ b/core/kits/manager.php
@@ -1,6 +1,7 @@
 <?php
 namespace Elementor\Core\Kits;
 
+use Elementor\Controls_Stack;
 use Elementor\Core\Kits\Controls\Repeater;
 use Elementor\Core\Kits\Documents\Tabs\Colors_And_Typography;
 use Elementor\Plugin;
@@ -183,21 +184,22 @@ class Manager {
 	 * Convert Scheme to Default Global
 	 *
 	 * If a control has a scheme property, convert it to a default Global.
+	 * The $element parameter is passed for the deprecated argument function message.
 	 *
-	 * @param $scheme - Control scheme property
+	 * @param $args - Control scheme property
+	 * @param Controls_Stack $element
 	 * @return array - Control/group control args
 	 * @since 3.0.0
 	 * @access public
-	 *
 	 */
-	public function convert_scheme_to_global( $scheme ) {
-		if ( isset( $scheme['type'] ) && isset( $scheme['value'] ) ) {
-			//_deprecated_argument( $args['scheme'], '3.0.0', 'Schemes are now deprecated - use $args[\'global\'] instead.' );
-			return $this->map_scheme_to_global( $scheme['type'], $scheme['value'] );
+	public function convert_scheme_to_global( $args, Controls_Stack $element ) {
+		if ( isset( $args['scheme']['type'] ) && isset( $args['scheme']['value'] ) ) {
+			//_deprecated_argument( $args['scheme'], '3.0.0', 'Element ' . $element->get_name() . ' - Control \'' . $args['name'] . '\' - Schemes ($args[\'scheme\']) are now deprecated - use $args[\'global\'] instead.' );
+			return $this->map_scheme_to_global( $args['scheme']['type'], $args['scheme']['value'] );
 		}
 
 		// Typography control 'scheme' properties usually only include the string with the typography value ('1'-'4').
-		return $this->map_scheme_to_global( 'typography', $scheme );
+		return $this->map_scheme_to_global( 'typography', $args['scheme'] );
 	}
 
 	public function register_controls() {
--- a/core/kits/manager.php
+++ b/core/kits/manager.php
@@ -1,7 +1,6 @@
 <?php
 namespace Elementor\Core\Kits;
 
-use Elementor\Controls_Stack;
 use Elementor\Core\Kits\Controls\Repeater;
 use Elementor\Core\Kits\Documents\Tabs\Colors_And_Typography;
 use Elementor\Plugin;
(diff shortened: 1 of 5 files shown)


includes/base/controls-stack.php
```diff
@@ -707,6 +716,28 @@ public function get_stack() {
...
+	 * If a control has a scheme property, convert it to a default Global.
+	 *
+	 * @param $args - control/group control args
+	 * @return array - control/group control args
+	 * @since 3.0.0
+	 * @access public
+	 *
+	 */
+	public function convert_scheme_to_default_global( $args ) {
+		if ( isset( $args['scheme']['type'] ) && isset( $args['scheme']['value'] ) ) {
```
matipojo, line 731: Please prepare a deprecation notice with `_deprecated_argument`.

matipojo (changes requested):

includes/controls/groups/base.php
```diff
@@ -538,11 +538,15 @@ function( $key ) use ( $args ) {
...
-	private function start_popover( Controls_Stack $element ) {
+	private function start_popover( Controls_Stack $element, $user_args ) {
 		$popover_options = $this->get_options( 'popover' );
 
 		$settings = $this->get_args();
 
+		if ( isset( $user_args['scheme'] ) ) {
+			$settings['global']['default'] = $element->convert_scheme_to_default_global( $user_args );
+		}
+
```
kobizz, line 549: Do it inside the `init_args` method

kobizz (changes requested):

includes/base/controls-stack.php
udidol: Not sure I understand, this is a new method

matipojo: A developer that uses the `scheme` argument (on `add_control`) should see a notice that it's a deprecated argument.
Our standard is to prepare the notice in a comment, and after 4 versions we show it.

udidol: Do I put this in every location where $args['scheme'] is used?

matipojo: Only here, because every use of `scheme` run's this method. I suggest to pass additional arguments, the element and control names in order to add theme in the notice content.

kobizz (dismissed):

core/kits/manager.php
```diff
@@ -151,6 +151,55 @@ public function get_kit_for_frontend() {
...
+	 *
+	 * @param $scheme - Control scheme property
+	 * @return array - Control/group control args
+	 * @since 3.0.0
+	 * @access public
+	 *
+	 */
+	public function convert_scheme_to_global( $scheme ) {
+		if ( isset( $scheme['type'] ) && isset( $scheme['value'] ) ) {
+			//_deprecated_argument( $args['scheme'], '3.0.0', 'Schemes are now deprecated - use $args[\'global\'] instead.' );
```
matipojo, line 195: Add element name as an args for the deprecation notice.

matipojo, line 195: Is it added to the deprecation sheet?

includes/base/controls-stack.php
```diff
@@ -3,6 +3,7 @@
 
 use Elementor\Core\Base\Base_Object;
 use Elementor\Core\DynamicTags\Manager;
+use Elementor\Core\Kits\Documents\Tabs\Colors_And_Typography;
```
matipojo, line 6: Is it in use?

includes/controls/groups/base.php
```diff
@@ -396,6 +396,10 @@ private function init_options() {
 	 */
 	protected function init_args( $args ) {
 		$this->args = array_merge( $this->get_default_args(), $this->get_child_default_args(), $args );
+
+		if ( isset( $this->args['scheme'] ) ) {
```
matipojo, line 400: Is it needed? without it it will happen on thr `add_control`, no?

matipojo (changes requested):

includes/base/controls-stack.php
udidol: No, will remove thanks

core/kits/manager.php
udidol: There's no template for PHP Vars in the deprecation sheet. Should I add one?

udidol: Not sure I understand

matipojo: Yes, add one.

matipojo: The notice should include the element\control name: like "Heading/title_color: Schemes are now deprecated - use $args['global'] instead."

udidol: Done

includes/controls/groups/base.php
udidol: Yes, because there are group control methods that process scheme values (They now process globals) (e.g. `start_popover()`)

includes/base/controls-stack.php
```diff
@@ -370,6 +370,14 @@ public function add_control( $id, array $args, $options = [] ) {
 			'position' => null,
 		];
 
+		if ( isset( $args['scheme'] ) ) {
+			$args['global'] = [
+				'default' => Plugin::$instance->kits_manager->convert_scheme_to_global( $args['scheme'] ),
```
matipojo, line 375: Missing element arg?

matipojo (changes requested):

matipojo (dismissed):

kobizz (approved):

core/kits/manager.php
kobizz: We can't add an extra parameter for BC reasons (It's incompatible with classes that extend this method)

Additionally, it's unlikely to send an extra parameter for all these methods for better clarification only

We can discuss it later, in the version when we need to actually do this deprecation

matipojo (approved):

core/kits/manager.php
```diff
@@ -151,6 +151,54 @@ public function get_kit_for_frontend() {
...
+	/**
+	 * Map Scheme To Global
+	 *
+	 * Convert a given scheme value to its corresponding default global value
+	 *
+	 * @param string $type 'color'/'typography'
+	 * @param $value
+	 * @return mixed
+	 */
+	private function map_scheme_to_global( $type, $value ) {
```
KingYes, line 163: maybe isset here?

KingYes (changes requested):

core/kits/manager.php
kobizz: Actually there is no reason that the old scheme will point to a value that doesn't exist

KingYes (approved):

PR #11913, opened 2020-07-21
Kit: Fix - Empty logo caused a PHP error.

matipojo:


core/kits/documents/tabs/site-identity.php
```diff
@@ -21,6 +21,7 @@ public function get_title() {
 
 	public function register_tab_controls() {
 		$custom_logo_id = get_theme_mod( 'custom_logo' );
+		$custom_logo_src = wp_get_attachment_image_src( $custom_logo_id, 'full' );
```
kobizz, line 24: Is there a situation that the id exists, but the src doesn't?

kobizz (changes requested):

core/kits/documents/tabs/site-identity.php
matipojo: Yes, a delete attachment.

kobizz: In this case, you shouldn't print the id at all

kobizz: In this case, do you need to print the ID?

matipojo: yes, in order the media control will show that a value is exist.

kobizz: There are still some conflicts that need to be resolved

matipojo: Done

kobizz (approved):

PR #11935, opened 2020-07-23
JS API/Core/Data: Fix - Clear cache on start (Fix PT#816).

iNewLegend:


kobizz (approved):

shilo-ey:
Global Colors | Missing defaults values 'Secondary', 'Text' and 'Accent' in the drop-down list

PR #11967, opened 2020-07-28
Fix: Empty global values in CSS

kobizz:


core/files/css/base.php
```diff
@@ -762,6 +763,11 @@ private function get_selector_global_value( $control, $global_key ) {
 
 		if ( ! empty( $control['groupType'] ) ) {
 			$property_name = str_replace( [ $control['groupPrefix'], '_tablet', '_mobile' ], '', $control['name'] );
+
+			if ( empty( $data['value'][ Global_Typography::TYPOGRAPHY_GROUP_PREFIX . $property_name ] ) ) {
```
matipojo, line 767: It doesn't cover `multiple` controls like `font-size`.

matipojo (changes requested):

core/files/css/base.php
kobizz: The data manager should return only changed values, without defaults

matipojo: It did it. If I understand correctly you only meant to check if `isset`?

matipojo:
```suggestion
			if ( ! isset( $data['value'][ Global_Typography::TYPOGRAPHY_GROUP_PREFIX . $property_name ] ) ) {
```

matipojo (approved):

PR #11987, opened 2020-07-29
Site Settings: Show Toaster when Kit settings are saved

udidol:


kobizz (dismissed):

core/kits/assets/js/manager.js
```diff
@@ -37,7 +37,11 @@ export default class extends elementorModules.editor.utils.Module {
 			icon: 'eicon-global-settings',
 			title: elementor.translate( 'site_settings' ),
 			type: 'page',
-			callback: () => $e.route( 'panel/global/menu' ),
+			callback: () => {
+				$e.run( 'panel/global/open', {
+					route: $e.routes.getHistory( 'panel' ).reverse()[ 0 ].route,
```
matipojo, line 42: Why route to the history item?

matipojo (changes requested):

core/kits/assets/js/manager.js
udidol: @iNewLegend can explain, I did this part with his instruction

kobizz (approved):

core/kits/assets/js/manager.js
iNewLegend: @udidol, you can choose any route as you wish, maybe @matipojo means that the route to back to, can be static?

matipojo: @udidol the `route` parameter is not relevant to here, please remove it.

udidol: @matipojo @iNewLegend When I remove this parameter, the "Back to Editor" link does not bring me back to the last place I was in before entering Site Settings. With this parameter, it does redirect me back to where I came from (Editing a widget or the hamburger menu - the two places where you can access global settings).

matipojo (approved):

PR #12007, opened 2020-08-01
Add visual bootstrap grid

promocot:
Hello please add visual bootstrap and baseline grid for better transfer project from photoshop or any program in to a elementor.

Best reference is 
BOOTSTRAP: Designer powerup for elementor
BASELINE GRID: Ooohboi steroid for elementor (and have more feature dor elementor) 

for simple site it is enough that there is, however, to create complex sites you need a grid, I was personally convinced by making my site how very good grids help in the speed of work and accurate transfer of the project. I and many who have already bought a full license would like to get more ease of use from what they already have, and not buy more plugins just for the sake of one function. The grid is important, it will take you to another level of design, I can give you an example of how and why you need a grid.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 **Requires at least:** 5.0  
 **Tested up to:** 5.4  
 **Requires PHP:** 5.6  
-**Stable tag:** 2.9.12  
+**Stable tag:** 2.9.13  
 **License:** GPLv3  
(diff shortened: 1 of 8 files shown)


matipojo:
Please open an [issue](https://github.com/elementor/elementor/issues/new/choose).

PR #12056, opened 2020-08-07
Tweak: Add custom Columns Gap in section

arielk:
Screen Recording 2020-08-09 at 0 13 41(1)


rodolphebertozzo:
Can you take a look at this request for improvement? https://github.com/elementor/elementor/issues/12058

includes/elements/section.php
```diff
@@ -288,6 +288,40 @@ protected function _register_controls() {
...
+						'max' => 100,
+					],
+					'vw' => [
+						'min' => 0,
+						'max' => 100,
+					],
+				],
+				'size_units' => [ 'px', '%', 'vh', 'vw' ],
+				'selectors' => [
+					'{{WRAPPER}} .elementor-column-gap-custom > .elementor-column > .elementor-element-populated' => 'padding: {{SIZE}}{{UNIT}};',
```
kobizz, line 321: 1. If there is an inner section, it will be affected by this rule.
2. You don't need the `.elementor-column-gap-custom` selector, since this selector is active only when `custom` is selected.

kobizz (changes requested):

includes/elements/section.php
```diff
@@ -288,6 +288,40 @@ protected function _register_controls() {
 					'extended' => __( 'Extended', 'elementor' ),
 					'wide' => __( 'Wide', 'elementor' ),
 					'wider' => __( 'Wider', 'elementor' ),
+					'custom' => __( 'Custom', 'elementor' ),
+				],
+			]
+		);
+
+		$this->add_responsive_control(
+			'gap_columns-custom',
```
kobizz, line 297: A control name with a `-` (dash) character might cause unexpected behavior.

kobizz (changes requested):

includes/elements/section.php
rodolphebertozzo, line 297: $this=>add_responsive_control(

arielk: Thanks @kobizz fix: https://github.com/elementor/elementor/pull/12056/commits/ceade3b5a65b092c9cb8371ccd3820e995e68a2f

includes/elements/section.php
arielk: 1. No, it does not affect the inner section.

2. We must use this class, when we remove it, the default gap is applied

kobizz (approved):

PR #12084, opened 2020-08-10
Site Editor - Fix - ref can be passed to Card component.

Nevoss:


matipojo (approved):

core/app/assets/js/ui/card/card.js
```diff
@@ -17,3 +17,7 @@ Card.propTypes = {
 Card.defaultProps = {
 	className: '',
 };
+
+Card.displayName = 'Card';
+
+export default Card;
```
kobizz, line 23: Why do you need these?

kobizz (changes requested):

core/app/assets/js/ui/card/card.js
Nevoss: When the functional component is not using the keyword `function` to declare the function (like this case), `eslint` requires the property `displayName` on the function.
We cannot use the keyword `function` for declaration because of `React.forwardRef()` that wraps our component to create another component that can accept a `ref` as a prop.

shilo-ey: @kobizz

kobizz (approved):

PR #12095, opened 2020-08-11
Kit controls initialization fix

kobizz:


includes/controls/groups/typography.php
```diff
@@ -86,7 +88,15 @@ public static function get_type() {
...
-		$default_fonts = Plugin::$instance->kits_manager->get_current_settings( 'default_generic_fonts' );
+		$kit = Plugin::$instance->kits_manager->get_active_kit_for_frontend();
+
+		/**
+		 * Retrieve the settings directly from DB, because of an open issue when a controls group is being initialized
+		 * from within another group
+		 */
+		$kit_settings = $kit->get_meta( PageManager::META_KEY );
+
+		$default_fonts = $kit_settings['default_generic_fonts'];
```
matipojo, line 99: Default value is missing in meta. maybe add it here manually.

matipojo (changes requested):

includes/controls/groups/typography.php
kobizz: It's not missing because it's a text control

matipojo: For a new install it's missing.

matipojo (approved):

PR #12100, opened 2020-08-11
Site Editor - New - Title redirects to main page.

Nevoss:


core/app/assets/js/layout/header.js
```diff
@@ -2,16 +2,22 @@ import HeaderButtons from './header-buttons';
...
 	} );
 
 	return (
 		<Grid container alignItems="center" justify="space-between" className="eps-app__header">
-			<span className="eps-app__logo-title-wrapper">
+			<TitleTag className="eps-app__logo-title-wrapper" { ...titleAttrs }>
 				<i className="eps-app__logo eicon-elementor"/>
 				<h1 className="eps-app__title">{ props.title }</h1>
-			</span>
+			</TitleTag>
```
matipojo, line 20: You can attach the onclick to the span.

matipojo (changes requested):

core/app/assets/js/layout/header.js
Nevoss: Yeah, just want to ensure that the code is semantically correct.
I will change it if this is not Important.

matipojo: It can always be an `a` that makes the redirect, without the handler as a prop. (override current commit)

matipojo (changes requested):

core/app/assets/js/layout/header.js
```diff
@@ -6,19 +6,37 @@ export default function Header( props ) {
...
-			</span>
+			</a>
 			<HeaderButtons buttons={ props.buttons } />
 		</Grid>
 	);
 }
 
 Header.propTypes = {
 	title: PropTypes.string,
+	titleRedirectTo: PropTypes.string,
```
matipojo, line 39:
```suggestion
	titleRedirectRoute: PropTypes.string,
```

default = ''

matipojo (changes requested):

matipojo (approved):

joshmarom (approved):

kobizz (approved):

PR #12111, opened 2020-08-13
Theme Builder - Debug - New - Load debug util.

Nevoss:


matipojo (dismissed):

core/app/assets/js/index.js
```diff
@@ -1,6 +1,8 @@
 import App from './app';
 import { Module as SiteEditor } from '@elementor/site-editor';
 
+import 'elementor-editor/utils/debug';
```
kobizz, line 4: Export it to elementor-common-modules and use it from there

kobizz (changes requested):

core/common/assets/js/common.js
```diff
@@ -35,6 +36,8 @@ class ElementorCommonApp extends elementorModules.ViewModule {
 
 		this.api = new API();
 
+		this.debug = new Debug();
```
matipojo, line 39: Should be the first thing.

matipojo (changes requested):

core/common/assets/js/common.js
Nevoss: Just did it, and then realize that the debug script depends on the "ajax" module that registered in "initModules" method.

matipojo (dismissed):

assets/dev/js/editor/editor-base.js
```diff
@@ -28,7 +28,7 @@ export default class EditorBase extends Marionette.Application {
 
 	helpers = require( 'elementor-editor-utils/helpers' );
 	imagesManager = require( 'elementor-editor-utils/images-manager' ); // TODO: Unused.
-	debug = require( 'elementor-editor-utils/debug' );
+	debug = elementorCommon.debug;
```
kobizz, line 31: 1. Add deprecation to this property (use getter with deprecation notice)
2. Change all internal cases to use `elementorCommon.debug`

kobizz (changes requested):

matipojo (approved):

kobizz (approved):

PR #12112, opened 2020-08-13
Theme Builder - Fix - Last location shouldn't be saved.

Nevoss:
Remove the iframe from dom.

### DIFF
--- a/core/app/assets/js/loader/commands/close.js
+++ b/core/app/assets/js/loader/commands/close.js
@@ -6,8 +6,8 @@ export class Close extends CommandBase {
 			return false;
 		}
 
-		this.component.iframe.style.display = 'none';
-		document.body.style.overflow = '';
+		this.component.iframe.remove();
+		this.component.iframe = null;
 
 		return true;
 	}


matipojo (approved):

core/app/assets/js/loader/commands/close.js
```diff
@@ -6,8 +6,8 @@ export class Close extends CommandBase {
 			return false;
 		}
 
-		this.component.iframe.style.display = 'none';
-		document.body.style.overflow = '';
+		this.component.iframe.remove();
```
kobizz, line 9: Route to the default screen instead of having to reload the whole content

kobizz (changes requested):

core/app/assets/js/loader/commands/close.js
matipojo: Currently the editor and the app are not fully synchronized, so, we need a full reload for cases that the data has been changed by the editor.

kobizz (approved):

PR #12206, opened 2020-08-24
Utils - Change - move scroll util from pro to core

Nevoss:


Nevoss:
Depends on https://github.com/elementor/elementor/pull/12261

kobizz (approved):

PR #12186, opened 2020-08-22
Fixes #12185

ibndawood:
## PR Checklist

- [x] The commit message follows our guidelines:  https://github.com/elementor/elementor/blob/master/.github/CONTRIBUTING.md

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## Summary

This PR can be summarized in the following changelog entry:

* Moved the code calling the parent to the bottom to make sure the action is called after attributes are added.

## Description
An explanation of what is done in this PR

* I've explained the issue here: https://github.com/elementor/elementor/issues/12185 I simply moved the code calling the parent to the bottom to make sure the action in the parent fired after attributes are added.

## Quality assurance

- [x] I have tested this code to the best of my abilities
- [x] I have added unittests to verify the code works as intended
- [ ] Docs have been added / updated (for bug fixes / features)

Fixes #12185

PR #12277, opened 2020-08-27
Required wp_slash() if WordPress importer is NOT exist.

### LINKED ISSUE #11466: Native wordpress import/export does not work
Hello,

as the title says, the native wordpress import/export does not work for pages created by elementor.  Am I missing something.  Isn't elementor supposed to be compatible with the native wordpress builder?

**Steps to reproduce**
1. Create a page with elementor
2. Tools > Export the full site/content
3. Import to another domain/wordpress install using the native wordpress importer (Tools > Import)
4. Layout is broken

I searched for similar issues but could not find any help.  This has been tested on a fresh install with no plugins and the default theme.

**EDIT**
There is something wrong with the __elementor_data_ meta field.

On the export file the elementor_data field is the following:
`[{"id":"cf381fa","elType":"section","settings":{"structure":"20"},"elements": ....`

After the import on a new site it has backslashes on it:
`[{\"id\":\"cf381fa\",\"elType\":\"section\",\"settings\":{\"structure\":\"20\"},\"elements\":`

Thats the reason why it is not displaying the elementor structure of the page.  How to solve that.  Am I missing something?

maheshwaghmare:
#12186  PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

Slashes are required if WordPress importer is NOT exist.

*

## Description

Regarding the issue https://github.com/elementor/elementor/issues/11466 the fix is valid.

But, There is one condition missing which is below:

```
if( empty( $wp_importer ) ) {
	return true;
}
```

If there is WordPress importer exist then we can skip the **wp_slash()**.

But, If it is NOT exist then we need **wp_slash()** for the `_elementor_data`.

*

## Test instructions

*

## Quality assurance

- [] I have tested this code to the best of my abilities
- [] I have added unittests to verify the code works as intended
- [] Docs have been added / updated (for bug fixes / features)

Fixes #11466

### DIFF
--- a/includes/compatibility.php
+++ b/includes/compatibility.php
@@ -357,12 +357,14 @@ class Compatibility {
 	public static function is_wp_importer_before_0_7() {
 		$wp_importer = get_plugins( '/wordpress-importer' );
 
-		if ( ! empty( $wp_importer ) ) {
-			$wp_importer_version = $wp_importer['wordpress-importer.php']['Version'];
+		if( empty( $wp_importer ) ) {
+			return true;
+		}
 
-			if ( version_compare( $wp_importer_version, '0.7', '<' ) ) {
-				return true;
-			}
+		$wp_importer_version = $wp_importer['wordpress-importer.php']['Version'];
+
+		if ( version_compare( $wp_importer_version, '0.7', '<' ) ) {
+			return true;
 		}
 
 		return false;


udidol:
Hi @maheshwaghmare,

This method is only run when the `wp_import_post_meta` or the `wxr_importer.pre_process.post_meta` hooks are called. These hooks are called by the WP Importer (or extensions of it). 

The compatibility functions that use the `is_wp_importer_before_0_7()` method are there to handle compatibility issues with WP Importer, so if WP importer does not exist, they should not intervene. That is why we CANNOT return `true` if WP Importer is not found.

If you are sending this PR because you came across a problem with importing Elementor content, please feel free to open a ticket in the Issues tab with a detailed explanation of your problem.

Thank you for your interest in contributing!

maheshwaghmare:
Actually, In "[Starter Templates](https://wordpress.org/plugins/astra-sites/)" plugin we have used the extended version [WXR_Importer](https://github.com/brainstormforce/astra-sites/blob/master/inc/importers/wxr-importer/class-wxr-importer.php#L13) of class WP_Importer. 

The WXR_Importer has the same filters which are used by default WordPress Importer.

Now, In Elementor version 3.0.0 and above the filters  `wp_import_post_meta` and `wxr_importer.pre_process.post_meta` add the slashes depending on the WordPress Importer plugin.

We have also added the site import support through WP CLI. While importing the site with the WP CLI command, we face the same issue. Because In Elementor the the filter `wp_import_post_meta` and `wxr_importer.pre_process.post_meta` works for only `is_admin() || defined( 'WP_LOAD_IMPORTERS' )`. So, For WP CLI site import, we have added the same filter into our compatibility file.

Now, To add compatibility for Elementor version 3.0.0 and above, we have added the quick fix in [/astra-sites/../class-astra-sites-compatibility-elementor.php#L58](https://github.com/brainstormforce/astra-sites/blob/next-release/inc/classes/compatibility/elementor/class-astra-sites-compatibility-elementor.php#L58).

For now, We have added the fix for the Elementor version 3.0.0 and above, and it works as expected. But, If some other developers also use the same way, then this may create conflict.

udidol:
I see.
Since the manipulation we do using those hooks is only done on Elementor content, then any external plugins that deal with Elementor content would indeed have to make sure they are compatible with changes we make to Elementor, including things like imports.

If you are not signed up to our mailing list, I recommend it - at [](https://developers.elementor.com/). Any major changes made are published here and any important announcements are sent to our mailing list.

PR #12317, opened 2020-08-29
frontend.js fix:  Uncaught ReferenceError: elementorCommon is not defined

shimondoodkin:
## PR Checklist

- [x ] The commit message follows our guidelines:  https://github.com/elementor/elementor/blob/master/.github/CONTRIBUTING.md

## PR Type
What kind of change does this PR introduce?

- [ x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## Summary

frontend.js fix:  Uncaught ReferenceError: elementorCommon is not defined

*

## Description

4 months ago in frontend.js there as added a depreciation message using functions that were only available when the user was signed in and had permission 'manage_options' as admin. 

this was causing a tricky frustrating error:  while logged in the website was viewed ok but while not logged in the website was viewed incorrectly. on some themes it was working, on pojo.me themes with additional pojo plugins, it was not working.
the video in the content was working during design, but not while viewed by the end-user.

after many hours of debugging php, I found that the js variable elementorCommon in only available while logged-in and has permission 'manage_options'.

a frustrating bug, but I fixed it.  :)

*

## Test instructions
This PR can be tested by following these steps:

install fully a free pojo theme :

    # install there should be elementor plugin
    # install all suggested plugins by the theme
    # install demo content

open an incognito window and browse to the website
open the javascript inspector and see the error message.

frontend.js?ver=3.0.3:6157 Uncaught ReferenceError: elementorCommon is not defined
    at Frontend.getGeneralSettings (frontend.js?ver=3.0.3:6157)
    at Object.cacheElements (app.min.js?ver=5.5:2)
    at Object.init (app.min.js?ver=5.5:2)
    at HTMLDocument.<anonymous> (app.min.js?ver=5.5:2)
    at i (jquery.js?ver=1.12.4-wp:2)
    at Object.fireWith [as resolveWith] (jquery.js?ver=1.12.4-wp:2)
    at Function.ready (jquery.js?ver=1.12.4-wp:2)
    at HTMLDocument.J (jquery.js?ver=1.12.4-wp:2)

*

## Quality assurance

- [x ] I have tested this code to the best of my abilities
- [ ] I have added unittests to verify the code works as intended
- [ ] Docs have been added / updated (for bug fixes / features)

I have checked my changes, by manually applying them to my installed version, and it was working.

[...]

PR #12323, opened 2020-08-30
Frontend: Fix - Console error`elementorCommon` is undefined.

matipojo:


shimondoodkin:
please see i made 3 commits.
what you did is not enough.

kobizz (approved):

shimondoodkin:
the if rule is errorious it should be not if edit mode 
rather: 
if not edit mode and not preview mode and user can manage options

or just check if library is loaded, in this window or parent window


https://github.com/elementor/elementor/pull/12317/commits/a63b6436ac9d0aeae030f9fe62b6d3ee6e26e310


 because commons are not loaded in many cases:
https://github.com/elementor/elementor/blob/release/3.1.0/includes/frontend.php#L212
https://github.com/elementor/elementor/blob/release/3.1.0/includes/frontend.php#L224

matipojo:
> the if rule is errorious it should be not if edit mode
> rather:
> if not edit mode and not preview mode and user can manage options
> 
> or just check if library is loaded, in this window or parent window
> 
> [a63b643](https://github.com/elementor/elementor/commit/a63b6436ac9d0aeae030f9fe62b6d3ee6e26e310)
> 
> because commons are not loaded in many cases:
> https://github.com/elementor/elementor/blob/release/3.1.0/includes/frontend.php#L212
> https://github.com/elementor/elementor/blob/release/3.1.0/includes/frontend.php#L224

You are right we will fix it. thanks!

PR #12334, opened 2020-08-30
Rest API - Change - Extend endpoint base class.

Nevoss:


data/base/endpoint.php
```diff
@@ -285,22 +285,30 @@ public function delete_items( $request ) {
...
-		if ( isset( $args['id'] ) && $args['id'] ) {
-			$route .= '(?P<id>[\w]+)/';
+	public function register_item_route(
+		$methods = WP_REST_Server::READABLE,
+		$args = [],
+		$route = '/',
+		$should_add_id_arg = true
+	) {
+		if ( $should_add_id_arg ) {
+			$args = array_merge( [
```
iNewLegend, line 302: I think its better to check if $args is empty then add id if not use $args as it.

PR #12347, opened 2020-08-31
Data/Base/Endpoint: Fix - Unable to extends args of `register_items_r…

iNewLegend:


iNewLegend:
Handled here: https://github.com/elementor/elementor/pull/12820

PR #12379, opened 2020-09-02
App: Feature - Wrap layout with context that holds basic elements refs.

Nevoss:


matipojo:
Not relevant yet. It will be opened in the future,

PR #12415, opened 2020-09-04
Add 'em' button border options

### LINKED ISSUE #11561: No "em" Border Radius option in Button Widget Styling
I think this might be a bug, because I can't see any reason why you wouldn't enable the ability to set the button border radius by emphemeral unit ("em"). Allowing only pixels and percentage values to be set, does not allow for the border-radius of buttons to easily scale.

rodolphebertozzo:
fix: https://github.com/elementor/elementor/issues/11561


arielk (approved):

rodolphebertozzo:
@arielk, @KingYes. Perhaps also add it for the borders of the images?

PR #12423, opened 2020-09-06
Templates: Fix - duplicate Single in new template modal.

Nevoss:


matipojo (dismissed):

includes/admin-templates/new-template.php
```diff
@@ -2,6 +2,7 @@
 namespace Elementor;
 
 use Elementor\Core\Base\Document;
+use ElementorPro\Modules\ThemeBuilder\Documents\Single;
```
kobizz, line 5: You can't explicitly filter pro classes here

kobizz (changes requested):

matipojo (approved):

PR #12424, opened 2020-09-06
Kit: Fix - PHP error undefined method `add_repeater_row`. (Fix #12305) )

### LINKED ISSUE #12305: Elementor Data Updater Issue ( v3.0.3 )
## Prerequisites

## Description

The following notice is stuck since 3.0.3
Elementor Data Updater – Database update process is running in the background.

When " Click here to run it now " it then give a error: There has been a critical error on your website.

## Isolating the problem

- [X] This bug happens with a default WordPress theme active.
- [X] I've rolled back versions and re-updated and the message still visible.

## Environment

WP: 5.5
PHP: 7.4.5

matipojo:
In some cases the a kit loses his "template type" and the instance of `get_active_kit`returns a document that doesn't have the `add_repeater_row` method.

### DIFF
--- a/core/kits/manager.php
+++ b/core/kits/manager.php
@@ -21,16 +21,14 @@ class Manager {
 
 	public function get_active_id() {
 		$id = get_option( self::OPTION_ACTIVE );
-		$kit_post = null;
 
-		if ( $id ) {
-			$kit_post = get_post( $id );
-		}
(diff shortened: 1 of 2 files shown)


kobizz (approved):

rodolphebertozzo:
Hi, I saw that version 3.0.6 normally fixed this problem, except that this PR is not merged, is that normal? @shilo-ey

PR #12457, opened 2020-09-09
Kit: Fix - Globals are missing on update kit.

matipojo:
There is an hook that deletes the cache but it needs also to refresh it from server for a case the user goes to the "Theme Style" tab.

### DIFF
--- a/core/kits/assets/js/hooks/ui/document/save/save/delete-globals-cache.js
+++ b/core/kits/assets/js/hooks/ui/document/save/save/delete-globals-cache.js
@@ -16,7 +16,7 @@ export class KitDeleteGlobalsCache extends After {
 
 	apply() {
 		// After kit updates - remove globals from cache and force re-request from server.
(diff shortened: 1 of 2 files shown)


core/kits/assets/js/hooks/ui/document/save/save/delete-globals-cache.js
```diff
@@ -16,7 +16,7 @@ export class KitDeleteGlobalsCache extends After {
 
 	apply() {
 		// After kit updates - remove globals from cache and force re-request from server.
-		$e.data.deleteCache( $e.components.get( 'globals' ), 'globals/index' );
+		$e.components.get( 'globals' ).refreshGlobalData();
```
kobizz, line 19: Now the default globals rendering will unnecesserily run each time the cache is deleted

kobizz (changes requested):

kobizz (approved):

fabianenzensperger:
@KingYes Ok, now solved? When can we have the update?

PR #12494, opened 2020-09-13
Fix global color issue with divider widget

joshmarom:


includes/widgets/divider.php
```diff
@@ -971,12 +971,11 @@ private function build_svg() {
 		];
 
 		if ( 'line' !== $selected_pattern['group'] ) {
-			$attr['fill'] = $settings['color'];
+			$attr['fill'] = 'black';
 			$attr['stroke'] = 'none';
 		} else {
-			$attr['stroke'] = $settings['color'];
+			$attr['stroke'] = 'black';
```
kobizz, line 977: Do we need 'black' values?

kobizz (changes requested):

kobizz (approved):

PR #12506, opened 2020-09-14
JS API/Commands: Fix - Promise commands leave `$e.{commands}.current`…

iNewLegend:


matipojo:
It will be fixed via `commands-new-bases` PR.

PR #12553, opened 2020-09-16
Kits: HotFix - When updating blogname or blogdescription it removes the kit settings.

Nevoss:

### DIFF
--- a/core/base/document.php
+++ b/core/base/document.php
@@ -11,6 +11,7 @@ use Elementor\User;
 use Elementor\Core\Settings\Manager as SettingsManager;
 use Elementor\Utils;
 use Elementor\Widget_Base;
+use Elementor\Core\Settings\Page\Manager as PageManager;
 
 if ( ! defined( 'ABSPATH' ) ) {
 	exit; // Exit if accessed directly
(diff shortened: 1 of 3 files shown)


matipojo (dismissed):

core/base/document.php
```diff
@@ -586,6 +587,27 @@ public function save( $data ) {
...
+	 * @return $this
+	 */
+	public function update_settings( $new_settings ) {
+		$document_settings = $this->get_meta( PageManager::META_KEY );
+
+		if ( ! $document_settings ) {
+			$document_settings = [];
+		}
+
+		SettingsManager::get_settings_managers( 'page' )
```
kobizz, line 602: What will happen if the new settings include external settings (such as post title, excerpt, status, etc.)?

kobizz (changes requested):

core/base/document.php
matipojo: It will be included as current behavior.

core/base/document.php
```diff
@@ -586,6 +587,27 @@ public function save( $data ) {
 		return true;
 	}
 
+	/**
+	 * @param $new_settings
+	 *
+	 * @return $this
```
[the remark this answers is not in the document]
matipojo: Use `self`.

matipojo (dismissed):

core/base/document.php
kobizz: This function should call the `save_settings` method with the merged settings

kobizz (approved):

matipojo (approved):

PR #12600, opened 2020-09-21
Globals: Fix - When global value not found it throws an error.

Nevoss:

### DIFF
--- a/core/common/assets/js/api/modules/command-base.js
+++ b/core/common/assets/js/api/modules/command-base.js
@@ -189,6 +189,16 @@ export default class CommandBase extends ArgsObject {
 	}
 
 	/**
+	 * Run all the catch hooks.
+	 *
+	 * @param {Error} e
+	 */
+	runCatchHooks( e ) {
+		$e.hooks.runDataCatch( this.currentCommand, this.args, e );
+		$e.hooks.runUICatch( this.currentCommand, this.args, e );
+	}
+
+	/**
 	 * Function onBeforeRun.
 	 *
 	 * Called before run().
@@ -242,10 +252,8 @@ export default class CommandBase extends ArgsObject {
 	 * @param {Error} e
 	 */
 	onCatchApply( e ) {
-		$e.hooks.runDataCatch( this.currentCommand, this.args, e );
+		this.runCatchHooks( e );
 
 		elementorCommon.helpers.consoleError( e );
-
-		$e.hooks.runUICatch( this.currentCommand, this.args, e );
 	}
 }
--- a/core/common/assets/js/api/modules/command-data.js
+++ b/core/common/assets/js/api/modules/command-data.js
@@ -1,4 +1,5 @@
 import CommandBase from './command-base';
+import * as errors from './errors';
 
 export default class CommandData extends CommandBase {
 	/**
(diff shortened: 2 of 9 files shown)


core/common/assets/js/api/modules/command-base.js
```diff
@@ -188,6 +188,16 @@ export default class CommandBase extends ArgsObject {
 		return result;
 	}
 
+	/**
```
matipojo, line 191: Please add a description to the commit with details about the technical problem.

matipojo (changes requested):

matipojo (dismissed):

core/common/assets/js/api/modules/errors/base-error.js
```diff
@@ -0,0 +1,64 @@
+export default class BaseError extends Error {
+	/**
+	 * Status code of the response
+	 *
+	 * @type {number}
+	 */
+	status = 0;
```
kobizz, line 7: What is the difference between `status` and `getStatus`?

core/common/assets/js/api/modules/errors/internal-server-error.js
```diff
@@ -0,0 +1,9 @@
+import BaseError from './base-error';
```
kobizz, line 1: Why do you need this error if it doesn't change anything?

core/common/assets/js/api/modules/command-data.js
```diff
@@ -177,4 +178,21 @@ export default class CommandData extends CommandBase {
...
 		return data;
 	}
+
+	/**
+	 * Called after apply() failed.
+	 *
+	 * @param e
+	 */
+	onCatchApply( e ) {
+		const status = e?.data?.status || 0,
```
kobizz, line 188: In what situation there isn't `e`?

core/kits/assets/js/globals/global-select-behavior.js
```diff
@@ -66,6 +66,13 @@ export default class GlobalControlSelect extends Marionette.Behavior {
 				this.view.applySavedValue();
 
 				return globalData.data;
+			} ).catch( ( e ) => {
+				// TODO: Need to be replaced by "e instanceof NotFoundError"
+				if ( 404 !== e?.data?.status ) {
```
kobizz, line 71: In what situation there isn't an `e`?

kobizz (changes requested):

core/common/assets/js/api/modules/errors/base-error.js
Nevoss: You are right, will remove it.

core/common/assets/js/api/modules/errors/internal-server-error.js
Nevoss: This was an example of a normalize error that returns from the server.
for this PR this is not required, I will remove it.

core/common/assets/js/api/modules/command-data.js
Nevoss: The problem here is that "e" not always has "data" and "status".
in the current time, we do not have a consistent structure of errors that returns from the server, in the future when we will fix this issue, it can be removed (I will add a TODO).

core/kits/assets/js/globals/global-select-behavior.js
Nevoss: Same as the comment above.

matipojo (dismissed):

kobizz (dismissed):

Issue #12602, opened 2020-09-21
Image Box widget - allow image to stay on the left/right when on mobile


viveknath322:
Hi

Image Box widget is working fine on desktop but there is an issue when viewing from mobile.

If we set the image position to left or right and when viewed from mobile the image will be stacked to the top. A similar issue is happening with the icon box also.
Any workaround to keep the image to stay left even on mobile?

You can view the icon box issue here elementor/elementor#3040

viveknath322:
I used the same code provided by @gerardreches for iconbox fix in the tread https://github.com/elementor/elementor/issues/3040 . Replaced icon with images. That's it.

`@media (max-width: 767px) {
    .elementor-widget-image-box.elementor-position-left .elementor-image-box-wrapper, .elementor-widget-image-box.elementor-position-right .elementor-image-box-wrapper {
        display: flex;
    }
    .elementor-widget-image-box.elementor-position-right .elementor-image-box-wrapper{
        flex-direction: row-reverse;
    }
    .elementor-widget-image-box.elementor-position-left .elementor-image-box-image {
        margin-right: 15px !important;
    }
    .elementor-widget-image-box.elementor-position-right .elementor-image-box-image {
        margin-left: 15px !important;
    }
}

/**To add a space between image and text**/

selector.elementor-widget-image-box .elementor-image-box-content {
    margin-left: 8px;
}`

DorShahar:
Hi @viveknath322 
I'm glad you found a workaround. 

Closing this issue.

jeffrschneider:
@DorShahar You closed the issue? Having a hack isn't the same as FIXING the issue. Please re-open and _actually_ fix it.

DorShahar:
Merged issue

I merged this issue with  #3040, which is on the same subject so we could deal with it more efficiently. 
Please feel free to continue the conversation there.
Thanks for your feedback!

heyhans77:
I realise that the image box widget for elementor doesn't change for mobile when you select align left.

You can enter the CSS below in the widget, under advanced tab and custom css:

/* 700px and smaller screen sizes */
@media (max-width:768px){

	/* Image box wrapper */
	.elementor-widget-image-box .elementor-image-box-wrapper{
		display:flex;
		text-align:left;
	}
	
	/* Image box content */
	.elementor-widget-image-box .elementor-image-box-content{
		margin-left:16px;
		
	}
	
}

Hope it works!

JacobBannier:
[...]

PR #12612, opened 2020-09-21
Allow image to stay on the left/right when on mobile

### LINKED ISSUE #3040: [Improvement] Icon Box widget - allow icon to stay on the left/right when on mobile
Hi,

The current implementation of the Icon Box widget always puts the icon centered, with the title and the text below it, when on mobile. The setting "Icon Position" is not taken into account.

This is because in frontend.css there is a @media (min-width: 768px) query which switches to flex layout but only for desktop and tablet, for mobile it is block layout.

On my website I need to have the icon on the left always, even on mobile.

Could you please allow icon positioning for mobile as well?

Thank you and best regards,
Drago

rodolphebertozzo:
fix #3040, #12602

### DIFF
--- a/assets/dev/scss/frontend/widgets/icon-box.scss
+++ b/assets/dev/scss/frontend/widgets/icon-box.scss
@@ -97,4 +97,15 @@
 	.elementor-icon-box-description {
 		margin: 0;
 	}
+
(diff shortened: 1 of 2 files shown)


rodolphebertozzo:
@arielk

arielk:
@rodolphebertozzo It's not a fix, it's a custom workaround for specific cases. we need to handle with it in other way.

PR #12688, opened 2020-09-29
Added system info cli command

KingYes:


modules/wp-cli/command.php
```diff
@@ -59,6 +59,22 @@ public function flush_css( $args, $assoc_args ) {
...
+	 *
+	 *  1. wp elementor system-info
+	 *      - This will print the System Info in JSON format
+	 *
+	 * @since 3.0.11
+	 * @access public
+	 * @alias system-info
+	 */
+	public function system_info() {
+		echo wp_json_encode( \Elementor\Tracker::get_tracking_data() );
```
matipojo, line 75: Maybe add a `format` arg? that accept `raw|json`?

matipojo (changes requested):

modules/wp-cli/command.php
KingYes: We can use with this: https://make.wordpress.org/cli/handbook/references/internal-api/wp-cli-utils-format-items/

But we do not have data as table format.
Let's offer ideas?

PR #12847, opened 2020-10-16
Add dynamic for some controls

arielk:


kobizz (changes requested):

kobizz (approved):

includes/widgets/tabs.php
```diff
@@ -111,9 +114,6 @@ protected function _register_controls() {
 				'placeholder' => __( 'Tab Content', 'elementor' ),
 				'type' => Controls_Manager::WYSIWYG,
 				'show_label' => false,
-				'dynamic' => [
```
matipojo, line 114: Backwards Compatibility?

matipojo (changes requested):

includes/widgets/tabs.php
arielk: @matipojo The wysiwyg control comes with [`dynamic = true` by default](https://github.com/arielk/elementor/blob/005a21181573db73f2d188d058d10a050f8d60f4/includes/controls/wysiwyg.php#L70), so until v3.0 we did not support this option in repeater, so that why we add `dynamic = false` in the past, in this PR I remove it so we can support the dynamic option even in repeater control.

matipojo (approved):

KingYes (approved):

PR #12955, opened 2020-10-29
Core/Forms: Fix - Cannot send forms from templates when the post is empty.

Nevoss:

### DIFF
--- a/tests/phpunit/base-class.php
+++ b/tests/phpunit/base-class.php
@@ -1,10 +1,17 @@
 <?php
 namespace Elementor\Testing;
 
+use Elementor\Plugin;
 use Elementor\Testing\Traits\Base_Elementor;
 use Elementor\Testing\Traits\Extra_Assertions;
 
 class Elementor_Test_Base extends \WP_UnitTestCase {
-
 	use Base_Elementor, Extra_Assertions;
+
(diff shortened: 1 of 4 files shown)


tests/phpunit/base-class.php
```diff
@@ -1,10 +1,17 @@
...
 use Elementor\Testing\Traits\Extra_Assertions;
 
 class Elementor_Test_Base extends \WP_UnitTestCase {
-
 	use Base_Elementor, Extra_Assertions;
+
+	public function tearDown() {
+		parent::tearDown();
+
+		Plugin::$instance->documents->restore_document();
```
matipojo, line 14: It needs to reset the `switched_data` & `current_doc`. Maybe try to re-initialize the manager:
```suggestion
		Plugin::$instance->documents = new Documents_Manager();
```

matipojo (changes requested):

tests/phpunit/base-class.php
Nevoss: Try it, and some errors fail.

matipojo (approved):

kobizz (approved):

KingYes (approved):

PR #13440, opened 2020-12-30
Structure update to the 'All Parts' menu item [ED-1389]

joshmarom:


shilo-ey:
Handled by a different PR #13455

PR #13464, opened 2021-01-03
Elementor ScreenShotter: Change - add config file and change yml for package visual regression testing (ED-946)

ghost:


matipojo (changes requested):
Fix conflict

PR #13503, opened 2021-01-06
Data: New - Allow the option to abort data request.

Nevoss:
A developer can send an AbortController signal and abort the request when the request is still pending.

ref: ED-1505

### DIFF
--- a/core/common/assets/js/api/core/data.js
+++ b/core/common/assets/js/api/core/data.js
@@ -265,6 +265,7 @@ export default class Data extends Commands {
 		const type = requestData.type,
 			nonce = wpApiSettings.nonce,
 			params = {
+				signal: requestData.args?.options?.signal,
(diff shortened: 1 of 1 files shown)


matipojo (changes requested):
Fix QUnit error

matipojo (dismissed):

core/common/assets/js/api/core/data.js
```diff
@@ -265,6 +265,7 @@ export default class Data extends Commands {
 		const type = requestData.type,
 			nonce = wpApiSettings.nonce,
 			params = {
+				signal: requestData?.args?.options?.signal,
```
kobizz, line 268: Where is it used?

kobizz (changes requested):

kobizz (approved):

PR #13541, opened 2021-01-09
Elementor vs Oxygen page builder

promocot:
Hello, please optimize your HTML and CSS code, for exemple Oxygen generate small code, while the element creates a lot of extra lines, oxygen and without a caching plugin shows a high page load speed, why can't you have this, you have a huge number of solutions, you are almost perfect but there is a problem with the code, why you do not solve it ?


shilo-ey:
Thanks for the feedback,
Please follow guidelines for opening a new [feature request](https://github.com/pojome/elementor/issues/new?template=Feature_request.md), so we can help you better.

* Prerequisites.
* Description and examples.
* One request per thread

Thanks.

PR #13560, opened 2021-01-11
Elements/Views/Widget: Fix - WP 5.6 widget empty state disappears. ED-1475

iNewLegend:


matipojo (dismissed):

assets/dev/js/editor/elements/views/widget.js
```diff
@@ -228,7 +228,7 @@ WidgetView = BaseElementView.extend( {
 		// TODO: Find a better way to detect if all the images have been loaded
 		self.$el.imagesLoaded().always( function() {
 			setTimeout( function() {
-				if ( 1 > self.$el.children( '.elementor-widget-container' ).outerHeight() ) {
```
kobizz, line 231: What was changed here?

kobizz (changes requested):

assets/dev/js/editor/elements/views/widget.js
iNewLegend: Please look at ED.

matipojo: jQuery 1.12: `self.$el.children( '.elementor-widget-container' ).outerHeight() === null` 
jQuery 3.5: `self.$el.children( '.elementor-widget-container' ).outerHeight() === undefined`

kobizz: So a better check is just `if ( self.$el.children( '.elementor-widget-container' ).outerHeight()`

kobizz (changes requested):

assets/dev/js/editor/elements/views/widget.js
kobizz: So a better check is just `if ( self.$el.children( '.elementor-widget-container' ).outerHeight()`

iNewLegend: @kobizz, its not because: it will never work...
image
i think it was uses `1 >` for the situations like bigger then `1` or bigger then `null`.

matipojo (approved):

assets/dev/js/editor/elements/views/widget.js
kobizz: It will work because both `null` and `undefined` won't pass the condition

kobizz (approved):

PR #13634, opened 2021-01-21
Data: Fix - Extend base methods in base controller and endpoint.

Nevoss:
ref: ED-1840

### DIFF
--- a/data/base/controller.php
+++ b/data/base/controller.php
@@ -136,6 +136,39 @@ abstract class Controller extends WP_REST_Controller {
 	}
 
 	/**
+	 * Creates multiple items.
+	 *
+	 * @param \WP_REST_Request $request Full data about the request.
+	 *
+	 * @return \WP_Error|\WP_REST_Response Response object on success, or WP_Error object on failure.
+	 */
+	public function create_items( $request ) {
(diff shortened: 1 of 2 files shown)


iNewLegend (dismissed):

matipojo (dismissed):

matipojo (dismissed):

data/base/controller.php
```diff
@@ -135,6 +135,39 @@ public function get_items( $request ) {
...
+	 * Delete multiple items.
+	 *
+	 * @param \WP_REST_Request $request Full data about the request.
+	 *
+	 * @return \WP_Error|\WP_REST_Response Response object on success, or WP_Error object on failure.
+	 */
+	public function delete_items( $request ) {
+		return new \WP_Error( 'invalid-method', sprintf( "Method '%s' not implemented. Must be overridden in subclass.", __METHOD__ ), array( 'status' => 405 ) );
+	}
+
```
kobizz, line 170: Use abstract methods

kobizz (changes requested):

data/base/controller.php
matipojo: It's a class that extends `WP_REST_Controller`  that uses this approach for the other methods (like `create_item` etc.)
so it should work the same,

kobizz: At least use our code standard (short array syntax)

matipojo (approved):

kobizz (approved):

PR #13636, opened 2021-01-21
Responsive Panel: Add - change the widget box to responsive (ED-1504)

ghost:


matipojo:
We used the CSS solution: #13650

PR #13650, opened 2021-01-24
Editor/Panel: New - Make widgets panel responsive (ED-1504)

StyleShit:


matipojo (approved):

assets/dev/scss/editor/panel/_categories.scss
```diff
@@ -52,7 +52,12 @@
...
 		overflow: hidden;
-		padding: 0 5px;
-		display: none;
+		padding: 0 10px;
+		display: grid;
+		gap: 10px;
+		// The uppercase M in `Min` is used in order to avoid SASS conflicts with CSS `min` function.
+		// Ref: https://css-tricks.com/when-sass-and-new-css-features-collide/
+		// Because the gap is 10px, half of the total container width is ( 100% - 10px ) / 2 === ( 50% - 5px ).
+		grid-template-columns: repeat( auto-fit, minmax( Min( 135px, calc( 50% - 5px ) ), 1fr ) );
```
kobizz, line 61: Why do you use the sass function?

kobizz (changes requested):

assets/dev/scss/editor/panel/_categories.scss
StyleShit: What SASS function? I don't use any.. That's pure CSS.
As I mentioned, I used `Min()` with a capital M because the SASS compiler interprets `min()` as a SASS function, but it ignores `Min()` ( or any other variation which is not all lower-cased ), which interprets as a pure CSS function.
See the [reference to CSS Tricks](https://css-tricks.com/when-sass-and-new-css-features-collide/).

kobizz (approved):

PR #13652, opened 2021-01-24
Add google sheet integration

promocot:
Hello, please add the ability to upload data for filling out forms in a google sheet. Today this is possible only through special services or add-ons, which are mostly useless and needed only for the sake of one function. Side decision is unnecessary waste and influence of third-party code on the site. 

How to do it ? in the same way as you added api for other services


matipojo:
Please open an issue.

PR #13653, opened 2021-01-24
Add global section settings

promocot:
Add please global setting for column margin for any devices. Now i can set in section column margin default, no gap and any. but if i want to create for mobile gap from left and right 5% i must set for any section, choose devices and set this, this is the time spent on setting up, but if it were possible to establish how it is implemented in other builders it could save a lot of time


matipojo:
Please open an Issue.

PR #13729, opened 2021-01-28
🛠️ Fix: Social Icons widget compatibility issue with HTML minification done by optimization plugins [ED-1922]

joshmarom:


includes/widgets/social-icons.php
```diff
@@ -576,6 +576,7 @@ protected function render() {
 				$migrated = isset( $item['__fa4_migrated']['social_icon'] );
 				$is_new = empty( $item['social'] ) && $migration_allowed;
 				$social = '';
+				$whitespace = count( $settings['social_icon_list'] ) !== $index + 1 ? '&nbsp;' : '';
```
kobizz, line 579: Move the `count` before the loop

kobizz (changes requested):

kobizz (approved):

PR #13732, opened 2021-01-28
Navigator: Fix - Docking not working (ED-1652)

StyleShit:


assets/dev/js/editor/regions/navigator/navigator.js
```diff
@@ -11,6 +11,7 @@ export default class extends BaseRegion {
 		this.component = $e.components.register( new Component( { manager: this } ) );
 
 		this.isDocked = false;
+		this.storage.size.width = this.$el.css( 'width' );
```
matipojo, line 14: What happens if the navigator is hidden on page load?

matipojo (changes requested):

assets/dev/js/editor/regions/navigator/navigator.js
StyleShit: It works as intended because `JQuery.css()` uses `getComputedStyle()` which retrieves the width from the css rules.

kobizz, line 14: If a user has already changed the navigator width in the previous session, this fix will override it

kobizz (changes requested):

assets/dev/js/editor/regions/navigator/navigator.js
StyleShit: Fixed.

matipojo (dismissed):

kobizz (dismissed):

matipojo (approved):

PR #13760, opened 2021-01-31
🛠️ Fix: Social Icons widget compatibility issue with HTML minification done by optimization plugins [ED-1922]

KingYes:
Reverts elementor/elementor#13759


shilo-ey:
Irrelevant PR

PR #13912, opened 2021-02-15
Utils/ReplaceURL: Fix - Page settings -> Background images does not replaced. ED-1334, ED-1630

### LINKED ISSUE #10382: Elementor Pro: Custom fonts file URLS are static and are not changed when using URL-Replacer
## Prerequisites

## Description
When creating a custom font, Elementor Pro links the fonts with a static url. When changing URL it is not affected by Elementors own URL-Replacer, that for with every domain change, the links break. The only way to fix this as the moment, is to delete each font-file link and readd them, after the new URL has been defined

## Isolating the problem

- [X] This bug happens with a default WordPress theme active.
- [X] I can reproduce this bug consistently using the steps above.

iNewLegend:
Prepare to fix #10382

### DIFF
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -136,7 +136,9 @@ class Utils {
 			'WHERE `meta_key` = \'' . Manager::META_KEY . '\''
 		);
 
-		$rows_affected = false === $second_rows_affected ? false : $rows_affected + $second_rows_affected;
+		if ( $second_rows_affected ) {
+			$rows_affected += $second_rows_affected;
+		}
 
 		if ( false === $rows_affected ) {
 			throw new \Exception( __( 'An error occurred', 'elementor' ) );
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -1,6 +1,8 @@
 <?php
 namespace Elementor;
 
+use Elementor\Core\Settings\Page\Manager;
+
 if ( ! defined( 'ABSPATH' ) ) {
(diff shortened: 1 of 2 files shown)


includes/utils.php
```diff
@@ -128,10 +130,21 @@ public static function replace_urls( $from, $to ) {
 			"WHERE `meta_key` = '_elementor_data' AND `meta_value` LIKE '[%' ;" ); // meta_value LIKE '[%' are json formatted
 		// @codingStandardsIgnoreEnd
 
+		$second_rows_affected = $wpdb->query(
+			"UPDATE {$wpdb->postmeta} " .
+			$wpdb->prepare( 'SET `meta_value` = REPLACE(`meta_value`, %s, %s) ', $from, $to ) .
+			'WHERE `meta_key` = \'' . Manager::META_KEY . '\''
+		);
+
+		$rows_affected = false === $second_rows_affected ? false : $rows_affected + $second_rows_affected;
```
matipojo, line 139: The second query shouldn't break the action.
```suggestion
		if ( $second_rows_affected ) {
			$rows_affected += $second_rows_affected;
		}
```

matipojo (changes requested):
Nice Work!

includes/utils.php
iNewLegend: Handled.

matipojo (approved):

kobizz (approved):

PR #13915, opened 2021-02-15
Add `elementor-image` & `elementor-text-editor` classes to DOM optimization experiment [ED-2095]

arielk:


assets/dev/scss/frontend/widgets/image.scss
```diff
@@ -4,27 +4,16 @@
 .elementor-widget-image {
 	text-align: center;
 
-	.elementor-image {
+	a {
```
kobizz, line 7: This softer selector might allow other selectors to override it. Could you keep it like before?
```
> a,
figure > a
```

kobizz (changes requested):

includes/widgets/image.php
```diff
@@ -777,8 +779,10 @@ protected function content_template() {
 				link_url = settings.image.url;
 			}
 
-			#><div class="elementor-image{{ settings.shape ? ' elementor-image-shape-' + settings.shape : '' }}"><#
-			var imgClass = '';
+			<?php if ( ! Plugin::$instance->experiments->is_feature_active( 'e_dom_optimization' ) ) { ?>
+				#><div class="elementor-image{{ settings.shape ? ' elementor-image-shape-' + settings.shape : '' }}"><#
+				var imgClass = '';
```
kobizz, line 784: This line should be outside of the condition

kobizz (changes requested):

includes/widgets/image.php
```diff
@@ -777,8 +779,10 @@ protected function content_template() {
...
+			<?php } ?>
```
kobizz, line 785:
```suggestion
			<?php } ?>
				var imgClass = '';
```

kobizz (changes requested):

includes/widgets/image.php
arielk: Fixed: https://github.com/elementor/elementor/pull/13915/commits/c40c6d9c151cd3c73948306b667b9a0e8ef3c689

kobizz (approved):

PR #13941, opened 2021-02-17
Google Maps widget: Change - Add more height units options (ED-2112)

ghost:


matipojo (approved):

includes/widgets/google-maps.php
```diff
@@ -150,7 +150,16 @@ protected function register_controls() {
...
 						'max' => 1440,
 					],
+					'vh' => [
+						'min' => 0,
+						'max' => 100,
+					],
+					'vw' => [
+						'min' => 0,
+						'max' => 100,
+					],
```
kobizz, line 160: `vw` unit is usually irrelevance for height values. Is there a specific reason to add it here?

kobizz (changes requested):

includes/widgets/google-maps.php
matipojo: @avivu?

avivu: We can remove the VW

PR #13953, opened 2021-02-18
Editor/Panel: New - Element drop area POC (ED-2130)

StyleShit:


assets/dev/js/editor/elements/views/widget.js
```diff
@@ -234,6 +234,35 @@ WidgetView = BaseElementView.extend( {
...
+					elementor.getPreviewView().onPanelElementDragEnd();
+
+					const elementView = elementor.channels.panelElements.request( 'element:selected' );
+					const widgetType = elementView.model.get( 'widgetType' );
+					const targetId = this.getContainer().id;
+					const tabId = event.currentTarget.dataset.tab;
+
+					console.log( `Dropped: ${ widgetType }` );
+					console.log( `Target: ${ targetId }` );
+					console.log( `Inner Element: ${ tabId }` );
```
matipojo, line 261: We need the repeater item id, in order to make it generic, try to add the item id as a data attribute on the `elementor-drop-area` element.

assets/dev/js/editor/elements/views/widget.js
```diff
@@ -234,6 +234,35 @@ WidgetView = BaseElementView.extend( {
...
+					console.log( `Side: ${ side }` );
```
matipojo, line 262: After the drop, add an element with the widget type name. in order to try dragging another widget and get the index of the new item (e.g. 0 if the new element is the first)

matipojo (changes requested):

assets/dev/js/editor/elements/views/widget.js
StyleShit: You mean something like `<div>button</div>`?

PR #13621, opened 2021-01-19
Editor/Panel: New - Shape masking (ED-2137)

StyleShit:
Can be applied to any type of widget.

## Image:

image

image

## Video:

image

## Google Map:

image

## Pricing tables:

image

PR #13988, opened 2021-02-22
Editor/Panel: New - Shape masking (ED-2137)

StyleShit:
Moved to #13621


assets/dev/js/editor/controls/media.js
```diff
@@ -28,6 +28,23 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
 		return this.model.get( 'media_type' );
 	},
 
+	/**
+	 *
```
matipojo, line 32: Remove empty line.

assets/mask-shapes/arrow.svg
```diff
@@ -0,0 +1 @@
+<svg id="Layer_1" data-name="Layer 1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 127.75 81.62"><defs><style>.cls-1{fill:#1b1b1a;}</style></defs><polygon class="cls-1" points="70.68 26.54 70.68 0 35.34 20.4 0 40.81 35.34 61.21 70.68 81.61 70.68 55.08 127.75 55.08 127.75 26.54 70.68 26.54"/></svg>
```
matipojo, line 1: Do we need the `id="Layer_1" data-name="Layer 1"`? the `cls-1` CSS & class name?

includes/controls/media.php
```diff
@@ -210,6 +210,7 @@ protected function get_default_settings() {
 			'label_block' => true,
 			'media_type' => 'image',
 			'should_include_svg_inline_option' => false,
+			'library_type' => '',
```
matipojo, line 213: Why?

includes/widgets/common.php
```diff
@@ -43,6 +43,51 @@ public function show_in_panel() {
...
+	private function get_shape_url( $shape ) {
+		return ELEMENTOR_ASSETS_URL . 'mask-shapes/' . $shape . '.svg';
+	}
+
+	/**
+	 * @return array Array of shapes with their URL as key.
+	 */
+	private function get_shapes() {
+		$shapes = [
+			$this->get_shape_url( 'circle' ) => __( 'Circle', 'elementor' ),
```
matipojo, line 60: Can't use the URL as a key because the site URL may be changed. use the name instead.
```suggestion
			'circle' => __( 'Circle', 'elementor' ),
```

includes/widgets/common.php
```diff
@@ -43,6 +43,51 @@ public function show_in_panel() {
 		return false;
 	}
 
+	/**
+	 * @param $shape String Shape name.
+	 *
+	 * @return string The shape path in the assets folder.
+	 */
+	private function get_shape_url( $shape ) {
+		return ELEMENTOR_ASSETS_URL . 'mask-shapes/' . $shape . '.svg';
```
matipojo, line 52: Maybe use one JSON file for all shapes? like here https://github.com/elementor/elementor/pull/13876/files#diff-9f05d4843d392f8d215c6a90fb7717b02f363928fca1cb60419fddaa5145fe6dR251-R252

includes/widgets/common.php
```diff
@@ -364,6 +409,247 @@ protected function register_controls() {
...
+		] );
+
+		$this->add_control(
+			'_mask_image',
+			[
+				'label' => __( 'Image', 'elementor' ),
+				'type' => Controls_Manager::MEDIA,
+				'media_type' => 'image',
+				'should_include_svg_inline_option' => true,
+				'library_type' => 'image/svg+xml',
```
matipojo, line 454: The user can still choose a png file?

includes/widgets/common.php
```diff
@@ -364,6 +409,247 @@ protected function register_controls() {
...
+					'_mask_shape' => 'custom',
+				],
+			]
+		);
+
+		$this->add_control(
+			'_mask_notice',
+			[
+				'type' => Controls_Manager::RAW_HTML,
+				'raw' => __( 'Need More Shapes?', 'elementor' ) . '<br>' . sprintf( __( 'Explore additional Premium Shape packs and use them in your site. <a target="_blank" href="%s">Learn More</a>', 'elementor' ), 'https://elementor.com' ),
```
matipojo, line 470: @shilo-ey please provide the final URL.

matipojo (changes requested):

assets/mask-shapes/arrow.svg
StyleShit: Yeah, I should remove it, but the shapes will get changed anyway.

includes/controls/media.php
StyleShit: Because I've added a `library_type` support in the JS, so I set a default value here.

includes/widgets/common.php
StyleShit: Why does it matter if the URL has changed? The key will get changed as well.

includes/widgets/common.php
StyleShit: It has to return a URL in order to be able to pass it to the `mask-image`. 
If we were injecting the SVG markup into the page that would have been a good solution.

includes/widgets/common.php
StyleShit: Yes, there are 2 buttons:

image

The left one will accept PNGs

matipojo (approved):

PR #13995, opened 2021-02-22
JS API: New - Add Ability to transform hash (#) to commands. ED-2181

iNewLegend:


iNewLegend:
Handled: https://github.com/elementor/elementor/pull/14001

PR #14065, opened 2021-02-28
Document: Fix - User can send unsafe settings.

matipojo:


core/base/document.php
```diff
@@ -1375,4 +1390,58 @@ private function add_handle_revisions_changed_filter() {
...
+	 * Go over all Select & Select2 controls in order to handle also 3rd party settings.
+	 *
+	 * @param array $settings The setting to validate.
+	 * @param array $controls The controls config for the settings set.
+	 *
+	 * @return array
+	 */
+	private function validate_controls( array $settings, array $controls ) {
+		foreach ( $controls as $control_id => $control_config ) {
+			if ( Controls_Manager::SELECT !== $control_config['type'] && Controls_Manager::SELECT2  !== $control_config['type'] ) {
```
rodolphebertozzo, line 1426:
```suggestion
			if ( Controls_Manager::SELECT !== $control_config['type'] && Controls_Manager::SELECT2 !== $control_config['type'] ) {
```

rodolphebertozzo (changes requested):

core/base/document.php
```diff
@@ -1375,4 +1390,59 @@ private function add_handle_revisions_changed_filter() {
...
+	/**
+	 * Validate elements settings.
+	 *
+	 * Iterate all elements via the `validate_controls` method.
+	 *
+	 * @param array $elements
+	 *
+	 * @return array
+	 */
+	private function validate_elements_settings( array $elements ) {
```
kobizz, line 1403: Please make it a general loop so we can use it for other purposes (such as saving assets used by the page)

core/base/document.php
```diff
@@ -1375,4 +1390,59 @@ private function add_handle_revisions_changed_filter() {
...
+	 *
+	 * Avoid change settings like `html_tag` & `title_size` to a `script` tag.
+	 * Go over all Select & Select2 controls in order to handle also 3rd party settings.
+	 *
+	 * @param array $settings The setting to validate.
+	 * @param array $controls The controls config for the settings set.
+	 *
+	 * @return array
+	 */
+	private function validate_controls( array $settings, array $controls ) {
```
kobizz, line 1424: It's better to add a method inside the control class for validating saved values

kobizz (changes requested):

kobizz (approved):

Issue #14098, opened 2021-03-03
3.1.2 update breaks the templates functionality


azkalex:
## Prerequisites

- [ x] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [x ] The issue still exists against the latest stable version of Elementor.

## Description

The last update on the free version (3.1.2) breaks the templates functionality. They can be seen in the editor but not on the actual pages.

## Steps to reproduce

Create a new template with any widget, heading for example, then create a page, add the template widget and select the template with the heading created earlier. It will be shown in the editor but not on the actual page without any errors in the dev tools. This happens on a blank WP installation over the hello theme without any plugins so it's not a conflict.

## Isolating the problem

- [x] This bug happens with only Elementor plugin active (and Elementor Pro).
- [ ] This bug happens with a default WordPress theme active.
- [x] I can reproduce this bug consistently using the steps above.

## Environment

<details>
<summary>System Info</summary>
```
<!-- For bug reports, let us know about your system environment: Copy and paste the system info report from Elementor => System info, and paste it here or in http://pastebin.com/ -->
```
</details>

shilo-ey:
Thanks, we are aware of this issue and working on a fix

alfredo49:
same issue

Worldwebscapes:
I can replicate it here as well.

glgrazi:
I have the paid version of Elementor. Anda I also have the bug. Should I wait your update for fixing the bug in new pages ?

mcrowell:
+1

ttufox:
> I have the paid version of Elementor. Anda I also have the bug. Should I wait your update for fixing the bug in new pages ?

You can rollback to 3.1.1 if you re in production. Hope it ll solve your problem.

shilo-ey:
This issue has been resolved in Elementor v3.1.3

Feel free to update

Thanks!

gerardreches:
> This issue has been resolved in Elementor v3.1.3
> 
> Feel free to update
> 
> Thanks!

Notice that after updating you will have to set again the saved section that your template widget has to load, since values might have been erased.

anghel-icu:
I noticed that version 3.1.2 breaks the templates and I replaced with shortcodes to have a quick fix for this issue. Now I am glad that it was resolved in the last version.

PR #14100, opened 2021-03-03
ED-2276 Fix: Template widget clears control value after save & reload

matipojo:
Some plugins add the multi-select values via ajax. so the validation against the options will always fail. This validation has been removed because there is no real reason to validate the multi-select because it's not designed for print raw HTML.


rodolphebertozzo:
ref : https://github.com/elementor/elementor/issues/14098

kobizz (approved):

PR #14151, opened 2021-03-07
Helpers: New - Method, validate HTML tag against allowed list. ED-2291

iNewLegend:
ref: ED-2229


matipojo:
Already exist in the Utils: https://github.com/elementor/elementor/blob/master/includes/utils.php#L23

PR #14157, opened 2021-03-08
Upload Handler: Fix - Harden SVG upload (ED-2212) 

StyleShit:


core/files/assets/files-upload-handler.php
```diff
@@ -89,7 +89,9 @@ public function handle_upload_prefilter( $file ) {
 	}
 
 	protected function is_file_should_handled( $file ) {
-		return $this->is_elementor_media_upload() && $this->get_mime_type() === $file['type'];
+		$ext = pathinfo( $file['name'], PATHINFO_EXTENSION );
+
+		return $this->is_elementor_media_upload() && $this->get_mime_type() === $file['type'] && $this->get_file_type() === $ext;
```
matipojo, line 94: Ask @rotemee if we can remove the mime-type check.

matipojo (changes requested):
Nice catch and test!

core/files/assets/files-upload-handler.php
StyleShit: Well, It doesn't hurt...

matipojo (approved):

core/files/assets/files-upload-handler.php
```diff
@@ -45,15 +45,15 @@ final public static function is_enabled() {
...
+		 * @param bool $enabled Whether upload is enabled or not.
 		 */
 		$enabled = apply_filters( 'elementor/files/allow_unfiltered_upload', $enabled );
 
 		return $enabled;
 	}
 
 	final public function support_unfiltered_files_upload( $existing_mimes ) {
-		if ( $this->is_elementor_media_upload() ) {
+		if ( $this->is_elementor_media_upload() && $this->is_enabled() ) {
```
kobizz, line 56: Isn't it handled [here](https://github.com/elementor/elementor/blob/master/core/files/assets/files-upload-handler.php#L83)?

kobizz (changes requested):

core/files/assets/files-upload-handler.php
StyleShit: Looks like it doesn't 🤷‍♂️ 
I have no idea why, it LGTM.

kobizz (approved):

PR #14175, opened 2021-03-09
ED-2292 Fix: Solved bug that gradients did not work when using global colors

KingYes:
The CSS Parser didn't parse the global default value


matipojo:
Fixed by https://github.com/elementor/elementor/pull/14484

PR #14516, opened 2021-04-11
Fix: App/Styes: App-imports, UI glitch in the "All parts". ED-2492

iNewLegend:


Nevoss (approved):

matipojo (changes requested):
Please remove the `buttons.scss` from the auto import

Nevoss:
> Please remove the `buttons.scss` from the auto import

Why remove? is it not have to stay in the imports?

matipojo:
> > Please remove the `buttons.scss` from the auto import
> 
> Why remove? is it not have to stay in the imports?

No, the auto-import is designed for CSS that not dependent on the order.

PR #14602, opened 2021-04-19
Fix: JS API/Editor: Fix - Move 'Editor/Documents' to components folder. ED-2724

iNewLegend:


assets/dev/js/editor/components/documents/models/editor.js
```diff
@@ -0,0 +1,43 @@
+export default class Editor {
```
StyleShit, line 1: File name should be `models/editor.js` or `modules/editor.js`?

StyleShit (changes requested):

assets/dev/js/editor/components/documents/models/editor.js
iNewLegend: its model.

StyleShit (approved):

PR #14617, opened 2021-04-20
Fix: Slides widget not working (ED-2715)

StyleShit:


assets/dev/js/editor/container/container.js
```diff
@@ -409,7 +409,7 @@ export default class Container extends ArgsObject {
 	}
 
 	getGlobalDefault( controlName ) {
-		const controlGlobalArgs = this.controls[ controlName ].global;
+		const controlGlobalArgs = this.controls[ controlName ]?.global;
```
matipojo, line 412: The bug was caused by a repeater container, so please check why the control didn't exist for this type of container.

matipojo (changes requested):

PR #14619, opened 2021-04-20
Fix: SystemInfo/Reporters - User enable only if there is current user. ED-2507

iNewLegend:


modules/system-info/reporters/user.php
```diff
@@ -15,6 +15,10 @@
  */
 class User extends Base {
 
+	public function is_enabled() {
+		return (bool) count( (array) wp_get_current_user()->data );
```
matipojo, line 19: `wp_get_current_user` returns `0`  if the user isn't logged-in.
```suggestion
		return wp_get_current_user();
```

modules/system-info/reporters/user.php
```diff
@@ -15,6 +15,10 @@
  */
```
matipojo, line 15: Did It affects the usage schema?

matipojo (changes requested):

modules/system-info/reporters/user.php
iNewLegend: Handled.

modules/system-info/reporters/user.php
iNewLegend: Yes but i think it will be synced ( merged ) if it comes first then usage will trigger about and else too.

StyleShit (approved):

matipojo (changes requested):
The tests fail.

matipojo (approved):

matipojo (changes requested):
Please start from 3.4

matipojo (approved):

PR #14632, opened 2021-04-20
New: Upgrades - Recalculate elements usage each upgrade. [ED-2512]

iNewLegend:


core/upgrade/upgrades.php
```diff
@@ -3,8 +3,6 @@
 
 use Elementor\Core\Breakpoints\Manager as Breakpoints_Manager;
 use Elementor\Core\Experiments\Manager as Experiments_Manager;
-use Elementor\Core\Kits\Documents\Tabs\Settings_Layout;
-use Elementor\Core\Responsive\Responsive;
```
StyleShit, line 7: Why?
I assume that those imports weren't in use so the IDE removed them automatically, but just making sure...

iNewLegend: I don't think it's the IDE, because its looks more like my behavior deleting unused namespaces.
 Anyways I ensured how it was before me:
 
image

StyleShit: It looks grayed out, which means that's not in use, maybe you should remove it.
@matipojo  ?

StyleShit (approved):

core/upgrade/upgrades.php
matipojo: Yes. Please remove the unused `use`.

matipojo (approved):
Nice Work!

PR #14641, opened 2021-04-21
New: Add remote Elementor CI Tests (ED-2305)

StyleShit:
@KingYes 

We need to add the remote repo (`owner/repo-name`) in the secrets (`secrets.REMOTE_TESTS_REPO`).


.github/workflows/elementor-tests.yml
```diff
@@ -0,0 +1,25 @@
...
+    steps:
+      - uses: actions/checkout@master
+      - name: Block non-permitted users
+        uses: StyleShit/action-whitelist-users@master
+        with:
+          whitelist-file: '.github/workflows/whitelist.txt'
+      - name: Elementor Tests / Init
+        env:
+          TOKEN: ${{ secrets.DEVOPS_TOKEN }}
+          REMOTE_REPO: ${{ secrets.REMOTE_REPO }}
```
matipojo, line 19:
```suggestion
          REMOTE_REPO: ${{ secrets.REMOTE_TESTS_REPO }}
```

.github/workflows/elementor-tests.yml
```diff
@@ -0,0 +1,25 @@
...
+          HEAD_REPO: ${{ github.event.pull_request.head.repo.full_name || github.repository }}
```
matipojo, line 20: Explain

matipojo (changes requested):

.github/workflows/elementor-tests.yml
StyleShit: Done.

.github/workflows/elementor-tests.yml
StyleShit: Done.

.github/workflows/elementor-tests.yml
```diff
@@ -0,0 +1,25 @@
...
+
+jobs:
+  run:
+    runs-on: ubuntu-16.04
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@master
+      - name: Block non-permitted users
+        uses: StyleShit/action-whitelist-users@master
```
ronkelementor, line 13: please move to `elementor` repo and change to commit sha instead of branch/tag name

.github/workflows/elementor-tests.yml
```diff
@@ -0,0 +1,25 @@
...
+        with:
+          whitelist-file: '.github/workflows/whitelist.txt'
+      - name: Elementor Tests / Init
+        env:
+          TOKEN: ${{ secrets.DEVOPS_TOKEN }} # GitHub access token to access the remote repo.
```
ronkelementor, line 18:
```suggestion
          TOKEN: ${{ secrets.MAINTAIN_TOKEN }} # GitHub access token to access the remote repo.
```

.github/workflows/elementor-tests.yml
```diff
@@ -0,0 +1,25 @@
...
+          REMOTE_TESTS_REPO: ${{ secrets.REMOTE_TESTS_REPO }} # Remote tests repo, e.g. `owner/repo-name`.
+          HEAD_REPO: ${{ github.event.pull_request.head.repo.full_name || github.repository }} # The HEAD repo that initiated the push/PR.
```
ronkelementor, line 20: if the `github.event.pull_request.head.repo.full_name` is not defined we don't need to run the workflow

.github/workflows/elementor-tests.yml
```diff
@@ -0,0 +1,25 @@
...
+        run: |
+          curl -X POST https://api.github.com/repos/${{ env.REMOTE_TESTS_REPO }}/dispatches \
```
ronkelementor, line 22: I prefer to use GitHub Actions component instead of `curl` call

.github/workflows/elementor-tests.yml
```diff
@@ -0,0 +1,25 @@
+name: Elementor Tests / Init
+
+on: [push, pull_request]
```
ronkelementor, line 3: why do you want on `push` and `pull_request` and not only `pull_request` event?

matipojo: In case it's a direct merge like after fix merge conflicts.

.github/workflows/elementor-tests.yml
matipojo, line 13: Please use the SHA ID: StyleShit/action-whitelist-users@SHA-ID

matipojo (changes requested):

.github/workflows/elementor-tests.yml
StyleShit: Done.

.github/workflows/elementor-tests.yml
StyleShit: That's empty only if it's a push and not a PR...

.github/workflows/elementor-tests.yml
StyleShit: Done.

.github/workflows/elementor-tests.yml
StyleShit: Done.

.github/workflows/elementor-tests.yml
StyleShit: Looks like the token doesn't exist.
https://github.com/elementor/elementor/runs/2466428236?check_suite_focus=true

(Open to see the parameters, they are empty)

.github/workflows/elementor-tests.yml
ronkelementor: so you don't need to check if the event type is `pull_request `?

StyleShit: I don't need to because I know that if the value is null, it has to be a push event.

ronkelementor (approved):

matipojo (approved):

PR #14707, opened 2021-04-26
Fix: Activation bug for IDN domains (ED-886)

StyleShit:


core/utils/str.php
```diff
@@ -0,0 +1,22 @@
+<?php
+namespace Elementor\Core\Utils;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly
+}
+
+class Str {
```
matipojo, line 8:
```suggestion
class String {
```

matipojo (changes requested):
Please fix the lint errors + Add a unit test.

core/utils/str.php
StyleShit: I'm not sure that's a good idea to use a variable type as class name. 
Plus, `Str` is the class name that Laravel uses for their string util so me and @Nevoss thought that's a better fit.

matipojo (approved):

PR #14820, opened 2021-05-05
Fix: Elementor CI Tests will be triggered only on the master repository [ED-3122]

StyleShit:


matipojo (approved):

.github/workflows/elementor-tests.yml
```diff
@@ -4,6 +4,7 @@ on: [push, pull_request_target]
 
 jobs:
   run:
+    if: github.repository_owner == 'elementor'
```
ronkelementor, line 7: did you checked if it working with `pull_request_target` event?

StyleShit: Yes

ronkelementor (approved):

PR #14885, opened 2021-05-10
Tweak: PHPCS Fixes [ED-3165]

kobizz:


modules/system-info/templates/raw.php
```diff
@@ -34,15 +34,15 @@
 		$sub_indent = str_repeat( "\t", $tabs_count );
 
 		if ( $is_plugins ) {
-			echo "== {$field['label']} ==" . PHP_EOL;
+			echo "== {$field['label']} ==" . PHP_EOL; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
udidol, line 37: Shouldn't `$field['label']` be escaped?

modules/system-info/templates/raw.php
```diff
@@ -53,7 +53,7 @@
 			if ( ! empty( $label ) ) {
 				$label .= ': ';
 			}
-			echo "{$sub_indent}{$label}{$field['value']}" . PHP_EOL;
+			echo "{$sub_indent}{$label}{$field['value']}" . PHP_EOL; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
udidol, line 56: Shouldn't `$label` and `$field['value']` be escaped?

udidol (changes requested):

PR #14898, opened 2021-05-11
Experiment: Phpcs with composer [ED-3182]

Nevoss:


Nevoss:
@ronkelementor

.github/workflows/lint.yml
```diff
@@ -31,14 +31,11 @@ jobs:
         uses: actions/checkout@master
       - name: Install Dependencies
         run: |
-          composer global require "squizlabs/php_codesniffer=*"
-          composer global require wp-coding-standards/wpcs
+          composer install
```
matipojo, line 34: Can you install only the necessary packages for the linting?

matipojo (changes requested):

.github/workflows/lint.yml
Nevoss: cannot exclude packages. myabe the install should be in the build process and not in the lint proccess.

matipojo: @ronkelementor please advise.

ronkelementor: why do you want to exclude something?

Nevoss: @matipojo @ronkelementor 
I think it is ok to install all the dependencies for the lint. 
we can later cache Composer's vendor folder, for now, it takes only a few seconds to install all the packages.

matipojo (approved):

PR #14977, opened 2021-05-18
Fix: Fetching colors from links using eyedropper opens the link [ED-3178]

StyleShit:


Nevoss (approved):
Seems good to me, but it seems like something happens to the Screenshotter.

PR #14980, opened 2021-05-18
Tweak: Kit-Library - favorite kits empty state screen [ED-3131]

Nevoss:


core/app/modules/kit-library/assets/js/hooks/use-kits.js
```diff
@@ -149,6 +150,13 @@ export default function useKits( initialQueryParams = {} ) {
 		[ query.data, queryParams ]
 	);
 
+	const selectedTaxonomies = useSelectedTaxonomies( queryParams.taxonomies );
+
+	const isFilterActive = useMemo(
+		() => queryParams.search || !! selectedTaxonomies.length,
+		[ queryParams ]
+	);
+
```
StyleShit, line 159: Maybe put it in the `useSelectedTaxonomies` hook and return an object? 
 ```JS
const { selectedTaxonomies , isFilterActive } = useSelectedTaxonomies( queryParams );
````

P.S. I don't really know if they are related, so maybe you shouldn't...

File: core/app/modules/kit-library/assets/js/pages/index/index-no-results.js [hunk 2]
```diff
@@ -1,8 +1,41 @@
...
+	const navigate = useNavigate();
+	const location = useLocation();
+
+	return useMemo( () => {
+		const content = [
+			{
+				title: __( 'No favorites here yet...', 'elementor' ),
+				description: __( 'Use the heart icon to save kits that inspire you. You\'ll be able to find them here', 'elementor' ),
+				CTAText: __( 'Continue Browsing', 'elementor' ),
+				CTAAction: () => navigate( '/kit-library' ),
```
StyleShit, line 17: Maybe `ctaAction` / `ctaText`?
Since those are the only text & action, is that even necessary to use "CTA" in the variable?

core/app/modules/kit-library/assets/js/pages/index/index-no-results.js
```diff
@@ -1,8 +1,41 @@
 import { Heading, Text, Grid, Button } from '@elementor/app-ui';
+import { useMemo } from 'react';
+import { useNavigate, useLocation } from '@reach/router';
 
 import './index-no-results.scss';
 
+function useNoResultContent( isFilterActive, clearFilter ) {
```
StyleShit, line 7: IDK what's the standard here, but shouldn't it be in an external file?
( Well... you probably won't need that in other places, so it might be the right place )

StyleShit (changes requested):

core/app/modules/kit-library/assets/js/hooks/use-kits.js
Nevoss: Not related in this case, the filter can be active if the search is active or if the taxonomies filter is active.

core/app/modules/kit-library/assets/js/pages/index/index-no-results.js
Nevoss: Changed to lower case. 
and yes we should recognize that this content is for the "call to action" button and not for the content of the page.

core/app/modules/kit-library/assets/js/pages/index/index-no-results.js
Nevoss: I just create a function to make sure the component itself is clean. and the logic of the content comes from another function. and because the code itself is necessary only for the current component I put it in the same file.

Not sure what is the standard but it looks OK to me. let me know if you find anything else.

StyleShit (approved):

core/app/modules/kit-library/assets/js/app.js
```diff
@@ -26,7 +25,7 @@ export default function App() {
 				<SettingsProvider value={ elementorAppConfig[ 'kit-library' ] }>
 					<Router>
 						<Index path="/" initialQueryParams={{}}/>
-						<Index path="/favorites" initialQueryParams={{ favorite: true }} noResultComponent={IndexFavoriteNoResult}/>
+						<Index path="/favorites" initialQueryParams={{ favorite: true }}/>
```
matipojo, line 28: Separate into 2 components. from the favorites return the index with all favorites-related props.

matipojo (changes requested):

core/app/modules/kit-library/assets/js/pages/favorites/favorites.js
```diff
@@ -0,0 +1,30 @@
+import Index from '../index/index';
+import IndexNoResults from 'elementor/core/app/modules/kit-library/assets/js/pages/index/index-no-results';
```
matipojo, line 2:
```suggestion
import IndexNoResults from '../index/index-no-results';
```

core/app/modules/kit-library/assets/js/pages/favorites/favorites.js
```diff
@@ -0,0 +1,30 @@
...
+					description={ __( 'Use the heart icon to save kits that inspire you. You\'ll be able to find them here', 'elementor' ) }
+					button={ {
+						text: __( 'Continue Browsing', 'elementor' ),
+						action: () => navigate( '/kit-library' ),
+					} }
+				/>;
+			}
+
+			return defaultComponent;
+		} }
```
matipojo, line 24: Extract the component to a variable

matipojo (changes requested):

matipojo (approved):

PR #15004, opened 2021-05-19
Tweak:  Add "Recreate Kit" button in Elementor settings.  [ED-3074] 

nisan250:


Nevoss (commented):
@matipojo Seems like for now it already create a kit automatically. and the button will not show never.

core/kits/manager.php
```diff
@@ -97,7 +97,7 @@ public function create( array $kit_data = [], array $kit_meta_data = [] ) {
 		return $kit->get_id();
 	}
 
-	private function create_default() {
+	public function create_default() {
```
Nevoss, line 100: I think `create_default` should also update the active kit option.

Nevoss (changes requested):

core/kits/manager.php
nisan250: Replaced private to public/
About the suggestion I will wait with it until I will see if this PR will be merged since today the exist code always create kit automatically when deleted.

includes/settings/tools.php
```diff
@@ -60,6 +62,25 @@ public function ajax_elementor_clear_cache() {
...
+	 *
+	 * Fired by `wp_ajax_elementor_recreate_kit` action.
+	 *
+	 * @since 1.0.0
+	 * @access public
+	 */
+	public function ajax_elementor_recreate_kit() {
+		check_ajax_referer( 'elementor_recreate_kit', '_nonce' );
+
+		$id = Plugin::$instance->kits_manager->create_default();
```
matipojo, line 78: Check if there is an active kit because a user can click the "button" many times.

matipojo (changes requested):

includes/settings/tools.php
nisan250: Done.

includes/settings/tools.php
```diff
@@ -60,6 +62,30 @@ public function ajax_elementor_clear_cache() {
...
+	 *
+	 * Fired by `wp_ajax_elementor_recreate_kit` action.
+	 *
+	 * @since 1.0.0
+	 * @access public
+	 */
+	public function ajax_elementor_recreate_kit() {
+		check_ajax_referer( 'elementor_recreate_kit', '_nonce' );
+
+		$id = get_option( Manager::OPTION_ACTIVE );
```
matipojo, line 78:
```suggestion
		$id = Plugin::$instance->kits_manager->get_active_id()
```

includes/settings/tools.php
```diff
@@ -60,6 +62,30 @@ public function ajax_elementor_clear_cache() {
...
+
+		if ( ! $id ) {
+			update_option( Manager::OPTION_ACTIVE, Plugin::$instance->kits_manager->create_default() );
```
matipojo, line 81: update the option only if it's created.

includes/settings/tools.php
```diff
@@ -60,6 +62,30 @@ public function ajax_elementor_clear_cache() {
...
+
+			wp_send_json_success();
+		}
+
+		wp_send_json_error();
```
matipojo, line 86: Send error only if it's not created.

matipojo (changes requested):

assets/dev/js/admin/admin.js
```diff
@@ -139,6 +139,21 @@ import environment from '../../../../core/common/assets/js/utils/environment';
...
+
+				$thisButton.removeClass( 'success' ).addClass( 'loading' );
+
+				$.post( ajaxurl, {
+					action: 'elementor_recreate_kit',
+					_nonce: $thisButton.data( 'nonce' ),
+				} )
+					.done( function() {
+						$thisButton.removeClass( 'loading' ).addClass( 'success' );
+					} );
```
matipojo, line 154: Catch errors and display them.

matipojo (changes requested):

PR #15062, opened 2021-05-24
Fix: Widgets search not working when pasting [ED-3145]

StyleShit:


assets/dev/js/editor/regions/panel/pages/elements/views/search.js
```diff
@@ -13,12 +13,17 @@ PanelElementsSearchView = Marionette.ItemView.extend( {
 
 	events: {
 		'keyup @ui.input': 'onInputChanged',
+		'input @ui.input': 'triggerInputChange',
```
matipojo, line 16: I think this line is enough.
```suggestion
		'input @ui.input': 'onInputChanged',
```

matipojo (changes requested):

assets/dev/js/editor/regions/panel/pages/elements/views/search.js
StyleShit: Well, you're right... It didn't work for me before for some reason, probably had some caching issue or something...

matipojo (approved):

PR #15196, opened 2021-06-03
Internal: Library - Loading templates performance improvements [ED-3149]

Nevoss:


matipojo:
Moved to https://github.com/elementor/elementor/pull/15662

PR #15229, opened 2021-06-06
Tweak: Synchronous elements initialization may produce long JavaScript tasks [ED-8112]

### LINKED ISSUE #15228: Synchronous elements initialization may produce long JavaScript tasks
## Prerequisites

- [x] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [x] The issue still exists against the latest stable version of Elementor.

## Description

Elementor initializes itself at `jQuery.ready` moment. During process of its initialization it also performs initialization of all page elements, that are built using Elementor. 

In a case of large amount of elements to initialize, or in a case of long initialization process of some element - this initialization process may take quite some time to complete.

Nothing is wrong there except for one small nuance: this initialization process is **synchronous**. And since JavaScript by itself runs into main thread - long synchronous task blocks whole main thread, effectively making browser less responsive for user interactions. Another visible consequence is lower score from Google's [PageSpeed](https://developers.google.com/speed/pagespeed/insights/) which tracks [main thread load](https://web.dev/mainthread-work-breakdown/) and treats significant load as severe negative factor. This is especially important in a case of mobile devices.

At the same time there is no any practical reason to have this initialization process to be synchronous because it performs initialization of a set of independent elements.

Digging into Elementor code shows that [`runElementsHandlers()`](https://github.com/elementor/elementor/blob/92d09f924db0ea16bee595d446f41c6f10949dbf/assets/dev/js/frontend/document.js#L40-L42) method is the best candidate for a place to break this solid synchronous process into smaller pieces.

There is a very simple (one line) solution for the issue: it is just required to run every handler as separate JavaScript event cycle, which can be achieved by wrapping `runReadyTrigger()` call in `setTimeout()`:

```js
runElementsHandlers() {
	this.elements.$elements.each( ( index, element ) => setTimeout( () => elementorFrontend.elementsHandler.runReadyTrigger( element ) ) );
```

## Steps to reproduce

1. Put multiple (10+) elements of different types that will require JavaScript initialization on the page
2. Open page in browser. Chrome is used by me for test purposes because it closely resembles PageSpeed test
3. Open "Performance" tab in Chrome DevTools
4. Measure page loading performance by running performance information gathering
5. Take a look at "Main" line at a time of `DCL` event

## Isolating the problem

- [x] This bug happens with only Elementor plugin active (and Elementor Pro).
- [x] This bug happens with a default WordPress theme active.
- [x] I can reproduce this bug consistently using the steps above.

## Example

Here is and example of performance report from actual site of one of ours client:
image

As can be seen - initialization process took about 400ms on my notebook and main thread remains blocked for almost all time.

## Test of proposed solution

Here is example of performance report from same site with proposed patch applied:
image

As can be seen - it looks much better now with main thread blocking to be much less. They're caused by long initialization of elements which are caused by third-party plugin. It, of course, also improves PageSpeed rank.

FlyingDR:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

This PR fixes issue, described in #15228 which may cause browser's main thread to be blocked for some time in a case of long Elementor's initialization process.

* Improve browser responsiveness during Elementor's initialization process

## Description

This PR provides fix, explained into #15228 which enforces initialization of every page element to be done into separate main thread event cycle which effectively breaks code into smaller chunks, reducing chance for long JavaScript tasks to be run.

## Test instructions

1. Put multiple (10+) elements of different types that will require JavaScript initialization on the page
2. Open page in browser which allows capturing performance metrics
3. Open "Performance" tab in browser's DevTools
4. Measure page loading performance by running performance information gathering
5. Take a look at "Main" line at a time of DCL event, ensure that there is no, or much less blocking time of browser's main thread.

## Quality assurance

Fixes #15228


Camilo517:
fixed?

KingYes (approved):

PR #15233, opened 2021-06-07
Tweak: JS API/Hooks  - Make data hooks runnable when history is off. [ED-3562]

iNewLegend:
__NOTE__: It will decrease the performance.


matipojo:
Use UI Hook

PR #15239, opened 2021-06-07
Fix: Kit-Library - Page doc type [ED-3572]

Nevoss:


core/app/modules/kit-library/assets/js/hooks/use-content-types.js
```diff
@@ -21,7 +21,7 @@ function fetchContentTypes() {
 		{
 			id: 'page',
 			label: __( 'Pages', 'elementor' ),
-			doc_types: [ 'wp-page' ],
+			doc_types: [ 'page' ],
```
matipojo, line 24: Use the post type instead of the doc type.

matipojo (changes requested):

core/app/modules/kit-library/assets/js/hooks/use-content-types.js
Nevoss: Done

matipojo (approved):

PR #15261, opened 2021-06-08
Fix: Inline editing not working when the Optimized DOM experiment is on [ED-3081]

StyleShit:


includes/widgets/text-editor.php
```diff
@@ -408,21 +408,22 @@ protected function register_controls() {
...
 	 */
 	protected function render() {
-		$editor_content = $this->get_settings_for_display( 'editor' );
+		$is_edit_mode = Plugin::$instance->editor->is_edit_mode();
 
+		$editor_content = $this->get_settings_for_display( 'editor' );
 		$editor_content = $this->parse_text_editor( $editor_content );
 
-		if ( ! Plugin::$instance->experiments->is_feature_active( 'e_dom_optimization' ) ) {
+		if ( $is_edit_mode ) {
```
matipojo, line 416: Where do you check if the Experiment is active? (PHP & JS)

matipojo (changes requested):

includes/widgets/text-editor.php
StyleShit: I don't.
After thinking about it, I've realized that it doesn't matter, we want this div only in edit mode, and we don't care about the DOM experiment.

matipojo: I can't revert it, please roll it back. it's for BC.

StyleShit: Why? Looks like it should work for BC anyway.

matipojo: If the experiment is inactive the div should be printed.

StyleShit: Done.

matipojo (approved):

PR #15277, opened 2021-06-09
New: PHP Lint Util functions [ED-3622]

StyleShit:
@matipojo Added also the `esc_html`s and PHPCS comments where I thought it's necessary.

### DIFF
--- a/includes/base/widget-base.php
+++ b/includes/base/widget-base.php
@@ -680,6 +680,26 @@ abstract class Widget_Base extends Element_Base {
 	}
 
 	/**
+	 * Print a setting content without escaping.
+	 *
+	 * Script tags are allowed on frontend according to the WP theme securing policy.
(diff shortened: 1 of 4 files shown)


Nevoss (approved):

includes/base/element-base.php
```diff
@@ -676,7 +676,8 @@ public function print_element() {
 			}
 
 			$this->before_render();
-			echo $content;
+			// PHPCS - The content has already been escaped by the `render` method.
```
matipojo, line 679: Please include only the new functions

matipojo (changes requested):

includes/base/element-base.php
StyleShit: Done.

matipojo (approved):

PR #15280, opened 2021-06-09
Fix: Kit-Library - import not working [ED-3436]

Nevoss:


matipojo (approved):

core/app/modules/import-export/assets/js/pages/import/import-process/import-process.js
```diff
@@ -22,6 +22,9 @@ export default function ImportProcess() {
 
 			if ( fileURL || context.data.fileResponse ) {
 				if ( fileURL ) {
+					fileURL[ 0 ] = decodeURIComponent( fileURL[ 0 ] );
```
kobizz, line 25: Why do you need to decode also `fileURL[ 0 ]`?

Nevoss: Done

kobizz (approved):

PR #15337, opened 2021-06-14
Fix: Setting padding with relative units (%) causes incorrect padding [ED-3707]

StyleShit:


iNewLegend (approved):

Aviad-Herman (changes requested):
please re-add the div we removed

PR #15380, opened 2021-06-16
Fix: WP audio widget only shows correct styling in live preview [ED-855]

nisan250:


StyleShit (approved):
Fix the linting issues

PR #15420, opened 2021-06-20
Fix: Recreate kit button is not showing when deleting default kit [ED-3843]

Nevoss:


matipojo (approved):

includes/settings/tools.php
```diff
@@ -362,7 +362,7 @@ protected function create_tabs() {
 			],
 		];
 
-		if ( ! Plugin::$instance->kits_manager->get_active_kit() ) {
+		if ( ! Plugin::$instance->kits_manager->get_active_kit()->get_id() ) {
```
matipojo, line 365: Tests?

matipojo (changes requested):

includes/settings/tools.php
Nevoss: Done

matipojo (approved):

PR #15421, opened 2021-06-20
Fix: Lint issue in create new template screen. [ED-3802]

iNewLegend:


includes/admin-templates/new-template.php
```diff
@@ -38,7 +38,7 @@
 ?>
 <script type="text/template" id="tmpl-elementor-new-template">
 	<div id="elementor-new-template__description">
-		<div id="elementor-new-template__description__title"><?php echo esc_html__( 'Templates Help You <span>Work Efficiently</span>', 'elementor' ); ?></div>
+		<div id="elementor-new-template__description__title"><?php Utils::print_unescaped_internal_string ( __( 'Templates Help You <span>Work Efficiently</span>', 'elementor' ) ); ?></div>
```
matipojo, line 41: It should be escaped. use wp_kses_post

matipojo (changes requested):

includes/admin-templates/new-template.php
iNewLegend: Please explain why... i see hard coded internal string.

matipojo: In fact, for non-English sites - the translation is printed. and since we do not check and validate the translations they need to be escaped. Anyway, we need a solution across the code, so, I close the PR.

PR #15445, opened 2021-06-23
Fix: Promotion - The 'Go Pro' link appears even if the Pro is active [ED-3917]

StyleShit:

### DIFF
--- a/includes/managers/controls.php
+++ b/includes/managers/controls.php
@@ -944,7 +944,8 @@ class Controls_Manager {
 		// If the user has a Pro but the experiment is off.
 		// (Don't need to check if the experiment is off because if it's on, the Pro will override this section anyway).
 		if ( Utils::has_pro() ) {
-			$messages[] = __( 'This feature is currently an experiment, you can turn it on in Elementor --> Settings --> Experiments.', 'elementor' );
+			$link = sprintf( '<a href="%s" target="_blank">%s</a>', admin_url( 'admin.php?page=elementor#tab-experiments' ), __( 'Experiments', 'elementor' ) );
+			$messages[] = sprintf( __( 'This feature is currently an experiment, you can turn it on in Elementor --> Settings --> %s.', 'elementor' ), $link );
 		}
 
 		if ( $additional_messages ) {
--- a/includes/managers/controls.php
+++ b/includes/managers/controls.php
@@ -941,13 +941,6 @@ class Controls_Manager {
 			__( 'Page Transitions let you style entrance and exit animations between pages as well as display loader until your page assets load.', 'elementor' ),
 		];
 
-		// If the user has a Pro but the experiment is off.
-		// (Don't need to check if the experiment is off because if it's on, the Pro will override this section anyway).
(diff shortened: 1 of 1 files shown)


includes/managers/controls.php
```diff
@@ -941,6 +941,12 @@ public function add_page_transitions_controls( Controls_Stack $controls_stack, $
 			__( 'Page Transitions let you style entrance and exit animations between pages as well as display loader until your page assets load.', 'elementor' ),
 		];
 
+		// If the user has a Pro but the experiment is off.
```
iNewLegend, line 944: @StyleShit, just wonder is there no any other way without asking if the pro is loaded?

StyleShit: How do you check if there is Pro without checking if there is Pro? 😅

iNewLegend: Maybe the side effects of the pro

iNewLegend (approved):

includes/managers/controls.php
```diff
@@ -941,6 +941,12 @@ public function add_page_transitions_controls( Controls_Stack $controls_stack, $
...
+		// (Don't need to check if the experiment is off because if it's on, the Pro will override this section anyway).
+		if ( Utils::has_pro() ) {
+			$messages[] = __( 'This feature is currently an experiment, you can turn it on in Elementor --> Settings --> Experiments.', 'elementor' );
```
matipojo, line 947: Add a link with target=_blank. but exclude the HTML tag from the translation.
```
sprintf( __( 'This feature is currently an experiment, you can turn it on in %s' ), sprintf( '<a...</a>', __( 'Elementor' ), __( 'Settings' )...  )
```

matipojo (changes requested):

includes/managers/controls.php
StyleShit: I was actually about to do that, IDK why I didn't 😅

includes/managers/controls.php
```diff
@@ -970,7 +970,8 @@ public function get_teaser_template( $texts ) {
 				<div class="elementor-nerd-box-message"><?php echo $message; ?></div>
 			<?php }
 
-			if ( $texts['link'] ) { ?>
+			// Show a `Go Pro` button only if the user doesn't have Pro.
```
matipojo, line 973: Maybe add support for an experiment check. the method will receive the experiment name:
```
get_teaser_template( $texts, $experiment )
```
and if the experiment is off, it will show a message where to turn it on.

matipojo (changes requested):

includes/managers/controls.php
matipojo: ?

StyleShit: But what happens when the Pro is the one who registers the experiment, but the user uses an older Pro version with the new Core version?
He's gonna have a message to turn on the experiment even if he doesn't have Pro, or he has an old Pro version.

matipojo (approved):

PR #15610, opened 2021-07-11
Tweak: Document - JS API/Docs/Components, Explanation of defaultUtils. [ED-4162]

iNewLegend:


docs/core/common/assets/js/api/core/components.md
```diff
@@ -111,5 +113,30 @@ The components are extensible so a 3rd party plugin can add some routes, command
     
     export default class Component;
     ```
+* Placement of component methods, and usage of `defaultUtils`
+	* Assuming you method which used from the commands across your component *namespace*, those methods can be placed as methods at the components level, see: `createAlert`.
+	* If the method is used outside the component you should use `defaultUtils`, see at code below.
```
StyleShit, line 118:
```suggestion
	* If the method is being used outside of the component, you should use `defaultUtils`. See the example below:
```

StyleShit (changes requested):

docs/core/common/assets/js/api/core/components.md
matipojo: I didn't understand why. there are more conventional ways to separate between public/private methods.

iNewLegend: > I didn't understand why. there are more conventional ways to separate between public/private methods.

I didn't understand what you mean, the difference in placement of method in component vs defaultUtils.

You can create methods in component and use them only if they being called from the namespace scope.

If the method is used outside the namespace scope, we can use 'defaultUtils'.

If you have alternative or explanations, please suggest.

matipojo: If a utility does not belong to the component it should be written as a utilility in the utils folder.

matipojo (changes requested):

docs/core/common/assets/js/api/core/components.md
matipojo: Public utils should be exported via ES6 `export`, otherwise it should be a native method. + remove the `defaultUtils`

matipojo (changes requested):
See the comments

PR #15662, opened 2021-07-15
Tweak: Library - On open it queries all templates [ED-3149]

iNewLegend:


includes/template-library/data/endpoints/templates.php
```diff
@@ -0,0 +1,20 @@
...
+	exit; // Exit if accessed directly.
+}
+
+class Templates extends Endpoint {
+	public function get_name() {
+		return 'templates';
+	}
+
+	public function get_items( $request ) {
+		return Plugin::$instance->templates_manager->get_library_data( [ 'only_sources', $request->get_param( 'source' ) ] );
```
Nevoss, line 17: What does this string 'only sources' mean? should it be only `[$request->get_param('source')]`? or `[ 'only_sources' => $request->get_param('source') ]`?

assets/dev/js/editor/components/template-library/commands-data/templates.js
```diff
@@ -0,0 +1,9 @@
+import CommandData from 'elementor-api/modules/command-data';
+
+export class Templates extends CommandData {
+	static getEndpointFormat() {
+		return 'template-library/templates';
```
Nevoss, line 5: The command is called 'templates' and the endpoint is 'template-library'. when you remove cache it is only 'library'. why?

includes/template-library/manager.php
```diff
@@ -54,6 +56,8 @@ class Manager {
 	 * @access public
 	 */
 	public function __construct() {
+		Data_Manager::instance()->register_controller( Controller::class );
```
Nevoss, line 59: Should we use `Plugin::$instance->data_manager` or this?

includes/template-library/manager.php
```diff
@@ -200,15 +204,18 @@ public function get_source( $id ) {
...
 	 *
-	 * @return array Templates array.
+	 * @return array
 	 */
-	public function get_templates() {
+	public function get_templates( $only_sources = [] ) {
 		$templates = [];
 
 		foreach ( $this->get_registered_sources() as $source ) {
+			if ( ! empty( $only_sources ) && ! in_array( $source->get_id(), $only_sources, true ) ) {
```
Nevoss, line 215: Just a suggestion: use the Collection `filter` method to remove the sources if needed and then use the method `flatten ` which does not exist yet in our code but can be added to the collection: 
https://laravel.com/docs/8.x/collections#method-flatten
https://github.com/illuminate/collections/blob/a9a225f0378fe64204b09e5897c5950605bc8dd4/Arr.php#L211

includes/template-library/manager.php
```diff
@@ -234,7 +241,9 @@ public function get_library_data( array $args ) {
 		Plugin::$instance->documents->get_document_types();
 
 		return [
-			'templates' => $this->get_templates(),
+			'templates' => $this->get_templates(
+				isset( $args['only_sources'] ) ? $args['only_sources'] : []
```
Nevoss, line 245: maybe use default args:
```php
$args = wp_parse_args( $args, [ 'only_sources' => [] ] );

$this->get_templates( $args );
```

assets/dev/js/editor/components/template-library/manager.js
```diff
@@ -297,15 +297,28 @@ TemplateLibraryManager = function() {
...
-			},
+		const query = { source: this.getFilter( 'source' ) },
+			options = {};
+
+		// TODO: Remove - it when all the data commands is ready, manage the cache!.
+		if ( 'local' === query.source ) {
+			options.refresh = true;
+		}
+
+		$e.data.get( 'library/templates', query, options ).then( ( result ) => {
```
Nevoss, line 310: Maybe should return the `promise` and remove the `onUpdate` method. but it should be checked

Nevoss (changes requested):

assets/dev/js/editor/components/template-library/manager.js
iNewLegend: Will be out of scope.

assets/dev/js/editor/components/template-library/commands-data/templates.js
iNewLegend: Removing the whole component data, since the component name is 'library'.

includes/template-library/manager.php
iNewLegend: Would be great, but I think guess it will be out of scope. Maybe we need to create a task for that.

includes/template-library/manager.php
iNewLegend: Handled.

includes/template-library/manager.php
iNewLegend: Handled.

Nevoss (approved):

matipojo (changes requested):

assets/dev/js/editor/components/template-library/commands-data/templates.js
iNewLegend: Yes, Please see the next file in the 'Files Changed':
image

matipojo (approved):

PR #15691, opened 2021-07-18
Fix: Bug in global widgets cause in 3.3.0 core [ED-4223]

iNewLegend:
The `handleChildrenRecursive` is unable to work without a parent


matipojo (changes requested):
Please pull from develop

matipojo (approved):

PR #15692, opened 2021-07-18
Internal: Allow delete $e.data cache from server-side [ED-4270]

iNewLegend:


core/router/manager.php
```diff
@@ -0,0 +1,51 @@
...
+namespace Elementor\Core\Router;
+
+use Elementor\Core\Base\Module as BaseModule;
+use Elementor\Core\Utils\Collection;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+class Manager extends BaseModule {
```
matipojo, line 11: Please try the cache-key solution

matipojo (changes requested):

PR #15778, opened 2021-07-27
Fix: ESC button should direct to the Menu panel in the Editor [ED-4352]

StyleShit:


modules/elements-color-picker/assets/js/editor/commands/end.js
```diff
@@ -11,6 +11,8 @@ export class End extends CommandBase {
 	 * @returns {void}
 	 */
 	apply() {
+		this.component.inactivate();
```
matipojo, line 14: Explain

matipojo (changes requested):

modules/elements-color-picker/assets/js/editor/commands/end.js
StyleShit: Done.

matipojo (approved):

PR #15794, opened 2021-07-28
Fix: Browser's autocomplete is taking over the results in the Finder [ED-4427]

StyleShit:


matipojo (approved):

matipojo (commented):
Please pull from 3.4 and send a fix for the conflict

PR #15850, opened 2021-08-01
Internal: Added phpDocs to filter hooks [ED-4619]

ramiy:
## Summary

* Tweak: add phpDocs to filter hooks and enforces code convention.

## Description

This PR adds phpDocs to filter hooks and also make small code changes to enforces code convention applied on filter hooks.

Elementor filter hooks have a code convention that was not strongly enforced, this PR enforces the code convention.

```php
// Set variable initial value
$xxx = [];

// Filter the variable
$xxx = apply_filters( 'filter-name', $xxx );

// Return the variable
return $xxx;
```

In most places the filter hooks use this syntax, but many new filter hooks are not. This PR fixes them.


core/admin/admin-notices.php
```diff
@@ -40,7 +40,19 @@ private function get_notices() {
...
 		];
 
-		return apply_filters( 'elementor/core/admin/notices', $notices );
+		/**
+		 * Admin notices.
+		 *
+		 * Filters Elementor admin notices.
+		 *
+		 * This hook can be used by external developers to remove existing
+		 * admin notice or to add new notices for Elementor addons.
```
matipojo, line 49:
```suggestion
		 * This hook can be used by external developers to add new notices for Elementor addons.
```

core/files/assets/svg/svg-handler.php
```diff
@@ -336,7 +336,19 @@ private function get_allowed_attributes() {
 			'textlength',
 		];
 
-		return apply_filters( 'elementor/files/svg/allowed_attributes', $allowed_attributes );
+		/**
+		 * Allowed attributes in SVG file.
```
matipojo, line 340: Since the SVG can run JS code that may inject malicious code - all attributes are removed except the allowed attributes. please consider before adding new attributes.

matipojo (changes requested):

matipojo (approved):

PR #15941, opened 2021-08-10
Fix: Widescreen breakpoint preview didn't fit the screen [ED-4504]

guyc-E:


assets/dev/js/editor/regions/responsive-bar/view.js
```diff
@@ -122,11 +123,14 @@ export default class View extends Marionette.ItemView {
 	}
 
 	onBreakpointSelected( e ) {
-		const currentDeviceMode = elementor.channels.deviceMode.request( 'currentMode' ),
-			selectedDeviceMode = e.target.value;
+		const selectedDeviceMode = e.target.value;
+
+		elementor.changeDeviceMode( selectedDeviceMode, false );
```
guyc-E, line 128: I deleted the if statment because it's always true. The function will never be triggered when you click on the same breakpoint.

assets/dev/js/editor/regions/responsive-bar/view.js
```diff
@@ -122,11 +123,14 @@ export default class View extends Marionette.ItemView {
...
 
-		if ( currentDeviceMode !== selectedDeviceMode ) {
-			elementor.changeDeviceMode( selectedDeviceMode, false );
+		if ( 'widescreen' === selectedDeviceMode ) {
+			this.autoScale();
+		} else {
+			this.resetScale();
```
kobizz, line 133: Always do autoscale. Change the autoscale functionality to set width to 100% when the screen is not smaller than the preview

kobizz (changes requested):

PR #15950, opened 2021-08-11
Internal: Allow Plugin::$instance->breakpoints->get_active_devices_list() to return a reversed list (ED-4410)

udidol:
## PR Type

- [ ] Bugfix
- [ ] Feature
- [X] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes

## Summary

* Internal: Allow Plugin::$instance->breakpoints->get_active_devices_list() to return a reversed list

### DIFF
--- a/core/breakpoints/manager.php
+++ b/core/breakpoints/manager.php
@@ -114,9 +114,10 @@ class Manager extends Module {
 	 *
 	 * @since 3.2.0
 	 *
+	 * @param $reverse
 	 * @return array
 	 */
-	public function get_active_devices_list() {
+	public function get_active_devices_list( $reverse = null ) {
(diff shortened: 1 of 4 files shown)


core/breakpoints/manager.php
```diff
@@ -114,9 +114,10 @@ public function get_active_breakpoints( $breakpoint_name = null ) {
 	 *
 	 * @since 3.2.0
 	 *
+	 * @param $reverse
 	 * @return array
 	 */
-	public function get_active_devices_list() {
+	public function get_active_devices_list( $reverse = null ) {
```
kobizz, line 120: Make it similar to the JS implementation as an array of parameters

kobizz (changes requested):

PR #15960, opened 2021-08-12
Fix: Added Custom Code Promotion. [ED-508]

iNewLegend:


includes/settings/settings.php
```diff
@@ -642,7 +642,7 @@ private function elementor_custom_code() {
 		?>
 		<div class="wrap">
 			<div class="elementor-blank_state">
-				<img src="<?php esc_url( ELEMENTOR_ASSETS_URL . 'images/go-pro-wp-dashboard.svg' ); ?>"  alt="go-pro-wp-dashboard" />
+				<img src="<?php Utils::print_unescaped_internal_string( ELEMENTOR_ASSETS_URL ); ?>images/go-pro-wp-dashboard.svg" />
```
Nevoss, line 645: Why was `alt` removed?

Nevoss (changes requested):

includes/settings/settings.php
iNewLegend: I copied and pasted from all other 'Code promotions'

Nevoss (approved):

PR #16002, opened 2021-08-17
Fix: Gradient control doesn't work on frontend when using Global Colors (ED-3517)

### LINKED ISSUE #13288: Bug with global values in linear-gradient css function
```php
'selectors' => [
],
```
```css
```
```css
```  

Expected output would be:
```css
```

udidol:
Fixes #13288

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

* Fix: Gradient control doesn't work on frontend when using Global Colors

### DIFF
--- a/core/files/css/base.php
+++ b/core/files/css/base.php
@@ -700,11 +700,11 @@ abstract class Base extends Base_File {
 			$value = $control['selectors_dictionary'][ $value ];
 		}
 
-		if ( ! $value && ! empty( $control['global']['default'] ) ) {
-			return $this->get_selector_global_value( $control, $control['global']['default'] );
-		}
-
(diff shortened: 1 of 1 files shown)


core/files/css/base.php
```diff
@@ -700,6 +700,10 @@ private function get_style_control_value( array $control, array $values ) {
 			$value = $control['selectors_dictionary'][ $value ];
 		}
 
+		if ( ! $value && ! empty( $control['global']['default'] ) ) {
```
kobizz, line 703: Move it inside the following condition

kobizz (changes requested):

core/files/css/base.php
udidol: Done.

kobizz (approved):

PR #16029, opened 2021-08-18
Tweak: Added an option for Storke in multiple widgets [ED-3914]

guyc-E:


includes/controls/groups/text-stroke.php
```diff
@@ -0,0 +1,121 @@
...
+namespace Elementor;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+/**
+ * Elementor text stroke control.
+ *
+ * A base control for creating text stroke control. Displays input fields to define
```
udidol, line 11:
```suggestion
 * A group control for creating a stroke effect on text. Displays input fields to define
```

includes/controls/groups/text-stroke.php
```diff
@@ -0,0 +1,121 @@
...
+	/**
+	 * Fields.
+	 *
+	 * Holds all the text stroke control fields.
+	 *
+	 * @since 3.5.0
+	 * @access protected
+	 * @static
+	 *
+	 * @var array Box shadow control fields.
```
udidol, line 27:
```suggestion
	 * @var array Text Stroke control fields.
```

includes/controls/groups/text-stroke.php
```diff
@@ -0,0 +1,121 @@
...
+	 */
+	protected static $fields;
+
+	/**
+	 * Get text stroke control type.
+	 *
+	 * Retrieve the control type, in this case `box-shadow`.
```
udidol, line 34: Fix all PHPDocs

includes/controls/groups/text-stroke.php
```diff
@@ -0,0 +1,121 @@
...
+				'rem' => [
+					'min' => 0,
+					'max' => 1,
+					'step' => 0.01,
+				],
+			],
+			'selector' => '{{WRAPPER}}',
+			'selectors' => [
+				'{{SELECTOR}}' => '-webkit-text-stroke: {{SIZE}}{{UNIT}}; stroke-width: {{SIZE}}{{UNIT}}',
+			],
```
udidol, line 82: Is this necessary?
Does
```suggestion
			'selectors' => [
				'{{WRAPPER}}' => '-webkit-text-stroke: {{SIZE}}{{UNIT}}; stroke-width: {{SIZE}}{{UNIT}}',
			],
```
Not work?

modules/shapes/widgets/text-path.php
```diff
@@ -295,38 +296,14 @@ protected function register_style_tab() {
...
-						],
-					],
-				],
+			Group_Control_Text_Stroke::get_type(),
+			[
+				'name' => 'text_stroke',
+				'selector' => '{{WRAPPER}} textPath',
+				// 'selectors' => [
+				// 	'{{WRAPPER}} textPath' => 'stroke-width: {{SIZE}}{{UNIT}}; stroke: {{VALUE}};',
+				// ],
```
udidol, line 306: Should this be deleted?

udidol (changes requested):

includes/controls/groups/text-stroke.php
guyc-E: I added default for 'selector' because if the developer needs to change the selector when implementing this control, it will be easier to change only the 'selector' and not the whole CSS properties with 'selectors.'

modules/shapes/widgets/text-path.php
guyc-E: yes

udidol (approved):

Issue #16119, opened 2021-08-24
Cannot upload svg anymore - Post-processing error


brittaweller:
## Prerequisites

- [x] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [x] The issue still exists against the latest stable version of Elementor.

## Description

I can't upload svg images anymore. And yes I have enabled it in elementor settings.
When I upload any svg it will fail to upload with this error message: "Post-processing of the image failed likely because the server is busy or does not have enough resources. Uploading a smaller image may help. Suggested maximum size is 2500 pixels."

## Steps to reproduce

Theme: Twenty One or Hello Elementor / Elementor activated 
Widget Icon -> Upload Svg => Not working

----
When I use a theme which allows svg upload than it works only if I deactivate Elementor

If I downgrade to Elementor 3.3.0 it is working as expected again!

## Isolating the problem

- [x] This bug happens with only Elementor plugin active (and Elementor Pro).
- [x] This bug happens with a default WordPress theme active.
- [x] I can reproduce this bug consistently using the steps above.

## Environment

<details>
<summary>System Info</summary>
https://pastebin.com/rcR63BFU				
</details>

d19dotca:
This is a duplicate of https://github.com/elementor/elementor/issues/16084 and should be closed.

brittaweller:
> This is a duplicate of #16084 and should be closed.

Yes you are right. Sorry I didn't find this before..

shilo-ey:
duplicate of #16084

shilo-ey:
This issue has been resolved in Elementor v3.4.3

Feel free to update

Thanks!

Issue #16088, opened 2021-08-20
Cant Upload SVG Files | /elementor/core/files/assets/svg/svg-handler.php::656


maurocarreraok:
## Prerequisites

- [X] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [X] The issue still exists against the latest stable version of Elementor.

## Description

Cant upload SVG Files.

![image](https://user-images.githubusercontent.com/68023943/130236744-deb5d6de-5fe8-4f74-b048-fb1156e06040.png)

![image](https://user-images.githubusercontent.com/68023943/130236847-5849b752-9fe3-4419-aeab-b1e29473f260.png)

## Isolating the problem

- [X] This bug happens with only Elementor plugin active (and Elementor Pro).
- [X] This bug happens with a default WordPress theme active.
- [X] I can reproduce this bug consistently using the steps above.

## Environment

<details>
<summary>System Info</summary>
```

					
== Server Environment ==
	Operating System: Linux
	Software: nginx/1.20.1
	MySQL version: MariaDB Server v10.6.4
	PHP Version: 7.3.29
	PHP Max Input Vars: 3000
	PHP Max Post Size: 8M
	GD Installed: Yes
	ZIP Installed: Yes
	Write Permissions: All right
	Elementor Library: Connected

== WordPress Environment ==
	Version: 5.8
	Site URL: https://reports..com
	Home URL: https://reports..com
	WP Multisite: No
	Max Upload Size: 8 MB
	Memory limit: 2048M
	Permalink Structure: /%postname%/
	Language: en-US
	Timezone: America/New_York
	Debug Mode: Inactive

== Theme ==
	Name:  Theme
	Version: 2.0.0
	Author: 
	Child Theme: Yes
	Parent Theme Name: Hello Elementor
	Parent Theme Version: 2.4.1
	Parent Theme Author: Elementor Team

== User ==
	Role: administrator
	WP Profile lang: en_US
	User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/92.0.4515.159 Safari/537.36

== Active Plugins ==
	Crocoblock Wizard
		Version: 1.1.0
		Author: Crocoblock

	Elementor
		Version: 3.4.2
		Author: Elementor.com

	Elementor Pro
		Version: 3.3.7
		Author: Elementor.com

	JetEngine
		Version: 2.8.10
		Author: Crocoblock

[...]

Issue #16084, opened 2021-08-20
Unable to upload SVG as of 3.4.2


chrissienoodle:
Hi,

As of Elementor 3.4.2, I am unable to upload SVGs. My site uses Elementor Pro 3.3.7 and the Hello Elementor 2.4.1 theme.  The option is enabled in settings and I also have a 'upload_mimes' hook in my child theme enabling the support.

When I rollback to 3.3, I can upload SVGs.

There is a series of 500 errors in the console. One for the async upload, the rest for the image sizes.  They all indicate a WP critical error but since the site is in production and doesn't use an email, I am unable to get more debug info.

Good Luck,

shilo-ey:
Thanks for reporting.

We will address this issue soon and release a patch.

Please follow this thread for updates.

Thanks!

hogash:
Having this issue too. 

Probably just validating `$xml` should do the trick to prevent trying to access an object as bool .

shilo-ey:
This issue has been resolved in Elementor v3.4.3

Feel free to update

Thanks!

pzeylstra:
The issue is still there. Elementor 3.4.4. active, option enabled in settings.

RicBaller:
I still have this issue. Currently the upload is enabled via the Elementor settings, but I still get the question to turn it on for one time when I open the editor. 
I've tried adding the SVG support plugin, but then I can only upload via the media page and then I get the error that the image is to large.

PR #16125, opened 2021-08-24
Fix: Can't upload SVG files using Elementor (#16084, #16119, #16088) [ED-4813]

guyc-E:


core/files/assets/svg/svg-handler.php
```diff
@@ -682,8 +682,8 @@ public function set_svg_meta_data( $data, $id ) {
 			// If the svg metadata are empty or the width is empty or the height is empty.
 			// then get the attributes from xml.
 			if ( empty( $data ) || empty( $data['width'] ) || empty( $data['height'] ) ) {
-
-				$xml = simplexml_load_file( wp_get_attachment_url( $id ) );
```
kobizz, line 686: Check if we can use `get_attached_file` instead

kobizz (changes requested):

kobizz (approved):

PR #16195, opened 2021-08-29
Fix: Data Updater causes fatal error due to DB corruption [ED-4839]

Nevoss:


modules/usage/module.php
```diff
@@ -570,6 +570,11 @@ private function save_document_usage( Document $document ) {
 
 				$this->add_to_global( $document->get_name(), $usage );
 			} catch ( \Exception $exception ) {
+				Plugin::$instance->logger->get_logger()->error( $exception->getMessage(), [
+					'document' => $document->get_id(),
+					'document_name' => $document->get_name(),
```
iNewLegend, line 575:
```suggestion
					'document_id' => $document->get_id(),
					'document_name' => $document->get_name(),
```

iNewLegend (changes requested):

modules/usage/module.php
Nevoss: Done

iNewLegend (approved):

modules/usage/module.php
```diff
@@ -570,6 +570,11 @@ private function save_document_usage( Document $document ) {
 
 				$this->add_to_global( $document->get_name(), $usage );
 			} catch ( \Exception $exception ) {
+				Plugin::$instance->logger->get_logger()->error( $exception->getMessage(), [
+					'document_id' => $document->get_id(),
+					'document_name' => $document->get_name(),
+				] );
+
 				return; // Do nothing.
```
matipojo, line 578:
```suggestion
				return;
```

matipojo (changes requested):

matipojo (approved):

PR #16198, opened 2021-08-30
Tweak: Height and width fields in Responsive bar cannot be edit in Desktop [ED-4529]

guyc-E:


assets/dev/js/editor/regions/responsive-bar/view.js
```diff
@@ -168,6 +171,22 @@ export default class View extends Marionette.ItemView {
 		this.ui.sizeInputHeight.val( Math.round( size.height ) );
 	}
 
+	onPreviewOpen() {
+		this.setWidthHeightInputsEditableState();
```
kobizz, line 175: Why does it happen on preview open? It should be done on every device mode change

kobizz (changes requested):

assets/dev/js/editor/regions/responsive-bar/view.js
guyc-E: It needs to happen both on device-mode-change and also on preview-open.
Because when you click on the "responsive-mode" button from the panel, the event of "on device-mode-change" is not triggered.

PR #16200, opened 2021-08-30
Fix: Lighthouse CI test is failing - removed unused css [ED-3698]

Nevoss:


includes/frontend.php
```diff
@@ -233,6 +233,10 @@ public function init() {
 		$document = Plugin::$instance->documents->get( $this->post_id );
 
 		if ( is_singular() && $document && $document->is_built_with_elementor() ) {
+			add_action( 'wp_enqueue_scripts', function () {
+				wp_dequeue_style( 'wp-block-library' );
+			} );
```
ItzhakAvm, line 238: Why not to add a function in the class for it?

Nevoss: It is a small function that is very straightforward, I don't see a reason to create a method for it.

YitzhakAvm (approved):

matipojo (approved):

PR #16258, opened 2021-09-05
Fix: When adding margin to column of media-carousel widget it breaks the UI [ED-4915]

guyc-E:


assets/dev/js/frontend/handlers/image-carousel.js
```diff
@@ -106,6 +108,9 @@ export default class ImageCarousel extends elementorModules.frontend.handlers.Sw
 
 		const Swiper = elementorFrontend.utils.swiper;
 
+		// Fix: When adding margin to column of media-carousel widget it breaks the UI [ED-4915]
+		this.elements.$widgetWrapper.width( 'calc( 100% - ( var( --e-column-margin-left, 0px ) + var( --e-column-margin-right, 0px ) ) )' );
```
rotemee, line 112: Why do we need this solution in JS? can't we put it in the CSS?

assets/dev/js/frontend/handlers/image-carousel.js
```diff
@@ -106,6 +108,9 @@ export default class ImageCarousel extends elementorModules.frontend.handlers.Sw
...
+
```
rotemee, line 113: @guyc-E - The problem is not just with the 'Image Carousel', it also happens in the 'Media Carousel'.
Please consider using "overflow: hidden/auto' on the same div that the column margins apply to.

guyc-E: I created a new task for "Media Carousel" and "Testimonial Carousel" for Pro.

assets/dev/js/frontend/utils/swiper-bc.js
```diff
@@ -33,6 +33,9 @@ export default class SwiperBC {
 			legacySwiper();
 		}
 
+		// Fix: When adding margin to column of carousel widgets, it breaks the UI.
```
rotemee, line 36: Why the: "// Fix:"?

assets/dev/js/frontend/utils/swiper-bc.js
```diff
@@ -33,6 +33,9 @@ export default class SwiperBC {
...
+		container.closest( '.elementor-widget-wrap' ).addClass( 'e-has-image-carousel-widget' );
```
rotemee, line 37: 'e-has-image-carousel-widget' -> it's not just the image carousel, it can be for other widgets, I would call it something like: 'e-swiper-container' or at least something that is related to 'swiper'.

assets/dev/scss/frontend/_grid.scss
```diff
@@ -137,6 +137,11 @@
 	> .elementor-element {
 		width: 100%;
 	}
+
+	// Fix: When adding margin to column of carousel widgets, it breaks the UI.
```
rotemee, line 141: Why the: "// Fix:"?
+
I would elaborate on the problem a bit, it's not understood what exactly breaks.

includes/elements/column.php
```diff
@@ -729,6 +729,7 @@ protected function register_controls() {
 				'size_units' => [ 'px', 'em', '%', 'rem' ],
 				'selectors' => [
 					'{{WRAPPER}} > .elementor-element-populated' => 'margin: {{TOP}}{{UNIT}} {{RIGHT}}{{UNIT}} {{BOTTOM}}{{UNIT}} {{LEFT}}{{UNIT}};',
+					'{{WRAPPER}}' => '--e-column-margin-right: {{RIGHT}}{{UNIT}}; --e-column-margin-left: {{LEFT}}{{UNIT}};',
```
rotemee, line 732: Why using another 'selector' item? can we just concatenate the variables to the first one?

assets/dev/scss/frontend/_grid.scss
```diff
@@ -137,6 +137,11 @@
...
+	&.e-has-image-carousel-widget {
```
rotemee, line 142: See the comment above regarding the class name.

rotemee (approved):

PR #16293, opened 2021-09-09
Fix: Section handler not reachable if Scroll snap is active [ED-4926]

guyc-E:


kobizz:
@guyc-E Change the target to `develop`

PR #16319, opened 2021-09-13
New: CLI - A command to generate all missing CSS files/cache

matipojo:


core/files/manager.php
```diff
@@ -133,6 +135,36 @@ public function clear_cache() {
 		do_action( 'elementor/core/files/clear_cache' );
 	}
 
+	/**
+	 * Generate all CSS cache or files.
+	 *
+	 * @access public
+	 */
+	public function generate_all_css() {
```
kobizz, line 143: Why can't you use `Plugin::$instance->files_manager->clear_cache();`?

kobizz (changes requested):

core/files/manager.php
matipojo: It doesn't create the new files. the purpose here is for a post-process on the generated files.

Webtica:
Hi @matipojo any idea when this will be released? Waiting eagerly for this feature! Great work!

PR #16388, opened 2021-09-23
Fix: Slides per view controls disappeared in multiple breakpoints in Testimonial Carousel [ED-5025]

udidol:
This has a complementary PR in E Pro.

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

* Fix: Slides per view controls disappeared in multiple breakpoints in Testimonial Carousel

## Description

* When a responsive control is updated, the `update_responsive_control()` or `add_responsive_control()` methods don't have any way of knowing if the original control needs to be duplicated or not (such as in cases where the control is `frontend_available`, has a `prefix_class` property or is `dynamic => active => true`). This PR adds a check for that before overwriting an updated responsive control.

### DIFF
--- a/includes/base/controls-stack.php
+++ b/includes/base/controls-stack.php
@@ -812,18 +812,21 @@ abstract class Controls_Stack extends Base_Object {
 			unset( $args['devices'] );
 		}
 
-		$responsive_duplication_mode = Plugin::$instance->breakpoints->get_responsive_control_duplication_mode();
-		$additional_breakpoints_active = Plugin::$instance->experiments->is_feature_active( 'additional_custom_breakpoints' );
-		$control_is_dynamic = ! empty( $args['dynamic']['active'] );
-		$is_frontend_available = ! empty( $args['frontend_available'] );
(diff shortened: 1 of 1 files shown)


includes/base/controls-stack.php
```diff
@@ -818,6 +818,13 @@ final public function add_responsive_control( $id, array $args, $options = [] )
 		$is_frontend_available = ! empty( $args['frontend_available'] );
 		$has_prefix_class = ! empty( $args['prefix_class'] );
 
+		if ( ! empty( $options['overwrite'] ) ) {
+			// If this is an updated control, check if it needs to be duplicated or not.
+			$is_duplicated_overwrite = $this->is_duplicated_overwrite( $id );
```
kobizz, line 823: Get the control to check with instead of duplicating the conditions

kobizz (changes requested):

includes/base/controls-stack.php
udidol: Done.

kobizz (approved):

PR #16449, opened 2021-10-01
disable some wigets and add new widgets

masihgh:
Hi i want disable some wigets and add new widgets
and disable some settings and add new settings

how can i do ?


matipojo:
Hi @Masihgh Please open an [issue](https://github.com/elementor/elementor/issues/new/choose).

PR #16492, opened 2021-10-06
Tweak: Added a quick Apply Kit option using a popup modal in Kit Library [ED-4049]

Nevoss:


core/app/assets/js/ui/dialog/dialog.js
```diff
@@ -12,7 +12,8 @@ export default function Dialog( props ) {
 		<DialogWrapper onSubmit={ props.onSubmit } onClose={ props.onClose }>
 			<DialogContent>
 				{ props.title && <DialogTitle>{ props.title }</DialogTitle> }
-				{ props.text && <DialogText>{ props.text }</DialogText> }
+				{ props.text && Array.isArray( props.text ) && props.text.map( ( text ) => <DialogText key={text}>{ text }</DialogText> ) }
+				{ props.text && ! Array.isArray( props.text ) && <DialogText>{ props.text }</DialogText> }
```
iNewLegend, line 16: I wonder... can it work like this?
```suggestion
{ 
	let text = null;
	if ( props.text ) {
		text = Array.isArray( props.text ) ? props.text.map( ( text ) => <DialogText key={text}>{ text }</DialogText> ) : <DialogText>{ props.text }</DialogText>
	}
	return text;
}
```

Nevoss: It can't, but I can extract it to another function

iNewLegend (approved):

matipojo (approved):

PR #16808, opened 2021-11-04
Fix: Nav menu not scrolling to the appropriate CSS ID if Scroll Snap is enabled [ED-5439]

guyc-E:


assets/dev/js/frontend/utils/anchors.js
```diff
@@ -84,7 +84,5 @@ module.exports = elementorModules.ViewModule.extend( {
 
 	onInit: function() {
 		elementorModules.ViewModule.prototype.onInit.apply( this, arguments );
-
-		this.bindEvents();
```
guyc-E, line 88: This method is already triggered from ViewModule parent class.
It causes the events to trigger twice.

assets/dev/js/frontend/utils/anchors.js
```diff
@@ -7,7 +7,7 @@ module.exports = elementorModules.ViewModule.extend( {
 			selectors: {
 				links: 'a[href*="#"]',
 				targets: '.elementor-element, .elementor-menu-anchor',
-				scrollable: 'html, body',
+				scrollable: isScrollSnapActive() ? 'body' : 'html, body',
```
guyc-E, line 10: If there are two elements, it causes the scrolling animation to run twice and it sometimes breaks scroll-snap. In line 75

assets/dev/js/frontend/utils/anchors.js
guyc-E: Before I fixed the bug, the animate function in line 75 ran four times.

assets/dev/js/frontend/utils/anchors.js
kobizz: Is the `html` still needed for when there is no scroll snap?

kobizz (changes requested):

assets/dev/js/frontend/utils/anchors.js
guyc-E: Yes, it works only with HTML, I thought maybe to delete the 'body' if scroll-snap is off, but I'm not sure if the 'html, body' together is fixing some sort of bug.

PR #16809, opened 2021-11-04
New: Drag image from desktop [ED-3817]

YitzhakAvm:


matipojo (changes requested):
Please fix the conflict

includes/widgets/video.php
```diff
@@ -1174,7 +1174,7 @@ private function render_hosted_video() {
 
 		$video_params = $this->get_hosted_params();
 		?>
-		<video class="elementor-video" src="<?php echo esc_url( $video_url ); ?>" <?php Utils::print_html_attributes( $video_params ); ?>></video>
+		<video class="elementor-video" src="<?php echo esc_attr( $video_url ); ?>" <?php Utils::print_html_attributes( $video_params ); ?>></video>
```
matipojo, line 1177: Why?

matipojo (changes requested):

includes/widgets/video.php
ItzhakAvm: Because it's not always a URL, when in preview mode - it's a base64 string.

matipojo: Please add a comment in the code

PR #16842, opened 2021-11-07
Fix: After downgrading from a version with Container - a PHP error is thrown [ED-5425]

StyleShit:


includes/elements/column.php
```diff
@@ -1015,13 +1015,17 @@ protected function add_render_attributes() {
...
 	 *
-	 * @return Element_Base Column default child type.
+	 * @return Element_Base|false Column default child type.
 	 */
 	protected function _get_default_child_type( array $element_data ) {
 		if ( 'section' === $element_data['elType'] ) {
 			return Plugin::$instance->elements_manager->get_element_types( 'section' );
 		}
 
+		if ( empty( $element_data['widgetType'] ) ) {
```
matipojo, line 1025: Explain

matipojo (changes requested):

includes/elements/column.php
StyleShit: Done

matipojo (approved):

PR #16888, opened 2021-11-10
Fix: Responsive choose control order is reversed in RTL websites [ED-5461]

shilo-ey:


assets/dev/scss/editor/panel/controls/_choose.scss
```diff
@@ -67,7 +67,8 @@
 
 	.elementor-control-text_align,
 	.elementor-control-align,
-	.elementor-control-position {
+	.elementor-control-position,
+	[class*="elementor-control-responsive-"] {
```
udidol, line 71:
```suggestion
	.elementor-control-position,
	.elementor-control-position[class*="elementor-control-responsive-"] {
```

udidol, line 71: This will target all responsive choose controls, which we shouldn't do.
We only want to target controls in which there is importance to the order of the choice-boxes, such as align and position controls (which is why these specific class names are targeted in the first place).

I would suggest something that will cover all of these controls:
```
.rtl {

	.elementor-control-type-choose {

		&[class*="elementor-control-text_align"],
		&[class*="elementor-control-align"],
		&[class*="elementor-control-position"] {
```

PR #16889, opened 2021-11-10
Tweak: Removed shortkey from panel tooltips from v.3.5.0 [ED-5476]

iNewLegend:


StyleShit (approved):

matipojo (changes requested):
Why not revert the original commit?
https://github.com/elementor/elementor/pull/16478

matipojo (approved):

PR #16910, opened 2021-11-11
Tweak: Prompt the user permission to allow unfiltered file uploads in Import Template flow [ED-5279]

udidol:
## Summary

* Tweak: Prompt the user permission to allow unfiltered file uploads in Import Template flow

### DIFF
--- a/includes/template-library/manager.php
+++ b/includes/template-library/manager.php
@@ -450,8 +450,15 @@ class Manager {
 	 * @return mixed Whether the export succeeded or failed.
 	 */
 	public function import_template( array $data ) {
+		// If the template is a JSON file, allow uploading it.
+		add_filter( 'elementor/allow-file-type/json', [ $this, 'enable_json_template_upload' ] );
+		add_filter( 'elementor/files/allow_unfiltered_upload', [ $this, 'enable_json_template_upload' ] );
+
 		// Imported templates can be either JSON files, or Zip files containing multiple JSON files
 		$upload_result = Plugin::$instance->uploads_manager->handle_elementor_upload( $data, [ 'zip', 'json' ] );
+
+		remove_filter( 'elementor/allow-file-type/json', [ $this, 'enable_json_template_upload' ] );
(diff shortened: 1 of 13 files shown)


includes/template-library/manager.php
```diff
@@ -450,9 +450,16 @@ public function direct_import_template() {
 	 * @return mixed Whether the export succeeded or failed.
 	 */
 	public function import_template( array $data ) {
+		// If the template is a JSON file, allow uploading it.
+		add_filter( 'elementor/allow-file-type/json', [ $this, 'enable_json_template_upload' ] );
+		add_filter( 'elementor/files/allow_unfiltered_upload', [ $this, 'enable_json_template_upload' ] );
```
kobizz, line 455: Remove this filter and handle deprecated methods

includes/template-library/manager.php
```diff
@@ -450,9 +450,16 @@ public function direct_import_template() {
 	 * @return mixed Whether the export succeeded or failed.
 	 */
 	public function import_template( array $data ) {
+		// If the template is a JSON file, allow uploading it.
+		add_filter( 'elementor/allow-file-type/json', [ $this, 'enable_json_template_upload' ] );
```
kobizz, line 454:
```suggestion
		add_filter( 'elementor/files/allow-file-type/json', [ $this, 'enable_json_template_upload' ] );
```

kobizz (changes requested):

includes/template-library/manager.php
udidol: The filter cannot be removed, as it is necessary for different usages in the product.
The specific file type filter is necessary for JSON files in order to enable importing single templates.
@kobizz

includes/template-library/manager.php
udidol: Done

includes/template-library/manager.php
udidol: Handling deprecated methods - done.

PR #16931, opened 2021-11-14
Internal: Wrong groups position in context-menu [ED-5535]

udidol:
## Summary

* Internal: wrong groups position in context-menu

### DIFF
--- a/assets/dev/js/editor/elements/views/behaviors/context-menu.js
+++ b/assets/dev/js/editor/elements/views/behaviors/context-menu.js
@@ -55,6 +55,27 @@ module.exports = Marionette.Behavior.extend( {
 			} );
 		}
 
+		// Adding custom groups will only be possible for elements' context menus (Section, Column, Widget).
+		if ( this.view.options.model?.get( 'elType' ) ) {
+			let customGroups = [];
+
+			/**
+			 * Filter Additional Context Menu Groups.
+			 *
+			 * This filter allows adding new context menu groups to elements.
+			 *
+			 * @since 3.5.0
+			 *
+			 * @param array customGroups - An array of group objects.
+			 * @param string elementType - The current element type.
+			 */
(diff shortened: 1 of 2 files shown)


assets/dev/js/editor/elements/views/behaviors/context-menu.js
```diff
@@ -55,6 +55,27 @@ module.exports = Marionette.Behavior.extend( {
...
+			 *
+			 * @since 3.5.0
+			 *
+			 * @param array customGroups - An array of group objects.
+			 * @param string elementType - The current element type.
+			 */
+			customGroups = elementor.hooks.applyFilters( 'elements/context-menu/groups', customGroups, this.view.options.model.get( 'elType' ) );
+
+			if ( customGroups.length ) {
+				contextMenuGroups.splice( afterGroupIndex + 1, 0, ...customGroups );
```
guyc-E, line 75: If in the future we will add more groups, the position "afterGroupIndex + 1" will not be before the "Delete" group unless we will change it to "afterGroupIndex + 2", maybe we need to check again the position of the "Delete"?

udidol: It shouldn't be a problem in this area, since groups are passed into the behavior by the view, and the index of the "delete" group wouldn't change unless the view changes it.

assets/dev/js/editor/elements/views/behaviors/context-menu.js
```diff
@@ -55,6 +55,27 @@ module.exports = Marionette.Behavior.extend( {
 			} );
 		}
 
+		// Adding custom groups will only be possible for elements' context menus (Section, Column, Widget).
+		if ( this.view.options.model?.get( 'elType' ) ) {
```
kobizz, line 59: @udidol Try to find a way to do that without having to do this explicit check here

kobizz (changes requested):

arielk:
@udidol is still relevant?

PR #16941, opened 2021-11-15
Fix: Font Awesome 4 support are wrong default value [ED-5404]

iNewLegend:


includes/managers/icons.php
```diff
@@ -471,7 +471,7 @@ public function register_admin_settings( Settings $settings ) {
 				'label' => esc_html__( 'Load Font Awesome 4 Support', 'elementor' ),
 				'field_args' => [
 					'type' => 'select',
-					'std' => 'yes',
+					'std' => '',
```
matipojo, line 474: Restore the old value, it shouldn't be empty.

matipojo (changes requested):

matipojo (approved):

PR #16952, opened 2021-11-16
Fix: Text path widget is not optimized and makes redundant file system calls [ED-5420]

StyleShit:


matipojo (changes requested):
Please fix the ScreenShotter error

StyleShit:
@matipojo 
Can't reproduce on my machine.
It's probably a delay in the fetch

PR #16959, opened 2021-11-17
Fix: D&D issues with right and left positioned elements [ED-5494]

StyleShit:


iNewLegend (approved):

tests/playwright/pages/wp-admin-page.js
```diff
@@ -1,58 +1,88 @@
...
+
+		const prefix = 'e-experiment';
+
+		for ( const [ id, state ] of Object.entries( experiments ) ) {
+			const selector = `#${ prefix }-${ id }`;
+
+			await this.page.selectOption( selector, state ? 'active' : 'inactive' );
+		}
+
+		await this.page.click( '#submit' );
```
StyleShit, line 81: @matipojo 

Only this one is mine.

tests/playwright/sanity/container.test.js
```diff
@@ -0,0 +1,44 @@
...
+
+	const buttonEl = await editor.getElementHandle( button ),
+		headingEl = await editor.getElementHandle( heading );
+
+	const elBeforeButton = await buttonEl.evaluate( ( node ) => node.previousElementSibling ),
+		elAfterHeading = await headingEl.evaluate( ( node ) => node.nextElementSibling );
+
+	// Assert.
+	// Test that the image is between the heading & button.
+	expect( elBeforeButton ).toBe( elAfterHeading );
```
matipojo, line 43: Try something like:
```suggestion
	// Test that the image is between the heading & button.
	expect( container.children[1].getype() ).toBe( 'image' );
```

matipojo (changes requested):

tests/playwright/sanity/container.test.js
StyleShit: For some reason, I can't get a reference to the Container object, since the test is running in another context from the page itself, and I get `undefined`.
Not sure why...

https://playwright.dev/docs/api/class-page#page-evaluate

image

PR #17030, opened 2021-11-24
Fix: Gradient background doesn't look correctly on edit mode while using global colors [ED-3916]

nisan250:
Before:
https://www.loom.com/share/b188c505abd54e2b8d31dbf981dedaeb

After:
https://www.loom.com/share/d2cd647a8e704751b5556abda6101140

### DIFF
--- a/assets/dev/js/editor/elements/views/base.js
+++ b/assets/dev/js/editor/elements/views/base.js
@@ -247,6 +247,9 @@ BaseElementView = BaseContainer.extend( {
 		_.defer( () => {
 			// Init container. Defer - in order to init the container after the element is fully initialized, and properties like `_parent` are available.
 			this.getContainer();
+
+			// Since the global colors were not ready, the styles are rendered again for the gradiant background color control.
+			this.renderStyles();
 		} );
 	},
 
--- a/assets/dev/js/editor/elements/views/base.js
+++ b/assets/dev/js/editor/elements/views/base.js
@@ -247,9 +247,6 @@ BaseElementView = BaseContainer.extend( {
 		_.defer( () => {
 			// Init container. Defer - in order to init the container after the element is fully initialized, and properties like `_parent` are available.
 			this.getContainer();
-
-			// Since the global colors were not ready, the styles are rendered again for the gradiant background color control.
-			this.renderStyles();
 		} );
 	},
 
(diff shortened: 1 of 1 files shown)


assets/dev/js/editor/elements/views/base.js
```diff
@@ -247,6 +247,9 @@ BaseElementView = BaseContainer.extend( {
 		_.defer( () => {
 			// Init container. Defer - in order to init the container after the element is fully initialized, and properties like `_parent` are available.
 			this.getContainer();
+
+			// Since the global colors were not ready, the styles are rendered again for the gradiant background color control.
+			this.renderStyles();
```
StyleShit, line 252: Maybe we should check why this is happening instead of putting a bandage on it making UI glitches?

nisan250: This solution is good enough. The global colors are available after the code in this class runs.

nisan250: The element is not rendered twice only once.

assets/dev/js/editor/elements/views/base.js
```diff
@@ -582,10 +582,14 @@ BaseElementView = BaseContainer.extend( {
 	},
 
 	renderUI: function() {
-		this.renderStyles();
 		this.renderCustomClasses();
 		this.renderCustomElementID();
 		this.enqueueFonts();
+
+		_.defer( () => {
+			// Since the global colors were not ready, the styles are rendered again for the gradiant background color control.
```
StyleShit, line 590:
```suggestion
			// Defer the styles render to make sure that the global colors are ready.
```

StyleShit (changes requested):

assets/dev/js/editor/elements/views/base.js
```diff
@@ -582,10 +582,14 @@ BaseElementView = BaseContainer.extend( {
 	},
 
 	renderUI: function() {
-		this.renderStyles();
 		this.renderCustomClasses();
 		this.renderCustomElementID();
 		this.enqueueFonts();
+
+		_.defer( () => {
```
StyleShit, line 589: @matipojo 

Maybe it'll be better to find the root cause for the global colors' unavailability?
Or do you think it's fine for now?

assets/dev/js/editor/elements/views/base.js
nisan250: Done

StyleShit (approved):

assets/dev/js/editor/elements/views/base.js
matipojo: I think it's OK for now

matipojo (approved):

PR #17092, opened 2021-12-01
New: Added a reusable button trait [ED-4597] (#17041)

matipojo:
This reverts commit 4862f066f5bf192ec5ada8d142194a608f146d9e.
The Trait as been merged to 3.5.0, and then reverted. this commit re-adds it to 3.6.0.


includes/widgets/traits/button-trait.php
```diff
@@ -0,0 +1,564 @@
...
+	 *
+	 * Render button widget text.
+	 *
+	 * @param \Elementor\Widget_Base|null $instance
+	 *
+	 * @since  3.4.0
+	 * @access protected
+	 */
+	protected function render_text( Widget_Base $instance = null ) {
+		if ( ! $instance ) {
```
matipojo, line 512: @StyleShit can you explain why?

StyleShit: It's the same as here:
https://github.com/elementor/elementor/blob/2094102d98b0d2735a853bdf7777e418a76cac8b/includes/widgets/traits/button-trait.php#L429-L432

StyleShit:
```suggestion
		// The default instance should be `$this` (a Button widget), unless the Trait is being used from outside of a widget (e.g. `Skin_Base`) which should pass an `$instance`.
		if ( empty( $instance ) ) {
```

StyleShit (changes requested):

PR #17121, opened 2021-12-06
Internal: Workflows - Use Node 14

chenele:
Temporary, use Node.js 14 and npm 6.


.github/workflows/lighthouse.yml
```diff
@@ -119,4 +119,3 @@ jobs:
         run: |
           MESSAGE_TEXT="@channel Repo: *$GITHUB_REPOSITORY* Workflow: *$GITHUB_WORKFLOW* is FAILED - <$GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID|link>"
           curl -X POST "https://slack.com/api/chat.postMessage" -d "username=${SLACK_BOT_NAME}&token=${SLACK_TOKEN}&channel=${SLACK_CHANNEL}&text=${MESSAGE_TEXT}&link_names=true"
```
matipojo, line 121:
```suggestion
          curl -X POST "https://slack.com/api/chat.postMessage" -d "username=${SLACK_BOT_NAME}&token=${SLACK_TOKEN}&channel=${SLACK_CHANNEL}&text=${MESSAGE_TEXT}&link_names=true"

```

.github/workflows/publish-patch.yml
```diff
@@ -173,4 +173,3 @@ jobs:
           SLACK_CHANNELS: ${{ secrets.SLACK_CHANNELS }}
         run: |
           bash "${GITHUB_WORKSPACE}/.github/scripts/slack-notify.sh"
```
matipojo, line 175:
```suggestion
          bash "${GITHUB_WORKSPACE}/.github/scripts/slack-notify.sh"

```

matipojo (changes requested):

PR #17134, opened 2021-12-08
Internal: Hash commands can accept arguments [ED-5518]

Nevoss:


tests/qunit/tests/core/common/assets/js/api/extras/hash-commands.spec.js
```diff
@@ -1,8 +1,18 @@
 import Component from './mock/component.spec';
 
 QUnit.module( 'File: core/common/assets/js/api/extras/hash-commands.js', ( hooks ) => {
+	let originalWarnFunction;
+
 	hooks.before( () => {
 		$e.components.register( new Component() );
+
+		// Make sure the original warn function will not be triggered to avoid printing on the console.
```
matipojo, line 9: Why not?

matipojo (changes requested):

PR #17181, opened 2021-12-14
Fix: Image Widget - avoid processing images with Jetpack's Photon

### LINKED ISSUE #22052: Enhancement: Site Accelerator ignores custom image size from Elementor
### Impacted plugin

Jetpack

### What

Make site accelerator respect the custom image size set in Elementor

### How

With Site Accelerator (Speed up image load times) disabled:

- Create a new page with Elementor and add an image with a custom size:

image

- View the page. The image is shown at the specified size:

image

- Now enable image "Speed up image load times" and view the page again. The image is not shown at the custom size:

image

Reported in https://wordpress.org/support/topic/image-rendering-3/

jeherve:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

* Image Widget: avoid conflicts with the Jetpack plugin's Image CDN feature

## Description

This change addresses [a conflict with the Jetpack plugin and its image CDN feature](https://github.com/Automattic/jetpack/issues/22052).
The image CDN feature, also known as Photon, attempts to process images as you display them on the frontend of your site, to serve them from a CDN instead of locally.

This should only happen on the frontend though, not when the images are being fetched before they're inserted.

## Test instructions

* Start with a site where you've activated Elementor, as well as the Jetpack plugin.
* Go to Jetpack > Settings > Performance, and enable the Site Accelerator feature.
* Go to Pages > Add New, and create a new page using the Elementor editor.
* In the Elementor editor, add an image widget.
* In the widget, upload a new image, and set the size to custom.
* Set a width of `500` and a height of `100`.
* Save your changes, and visit the page. You'll see that the image is smaller than that.
* Apply this patch.
* Refresh the page; the image will have the right size.

## Quality assurance

Fixes https://github.com/Automattic/jetpack/issues/22052

### DIFF
--- a/includes/controls/groups/image-size.php
+++ b/includes/controls/groups/image-size.php
@@ -105,7 +105,25 @@ class Group_Control_Image_Size extends Group_Control_Base {
 				'class' => trim( $image_class ),
 			];
 
+			/*
+			 * If you use the Jetpack plugin and its image CDN feature,
+			 * the image strings returned will use the CDN URL.
+			 * We don't want that.
+			 * The CDN URL will be added later on on the front end.
+			 *
+			 * @see https://ethitter.com/p/897/
+			 * @see https://github.com/Automattic/jetpack/issues/22052
+			 */
+			if ( class_exists( '\Automattic\Jetpack\Image_CDN\Image_CDN' ) ) {
+				remove_filter( 'image_downsize', array( \Automattic\Jetpack\Image_CDN\Image_CDN::instance(), 'filter_image_downsize' ) );
+			}
+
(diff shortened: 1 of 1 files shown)


alphavaletec:
Hello,

We use Elementor and JetPack on our websites in Brazil. And both are very good and have helped us a lot in the projects. Thank you very much!

--------

The problem is that the Elementor Custom Image is not loaded on the JetPack CDN.

We entered the code for correction, but without success.


Elementor custom images are very good to avoid unnecessary "Thumbnails".


Thank you for your attention and we look forward to an update!

PR #17252, opened 2021-12-20
Fix: System info file displays inaccurate WP memory limit. [ED-5717]

iNewLegend:


modules/system-info/reporters/wordpress.php
```diff
@@ -74,7 +74,7 @@ public function get_fields() {
 	 */
 	public function get_memory_limit() {
 		$result = [
-			'value' => ini_get( 'memory_limit' ),
+			'value' => (string) WP_MAX_MEMORY_LIMIT, // Use WordPress max memory limit, since its dynamic.
```
Nevoss, line 77: Are you sure the right value is  `WP_MAX_MEMORY_LIMIT` and not `WP_MEMORY_LIMIT`?

Nevoss (changes requested):

modules/system-info/reporters/wordpress.php
matipojo: We need to add all of them, in the past we already used the `WP_MEMORY_LIMIT` but on some servers it was different from the `ini_get( 'memory_limit' )`. https://github.com/elementor/elementor/commit/02f81ced71b7ce3c65e5d74377206afb014b6352
@iNewLegend please add the 3 values: `get_ini_memory_limit, get_wp_memory_limit, and, get_wp_max_memory_limit`
See https://github.com/WordPress/WordPress/blob/master/wp-includes/default-constants.php#L37

iNewLegend: `WP_MAX_MEMORY_LIMIT` is the most accurate since it provide the max memory limit for admin and see here:
https://github.com/WordPress/WordPress/blob/master/wp-includes/default-constants.php#L51
its favor the `wp_is_ini_value_changeable( 'memory_limit' )`.
`WP_MEMORY_LIMIT` is for non admin users.

In order to understand it, please provide me an example where `WP_MAX_MEMORY_LIMIT` is not correct.

StyleShit: Maybe just do like this:
```PHP
$memory_limit = min( [ WP_MAX_MEMORY_LIMIT, WP_MEMORY_LIMIT, ini_get( 'memory_limit' ) ] );
```

StyleShit (changes requested):

modules/system-info/reporters/wordpress.php
iNewLegend: Why?

StyleShit (approved):

modules/system-info/reporters/server.php
```diff
@@ -126,6 +128,45 @@ public function get_php_version() {
...
+	 *    @type bool   $warning        Whether to display a warning. True if the limit
+	 *                                 is below the recommended 64M, False otherwise.
+	 * }
+	 */
+	public function get_php_memory_limit() {
+		$result = [
+			'value' => (string) WP_Site_Health::get_instance()->php_memory_limit,
+		];
+
+		$min_recommended_memory = '64M';
```
iNewLegend, line 150:
```suggestion
	 *                                 is below the recommended 512M, False otherwise.
	 * }
	 */
	public function get_php_memory_limit() {
		$result = [
			'value' => (string) WP_Site_Health::get_instance()->php_memory_limit,
		];

		$min_recommended_memory = '512M';
```

@matipojo

modules/system-info/reporters/server.php
```diff
@@ -126,6 +128,45 @@ public function get_php_version() {
...
+	 *
+	 *    @type string $value          PHP memory limit.
+	 *    @type string $recommendation Recommendation memory limit.
+	 *    @type bool   $warning        Whether to display a warning. True if the limit
+	 *                                 is below the recommended 64M, False otherwise.
+	 * }
+	 */
+	public function get_php_memory_limit() {
+		$result = [
+			'value' => (string) WP_Site_Health::get_instance()->php_memory_limit,
```
matipojo, line 147: Use ini_get

modules/system-info/reporters/server.php
matipojo: "WP Memory limit of 128 MB or greater (256 MB or higher is preferred)"

https://elementor.com/help/requirements/

matipojo (changes requested):

modules/system-info/reporters/server.php
iNewLegend: Handled.

modules/system-info/reporters/server.php
iNewLegend: Handled.

PR #17320, opened 2021-12-28
Fix: Can’t drag&drop elements inside a container [ED-6077]

YitzhakAvm:


assets/dev/js/editor/utils/jquery-html5-dnd.js
```diff
@@ -354,6 +354,12 @@
 					},
 				} );
 			} else {
+				// Override the onDrop callback with a user-provided one if present.
```
matipojo, line 357: It should work with a hook, please consult with @iNewLegend

matipojo (changes requested):

assets/dev/js/editor/utils/jquery-html5-dnd.js
ItzhakAvm: We're not using any command here, where the hook should be placed?

PR #17373, opened 2022-01-04
Fix: Core SVG icons from template library are imported empty [ED-5980]

Nevoss:


nisan250 (approved):

StyleShit (approved):

includes/template-library/sources/remote.php
```diff
@@ -189,6 +189,9 @@ public function get_data( array $args, $context = 'display' ) {
 			return $data;
 		}
 
+		// Set the Request's state as an Elementor upload request, in order to support unfiltered file uploads.
+		Plugin::$instance->uploads_manager->set_elementor_upload_state( true );
```
matipojo, line 193: Set `false` after the import (maybe from the upload manager)

matipojo (changes requested):

includes/template-library/sources/remote.php
Nevoss: Done (The upload of the images is not from the upload manager, but it uses the upload manager to check if allow uploading, this is why I revert the `elementor_upload_state` in `remote.php` and not inside the upload manager)

Issue #16516, opened 2021-10-08
Google Font Update


hitsgee:
## Prerequisites

- [X] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [X] The issue still exists against the latest stable version of Elementor.

## Description

Could you please update the Google font library. Many fonts seem to be missing such as "Atkinson Hyperlegible" and many more I have found.

## Isolating the problem

- [X] This bug happens with only Elementor plugin active (and Elementor Pro).
- [X] This bug happens with a default WordPress theme active.
- [X] I can reproduce this bug consistently using the steps above.

shilo-ey:
This feature has been introduced in Elementor v3.6.0

Feel free to update

Thanks!

Issue #13501, opened 2021-01-05
Please add JetBrains Mono to Google Fonts list


mrtopf:
## Prerequisites

- [x] I have searched for similar features requests in both open and closed tickets and cannot find a duplicate.
- [x] The feature is still missing in the latest stable version of Elementor ( Elementor Pro. )

**What problem is your feature request going to solve? Please describe.**
JetBrains Mono is missing from the fonts list although it is a Google font. I added it myself to `elementor/includes/fonts.php` but that's not so update proof. There also might be more fonts missing (maybe there is a script to sync them?)

**Describe the solution you'd like**
Well, add it :-)

**Describe alternatives you've considered**
Uploading it manually.

shilo-ey:
This feature has been introduced in Elementor v3.6.0

Feel free to update

Thanks!

PR #17496, opened 2022-01-17
Tweak: Updated Google Fonts list [ED-6245]

KingYes:
#16516, #13501


includes/fonts.php
```diff
@@ -186,7 +188,7 @@ private static function get_native_fonts() {
 			'Amiri' => self::GOOGLE,
 			'Amita' => self::GOOGLE,
 			'Anaheim' => self::GOOGLE,
-			'Andada' => self::GOOGLE,
+			'Andada Pro' => self::GOOGLE,
```
matipojo, line 191: BC? What will happen for an element that already uses the `Andada` font?

matipojo (changes requested):

PR #17527, opened 2022-01-20
Tweak: Lightbox [ED-6293]

udidol:


assets/dev/js/frontend/utils/lightbox/lightbox.js
```diff
@@ -256,14 +256,19 @@ module.exports = elementorModules.ViewModule.extend( {
...
 			$videoElement = $( '<video>', videoParams );
 		} else {
-			let apiProvider = elementorFrontend.utils.baseVideoLoader;
+			let apiProvider;
 
 			if ( -1 !== options.url.indexOf( 'vimeo.com' ) ) {
 				apiProvider = elementorFrontend.utils.vimeo;
 			} else if ( options.url.match( /^(?:https?:\/\/)?(?:www\.)?(?:m\.)?(?:youtu\.be\/|youtube\.com)/ ) ) {
 				apiProvider = elementorFrontend.utils.youtube;
 			}
```
guyc-E, line 265: Maybe it will be shorter to just use:
```
else{
 return;
}
```
And I don't think we need the comment, it's obvious what the code does.

udidol: Good call, will change

guyc-E (approved):

PR #17528, opened 2022-01-20
Tweak: Change Developer Edition promotional notice triggers [ED-5562]

Nevoss:


tests/phpunit/elementor/core/admin/notices/test-elementor-dev-notice.php
```diff
@@ -49,10 +50,22 @@ public function test_should_print__when_plugin_is_installed( $plugin_name ) {
 		$this->assertTrue( $result );
 	}
 
-	public function test_should_print__when_one_experiment_in_active() {
+	public function promotion_plugins_data_provider() {
```
Nevoss, line 53: This method just moved from one place to another

PR #17588, opened 2022-01-26
Tweak: Added dynamic tag control to various core widgets and features [OBXT-384]

hein-obox:


udidol:
The screenshotter and playwright fails have nothing to do with this PR, they shouldn't prevent merge.

includes/widgets/icon-box.php
```diff
@@ -404,22 +404,38 @@ protected function register_controls() {
...
 					'unit' => 'deg',
+					'size' => '',
+				],
+				'tablet_default' => [
+					'unit' => 'deg',
+					'size' => '',
+				],
+				'mobile_default' => [
+					'unit' => 'deg',
+					'size' => '',
```
udidol, line 430: @hein-obox Mobile and Tablet defaults aren't enough.
Since we have the Additional Breakpoints experiment - you need to take into account all of the active breakpoints, in case the user activated the experiment and is using any additional breakpoints, such as Laptop.

includes/widgets/icon.php
```diff
@@ -366,6 +366,7 @@ protected function register_controls() {
 			]
 		);
 
+		// test comment
```
udidol, line 369: Remove this

udidol (changes requested):

udidol (approved):

PR #17734, opened 2022-02-07
Tweak: Add Lazy load to all the widgets using Swiper [ED-2409]

ghost:


assets/dev/js/frontend/handlers/background-slideshow.js
```diff
@@ -64,7 +71,8 @@ export default class BackgroundSlideshow extends elementorModules.frontend.handl
 			direction = 'slide_left' === elementSettings.background_slideshow_slide_transition ? 'ltr' : 'rtl',
 			$container = jQuery( '<div>', { class: classes.swiperContainer, dir: direction } ),
 			$wrapper = jQuery( '<div>', { class: classes.swiperWrapper } ),
-			kenBurnsActive = elementSettings.background_slideshow_ken_burns;
+			kenBurnsActive = elementSettings.background_slideshow_ken_burns,
+			lazyload = 'yes' === elementSettings.background_slideshow_lazyload;
```
guyc-E, line 75: Maybe we should create a method instead of repeating this code?
```
isLazyLoad(){
  return 'yes' === elementSettings.background_slideshow_lazyload;
}
```

guyc-E (changes requested):

assets/dev/js/frontend/handlers/background-slideshow.js
ManorHazaz: Can't use elementSettings outside the functions.

guyc-E (approved):

PR #17852, opened 2022-02-15
Internal: Nested Infra – Frontend Allow nested capabilities. [ED-6554]

iNewLegend:


assets/dev/js/frontend/handlers/base-tabs.js
```diff
@@ -181,13 +211,7 @@ export default class baseTabs extends elementorModules.frontend.handlers.Base {
 		this.activateDefaultTab();
 	}
 
-	onEditSettingsChange( propertyName ) {
```
Nevoss, line 184: Why removed?

assets/dev/js/frontend/handlers/base.js
```diff
@@ -33,7 +33,8 @@ module.exports = elementorModules.ViewModule.extend( {
 		var $mainElement = this.$element;
 
 		return $mainElement.find( selector ).filter( function() {
-			return jQuery( this ).closest( '.elementor-element' ).is( $mainElement );
+			// Start `closest` from parent since self can be `.elementor-element`.
+			return jQuery( this ).parent().closest( '.elementor-element' ).is( $mainElement );
```
Nevoss, line 37: are you trying to find the "tab" container? if yes can you assign to it some attr?

Nevoss (changes requested):

assets/dev/js/frontend/handlers/base-tabs.js
iNewLegend: Its not needed anymore.

assets/dev/js/frontend/handlers/base.js
iNewLegend: `findElement` does not take in advantage nested elements, and using old code, it will give him wrong handler.

matipojo (changes requested):
Please duplicate the tabs and change only the new one

PR #17869, opened 2022-02-17
Fix: Posts, Archive Posts - Widget appears empty while using PHP 8.1. [ED-6466]

iNewLegend:


includes/base/widget-base.php
```diff
@@ -254,12 +256,10 @@ public function hide_on_search() {
 	public function start_controls_section( $section_id, array $args = [] ) {
 		parent::start_controls_section( $section_id, $args );
 
-		static $is_first_section = true;
-
-		if ( $is_first_section ) {
+		if ( $this->is_first_section ) {
```
Nevoss, line 259: I think it cannot be as private property because it should be available for each widget render to avoid an infinite loop. please talk to me about it.

Nevoss (changes requested):

Nevoss (approved):

PR #17950, opened 2022-02-23
Internal: Container - Prepare `feature/container` to `infra-nested`  `elementsManager`. [ED-6589]

iNewLegend:


matipojo (changes requested):
Where do you register the element?

matipojo (approved):

PR #18110, opened 2022-03-09
Fix: Custom icons disappear at frontend if name contains numbers  [ED-1040]

TzviRabinovitch:
for pro (pr#1537)


includes/managers/icons.php
```diff
@@ -163,8 +163,7 @@ public static function get_icon_manager_tabs() {
 		 * @param array $additional_tabs Additional icon manager tabs. Default is an empty array.
 		 */
 		$additional_tabs = apply_filters( 'elementor/icons_manager/additional_tabs', $additional_tabs );
-
-		return array_merge( self::$tabs, $additional_tabs );
+		return array_replace( self::$tabs, $additional_tabs );
```
matipojo, line 166:
```suggestion

		return array_replace( self::$tabs, $additional_tabs );
```

PR #18150, opened 2022-03-13
Fix: Imported kit doesn't contain the global design [ED-6783]

guyc-E:


matipojo (approved):

matipojo (changes requested):
Please fix the checks errors

PR #18153, opened 2022-03-14
Internal: JS API - Merge `3.6.0` to `js-api-commands-refactor` [ED-2084]

iNewLegend:


core/base/document.php
```diff
@@ -1369,6 +1370,101 @@ public function __construct( array $data = [] ) {
 		parent::__construct( $data );
 	}
 
+	/**
+	 * Get document settings usage.
+	 *
+	 * @return array
+	 */
+	public function get_usage() {
```
matipojo, line 1378: It's not a part of 3.6.0. Please ensure only 3.6.0 is merged in the PR

matipojo (changes requested):

core/base/document.php
iNewLegend: Handled.

PR #18163, opened 2022-03-15
Internal: JS API - Fix refactor BC. [ED-2084]

iNewLegend:


modules/web-cli/assets/js/core/commands.js
```diff
@@ -19,6 +19,23 @@ export default class Commands extends CommandsBackwardsCompatibility {
...
+					'$e.commands.classes',
+					'3.7.0',
+					'$e.commands.getCommandClass(), $e.commandsInternal.getCommandClass(), $e.data.getCommandClass(), $e.routes.getCommandClass() according to the requested command infra-structure,'
+				);
+
+				return {
+					... $e.commands.commands,
+					... $e.commandsInternal.commands,
+					... $e.data.commands,
+					... $e.routes.commands,
```
matipojo, line 35: Why return all? Each one should return its own commands.

matipojo (changes requested):

modules/web-cli/assets/js/core/commands.js
iNewLegend: `$e.commands.classes` in BC included all the commands.

matipojo: I see, but why?

iNewLegend: Explained in private.

PR #18192, opened 2022-03-20
CLI import fix

ghost:


core/logger/manager.php
```diff
@@ -54,6 +54,10 @@ public function shutdown( $last_error = null, $should_exit = false ) {
 	}
 
 	public function rest_error_handler( $error_number, $error_message, $error_file, $error_line ) {
+		if ( Utils::is_wp_cli() ) {
```
udidol, line 57: Add comment that this is temporary until all errors are solved in core and pro

udidol (changes requested):

core/logger/manager.php
udidol: The comment should be a TODO comment: `// TODO: `

core/logger/manager.php
```diff
@@ -54,6 +54,10 @@ public function shutdown( $last_error = null, $should_exit = false ) {
...
+			return null;
+		}
+
```
udidol, line 60: Check lint fail

PR #18218, opened 2022-03-22
Internal: Updated "Share Link" library [ED-6816]

ghost:


assets/lib/share-link/share-link.js
```diff
@@ -16,6 +16,17 @@
 			$element.on( 'click', function() {
 				openShareLink( networkName );
 			} );
+
+			// Add "Enter" and "Space" event only if the element has role=button attribute.
+			if( 'button' === $element.attr( 'role' ) ) {
```
guyc-E, line 21: Add space

assets/lib/share-link/share-link.js
```diff
@@ -16,6 +16,17 @@
...
+				$element.on( 'keyup', ( event ) => {
+					if ( 13 === event.keyCode || 32 === event.keyCode ) {
+						event.preventDefault();
+
+						openShareLink( networkName );
+					}
+				});
```
guyc-E, line 28: Add space

assets/lib/share-link/share-link.js
```diff
@@ -16,6 +16,17 @@
 			$element.on( 'click', function() {
 				openShareLink( networkName );
 			} );
+
+			// Add "Enter" and "Space" event only if the element has role=button attribute.
+			if ( 'button' === $element.attr( 'role' ) ) {
```
guyc-E, line 21: I think the if statement is redundant because the role="button" is always exist.

assets/lib/share-link/share-link.js
```diff
@@ -16,6 +16,17 @@
...
+				$element.on( 'keyup', ( event ) => {
+					if ( 13 === event.keyCode || 32 === event.keyCode ) {
+						event.preventDefault();
```
guyc-E, line 24: I think the event.preventDefault() is redundant because the event is always on a div, and a div doesn't have a default behavior.

assets/lib/share-link/share-link.js
ManorHazaz: Currently the elements that using the library are divs, divs don't have role.

assets/lib/share-link/share-link.js
ManorHazaz: Just for future coverage, if the use will be with <a> or <button>.

assets/lib/share-link/share-link.js
udidol: So how will this ever work? @ManorHazaz

yossi-elementor (approved):

PR #18339, opened 2022-04-03
Fix: Container is not functioning as expected [ED-6845]

StyleShit:


rodolphebertozzo (approved):

assets/dev/scss/frontend/_container.scss
```diff
@@ -175,10 +175,11 @@
 			flex-basis: 100%;
 		}
 
-		// Fix media carousel widget broken (ED-6796).
-		// TODO: Find a better solution.
-		&-media-carousel {
+		// Fix swiper-based widgets broken (ED-6849).
+		// TODO: Use `:has( .swiper-container )` selector when available && remove this class from `frontend/utils/swiper.js`.
```
matipojo, line 179: Why not use the existing `e-swiper-container` class

matipojo (changes requested):

assets/dev/scss/frontend/_container.scss
StyleShit: Because it's attached to the column

PR #18409, opened 2022-04-11
Fix: Kit Library - See it live opens a blank page and not the Home Page [ED-6906]

guyc-E:


ghost (approved):

yossi-elementor (approved):

core/app/modules/import-export/module.php
```diff
@@ -400,7 +400,7 @@ private function get_elementor_document( $page_id ) {
 			return false;
 		}
 
-		return $document;
+		return $document->get_wp_preview_url();
```
ManorHazaz, line 403: Won't work anymore

ghost (changes requested):

PR #18495, opened 2022-04-25
Fix: Gradient control doesn't work on frontend when using Global Colors (#13288) [ED-3517]

### LINKED ISSUE #13288: Bug with global values in linear-gradient css function

udidol:
## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

* Fix: Gradient control doesn't work on frontend when using Global Colors (Fixes #13288)

### DIFF
--- a/core/files/css/base.php
+++ b/core/files/css/base.php
@@ -696,6 +696,22 @@ abstract class Base extends Base_File {
 
 		$value = $values[ $control['name'] ];
 
+		// If the control value is empty, and the control has a global default set, fetch the global value and use it.
+		if ( ! $value && ! empty( $control['global']['default'] ) ) {
+			$global_enabled = false;
+
+			if ( 'color' === $control['type'] ) {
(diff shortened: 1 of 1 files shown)


core/files/css/base.php
```diff
@@ -696,6 +696,22 @@ private function get_style_control_value( array $control, array $values ) {
 
 		$value = $values[ $control['name'] ];
 
+		// If the control value is empty, and the control has a global default set, fetch the global value and use it.
```
KingYes, line 699: Why not to move this code to another method?

KingYes (changes requested):

core/files/css/base.php
udidol: How would you do it? @KingYes

PR #18504, opened 2022-04-26
Tweak: Added responsive control to image position in Image Box widget [OBXT-661]

hein-obox:


assets/dev/scss/frontend/widgets/image-box.scss
```diff
@@ -3,44 +3,20 @@
 //
 .elementor-widget-image-box {
 
-	.elementor-image-box-content {
-		width: 100%; // Hack for Chrome in flex mode
-	}
-
-	@media (min-width: $screen-mobile-next) {
-
```
hein-obox, line 11: I have applied the image position to all breakpoints, but kept the default value for the mobile version.

includes/widgets/image-box.php
```diff
@@ -155,12 +155,13 @@ protected function register_controls() {
...
 
-		$this->add_control(
+		$this->add_responsive_control(
 			'position',
 			[
 				'label' => esc_html__( 'Image Position', 'elementor' ),
 				'type' => Controls_Manager::CHOOSE,
 				'default' => 'top',
+				'mobile_default' => 'top',
 				'options' => [
```
hein-obox, line 165: Default mobile version for BC reasons.

includes/widgets/image-box.php
```diff
@@ -175,7 +176,15 @@ protected function register_controls() {
...
-				'prefix_class' => 'elementor-position-',
+				'selectors_dictionary' => [
+					'left' => '--wrapper-display: flex; --wrapper-align-ltr: left; --wrapper-align-rtl: right; --wrapper-direction-ltr: row; --wrapper-direction-rtl: row-reverse; --image-margin-left: 0; --image-margin-right: unset; --image-margin-bottom: 0;',
+					'right' => '--wrapper-display: flex; --wrapper-align-ltr: right; --wrapper-align-rtl: left; --wrapper-direction-ltr: row-reverse; --wrapper-direction-rtl: row; --image-margin-left: unset; --image-margin-right: 0; --image-margin-bottom: 0;',
+					'top' => '--wrapper-display: block;  --wrapper-align-ltr: center; --wrapper-align-ltr: center; --image-margin-left: auto; --image-margin-right: auto; --image-margin-bottom: unset;',
+				],
+				'selectors' => [
+					'{{WRAPPER}}' => '{{VALUE}};',
+				],
+				'prefix_class' => 'elementor-position%s-',
```
hein-obox, line 187: The prefix class isn't used anymore, but I have kept it for BC reasons.

assets/dev/scss/frontend/widgets/image-box.scss
```diff
@@ -67,32 +43,38 @@
...
+	@media (max-width: $screen-mobile-max) {
+
+		&.elementor-position {
+
+			&-left, &-right {
+
+				// BC compatibility code since 3.7.0
+				.elementor-widget-container .elementor-image-box-wrapper .elementor-image-box-img {
+					margin-bottom: 0;
+				}
```
hein-obox, line 76: Preferably I would use margin-bottom: 15px, becasue that was the original code from line 76.
But that code was overwritten by the following code from the general styling.

.elementor .elementor-widget:not(.elementor-widget-text-editor):not(.elementor-widget-theme-post-content) figure {
    margin: 0;
}

I am not sure which value I should use.

assets/dev/scss/frontend/widgets/image-box.scss
```diff
@@ -67,32 +43,38 @@
...
-			margin-bottom: 15px;
-		}
-	}
-
 	.elementor-image-box-img {
 		display: inline-block; // For alignment content
 	}
 
+	&.elementor-element .elementor-widget-container .elementor-image-box-wrapper .elementor-image-box-img {
+		margin-left: var(--image-margin-left, var(--image-spacing, 15px));
```
hein-obox, line 51: I have added the additional classnames names for CSS specificity reasons. 

I need to overwrite this general styling code: 
.elementor .elementor-widget:not(.elementor-widget-text-editor):not(.elementor-widget-theme-post-content) figure {
    margin: 0;
}

louiswol94:
Hello @hein-obox Please check RTL.

Production (Image position left):
Screenshot 2022-04-26 at 10 44 17

This PR (Image position left):
Screenshot 2022-04-26 at 10 39 51

The spacing next your image is missing as it is directly next to the heading+text.

Also the position control says Left/Right because it is an old control. It does not say Start/End. So because it explicitly says "Left", I think your image on the wrong side of the screen on RTL. Your side would have been correct if the control said "Start". @avivu, correct?

avivu:
> Hello @hein-obox Please check RTL.
> 
> Production (Image position left): Screenshot 2022-04-26 at 10 44 17
> 
> This PR (Image position left): Screenshot 2022-04-26 at 10 39 51
> 
> The spacing next your image is missing as it is directly next to the heading+text.
> 
> Also the position control says Left/Right because it is an old control. It does not say Start/End. So because it explicitly says "Left", I think your image on the wrong side of the screen on RTL. Your side would have been correct if the control said "Start". @avivu, correct?

Yes, you are correct @louiswol94 
We should keep it as today.

hein-obox:
@louiswol94 

RTL Flex direction fixed.

includes/widgets/image-box.php
louiswol94: @hein-obox Maybe it is worth adding a `tablet_default`? I see most places in Elementor has default/tablet_default/mobile_default?

assets/dev/scss/frontend/widgets/image-box.scss
louiswol94: @hein-obox Please elaborate on the comment in the CSS.

louiswol94: @hein-obox The original code from line 76 has `margin-bottom: 15px;`?

includes/widgets/image-box.php
hein-obox: I am hesisant to do this for BC reasons. 

@avivu Please advise.

assets/dev/scss/frontend/widgets/image-box.scss
hein-obox: @louiswol94 

The margin-bottom of 15px is overwritten by the :not fix for Gutenberg. 

.elementor .elementor-widget:not(.elementor-widget-text-editor):not(.elementor-widget-theme-post-content) figure {
margin: 0;
}

I have followed the 0px, to keep it BC compatible. 

But I am not a 100% sure if we should stick to 0px or should use 15px. 

Aviv stressed the BC aspect.

includes/widgets/image-box.php
avivu: @hein-obox @louiswol94 
If there is a default today, we should keep it, but if there is not a default today, we should leave it as it is.

assets/dev/scss/frontend/widgets/image-box.scss
louiswol94: @hein-obox What I mean is the old code from this widget is `15px`. But you changed it to `0` here. Should you not keep the 15px?

hein-obox: @louiswol94 

When you inspect, you'll see this code: 
.elementor .elementor-widget:not(.elementor-widget-text-editor):not(.elementor-widget-theme-post-content) figure {
margin: 0;
}

That forcing a margin-bottom: 0; value, because of the CSS specificity. That is the BC version I am dealing with. That is how it works in the current production version.

I am not introducing the 0, I am merely formalising it. 

But preferably I would use the 15px version. 

Let's double check with @avivu: 
- Originally on mobile we had a margin bottom of 15px. 
- Last year this value changed to 0px, because of some Gutenberg styling (this seems like a bug to me, but this is what is currently how the production version works). 
- In my latest code I am using that margin-bottom: 0; value. Meaning: nothing will change on production, which is good from BC perspective. But now I am kind of formalising a bug.

louiswol94: @hein-obox I understand. But because our code was on 15px we should just leave it on 15px in my view. The 0px that was added in the global css probably affected this widget by mistake and does not mean we should `change` this widget's code to 0px.

includes/widgets/image-box.php
louiswol94: @hein-obox @avivu The control was not responsive before so `tablet_default` would not have been there.
What I meant was that Hein JUST added `mobile_default`, and nothing for `tablet_default` after making the control responsive, and I see in Elementor we usually specify BOTH. But I think it is okay, Udi will pick it up if it is a problem.
Thanks!

louiswol94: @hein-obox Did you see this question?

assets/dev/scss/frontend/widgets/image-box.scss
louiswol94: @hein-obox This is a tricky one. Because the original widget intended the `15px`. But the Global CSS of `0px` has been added 2 years ago already, but no one noticed it affected this widget also. So now I guess we have to stick to `0px` so that the Screenshotter can pass.

FYI: https://github.com/elementor/elementor/pull/11998

hein-obox: @louiswol94 

This is fascinating. 

Aviv gave me the instruction to use the margin-bottom: 15px;

@avivu 

It is up to you and your colleagues to decide about this.

hein-obox: @louiswol94 @avivu 
Final verdict from Aviv: we have decided to use 15px. 
Someone will need to overrule the screenshotter error.

PR #18516, opened 2022-04-27
Fix: Global Colors - Gradient control doesn't work on frontend [ED-3517] (Fix #13288)

### LINKED ISSUE #13288: Bug with global values in linear-gradient css function

udidol:
## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

* Fix: Gradient control doesn't work on frontend when using Global Colors (Fixes #13288)


core/files/css/base.php
```diff
@@ -673,6 +678,38 @@ protected function add_control_style_rules( array $control, array $values, array
...
+	 * fetches and returns the global default value. Otherwise, it returns null.
+	 *
+	 * @since 3.7.0
+	 * @access private
+	 *
+	 * @param $control
+	 * @return string|null
+	 */
+	private function get_control_global_default_value( $control ) {
+		if ( ! empty( $control['global']['default'] ) ) {
```
KingYes, line 694: Try `if ( empty( $control['global']['default'] ) ) {` and exit from the function early..

core/files/css/base.php
```diff
@@ -673,6 +678,38 @@ protected function add_control_style_rules( array $control, array $values, array
...
+			// If the control value is empty, and the control has a global default set, fetch the global value and use it.
+			$global_enabled = false;
```
KingYes, line 696: What's the different between `false` to `null` here?

KingYes (changes requested):

core/files/css/base.php
udidol: The difference is that the following methods return booleans that override this variable's value. So I prefer to instantiate it as a boolean as well to keep the typing identical.

core/files/css/base.php
udidol: Done

core/files/css/base.php
```diff
@@ -442,6 +442,11 @@ public function parse_property_placeholder( array $control, $value, array $contr
 			$value = call_user_func( $value_callback, $control );
 		}
 
+		// If the control value is empty, check for global default.
```
matipojo, line 445: Please add a screenshot test

matipojo (changes requested):

tests/screenshotter/config/buttons.json
```diff
@@ -1 +1 @@
...
...lements":[],"isInner":false,"widgetType":"text-editor","elType":"widget"},{"id":"502b7d2d","settings":{"text":"Gradient background","background_background":"gradient"},"elements":[],"isInner":false,"widgetType":"button","elType":"widget"},{"id":"4b0a1e2f","settings":{"text":"Radius","background_color":"#000000","border_radius":{"unit":"px","top":"50","right":"50","bottom":"50","left":"50","isLinked":true}},"elements":[],"isInner":false,"widgetType":"button","elType":"widget"},{"id":"cc111f6","settings":{"text":"Shadow","background_color":"#000000","button_box_shadow_box_shadow_type":"yes","button_box_shadow_box_shadow":{"horizontal":0,"vertical":0,"blur":10,"spread":5,"color":"rgba(0,0,0,0.5)"}},"elements":[],"isInner":false,"widgetType":"button","elType":"widget"},{"id":"774dc81b","settings":{"text":"Padding","background_color":"#000000","button_box_shadow_box_shadow":{"horizontal":0,"vertical":0,"blur":10,"spread":5,"color":"rgba(0,0,0,0.5)"},"text_padding":{"unit":"px","top":"30","right":"30","bottom":"30","left":"30","isLinked":true}},"elements":[],"isInner":false,"widgetType":"button","elType":"widget"}],"isInner":false,"elType":"column"}],"isInner":false,"elType":"section"}]}
```
matipojo, line 1: Please move to global setting test

matipojo (changes requested):

tests/screenshotter/config/buttons.json
udidol: Done.

PR #18518, opened 2022-04-27
Internal: Remove 3.7.0 deprecations from Core [ED-6921]

rami-elementor:


includes/base/element-base.php
```diff
@@ -170,24 +170,6 @@ final public function enqueue_styles() {
...
-	final public static function add_edit_tool() {}
-
-	/**
-	 * @since 2.2.0
-	 * @deprecated 2.6.0
-	 * @access public
-	 * @static
-	 */
-	final public static function is_edit_buttons_enabled() {
-		return get_option( 'elementor_edit_buttons' );
```
rami-elementor, line 188: @matipojo Should we also delete this option from the DB?

matipojo:
Currently, we will not remove PHP deprecated code, due to the risk for 3rd party plugins.

PR #18569, opened 2022-05-03
Fix: Container - Sometimes changing a control breaks the Editor [ED-7100]

StyleShit:


Nevoss (approved):

matipojo (changes requested):
Please cover with tests

PR #18566, opened 2022-05-03
Tweak: Uploads Manager - Added filters to allow modifying upload temp paths (#18565) [ED-7410]

patrick-leb:
## PR Checklist

- [X] The commit message follows our guidelines:  https://github.com/elementor/elementor/blob/master/.github/CONTRIBUTING.md

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [X] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## Summary

This PR can be summarized in the following changelog entry:

* adds filters in the uploads-manager to control temp paths

## Description
An explanation of what is done in this PR

* see #18565 - adds 2 filters to control where the temp file and temp dir are created.

## Test instructions
This PR can be tested by following these steps:

* `add_filter('elementor/files/temp-dir', function($path){ return '/my/new/path'; }`
* `add_filter('elementor/files/temp-file', function($filepath){ return '/my/new/file/path'; }`

## Quality assurance

- [X] I have tested this code to the best of my abilities
- [ ] I have added unittests to verify the code works as intended
- [ ] Docs have been added / updated (for bug fixes / features)

Fixes #18565

PR #18577, opened 2022-05-04
Tweak: System Info Server Reporter - Added filter for paths checked for write permissions [ED-7655]

### LINKED ISSUE #17255: Improper WordPress root write permissions check
ℹ️  ***For those impacted by this, see the workaround at the end of this report.***
 
 Please note this is not host-specific -- while hosts like WordPress.com and Pressable encounter this issue, it can occur on ANY hosting environment configured in a similar way.

## Prerequisites

- [x] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [x] The issue still exists against the latest stable version of Elementor.

## Description

Elementor checks for write permissions to the core WordPress directory using ABSPATH.  That specifically isn't the issue but there are some related problems that it causes.

https://github.com/elementor/elementor/blob/7d2ab814abea1424627a5b1d5dd63dafcf785820/modules/system-info/reporters/server.php#L259

### 1) Why does Elementor check for this? It shouldn't be necessary.  

Elementor shouldn't need write access to WordPress core files.

Ideally, this would be checking to ensure `WP_CONTENT_DIR` is writeable, not the core files directory via `ABSPATH`.

Some hosts symlink and protect the core WordPress directory but leave the rest of the web root and wp-content accessible and writeable.  This is all Elementor should need.

On hosts that do symlink the root, Elementor generally _**works without issue**_ but the system report displays an error. 

`There are some writing permissions issues with the following directories/files: – WordPress root directory`

Screen Shot on 2021-12-20 at 15:52:05

### 2) The check causes the Kit Library import functions fail.

Since the Kit Library importer requires the write permissions check to pass, the kit importer will fail with `Elementor is not authorized to read or write from this file. Contact your site host.`

https://github.com/elementor/elementor/blob/7d2ab814abea1424627a5b1d5dd63dafcf785820/core/app/modules/import-export/iterator.php#L51

## Workaround

If we **_manually bypass this check, the kit importer will work without issue_** -- all pages and media are imported and the error will be dismissed.

A quick way to bypass this is to edit `wp-content/plugins/elementor/modules/system-info/reporters/server.php` and remove the `ABSPATH => 'WordPress root directory',` line from `$paths_to_check` 

## Isolating the problem

- [x] This bug happens with only Elementor plugin active (and Elementor Pro).
- [x] This bug happens with a default WordPress theme active.
- [x] I can reproduce this bug consistently using the steps above.

## Environment

Various hosting environments that symlink or protect the WordPress core files.

gilzow:
Adds filter to adjust the initial list of directories to check for write permissions.
In some systems, ABSPATH will not be writable, nor does it need to be. In other situations, plugins for elementor may need to add directories to the write check. This adds a filter after initialization of the `$paths_to_check` variable to adjust its contents.

## PR Checklist

- [X] The commit message follows our guidelines:  https://github.com/elementor/elementor/blob/master/.github/CONTRIBUTING.md

## PR Type

- [X] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

* Adds filter `elementor/modules/system-info/reporters/server/initial_paths_to_check`  to adjust the initial list of directories to check for write permissions.

## Description

* In some systems, `ABSPATH` will not be writable, nor does it need to be. In other situations, plugins for elementor may need to add directories to the write check. This adds a filter after initialization of the `$paths_to_check` variable to adjust its contents.

## Quality assurance

Fixes #17255


patrick-leb:
@udidol Thank you for merging https://github.com/elementor/elementor/pull/18566 - can this one be looked at as well, it's really complementary?

Thanks.

modules/system-info/reporters/server.php
```diff
@@ -300,6 +300,8 @@ public function get_write_permissions() {
 		$paths_to_check = [
 			ABSPATH => 'WordPress root directory',
 		];
+		
+		$paths_to_check = apply_filters('elementor/modules/system-info/reporters/server/initial_paths_to_check', $paths_to_check);
```
udidol, line 304: It makes more sense to do this in line 325

udidol (changes requested):

modules/system-info/reporters/server.php
gilzow: I assume you actually meant [line 323](https://github.com/elementor/elementor/blob/master/modules/system-info/reporters/server.php#L323)? 

Sure? I can move the filter to occur right before you start looping through the array, if that's how you would prefer it. I just need some way to modify the list.

udidol: Yes I believe editing it once it's fully populated is the way to go.

udidol (changes requested):

modules/system-info/reporters/server.php
gilzow: Done. Please check updated version

modules/system-info/reporters/server.php
```diff
@@ -321,6 +321,13 @@ public function get_write_permissions() {
 			$paths_to_check[ $htaccess_file ] = '.htaccess file';
 		}
 
+		/**
+		 * Filters the list of files/directories that should be checked for write permissions
+		 *
+		 * @param array $paths_to_check 
+		 */
+		$paths_to_check = apply_filters('elementor/modules/system-info/reporters/server/initial_paths_to_check', $paths_to_check);
```
udidol, line 329:
```suggestion
		$paths_to_check = apply_filters('elementor/system-info/reporters/server/paths_to_check', $paths_to_check);
```

udidol (dismissed):

modules/system-info/reporters/server.php
udidol: @gilzow Please fix PHP Lint issues, see CI test runs at the bottom of the Conversation screen in this PR

gilzow:
Suggestion added, spacing added/removed per linter.

udidol:
@gilzow 
We have looked into it and decided to just replace this check in the Import/Export iterator with a specific check for the WP Uploads directory. I am leaving this PR open for now, once I see that the updated code is merged I will update here.

This way or the other, your issue will be taken care of soon.

udidol:
The problem addressed in this PR is solved in the upcoming Elementor Core v3.8.0.

PR #18614, opened 2022-05-10
 Internal: Docs - Container [ED-7047]

StyleShit:


docs/modules/container-converter/container-converter.md
```diff
@@ -0,0 +1,122 @@
...
+
+This is a migration tool for Elementor that helps users migrating from the older Sections/Columns layouts to the new and
+updated Container layouts.
+
+It's a standalone module that injects conversion buttons to each Section & Page, and will (or at least should) be
+removed in the future when the Container element will be a stable feature and become the default building block.
+
+It registers 2 commands:
+
+- `$e.run( 'container-converter/convert' )` - Converts a single "convertable" element and its descendants to a Containers.
```
ManorHazaz, line 17:
```suggestion
- `$e.run( 'container-converter/convert', { container } )` - Converts a single "convertable" element and its descendants to a Containers.
```

Nevoss (approved):

docs/includes/elements/container.md
```diff
@@ -74,3 +76,149 @@ Another aspect is the new elements hierarchy. We needed to decide which elements
...
+```
+
+This variables reset method is required in order to avoid style leakage as stated above.
+
+#### Guidelines:
+
+- Name the CSS variable after the corresponding CSS property.
+- Try to always use `{{WRAPPER}}` as the selector, and consume the variable under the proper selector in the SCSS file
+  instead of putting a long selector in the PHP file. 
+- Make sure the default value you set for the property is the real default according to
```
matipojo, line 129: If the element has a default value (for example, align=left) it should be used for the CSS variable initial value.

docs/modules/container-converter/container-converter.md
```diff
@@ -0,0 +1,122 @@
...
+
+## How It Works:
+
+The PHP side of things is fairly easy. It just handles injecting the conversion buttons and loading the assets.
+
+On the frontend from the other hand (pun intended), things get more complicated.
+
+We can generally divide that into 2 parts:
+
+1. **Migrator** - Handles the actual migration logic (duh...) using a simple API.
```
matipojo, line 30:
```suggestion
1. **Migrator** - Handles the actual migration logic  using a simple API.
```

matipojo (changes requested):

docs/includes/elements/container.md
StyleShit: Yup, done.

docs/modules/container-converter/container-converter.md
StyleShit: 🤣 
Done

PR #18680, opened 2022-05-18
Internal: Better handling safe-mode cookie [ED-7105]

nahmana:


modules/safe-mode/module.php
```diff
@@ -101,13 +101,13 @@ public function enable_safe_mode() {
...
 
-		$token = md5( wp_rand() );
+		$token = hash( 'sha256', wp_rand() );
 
 		// Only who own this key can use 'elementor-safe-mode'.
 		update_option( self::OPTION_TOKEN, $token );
 
 		// Save for later use.
-		setcookie( self::OPTION_TOKEN, $token, time() + HOUR_IN_SECONDS, COOKIEPATH );
+		setcookie( self::OPTION_TOKEN, $token, time() + HOUR_IN_SECONDS, COOKIEPATH, '', true, true );
```
StyleShit, line 110:
```suggestion
		setcookie( self::OPTION_TOKEN, $token, time() + HOUR_IN_SECONDS, COOKIEPATH, '', is_ssl(), true );
```

modules/safe-mode/module.php
```diff
@@ -131,7 +131,7 @@ public function disable_safe_mode() {
 		delete_option( 'elementor_safe_mode_created_mu_dir' );
 
 		delete_option( self::OPTION_TOKEN );
-		setcookie( self::OPTION_TOKEN, '', 1 );
+		setcookie( self::OPTION_TOKEN, '', 1, '', '', true, true );
```
StyleShit, line 134: same as above

StyleShit (changes requested):

modules/safe-mode/module.php
nahmana: Done.

modules/safe-mode/module.php
nahmana: done.

StyleShit (approved):

PR #18741, opened 2022-05-25
Fix: Insert library button is clickable while inserting a template [ED-7008]

ghost:


assets/dev/js/admin/admin.js
```diff
@@ -396,12 +398,15 @@ import FilesUploadHandler from '../editor/utils/files-upload-handler';
 				$( '#elementor-import-template-area' ).toggle();
 			} );
 
-			$importNowButton.on( 'click', ( event ) => {
```
yossi-elementor, line 399: Shouldn't the onlick event remain?

yossi-elementor (changes requested):

yossi-elementor (approved):

PR #18779, opened 2022-05-29
Tweak: Add "Skip and Deactivate" text button to deactivation survey

mark-elementor:


assets/dev/js/admin/admin-feedback.js
```diff
@@ -52,15 +52,13 @@
 								callback: self.sendFeedback.bind( self ),
 							} );
 
-							if ( ! elementorAdmin.config.feedback.is_tracker_opted_in ) {
```
ManorHazaz, line 55: Maybe we also should delete the `is_tracker_opted_in` from the config if we no longer using it?

ghost (changes requested):

ghost (approved):

core/admin/feedback.php
```diff
@@ -182,18 +181,6 @@ public function ajax_elementor_deactivate_feedback() {
 		wp_send_json_success();
 	}
 
-	/**
-	 * @since 2.3.0
-	 * @access protected
-	 */
-	protected function get_init_settings() {
```
yossi-elementor, line 189: If the whole method was removed, shouldn't we handle the places that invoked this method?

yossi-elementor (changes requested):

core/admin/feedback.php
yossi-elementor: Answered: no need.

yossi-elementor (approved):

PR #18784, opened 2022-05-29
Internal: Elements - A method to sanitize settings before save [ED-7250]

matipojo:


includes/base/element-base.php
```diff
@@ -543,6 +543,26 @@ public function get_raw_data( $with_html_content = false ) {
 		];
 	}
 
+	public function get_data_for_save() {
```
StyleShit, line 546: it reminds me the `get_settings_for_display()` thing that nobody really knows how it works...
maybe we need a better name?

includes/base/element-base.php
```diff
@@ -803,6 +823,14 @@ protected function get_initial_config() {
 		return $config;
 	}
 
+	/**
+	 * A Base method for sanitizing the settings before save.
+	 * This method is meant to be overridden by the element.
+	 */
+	protected function on_save( array $settings ) {
+		return $settings;
```
StyleShit, line 831: don't you want to add some generic sanitizations?

StyleShit (changes requested):

includes/base/element-base.php
matipojo: It's OK for now

includes/base/element-base.php
matipojo: It's an infra only PR

PR #18813, opened 2022-06-01
Fix: Nested Tabs - Duplicate a Tab [ED-7328]

iNewLegend:


matipojo (changes requested):
Please cover with tests

PR #18928, opened 2022-06-13
Internal: Hide 'nested-elements' experiment [ED-2256]

iNewLegend:


tests/playwright/pages/wp-admin-page.js
```diff
@@ -81,6 +81,16 @@ module.exports = class WpAdminPage extends BasePage {
 		for ( const [ id, state ] of Object.entries( experiments ) ) {
 			const selector = `#${ prefix }-${ id }`;
 
+			// Try make visible the element - Since some of the experiments are may be hidden for the user,
+			// but actually exist and need to be tested.
```
iNewLegend, line 85:
```suggestion
			// Try to make the element visible - Since some of the experiments are may be hidden for the user,
			// but actually exist and need to be tested.
```

PR #18961, opened 2022-06-19
Internal: ESLint - Implement a new set of rules [ED-7408]

maork-elementor:


.eslintrc.js
```diff
@@ -134,6 +134,8 @@ module.exports = {
 		} ],
 		'react/react-in-jsx-scope': 'off',
 		'babel/semi': 1,
+		'spaced-comment': [ 'error', 'always', { markers: [ '!' ] } ],
```
matipojo, line 137: Why not?
```suggestion
		'spaced-comment': 'error',
```

matipojo (changes requested):

.eslintrc.js
maork-elementor: @matipojo 

If "always" then the // or /* must be followed by at least one whitespace.
The "markers" value is an array of string patterns which are considered markers for docblock-style comments.
The difference between a marker and an exception is that a marker only appears at the beginning of the comment whereas exceptions can occur anywhere in the comment string.

What do you think us the right chiose for us?

maork-elementor:
PR open to 3.7

PR #19023, opened 2022-06-28
New: Nested Tabs - Style title section [ED-7333]

hein-obox:


modules/tabs-v2/widgets/tabs-v2.php
```diff
@@ -523,6 +526,281 @@ protected function register_controls() {
...
+						'{{WRAPPER}}' => '--tabs-v2-title-typography-letter-spacing: {{SIZE}}{{UNIT}};',
+					],
+				],
+				'word_spacing' => [
+					'selectors' => [
+						'{{WRAPPER}}' => '--tabs-v2-title-typography-word-spacing: {{SIZE}}{{UNIT}};',
+					],
+				],
+			],
+		] );
```
udidol, line 596: I like the creativity here and it would work, but I prefer to leave it with a simple selector.
The reason: If at any point we would want to add another control to the Typography group control (e.g. `text-stroke`), we would have to go to all the places that implement individual CSS variable selectors and add that control to all of them.

A single selector removes the need for that, at the price of not supporting CSS variables, I'm aware of that.

modules/tabs-v2/widgets/tabs-v2.php
```diff
@@ -523,6 +526,281 @@ protected function register_controls() {
...
+		$this->start_controls_tab(
+			'title_hover',
+			[
+				'label' => esc_html__( 'Hover', 'elementor' ),
+			]
+		);
+
+		$this->add_group_control( Group_Control_Typography::get_type(), [
+			'name' => 'title_typography_hover',
+			'fields_options' => [
```
udidol, line 636: Same as above

modules/tabs-v2/widgets/tabs-v2.php
```diff
@@ -523,6 +526,281 @@ protected function register_controls() {
...
+		$this->start_controls_tab(
+			'title_active',
+			[
+				'label' => esc_html__( 'Active', 'elementor' ),
+			]
+		);
+
+		$this->add_group_control( Group_Control_Typography::get_type(), [
+			'name' => 'title_typography_active',
+			'fields_options' => [
```
udidol, line 723: Same as above

udidol (changes requested):

modules/tabs-v2/widgets/tabs-v2.php
hein-obox: updated

modules/tabs-v2/widgets/tabs-v2.php
hein-obox: updated. 

Credits to Leonid! He had done this setup.

udidol (approved):

PR #19034, opened 2022-06-29
Fix: Fatal error while uploading & deleting files [ED-7394]

maork-elementor:


includes/managers/image.php
```diff
@@ -150,7 +150,9 @@ public function get_lightbox_image_attributes( $id ) {
 
 	private function delete_custom_images( $post_id ) {
 		$image_meta = wp_get_attachment_metadata( $post_id );
-
+		if ( empty( $image_meta ) ) {
+			return;
+		}
```
matipojo, line 155:
```suggestion
		}

```

PR #19015, opened 2022-06-27
Fix: Nested Elements - Sync repeater with containers [ED-7545]

Nevoss:

PR #19046, opened 2022-06-30
Fix: Nested Element - Lock tab items [ED-7546]

Nevoss:


Nevoss:
Depends on #19015

modules/nested-elements/assets/js/editor/nested-repeater/models/nested-model-base.js
```diff
@@ -22,7 +22,10 @@ export default class NestedModelBase extends elementor.modules.elements.models.E
 
 		return 'container' === childElType &&
 			'widget' === parentElType &&
-			isWidgetSupportNesting( this.get( 'widgetType' ) );
+			isWidgetSupportNesting( this.get( 'widgetType' ) ) &&
+			// When creating a container for the tabs widget specifcly from the repeater, the container should be locked,
+			// so only containers that are locked (created from the repeater) can be inside the tabs widget.
```
matipojo, line 27: // When creating a container for the tabs widget specifically from the repeater, the container should be locked,
			// so only containers that are locked (created from the repeater) can be inside the tabs widget.
```

PR #19065, opened 2022-07-03
Internal: CI - PHPCS not working [ED-7589]

Nevoss:


core/app/app.php
```diff
@@ -16,6 +16,7 @@ class App extends BaseApp {
 
 	const PAGE_ID = 'elementor-app';
 
+
```
udidol, line 19: remove

Nevoss, line 19:
```suggestion
```

Issue #18880, opened 2022-06-08
🐞 Bug in container (Accordion and toggle widget)


79ho3ein:
### Prerequisites

- [X] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [X] The issue still exists against the latest stable version of Elementor.

### Description

hi

If in a container, the direction is in the form of columns and Align Items in the center, the Accordion and toggle widget will have a problem. Please watch the video below :

https://user-images.githubusercontent.com/77163248/172606942-96112878-a9db-41fb-81ee-32168b23ad7f.mp4

As you can see in the video above, opening and closing the accordion encounters bug. **Please note that this bug is present in Accordion and toggle widgets.**

### Steps to reproduce

Video above

### Isolating the problem

- [X] This bug happens with only Elementor plugin active (and Elementor Pro).
- [X] This bug happens with a Blank WordPress theme active ([Hello theme](https://wordpress.org/themes/hello-elementor/)).
- [X] I can reproduce this bug consistently using the steps above.

### System Info

-

79ho3ein:
This bug is also present in the Price List widget

79ho3ein:
thank you @shilo-ey

shilo-ey:
Hey,

This issue has been resolved in Elementor v3.7.0

Feel free to update

Thanks!

sebask63:
Is it a must-have Elementor Pro for this feature?
I can't get if in the free version of Elemento v3.7

sebask63:
Am happy to share one of my works with Elementor; [www.katikatisafaris.com](www.katikatisafaris.com)

PR #19144, opened 2022-07-13
Fix: Widgets disappear when the direction is set to a column in Container (#18880) [ED-7524]

hein-obox:


includes/controls/groups/flex-container.php
```diff
@@ -47,8 +47,14 @@ protected function init_fields() {
 				],
 			],
 			'default' => '',
+			'selectors_dictionary' => [
+				'row' => '--flex-direction: row; --container-widget-width: initial;',
```
udidol, line 51: Add a comment here explaining that this is for handling widgets that get an undefined width in column mode.

includes/controls/groups/flex-container.php
```diff
@@ -47,8 +47,14 @@ protected function init_fields() {
 				],
 			],
 			'default' => '',
+			'selectors_dictionary' => [ // The `--container-widget-width` is used for handling widgets that get an undefined width in column mode (ED-7524 & ED-7604).
```
udidol, line 50:
```suggestion
			// The `--container-widget-width` CSS variable is used for handling widgets that get an undefined width in column mode.
			'selectors_dictionary' => [
```

udidol (approved):

PR #19153, opened 2022-07-14
Tweak: Widgets overlapping issues on mobile when using containers [ED-7604]

hein-obox:


assets/dev/scss/frontend/widgets/video.scss
```diff
@@ -47,3 +47,11 @@
 		}
 	}
 }
+
+.e-container {
+
+	> .elementor-widget-video {
+		// Fix - Video widget has an undefined height on mobile using 'flex column wrap' inside the flex container (ED-7604).
```
hein-obox, line 54: I have added the `>` symbol to make sure we are only targeting the direct child. 

This could be relevant e.g. for the nested tab widget, where we can have a multi levels of containers and widgets which need to be handled seperately.

includes/controls/groups/flex-container.php
```diff
@@ -47,8 +47,14 @@ protected function init_fields() {
 				],
 			],
 			'default' => '',
+			'selectors_dictionary' => [
+				'row' => '--flex-direction: row; --container-widget-width: initial;',
```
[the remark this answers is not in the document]
hein-obox: updated

includes/controls/groups/flex-container.php
```diff
@@ -47,8 +47,14 @@ protected function init_fields() {
 				],
 			],
 			'default' => '',
+			'selectors_dictionary' => [ // The `--container-widget-width` is used for handling widgets that get an undefined width in column mode (ED-7524 & ED-7604).
```
udidol, line 50:
```suggestion
			// The `--container-widget-width` CSS variable is used for handling widgets that get an undefined width in column mode.
			'selectors_dictionary' => [
```

udidol (approved):

PR #19197, opened 2022-07-19
Tweak: Added EM unit to Elements Gap in Container widget [ED-7529]

louiswol94:


TzviRabinovitch (dismissed):

includes/controls/groups/flex-container.php
```diff
@@ -167,7 +167,7 @@ protected function init_fields() {
 					'max' => 100,
 				],
 			],
-			'size_units' => [ 'px', '%', 'vw' ],
+			'size_units' => [ 'px', '%', 'vw', 'em' ],
 			'selectors' => [
```
hein-obox, line 171: Can we add a range e.g. from 0 to 50?

hein-obox (changes requested):

includes/controls/groups/flex-container.php
hein-obox: @avivu 

All unit sizes have range. I would expect a range for 'em' as well. What do you think?

louiswol94: Done

hein-obox (approved):

PR #19207, opened 2022-07-20
Fix: User language is not being considered and not acting as expected [ED-7748]

StyleShit:


tests/phpunit/elementor/core/utils/test-force-locale.php
```diff
@@ -0,0 +1,42 @@
...
+	public function test_force() {
+		// Arrange.
+		$force_locale = new Force_Locale( 'he_IL' );
+
+		// Act.
+		$force_locale->force();
+
+		// Assert.
+		$this->assertEquals( 'לחץ כאן', __( 'Click here', 'elementor' ) );
+	}
```
Nevoss, line 29: Please clear after the test, to remove the filter (just use your class reset)

core/utils/force-locale.php
```diff
@@ -0,0 +1,69 @@
...
+	exit; // Exit if accessed directly.
+}
+
+/**
+ * Force translation to use a specific locale.
+ *
+ * A hacky class to force any translation functions in the call-stack between the
+ * `force()` & `reset()` methods to use a specific locale.
+ */
+class Force_Locale {
```
Nevoss, line 14: Maybe add some doc block to the properties and the methods?

Nevoss (changes requested):

core/utils/force-locale.php
StyleShit: Done

Nevoss (dismissed):

includes/managers/widgets.php
```diff
@@ -374,6 +376,10 @@ public function ajax_get_widgets_default_value_translations( array $data = [] )
 				return ! empty( $widget['controls'] );
 			} )
 			->all();
+
+		$force_locale->restore();
```
matipojo, line 380: If you are able to force the language for a specific part, why do you need the AJAX?

matipojo (changes requested):

includes/managers/widgets.php
StyleShit: Because of the cache in `Controls_Stack::get_stack()` that will force us to add another parameter to the function, and it'll override the controls default values so we'll need to clean after that in order to prevent other `get_stack()`s from breaking.

PR #19211, opened 2022-07-20
Tweak: Add a hook to get manifest data in import CLI command [ED-7671]

udidol:
## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

* Tweak: Add a hook to get manifest data in Kit import CLI command

### DIFF
--- a/core/app/modules/import-export/wp-cli.php
+++ b/core/app/modules/import-export/wp-cli.php
@@ -155,6 +155,17 @@ class Wp_Cli extends \WP_CLI_Command {
 			$manifest_data = $this->get_manifest_data( $import_settings['session'] );
 			$manifest_data = $import->adapt_manifest_structure( $manifest_data );
 
+			/**
+			 * Import Export Manifest Data
+			 *
+			 * Allows 3rd parties to read and edit the kit's manifest before it is used.
+			 *
+			 * @since 3.7.0
+			 *
+			 * @param array $manifest_data The Kit's Manifest data
+			 */
+			$manifest_data = apply_filters( 'elementor/import-export/wp-cli/manifest_data', $manifest_data );
+
 			if ( isset( $manifest_data['plugins'] ) ) {
 				$successfully_imported_plugins = $this->import_plugins( $manifest_data['plugins'] );
 


core/app/modules/import-export/wp-cli.php
```diff
@@ -155,6 +155,17 @@ public function import( array $args, array $assoc_args ) {
...
+			/**
+			 * Import Export Manifest Data
+			 *
+			 * Allows 3rd parties to read and edit the kit's manifest before it is used.
+			 *
+			 * @since 3.8.0
+			 *
+			 * @param array $manifest_data The Kit's Manifest data
+			 */
+			apply_filters( 'elementor/import-export/manifest_data', $manifest_data );
```
ManorHazaz, line 167:
```suggestion
			apply_filters( 'elementor/import-export/wp-cli/manifest_data', $manifest_data );
```

Maybe you should be more specific that the filter is available only in the CLI?

ghost (changes requested):

core/app/modules/import-export/wp-cli.php
udidol: I agree. Done.

ghost (dismissed):

nivnoimanelementor (dismissed):
Usually a filter is used to return data, so in this case I would use an action, but other than that it seems fine to me. If it's okay by you, then great.

core/app/modules/import-export/wp-cli.php
```diff
@@ -155,6 +155,17 @@ public function import( array $args, array $assoc_args ) {
...
+			/**
+			 * Import Export Manifest Data
+			 *
+			 * Allows 3rd parties to read and edit the kit's manifest before it is used.
+			 *
+			 * @since 3.8.0
+			 *
+			 * @param array $manifest_data The Kit's Manifest data
+			 */
+			apply_filters( 'elementor/import-export/wp-cli/manifest_data', $manifest_data );
```
udidol, line 167:
```suggestion
			$manifest_data = apply_filters( 'elementor/import-export/wp-cli/manifest_data', $manifest_data );
```

udidol:
> Usually a filter is used to return data, so in this case I would use an action, but other than that it seems fine to me. If it's okay by you, then great.

You are right - I missed it. I added an overwrite of the manifest variable returning from the filter. good call @nivnoimanelementor

yossi-elementor (approved):

PR #19218, opened 2022-07-21
Fix: Elementor Top bar is not fully responsive when WP sidebar is collapsed [ED-7772]

StyleShit:


arielk:
@StyleShit @avivu what need to do with this PR?

PR #19226, opened 2022-07-21
Fix: Spacer widget is not working as expected [ED-7720]

hein-obox:


assets/dev/scss/frontend/widgets/spacer.scss
```diff
@@ -1,10 +1,35 @@
-.elementor-spacer-inner {
-	.e-container.e-container--row & {
-		width: var( --spacer-size );
```
hein-obox, line 3: I have redone this widget completely because the original .e-container.e-container--row/column classes are not responsive.

PR #19264, opened 2022-07-26
Tweak: Added new Layout section to the Widgets panel [ED-7528]

louiswol94:


includes/managers/elements.php
```diff
@@ -298,6 +298,14 @@ private function init_categories() {
 			],
 		];
 
+		if ( Plugin::instance()->experiments->is_feature_active( 'container' ) ) {
+			$this->categories = array_merge_recursive( [
+				'layout' => [
+					'title' => esc_html__( 'Layout', 'elementor' ),
+				],
```
hein-obox, line 305: I think for future use of this Layout outside of the context of the Flex Container, it might be an idea to remove the 'container experiment active' check, and to add this parameter: 

'hideIfEmpty' => true,

In this way the Layout will be hidden when the container is not active, but the usuage is extended beyond the container experiment.

includes/managers/elements.php
```diff
@@ -298,6 +298,14 @@ private function init_categories() {
 			],
 		];
 
+		if ( Plugin::instance()->experiments->is_feature_active( 'container' ) ) {
+			$this->categories = array_merge_recursive( [
```
hein-obox, line 302: Possibly remove 'if ( Plugin::instance()->experiments->is_feature_active( 'container' ) ) {'

includes/managers/elements.php
louiswol94: Works. Done.

includes/managers/elements.php
louiswol94: Done.

hein-obox (dismissed):

includes/managers/elements.php
```diff
@@ -298,6 +298,13 @@ private function init_categories() {
 			],
 		];
 
+		$this->categories = array_merge_recursive( [
```
udidol, line 301: Why add it like this and not directly to the array?

udidol (changes requested):
Also - please consult with @avivu about whether "Inner Section" should also be included in the "Layout" category, or just the Flex Container.

includes/managers/elements.php
louiswol94: @udidol Updated.

louiswol94:
> Also - please consult with @avivu about whether "Inner Section" should also be included in the "Layout" category, or just the Flex Container.

@udidol 
Feedback from Aviv:

Just the Container :slightly_smiling_face:
Inner will stay at the same place

udidol (approved):

PR #19267, opened 2022-07-26
Fix: Width and Elements gap values are not working [ED-7694]

hein-obox:


TzviRabinovitch (dismissed):

includes/elements/container.php
```diff
@@ -416,16 +434,29 @@ protected function register_container_layout_controls() {
...
+		}
+
 		$this->add_group_control(
 			Group_Control_Flex_Container::get_type(),
 			[
 				'name' => 'flex',
 				'selector' => '{{WRAPPER}}',
 				'fields_options' => [
 					'gap' => [
-						'label' => esc_html_x( 'Gap between elements', 'Flex Container Control', 'elementor' ),
```
louiswol94, line 426: @hein-obox Please just double check this label :-) I might be wrong, but I think you are changing it back. I had to update it to `Gap between elements` in another task. Maybe just pull release/3.7.0 in here to be sure.

hein-obox: @louiswol94 Fixed by merging

PR #19286, opened 2022-07-27
Fix: Widget width is not working as expected In Container [ED-7723]

hein-obox:


louiswol94 (changes requested):

assets/dev/scss/frontend/_global.scss
```diff
@@ -124,6 +124,27 @@
 	}
 }
 
+.elementor {
+
+	.elementor-widget.elementor-widget {
```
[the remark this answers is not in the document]
hein-obox: @louiswol94 

I have copied this trick from 
https://github.com/elementor/elementor/blob/7a4df168ee2a4c36e26f18835dd11337c97329bf/assets/dev/scss/frontend/_container.scss#L142. 

See https://stackoverflow.com/questions/29112807/does-repeating-a-classname-in-a-css-rule-increase-its-priority as well.

louiswol94 (dismissed):

assets/dev/scss/frontend/_global.scss
```diff
@@ -124,6 +124,17 @@
 	}
 }
 
+.elementor {
+
+	:is( .e-container ) {
+
+		// Hack to increase specificity and avoid using `!important`.
+		& > .elementor-widget.elementor-widget {
```
udidol, line 132: I am not sure if I like this. What happens in section/column? Why does it only happen in container?
Is anything different if you set the container to `flex-wrap: wrap`?

udidol (changes requested):

assets/dev/scss/frontend/_global.scss
hein-obox: @udidol 

Inside section/column: a max-width value has been set here: which is equal to the width. I really don't know why. I didn't want to change anything for the section/column.
https://github.com/marcelementor/elementor/blob/2fba924811524d429f1a83a8354bcc95c325abe6/includes/widgets/common.php#L259. 

It only happens when using `flex column`. Using `flex-wrap: wrap` doesn't make a difference.

assets/dev/scss/frontend/_container.scss
```diff
@@ -256,6 +256,13 @@
 			}
 		}
 	}
+
+	.elementor .elementor-section-wrap & {
+
```
hein-obox, line 261: In this way we can avoid the specifity hack with the repeated classname.

assets/dev/scss/frontend/_container.scss
```diff
@@ -256,6 +256,15 @@
 			}
 		}
 	}
+
+	// Hack to increase specificity and avoid using `!important`.
+	.elementor .elementor-section-wrap & {
+
+		// This was done to make sure that the widgets inside containers do not exceed the container's max width (ED-7723).
```
udidol, line 263:
```suggestion
		// This was done to make sure that the widgets inside containers do not exceed the container's max width.
```

udidol (approved):

PR #19294, opened 2022-07-28
Fix: Container outputs redundant CSS lines [ed-7727]

hein-obox:


includes/elements/container.php
```diff
@@ -294,9 +294,6 @@ protected function register_container_layout_controls() {
 			[
 				'type' => Controls_Manager::HIDDEN,
 				'default' => 'flex',
-				'selectors' => [
-					'{{WRAPPER}}' => '--display: {{VALUE}}',
-				],
 			]
```
hein-obox, line 297: I have decided not to remove the hidden field, because we will need in future when we add the display; grid; to this widget. 

For moment we could however decide to remove this field completely. It doesn't make much of difference.

TzviRabinovitch (dismissed):

includes/elements/container.php
udidol: You should remove the entire control. If we need it in the future we will use it.

udidol (changes requested):
@hein-obox 
1. Don't you need to replace the `--display` variable assignment by the control, which you are removing, with a default assignment in the CSS?
2. I see only `--display` removed in this PR. What about the other 2 properties? I don't see them handled here at all.

includes/elements/container.php
hein-obox: @udidol 

Removed

includes/elements/container.php
```diff
@@ -312,14 +301,8 @@ protected function register_container_layout_controls() {
 					'boxed' => esc_html__( 'Boxed', 'elementor' ),
 					'full' => esc_html__( 'Full Width', 'elementor' ),
 				],
-				'render_type' => 'ui',
-				'selectors' => [
-					'{{WRAPPER}}' => '{{VALUE}}',
-				],
```
hein-obox, line 318: I have used a prefix_class to avoid additional css. 
The container type isn't responsive, therefore I believe we can use a prefix. 
I don't think there is another way how we could reduce the css for the frontend selector.

The only alternative which I can think off: we can set 'boxed' or 'full width' as default value, and only add additional css for when the other option is selected.

udidol: No need for this. As I told you, we want to avoid prefix class.
All you have to do here is keep everything as is except remove the `default` value.
The CSS should only include a value of `--width: 100%` on the container by default

assets/dev/scss/frontend/_container.scss
```diff
@@ -90,6 +90,17 @@
 		pointer-events: none; // Prevent the placeholder from blocking the drop area.
 	}
 
+	&--width {
```
udidol, line 93: Not needed, see comment on `add_control()` call

includes/elements/container.php
```diff
@@ -428,9 +411,6 @@ protected function register_container_layout_controls() {
 						'placeholder' => $this->active_kit->get_settings_for_display( 'space_between_widgets' ),
 					],
 				],
-				'condition' => [
```
udidol, line 431: Why this?

udidol (changes requested):

includes/elements/container.php
hein-obox: This condition was link to the original hidden display: flex; control.

includes/elements/container.php
```diff
@@ -317,7 +306,6 @@ protected function register_container_layout_controls() {
 					'{{WRAPPER}}' => '{{VALUE}}',
 				],
 				'selectors_dictionary' => [
-					'boxed' => '--width: 100%;',
 					'full' => '--content-width: 100%;',
```
hein-obox, line 309: I have removed the default value 'boxed width 100%', because that value is set inside scss already.

includes/elements/container.php
hein-obox: Understood.

includes/elements/container.php
udidol: Another thing - currently, this control either sets `--width: 100%` or `--content-width: 100%`.
Now that `--width: 100%` is set by default in the CSS, choosing `full` will set the `--content-width` variable, but it won't modify the `--width` variable. 

The default CSS for containers is:
```
--content-width: Min( 100%, var( --container-max-width, 1140px ) );
--width: 100%;
```

So depending on the value of `--container-max-width`, there could be a clash here
Maybe it would be better that the `boxed` option would set the `--width` variable and also set `--content-width` to `inherit` or `initial` or whatever gives it its original value, and the opposite for the `full` option.

hein-obox: @udidol 

About your comment: 

> 
Maybe it would be better that the boxed option would set the --width variable and also set --content-width to inherit or initial or whatever gives it its original value, and the opposite for the full option.

For the `boxed` version we need this defition: `--content-width: Min( 100%, var( --container-max-width, 1140px ) );`
We shouldn't overwrite this.

hein-obox: @udidol 

For clarity: the only thing which I have removed is the `--width: 100%` for `boxed` generated by the widget controls. 
This won't change anything, because there is a default `--width` inside the html already.

udidol (approved):

PR #19320, opened 2022-08-01
Internal: Experiment - Introduced internal experiment [ED-7850]

louiswol94:


assets/dev/scss/admin/_experiments.scss
```diff
@@ -85,5 +85,10 @@
 			margin: 10px 0;
 			max-width: 820px;
 		}
+
+		&.elementor_experiment-loop {
+			display: none;
+			pointer-events: none;
+		}
```
hein-obox, line 92: Could we remove pointer-events?

PR #19354, opened 2022-08-04
Fix: Width and Elements gap values (ED-7694)

hein-obox:


includes/elements/container.php
```diff
@@ -289,7 +289,13 @@ protected function register_container_layout_controls() {
 			]
 		);
 
-		$min_affected_device = Breakpoints_Manager::BREAKPOINT_KEY_TABLET;
+		$active_breakpoints = Plugin::$instance->breakpoints->get_active_breakpoints();
+
+		if ( array_key_exists( 'mobile_extra', $active_breakpoints ) ) {
```
StuartElementor, line 294:
```suggestion
		if ( array_key_exists( Breakpoints_Manager::BREAKPOINT_KEY_MOBILE_EXTRA, $active_breakpoints ) ) {
```

TzviRabinovitch (dismissed):

udidol:
@hein-obox Change base to `develop`

includes/elements/container.php
```diff
@@ -338,6 +344,7 @@ protected function register_container_layout_controls() {
 				Breakpoints_Manager::BREAKPOINT_KEY_LAPTOP => $min_affected_device,
 				Breakpoints_Manager::BREAKPOINT_KEY_TABLET_EXTRA => $min_affected_device,
 				Breakpoints_Manager::BREAKPOINT_KEY_TABLET => $min_affected_device,
+				Breakpoints_Manager::BREAKPOINT_KEY_MOBILE_EXTRA => $min_affected_device,
```
udidol, line 347: I know this isn't you who did it, but I do not understand why create the `$width_control_settings` then use `array_merge()` in the `add_responsive_control()` call.
Just put everything in the `add_responsive_control()` please.

udidol (changes requested):

includes/elements/container.php
hein-obox: @udidol 

I believe the logic makes sense. 

The width controls are used both for the boxed width and for the full width version. The `array_merge()` makes sense, because in this way we only have to define the controls once, and they can be used twice.

udidol: I didn't see it used twice, OK.

udidol (approved):

PR #19380, opened 2022-08-08
Fix: Improter WordPress root write permissions check causes import failures [ED-5913]

ghost:


modules/system-info/reporters/server.php
```diff
@@ -410,4 +383,77 @@ public function get_elementor_library() {
...
+	 * @param array $paths_to_check
+	 * @param array $errors
+	 *
+	 * @return array {
+	 *      @type string value: Writing permissions status.
+	 *      @type bool   warning: Whether to display a warning. True if some required.
+	 * }
+	 */
+	public function check_write_permissions( array $paths_to_check, $errors = [] ) {
+		$write_problems = $errors;
```
udidol, line 402: No need for this renaming... You can just call the parameter `$write_problems`

udidol (changes requested):

modules/system-info/reporters/server.php
ManorHazaz: Done

modules/system-info/reporters/server.php
```diff
@@ -410,4 +383,75 @@ public function get_elementor_library() {
...
+	 *
+	 * @param array $paths_to_check
+	 * @param array $write_problems
+	 *
+	 * @return array {
+	 *      @type string value: Writing permissions status.
+	 *      @type bool   warning: Whether to display a warning. True if some required.
+	 * }
+	 */
+	public function check_write_permissions( array $paths_to_check, $write_problems = [] ) {
```
StyleShit, line 401: try to do something like this:
```PHP
function get_permissions( $paths ) {
    // ...

    return [
        'path-1' => [ 
            'read' => true, 
            'write' => false, 
            'execute' => true,
         ],
    ];
}
```

and change the original code to use this and return its old data

StyleShit (changes requested):

modules/system-info/reporters/server.php
```diff
@@ -410,4 +403,65 @@ public function get_elementor_library() {
 			'value' => 'Connected',
 		];
 	}
+
+	/**
+	 * Get permissions by keys. If empty keys, get all permissions.
+	 * Be aware that you can only use key paths that are defined in the get_paths() method.
+	 *
+	 * @since 3.7.0
```
udidol, line 411:
```suggestion
```

udidol (dismissed):

core/app/modules/import-export/iterator.php
```diff
@@ -46,10 +46,18 @@ public function __construct( array $settings ) {
 
 		$server = new Server();
 
-		$server_write_permissions = $server->get_write_permissions();
-
-		if ( $server_write_permissions['warning'] ) {
-			throw new \Error( self::NO_WRITE_PERMISSIONS_KEY );
+		$paths_to_check = [
+			$server::KEY_PATH_WP_CONTENT_DIR,
```
StyleShit, line 50:
```suggestion
			Server::KEY_PATH_WP_CONTENT_DIR,
```

easier to jump to the class

modules/system-info/reporters/server.php
```diff
@@ -291,8 +297,8 @@ public function get_mysql_version() {
 
 	/**
 	 * Get write permissions.
-	 *
-	 * Check whether the required folders has writing permissions.
+	 * Check whether the required paths have writing permissions.
+	 * Do not use this function if you want the check write permissions use get_permissions() instead.
```
StyleShit, line 301: Add deprecation notice instead

modules/system-info/reporters/server.php
```diff
@@ -410,4 +403,63 @@ public function get_elementor_library() {
 			'value' => 'Connected',
 		];
 	}
+
+	/**
+	 * Get permissions by keys. If empty keys, get all permissions.
+	 * Be aware that you can only use key paths that are defined in the get_paths() method.
```
StyleShit, line 409: it feels odd to me...
How do I check a specific path that I need?
Maybe we should have something like this that handles regular paths:

```PHP
public function get_permissions( $path ) { ... }

// ...

$server->get_permissions( 'path/to/file.txt' );
```

And something like this that handles any "system" paths (need to work on the name) that will always return everything (the filter logic will be on the consumer side):

```PHP
public function get_system_paths_permissions() { ... }

// ...

$permissions = $server->get_system_paths_permissions();

$permissions->filter( $key === Server::KEY_PATH_WP_DIR )
```
 What do you think?

modules/system-info/reporters/server.php
```diff
@@ -410,4 +403,63 @@ public function get_elementor_library() {
...
+			static::KEY_PATH_HTACCESS_FILE => file_exists( ABSPATH . '/.htaccess' ) ? ABSPATH . '/.htaccess' : '',
+			static::KEY_PATH_WP_CONTENT_DIR => WP_CONTENT_DIR,
+			static::KEY_PATH_UPLOADS_DIR => wp_upload_dir()['basedir'] ?? '',
+			static::KEY_PATH_ELEMENTOR_UPLOADS_DIR => ! empty( wp_upload_dir()['basedir'] ) ? wp_upload_dir()['basedir'] . '/elementor' : '',
+		];
+	}
+
+	/**
+	 * @param $path
+	 * @return array{read: bool, write: bool, execute: bool}
```
StyleShit, line 448: Great PHPDoc!

modules/system-info/reporters/server.php
```diff
@@ -410,4 +403,63 @@ public function get_elementor_library() {
...
+
+		return $permissions;
+	}
+
+	/**
+	 * Get all the paths that we might need to check permissions.
+	 *
+	 * @return array
+	 */
+	private function get_paths(): array {
```
StyleShit, line 436: If we're gonna use the method I said above, maybe change it to:
```suggestion
	private function get_system_paths(): array {
```

and it can also be public, might be useful

modules/system-info/reporters/server.php
```diff
@@ -410,4 +403,63 @@ public function get_elementor_library() {
...
+			static::KEY_PATH_UPLOADS_DIR => wp_upload_dir()['basedir'] ?? '',
+			static::KEY_PATH_ELEMENTOR_UPLOADS_DIR => ! empty( wp_upload_dir()['basedir'] ) ? wp_upload_dir()['basedir'] . '/elementor' : '',
+		];
+	}
+
+	/**
+	 * @param $path
+	 * @return array{read: bool, write: bool, execute: bool}
+	 */
+	private function get_path_permissions( $path ): array {
```
StyleShit, line 450: tests?

StyleShit (changes requested):

core/app/modules/import-export/iterator.php
```diff
@@ -46,10 +46,22 @@ public function __construct( array $settings ) {
...
+			Server::KEY_PATH_WP_CONTENT_DIR,
+			Server::KEY_PATH_UPLOADS_DIR,
+			Server::KEY_PATH_ELEMENTOR_UPLOADS_DIR,
+		];
+
+		$permissions = $server->get_system_paths_permissions();
+
+		$permissions = array_filter( $permissions, function ( $key ) use ( $paths_to_check ) {
+			return array_key_exists( $key, $paths_to_check );
+		}, ARRAY_FILTER_USE_KEY  );
```
matipojo, line 59: `array_intersect`?

modules/system-info/reporters/server.php
```diff
@@ -410,4 +406,56 @@ public function get_elementor_library() {
 			'value' => 'Connected',
 		];
 	}
+
+	/**
+	 * @return array []{read: bool, write: bool, execute: bool}
+	 */
+	public function get_system_paths_permissions() : array {
+		$paths = $this->get_system_paths();
```
matipojo, line 414: Maybe pass the exactly paths you need to check instead of checking all system paths and then filter it.
```suggestion
	public function get_system_paths_permissions( $paths ) : array {
```

modules/system-info/reporters/server.php
```diff
@@ -410,4 +406,56 @@ public function get_elementor_library() {
...
+
+		$permissions = [];
+
+		foreach ( array_keys( $paths ) as $key_path ) {
+			$path = $paths[ $key_path ];
```
matipojo, line 419:
```suggestion
		foreach ( $paths as $key_path => $path ) {
```

core/app/modules/import-export/iterator.php
```diff
@@ -46,10 +46,22 @@ public function __construct( array $settings ) {
...
+
+		foreach ( $permissions as $permission ) {
+			if ( ! $permission['write'] ) {
+				throw new \Error( self::NO_WRITE_PERMISSIONS_KEY );
+			}
```
matipojo, line 64: Maybe throw an error with all problematic paths

matipojo (changes requested):
Please cover with tests

core/app/modules/import-export/iterator.php
```diff
@@ -46,7 +46,15 @@ public function __construct( array $settings ) {
...
-		$server_write_permissions = $server->get_write_permissions();
+		$paths_to_check = [
+			WP_CONTENT_DIR => 'WordPress content directory',
+		];
+
+		$upload_directories = $server->get_upload_directories_paths();
+
+		$paths_to_check = array_merge( $paths_to_check, $upload_directories['paths'] );
+
+		$server_write_permissions = $server->check_write_permissions( $paths_to_check, $upload_directories['errors'] );
```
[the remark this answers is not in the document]
ManorHazaz: Done.

modules/system-info/reporters/server.php
ManorHazaz: Done

core/app/modules/import-export/iterator.php
ManorHazaz: Done

modules/system-info/reporters/server.php
ManorHazaz: It's not deprecated, it's just should not be used.

ManorHazaz: But I changed the comment anyway

modules/system-info/reporters/server.php
ManorHazaz: I changed the solution, inspired by your suggestion, what do you think?

modules/system-info/reporters/server.php
ManorHazaz: Done

core/app/modules/import-export/iterator.php
ManorHazaz: Won't work.

modules/system-info/reporters/server.php
ManorHazaz: I did this in your way the first time, @StyleShit suggested I will do it like the current solution.

modules/system-info/reporters/server.php
ManorHazaz: Oops :)

core/app/modules/import-export/iterator.php
ManorHazaz: A fix for all the errors and exceptions is currently being made by our product team. I think for now it's good enough just to throw a general error.

modules/system-info/reporters/server.php
ManorHazaz: I tried to add tests, but couldn't find a way to change permissions.

modules/system-info/reporters/server.php
matipojo: Every path check is a Disk Access so, we need to reduce it as much it's possible

matipojo (changes requested):

modules/system-info/reporters/server.php
```diff
@@ -423,18 +429,21 @@ public function get_system_paths_permissions() : array {
...
-	 * @return array
+	 * @param $path_key
+	 * @return string
 	 */
-	public function get_system_paths() : array {
-		return [
+	public function get_system_path( $path_key ) : string {
+		$system_paths = [
 			static::KEY_PATH_WP_ROOT_DIR => ABSPATH,
 			static::KEY_PATH_HTACCESS_FILE => file_exists( ABSPATH . '/.htaccess' ) ? ABSPATH . '/.htaccess' : '',
```
matipojo, line 440: Don't check if it exists until you need it

matipojo (changes requested):

modules/system-info/reporters/server.php
```diff
@@ -305,34 +310,32 @@ public function get_mysql_version() {
...
 
-		if ( is_dir( $elementor_uploads_path ) ) {
-			$paths_to_check[ $elementor_uploads_path ] = 'Elementor uploads directory';
-		}
+		$permissions = $this->get_paths_permissions( $paths_to_check );
 
-		$htaccess_file = ABSPATH . '/.htaccess';
+		$permissions = array_filter( $permissions, function ( $key ) use ( $paths_messages ) {
+			return array_key_exists( $key, $paths_messages );
+		}, ARRAY_FILTER_USE_KEY  );
```
matipojo, line 332: Do you still need it?

matipojo (changes requested):

modules/system-info/reporters/server.php
ManorHazaz: Yes, it matches the path permissions to the proper message.

ManorHazaz: Removed.

PR #19436, opened 2022-08-11
Internal: Experiment - Introduced internal experiment [ED-7922]

TzviRabinovitch:


tests/phpunit/elementor/core/experiments/test-manager.php
```diff
@@ -82,6 +83,29 @@ public function test_add_feature__ensure_wrap_core_dependency() {
...
+		self::assertNotEmpty( $feature );
+		self::assertTrue( $feature['hidden'] );
+	}
+
+	public function test_features_are_added_as_not_hidden_by_default() {
+		// Act
+		$this->add_test_feature();
+
+		// Assert
+		$features = $this->experiments->get_features();
```
udidol, line 102: Is the feature list empty before you add the test feature? our CI tests activate a bunch of features by default during the run, so it might not be empty regardless of whether you add a test feature here or not

udidol (changes requested):

udidol (approved):

PR #19437, opened 2022-08-11
Fix: Default Flex Grow affects the layout [ed-7869]

hein-obox:


assets/dev/scss/frontend/widgets/divider.scss
```diff
@@ -137,11 +137,28 @@
...
 }
 
-.e-container {
+.elementor {
 
-	& > .elementor-widget-divider {
-		// Fix - Divider widget has undefined width and height issues inside the flex container (ED-7722).
-		width: var( --container-widget-width, 100% );
-		flex-grow: 1;
+	.elementor-element {
```
StuartElementor, line 142:
```suggestion
	&-element {
```

assets/dev/scss/frontend/widgets/divider.scss
```diff
@@ -137,11 +137,28 @@
 	direction: rtl;
 }
 
-.e-container {
+.elementor {
```
StuartElementor, line 140: I wonder if this parent `.elementor` container is necessary.
Can you check if it is necessary - maybe it is to get the specificity that you are needing. but maybe it;s also just superflous and not needed, in which case you can remove it.

assets/dev/scss/frontend/widgets/divider.scss
```diff
@@ -137,11 +137,28 @@
...
-	& > .elementor-widget-divider {
-		// Fix - Divider widget has undefined width and height issues inside the flex container (ED-7722).
-		width: var( --container-widget-width, 100% );
-		flex-grow: 1;
+	.elementor-element {
+
+		&.elementor-section {
+
+			.elementor-divider-separator {
+				width: var( --divider-width, 100% );
```
StuartElementor, line 147: This is a really small but best we keep with the standard in mostly all of the El `.scss`'s
wrong: `var( --divider-width, 100% );`
right: `var(--divider-width, 100%);`

Can you chnage this for all the `var()`'s

Small one, but thanks

assets/dev/scss/frontend/widgets/divider.scss
hein-obox: @StuartElementor I have to increase the specificy for BC reasons.

assets/dev/scss/frontend/widgets/divider.scss
hein-obox: @StuartElementor That option isn't possible because it will write '.elementor-element' instead of '.elementor .elementor-element', which we need for specificity reasons. 

I have changed my base selector though, for clarity.

assets/dev/scss/frontend/widgets/divider.scss
hein-obox: @StuartElementor 
I have changed it. 

I have checked it with Udi in the passed, but I believe he indicated that both versions are allowed inside scss. There isn't clarity about this.

udidol (approved):

PR #19451, opened 2022-08-12
Fix: When copy-pasting a widget on a page, the widget is being pasted in a Section (..) [ed-7790]

hein-obox:


TzviRabinovitch (dismissed):

assets/dev/js/editor/document/elements/commands/paste.js
```diff
@@ -48,8 +48,23 @@ export class Paste extends $e.modules.editor.document.CommandHistoryBase {
 			let index = 'undefined' === typeof at ? targetContainer.view.collection.length : at;
 
 			data.forEach( ( model ) => {
-				switch ( model.elType ) {
-					case 'container':
+				const elementType = ( elementorCommon.config.experimentalFeatures.container && ( 'section' !== model.elType && 'column' !== model.elType ) )
```
louiswol94, line 51: @hein-obox Why does `model.elType` not exist when it is a `container`? Like why do you have to say "well if it's not a section or a column, then it must be a container" ... I believe it would be better if model.elType can be checked directly for 'container'. Then the switch can stay on `switch ( model.elType ) {`

louiswol94 (changes requested):

assets/dev/js/editor/document/elements/commands/paste.js
udidol: Agree

udidol (changes requested):

assets/dev/js/editor/document/elements/commands/paste.js
hein-obox: @louiswol94 
@udidol 

We have multiple element types, such as at least 'container', 'widget', 'section' and 'column'. I am not 100% if there could be other element types as well. 

But in this way I believe I have covered for all possible scenarios. 

I can't just check for the `model.elType`, because in case of a widget we want to know both the `elType` and if the container experiment is active.

udidol (dismissed):

tests/qunit/tests/assets/dev/js/editor/document/elements/commands/paste.spec.js
```diff
@@ -133,10 +133,10 @@ export const Paste = () => {
 				}
 			} );
 
-			QUnit.test( 'On preview container', ( assert ) => {
```
udidol, line 136: Why replace this column test with Container?

udidol: Also, if the container experiment is off, will this fail?

udidol (changes requested):

PR #19494, opened 2022-08-16
Fix: Horizontal scrolling appears in the edit area when viewing it with different devices [ED-7469]

maork-elementor:


matipojo (changes requested):
Please validate the scale in the Responsive Bar is working as expected
https://github.com/elementor/elementor/blame/master/assets/dev/scss/editor/_devices.scss#L26

PR #19598, opened 2022-08-24
Tweak: Add Dynamic Tags for Global Colors [ED-8050]

arielk:


udidol (approved):

tizum:
This update contains bugs as reported here.

https://github.com/elementor/elementor/issues/20368

tizum:
I believe that this merge is missing some functionality, as it does not work as expected, I did several tests, new installations from scratch, and none of them work, inside the constructor it works, it shows the colors, but when you go to the page in the front-end, the color doesn't render, I believe it's just some detail missing.

PR #19652, opened 2022-08-30
Fix: Translation

rodolphebertozzo:
The native translation method of Wordpress esc_html__() is missing for adding the HTML tag 'a' to the container


rodolphebertozzo:
@arielk

PR #19663, opened 2022-08-31
Fix: Inline Font Icons Experiment making the Hotspot Widget Label Icon height bigger [ED-7740]

StyleShit:
We need to always print the `viewBox` since the SVG element doesn't inherit that from the `symbol` tag and looses its aspect-ratio.

Before:
image

After:
image

### DIFF
--- a/includes/managers/icons.php
+++ b/includes/managers/icons.php
@@ -280,14 +280,14 @@ class Icons_Manager {
 
 		$attributes['class'][] = self::FONT_ICON_SVG_CLASS_NAME;
 		$attributes['class'][] = 'e-' . $icon_data['key'];
+		$attributes['viewBox'] = '0 0 ' . $icon_data['width'] . ' ' . $icon_data['height'];
 
 		/**
 		 * If in edit mode inline the full svg, otherwise use the symbol.
 		 * Will be displayed only after page update or widget "blur".
 		 */
 		if ( Plugin::$instance->editor->is_edit_mode() ) {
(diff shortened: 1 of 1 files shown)


includes/managers/icons.php
```diff
@@ -280,14 +280,14 @@ public static function get_font_icon_svg( $icon, $attributes = [] ) {
...
+		$attributes['viewBox'] = '0 0 ' . $icon_data['width'] . ' ' . $icon_data['height'];
 
 		/**
 		 * If in edit mode inline the full svg, otherwise use the symbol.
 		 * Will be displayed only after page update or widget "blur".
 		 */
 		if ( Plugin::$instance->editor->is_edit_mode() ) {
-			return '<svg xmlns="http://www.w3.org/2000/svg" ' . Utils::render_html_attributes( $attributes ) . '
-				viewBox="0 0 ' . esc_attr( $icon_data['width'] ) . ' ' . esc_attr( $icon_data['height'] ) . '">
+			return '<svg xmlns="http://www.w3.org/2000/svg" ' . Utils::render_html_attributes( $attributes ) . '">
```
StyleShit, line 290: @matipojo 
Open the lines below, see the `svg` + `use` prints

matipojo: Can you attach screenshots before/after for several icons?

StyleShit: Done

StyleShit:
Resolved by #19672

PR #19693, opened 2022-09-04
Internal: Remove unused code in Inline Fonts Icons experiment [ED-8139]

StyleShit:


includes/managers/icons.php
```diff
@@ -229,28 +227,6 @@ private static function is_font_icon_inline_svg() {
 		return Plugin::$instance->experiments->is_feature_active( 'e_font_icon_svg' );
 	}
 
-	/**
-	 * render_svg_symbols
-	 *
-	 */
-	public static function render_svg_symbols() {
```
matipojo, line 236: Mark as deprecated in order to avoid fatal errors, in case of using by a 3rd party plugin

matipojo (changes requested):

includes/managers/icons.php
StyleShit: Done

includes/managers/icons.php
```diff
@@ -229,26 +228,13 @@ private static function is_font_icon_inline_svg() {
...
-		}
-
-		$svg .= '</svg>';
+		/**
+		 * @var $deprecation Deprecation
+		 */
+		$deprecation = Plugin::$instance->modules_manager->get_modules( 'dev-tools' )->deprecation;
 
-		Utils::print_unescaped_internal_string( $svg );
+		$deprecation->deprecated_function( __METHOD__, '3.8.0' );
```
matipojo, line 237: Please mark it only via PHPDoc

matipojo (changes requested):

includes/managers/icons.php
StyleShit: Done

PR #19694, opened 2022-09-04
Tweak: Add more units to testimonial and image-box widgets [ED-8146]

rodolphebertozzo:
Tweak: Add more units to testimonial and image-box widgets


rodolphebertozzo:
@arielk

Issue #19546, opened 2022-08-21
🐞 Bug report: Fatal Error > Uncaught TypeError: Unsupported operand types: string * float [ED-8155]


andrewnotari:
### Prerequisites

- [X] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [X] The issue still exists against the latest stable version of Elementor.

### Description

Uncaught TypeError: Unsupported operand types: string * float
Better details: https://wordpress.org/support/topic/typeerror-unsupported-operand-types-string-float/

### Steps to reproduce

Set custom image size

### Isolating the problem

- [X] This bug happens with only Elementor plugin active (and Elementor Pro).
- [ ] This bug happens with a Blank WordPress theme active ([Hello theme](https://wordpress.org/themes/hello-elementor/)).
- [X] I can reproduce this bug consistently using the steps above.

### System Info

https://pastebin.com/Bgygyqfs

artuze:
For what it's worth, I've been having a bit of [a discussion](https://wordpress.org/support/topic/typeerror-unsupported-operand-types-string-float/) about this same issue over in the WordPress plugin support forum.

Basically, there's a regex in includes/controls/groups/image-size.php that's pulling the one or two dimensions out of the image size name and shoving that into the array of additional image sizes. Normally fine, but it's putting them in there as strings, so underlying WordPress functions are trying, somewhere, to do a comparison of image sizes and having an issue with the string vs. number comparison and throwing a TypeError.

Easiest fix going forward is a couple of `(int)`s thrown in there. But that doesn't fix the already broken values saved in the postmeta. For that, you'd need an upgrade script that searches them all out (`select * from wp_postmeta where meta_key="wp_attachment_metadata" and meta_value like '%s:5:"width";s:'` or something like that), changes the type of all the widths and heights, and resaves that postmeta.

andrewnotari:
Really thanks @artuze 

Confirm that it is an issue with PHP 8.X compatibility.
It’s not server configuration.
Confirm that use PHP 7.4 isn’t a solution to the issue.

bredington:
Just following up as I have a Wordpress site that is getting the same error. Is there a solution or fix incoming?

zohaibtahir:
I also face the same issue. please anybody response with the correct solution ASAP.

jetbautista:
I have same issue, my temporary solution is to downgrade my php version to 7.4.

ARod84:
[...]

PR #18560, opened 2022-05-02
Fix: Images - Thumbnail files not deleted upon deleting its main image/attachment [ED-6973]

nahmana:

PR #19722, opened 2022-09-06
Fix: Custom Image Size generates a fatal error after updating to PHP 8+ [ED-8165]

KingYes:
https://github.com/elementor/elementor/issues/19546

This PR resolves the new custom images for PHP8. Retro-saved custom images will be handled in a [separate PR](https://github.com/elementor/elementor/pull/19725).
For clarity, the bug only affects new custom images in Elementor 3.7+.


KingYes:
The bug starts from there: https://github.com/elementor/elementor/pull/18560

tests/phpunit/elementor/includes/managers/test-image.php
```diff
@@ -42,17 +42,17 @@ public function test_get_details__adds_metadata() {
 
 		$this->assertTrue( is_array( $new_metadata['sizes'] ) );
 
-		$this->assertEquals( [
+		$this->assertSame( [
 				'file' => 'test-image.png',
-				'width' => '300',
-				'height' => '300',
+				'width' => 300,
```
matipojo, line 47: Maybe add a test with empty width/height.

matipojo (changes requested):

PR #19735, opened 2022-09-07
Internal: Using the right way with the widgets API [ED-8170]

KingYes:


core/base/module.php
```diff
@@ -316,13 +316,11 @@ protected function get_widgets() {
 	/**
 	 * Initialize the module related widgets.
 	 */
-	public function init_widgets() {
-		$widget_manager = Plugin::instance()->widgets_manager;
-
+	public function init_widgets( Widgets_Manager $widgets_manager ) {
```
matipojo, line 319: It changes the method signature, it will break 3rd party plugins that extend this class
For example:
https://github.com/nerds-farm/e-addons-for-elementor/blob/ed8799496d9b83b744254db048ecd6cf80e44489/base/module-base.php#L232

matipojo (changes requested):

core/base/module.php
KingYes: Thanks!

PR #19736, opened 2022-09-07
Tweak: Soft deprecation for the wrong using with widgets API [ED-8171]

KingYes:
image


includes/managers/widgets.php
```diff
@@ -240,6 +240,12 @@ public function register_widget_type( Widget_Base $widget ) {
 	 */
 	public function register( Widget_Base $widget_instance ) {
 		if ( is_null( $this->_widget_types ) ) {
+			_doing_it_wrong(
+				__METHOD__,
+				esc_html__( 'Register widgets must running after `elementor/widgets/register` action.', 'elementor' ),
```
rami-elementor, line 245: typo

PR #19751, opened 2022-09-08
Fix: Elements Regressions - Font "swap" make the tests fail [ED-8189]

Nevoss:


tests/elements-regression/env-setup/actions/change-font-display.js
```diff
@@ -0,0 +1,26 @@
+const ElementorSettingsPage = require( '../../pages/elementor-settings-page' );
+
+let elementorFontDisplayOriginalValue;
+
+module.exports = {
+	beforeAll: async ( { page } ) => {
+		const settingsPage = new ElementorSettingsPage( page );
```
matipojo, line 7: Maybe just config it via `wp-cli` command?
https://github.com/elementor/elementor/blob/master/tests/wp-env/config/setup.sh#L10

matipojo (changes requested):

tests/elements-regression/env-setup/actions/change-font-display.js
Nevoss: But then it will affect only the CI, what about local testing?

matipojo: It can be tested locally? (Linux screenshot)

Nevoss: You can create your own photos + You can run on a container (e3 or `npm run test:elements-regression:update-snapshots`)

matipojo (changes requested):
Please resolve the conflicts

matipojo (changes requested):
Please resolve the conflicts

PR #19756, opened 2022-09-08
Tweak: Soft deprecation for wrong using the widgets API [ED-8171]

matipojo:
Reverts elementor/elementor#19755


includes/managers/widgets.php
```diff
@@ -240,6 +240,12 @@ public function register_widget_type( Widget_Base $widget ) {
 	 */
 	public function register( Widget_Base $widget_instance ) {
 		if ( is_null( $this->_widget_types ) ) {
+			_doing_it_wrong(
```
matipojo, line 243: if ( ! did_action( 'elementor/widgets/register' ) ) {
	_doing_it_wrong( ... );
}

KingYes: @matipojo You're wrong with it. This condition will change the logic of the code.

PR #19890, opened 2022-09-22
Internal: React History Panel

matipojo:


assets/dev/js/editor/regions/panel/footer.js
```diff
@@ -134,7 +134,7 @@ module.exports = Marionette.ItemView.extend( {
 	},
 
 	onHistoryClick() {
-		$e.route( 'panel/history/actions' );
+		$e.route( 'panel/history/actions-v2' );
```
StyleShit, line 137: in the future, we'll probably want to keep the original API (?), so maybe we should keep all those "old" API calls, hook into the route system and replace it to v2 conditionally (based on experiment or whatever)

WDYT?

modules/history/assets/js/history/panel-tab.js
```diff
@@ -1,7 +1,7 @@
 import ItemView from './item-view';
 import EmptyView from './empty';
 
-module.exports = Marionette.CompositeView.extend( {
+export default Marionette.CompositeView.extend( {
```
StyleShit, line 4:
```suggestion
export default class extends CompositeView {
```

package.json
```diff
@@ -8,6 +8,11 @@
     "prev_stable_version": "3.1.4",
     "author": "Elementor Team",
     "homepage": "https://elementor.com/",
+    "workspaces": {
+        "packages": [
+            "packages/editor/packages/*"
+        ]
+    },
```
StyleShit, line 15: why do we need 2 workspaces? let's just remove the other one...

assets/dev/js/editor/regions/panel/footer.js
matipojo: Done

modules/history/assets/js/history/panel-tab.js
matipojo: It's not related to the POC

package.json
matipojo: It's a temp workspace until the package will be published

Issue #19845, opened 2022-09-19
🐞 Bug Report : important bug in menu cart - container (The menu cart opens below the content)


79ho3ein:
### Prerequisites

- [X] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [X] The issue still exists against the latest stable version of Elementor.

### Description

If you create the header and content of the website using the container, if you use the menu cart widget in the header, after the menu cart is opened, this widget will be displayed below the content. Watch the following video :

https://user-images.githubusercontent.com/77163248/190972397-cbdac388-56f6-4030-8b86-aecf01a60b32.mp4

**please note**

If sticky mode is enabled for the header container, there is no problem. But if sticky is off, this bug will occur.

### Steps to reproduce

- Add a container for the header and place the menu cart widget inside it.
- Also create a page and put contents inside it. For example, a slider or several images.
- Now visit the website and click on the menu cart to open it.

 The bug is quite obvious. The menu cart widget opens below the content.

### Isolating the problem

- [X] This bug happens with only Elementor plugin active (and Elementor Pro).
- [X] This bug happens with a Blank WordPress theme active ([Hello theme](https://wordpress.org/themes/hello-elementor/)).
- [X] I can reproduce this bug consistently using the steps above.

### System Info

-

navarroido:
Try to add z-index for the container with the cart, and in addition to the cart itself.

79ho3ein:
Hi @navarroido 
Yes, that was the method @zeinnicholas recommended as well. But using this method brings problems to another part. I will give an example for it  : 

If we increase the z-index value in the container (in which the menu cart widget is located), it makes the handle of the first container unreachable. Because it is placed under the header container. See the following video :

https://user-images.githubusercontent.com/77163248/191133314-3ff21780-2f2a-4c43-9c83-7160ecea5ff8.mp4

So I think it would be much better if this problem is fixed in the menu cart widget by default and it works like before.

zeinnicholas:
Thanks for the feedback @79ho3ein !

We will pass this along to the people responsible and see what we can do to address the issue with the handle.

Until then, an alternative to the handle issue is to use the navigator to select the container below the header. It's not a proper solution, but at least it will let you get around these limitations. 😉

79ho3ein:
Thank you for your attention @zeinnicholas

mkiii333:
This one is related https://github.com/elementor/elementor/issues/19705

arielk:
[...]

Issue #19834, opened 2022-09-18
🐞 Bug Report: Container's css "isolation: isolate;" is an overkill and will likely cause problems (ED-8350)


hogash:
### Prerequisites

- [X] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [X] The issue still exists against the latest stable version of Elementor.

### Description

This pull request https://github.com/elementor/elementor/commit/57489eb30cc121bf46dc5314604d89ef037900bd fixed the display of the video's fallback image, but brought some negative z-indexes to the overlay and video container, plus the `isolation: isolate` property on the Container element.

This will likely create problems since it creates a new stack context and may lead to issues like other positions, transforms etc. I can't predict exactly everything, but for example a z-index of 100000 on an element inside the container will not leave the container unless the container also has that z-index. This creates a stacking problem.

The solution is pretty simple, just move the `:before` pseudo selector's properties on the video container. The overlay will stay inside the video container while being overlapped by the content of the Container.

Made a quick video recording using the alternative https://d.pr/v/gDT4oG . 

### Steps to reproduce

https://d.pr/v/gDT4oG

### Isolating the problem

- [X] This bug happens with only Elementor plugin active (and Elementor Pro).
- [X] This bug happens with a Blank WordPress theme active ([Hello theme](https://wordpress.org/themes/hello-elementor/)).
- [X] I can reproduce this bug consistently using the steps above.

### System Info

-

nicholaszein:
Hello @hogash! 

Thank you for submitting your issue! 🙏 

Please allow us some time to investigate it, and get back to you with our findings. 🧐 

Best regards

rally-lad:
Just to add to this.

This issue also exists in Elementor 3.7.7 with Elementor Forms V3 recaptcha hiding behind flex box containers.
No matter what the z-index is set to on the .grecaptcha-badge it falls behind a container.

nicholaszein:
Thank you @rally-lad! I'll add this to our tracker as well.

Cheers! 🥂

arielk:
This issue has been resolved in Elementor v3.8.0-beta-5

Feel free to update

Thanks!

Issue #19705, opened 2022-09-05
🐞 Bug report: Navigation (Dropdown) on Mobile  + Search Form Full Screen Isn't Working Properly 


mkiii333:
### Prerequisites

- [X] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [X] The issue still exists against the latest stable version of Elementor.

### Description

Another update, another bug! It feels like this is becoming a "new normal."

We're using Flex containers.

After recent updates, the drop-down menu (navigation) isn't working correctly on mobile (it feels like something is messed up with the z-index or priorities of how each element should be shown).

Another website uses Elementor version 3.6.8 and Elementor Pro Version 3.6.5 and does not have this issue. However, after updating to the recent Elementor version, this bug also occurs on that website.

https://monosnap.com/file/6YcXGZcHd3P3Pa38d4PtzPnNoOiW4r

In the live version of the website, it's even worse:
https://monosnap.com/file/UHXXIuEyeVWPdKarfrObazuQEoblHU

Another bug: Search Form Full Screen Isn't Working Properly 

I feel this is somehow related to the previous issue.

When you click on the search form (Desktop & Mobile), the full screen with search input doesn't appear.

https://monosnap.com/file/4QK0ilM89ebnm0gGmKaPeEDpU4Jz3h

It feels that each update generates a bunch of bugs. We stopped using the Sticky function (for the sidebar) because it's glitching as well... We're tired of this hit-or-miss approach with each update.

### Steps to reproduce

1. Use flex containers
2. Add navigation to a header, use drop-down hamburger, check mobile
3. Add search form widget and use full-screen skin.

### Isolating the problem

- [X] This bug happens with only Elementor plugin active (and Elementor Pro).
- [X] This bug happens with a Blank WordPress theme active ([Hello theme](https://wordpress.org/themes/hello-elementor/)).
- [X] I can reproduce this bug consistently using the steps above.

### System Info

					
== Server Environment ==
	Operating System: Linux
	Software: Apache/2.4.51 (Debian)
	MySQL version: mariadb.org binary distribution v10.6.3
	PHP Version: 8.1.8
	PHP Memory Limit: 1500M
	PHP Max Input Vars: 14900
	PHP Max Post Size: 200M
	GD Installed: Yes
	ZIP Installed: Yes
	Write Permissions: All right
	Elementor Library: Connected

== WordPress Environment ==
	Version: 6.0.2
	Site URL: https://.com
	Home URL: https://.com
[...]

PR #20014, opened 2022-10-10
Fix: Container isolation: isolate issues (#19834, #19845, #19705) [ED-8353]

louiswol94:


hein-obox (dismissed):

tests/playwright/sanity/container.test.js
```diff
@@ -261,4 +261,56 @@ test.describe( 'Container tests', () => {
...
+
+		const containerId = await editor.addElement( { elType: 'container' }, 'document' );
+
+		await wpAdmin.activatePanelTab( 'style' );
+		await page.locator( '[data-tooltip="Video"]' ).click();
+		await page.locator( '[data-setting="background_video_link"]' ).fill( videoURL );
+		await page.locator( '.eicon-plus-circle' ).nth( 3 ).click();
+		await page.locator( 'text=Media Library' ).click();
+		await page.waitForSelector( 'text=Insert Media' );
+		await page.waitForTimeout( 1000 );
```
TzviRabinovitch, line 294: Can you replace it with another check?

tests/playwright/sanity/container.test.js
```diff
@@ -261,4 +261,56 @@ test.describe( 'Container tests', () => {
...
+		await page.setInputFiles( 'input[type="file"]', './tests/playwright/resources/mountain-image.jpeg' );
+		await page.waitForLoadState( 'networkidle' );
+		await page.click( '.button.media-button' );
+		await page.locator( '.elementor-control-section_background_overlay' ).click();
+		await page.locator( '[data-tooltip="Classic"]' ).nth( 0 ).click();
```
TzviRabinovitch, line 299: nth is not the best option

tests/playwright/sanity/container.test.js
```diff
@@ -261,4 +261,56 @@ test.describe( 'Container tests', () => {
...
+		await editor.closeNavigatorIfOpen();
+
+		const container = editor.getFrame().locator( '.elementor-element-' + containerId );
+
+		expect( await container.screenshot( {
+			type: 'jpeg',
+			quality: 70,
+		} ) ).toMatchSnapshot( 'container-background.jpeg' );
+
+		await wpAdmin.setExperiments( {
```
TzviRabinovitch, line 312: I don't think it's needed to set it to false

tests/playwright/sanity/container.test.js
```diff
@@ -261,4 +261,56 @@ test.describe( 'Container tests', () => {
...
+		const videoURL = await page.locator( '.attachment-details-copy-link' ).inputValue();
+
+		const editor = await wpAdmin.useElementorCleanPost();
+
+		const containerId = await editor.addElement( { elType: 'container' }, 'document' );
+
+		await wpAdmin.activatePanelTab( 'style' );
+		await page.locator( '[data-tooltip="Video"]' ).click();
+		await page.locator( '[data-setting="background_video_link"]' ).fill( videoURL );
+		await page.locator( '.eicon-plus-circle' ).nth( 3 ).click();
```
TzviRabinovitch, line 291: nth is not the best option

tests/playwright/sanity/container.test.js
```diff
@@ -261,4 +261,56 @@ test.describe( 'Container tests', () => {
...
+		await page.goto( '/wp-admin/media-new.php' );
+
+		if ( await page.locator( 'text=browser uploader' ).nth( 1 ).isVisible() ) {
+			await page.locator( 'text=browser uploader' ).nth( 1 ).click();
+		}
+
+		await page.setInputFiles( 'input[name="async-upload"]', './tests/playwright/resources/video.webm' );
+		await page.locator( '#html-upload' ).click();
+		await page.waitForSelector( '.attachments-wrapper' );
+		await page.locator( 'ul.attachments li' ).nth( 0 ).click();
```
TzviRabinovitch, line 280: nth is not the best option

tests/playwright/sanity/container.test.js
```diff
@@ -261,4 +261,56 @@ test.describe( 'Container tests', () => {
...
+
+	test( 'Fallback image is not on top of background video', async ( { page }, testInfo ) => {
+		const wpAdmin = new WpAdminPage( page, testInfo );
+		await wpAdmin.setExperiments( {
+			container: true,
+		} );
+
+		await page.goto( '/wp-admin/media-new.php' );
+
+		if ( await page.locator( 'text=browser uploader' ).nth( 1 ).isVisible() ) {
```
TzviRabinovitch, line 273: nth is not the best option
also, try to avoid search by text if possible
can you extract the element to a const ?

TzviRabinovitch (changes requested):

tests/playwright/sanity/container.test.js
hein-obox: @TzviRabinovitch 
Following.

tests/playwright/sanity/container.test.js
hein-obox: @TzviRabinovitch We have been doing that everywhere. What are your considerations? I will be good to know, also for other tests.

tests/playwright/sanity/container.test.js
hein-obox: @TzviRabinovitch What is your consideration to avoid text? For RTL reasons or other?

louiswol94: @TzviRabinovitch Done.

tests/playwright/sanity/container.test.js
louiswol94: @TzviRabinovitch nth cannot be avoided here.

tests/playwright/sanity/container.test.js
louiswol94: @TzviRabinovitch Done

tests/playwright/sanity/container.test.js
louiswol94: @TzviRabinovitch I agree with Hein. You will see on every container test we set the experiment back to false at the end of the test.

tests/playwright/sanity/container.test.js
louiswol94: @TzviRabinovitch Done.

tests/playwright/sanity/container.test.js
louiswol94: @TzviRabinovitch Done.

tests/playwright/sanity/container.test.js
TzviRabinovitch: @hein-obox yes

tests/playwright/sanity/container.test.js
TzviRabinovitch: @hein-obox @louiswol94 
while in theory, this is a good practice it also wastes time 

the best approach would be to set the experiment on before running all the tests in a file and turning it off at the end

TzviRabinovitch (approved):

PR #20028, opened 2022-10-12
Internal: CI - Core build should work like the Pro [ED-8012]

maork-elementor:


matipojo (changes requested):
Please send to master

PR #20085, opened 2022-10-19
New: Kit defaults - Usage data [ED-8178]

StyleShit:


Nevoss (dismissed):

modules/kit-elements-defaults/module.php
```diff
@@ -45,5 +48,7 @@ public function __construct() {
 		} );
 
 		Plugin::$instance->data_manager_v2->register_controller( new Controller() );
+
+		( new Usage() )->register();
```
matipojo, line 52: if is_admin

modules/kit-elements-defaults/module.php
```diff
@@ -11,6 +11,9 @@
 }
 
 class Module extends BaseModule {
+
+	const META_KEY = '_elementor_elements_defaults_values';
```
matipojo, line 15:
```suggestion
	const META_KEY = '_elementor_elements_default_values';
```

matipojo (changes requested):

modules/kit-elements-defaults/module.php
StyleShit: Good catch.
It was copy-pasted from the controller, so we'll need to fix that in the other PRs @Nevoss

modules/kit-elements-defaults/module.php
StyleShit: I didn't find any other place in the code where we check that before sending the tracking data...

PR #20087, opened 2022-10-19
Internal: Restrict users from creating new Theme Builder parts when license is inactive [ED-7728]

mark-elementor:


includes/template-library/sources/local.php
```diff
@@ -1341,16 +1342,24 @@ public function print_blank_state_template( $current_type_label, $href, $descrip
...
+					</a>
+				<?php } else {
+					// TODO: Connect & Activate has to be here
+				} ?>
 			</div>
 		<?php
 	}
 
+	private function is_license_valid() {
+		return ( new License_Service )->is_valid();
```
davseve, line 1360: Why do you create a new instance of the License_Service? 
Haven't created a services class that registers all the services classes?

davseve (changes requested):

includes/template-library/sources/local.php
```diff
@@ -1718,6 +1732,7 @@ private function remove_elementor_cpt_from_sitemap( array $post_types ) {
 	public function __construct() {
 		parent::__construct();
 
+		$this->app_services = new AppServices();
```
ManorHazaz, line 1735: Are you creating a new instance of the services for each usage?
 Don't you think it will be better to create the service once the general app.php is loaded and consume it from there?
 EDITED
 I saw you created the instances both in the JS and in PHP, but you didn't use it here, why?

app/services/license/license-service.js
```diff
@@ -0,0 +1,17 @@
+export default class AppsLicenseService extends $e.modules.ComponentBase {
+	getNamespace() {
+		return 'app/services/license';
+	}
+
+	isActive() {
```
ManorHazaz, line 6: I'm assuming that you are planning to overwrite this from the pro, what will happen if the user won't update the pro plugin?

app/services/base-service.php
```diff
@@ -0,0 +1,5 @@
+<?php
+namespace Elementor\App\Services;
+
+abstract class Base_Service implements Service_Interface {
```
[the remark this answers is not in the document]
ManorHazaz: Agreed, one of them is also redundant in my opinion.
Not sure which is the best, again you need to test it with the pro implementation. What will happen in the Pro if you want to implement the interface or extend the base which will not be available since to user only updated the pro?

ghost (commented):
This is not a full CR.
Those are just a few comments and concerns about the implementation of the services.

app/services/services.php
```diff
@@ -0,0 +1,39 @@
...
+		$this->services = apply_filters( 'elementor/app/services/register', $this->services );
+	}
+
+	/**
+	 * Get a registered service
+	 *
+	 * @param $name
+	 * @return mixed|null
+	 */
+	public function get_service( $name ) {
```
ManorHazaz, line 32: Maybe put the public functions first? just for readability.

ghost (changes requested):

app/services/base-service.php
```diff
@@ -0,0 +1,7 @@
+<?php
+namespace Elementor\App\Services;
+
+abstract class Base_Service {
+	abstract function get_name() : string;
+	abstract function register();
```
ManorHazaz, line 6: Add phpdoc

app/app.php
```diff
@@ -16,9 +17,12 @@
 }
 
 class App extends BaseApp {
+	public static $instance;
```
ManorHazaz, line 20: Why?

ghost (changes requested):

app/app.php
mark-elementor: Explained

app/services/services.php
```diff
@@ -0,0 +1,39 @@
...
+	private $services = [];
+
+	public function __construct() {
+		add_action( 'elementor/common/after_register_scripts', function () {
+			$this->register_services();
+		} );
+	}
+
+	private function register_services() {
+		$this->services[ 'license' ] = ( new License_Service() )->register();
```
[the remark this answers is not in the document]
mark-elementor: Done

app/services/base-service.php
mark-elementor: Added.

PR #20118, opened 2022-10-24
Fix: Pasting an element into the Preview Container throws a JS error [ED-8181]

StyleShit:
Used the route args instead of the unreliable `elementor.getCurrentElement()` method.


TzviRabinovitch (approved):

Nevoss (changes requested):
tests?

Nevoss (approved):

PR #20124, opened 2022-10-24
Fix: Kits Defaults - Context menu item doesn't exist [ED-8488]

StyleShit:


Nevoss (changes requested):
Seems like it does not fix the problem

modules/kit-elements-defaults/module.php
```diff
@@ -34,7 +34,10 @@ private function enqueue_scripts() {
 		wp_enqueue_script(
 			'elementor-kit-elements-defaults-editor',
 			$this->get_js_assets_url( 'kit-elements-defaults-editor' ),
-			[],
+			[
+				'elementor-common',
+				'elementor-editor-modules',
```
Nevoss, line 39:
```suggestion
				'elementor-editor-modules',
				'elementor-editor-document',
```

Seems like this one helps, not sure what is the diffrence

Nevoss (changes requested):

Nevoss (approved):

PR #20151, opened 2022-10-27
New: A dynamic rating widget that supports SVG icons and Font Awesome [ED-7551]

maork-elementor:


includes/widgets/star-rating.php
```diff
@@ -123,10 +124,10 @@ protected function register_controls() {
 				'label' => esc_html__( 'Icon', 'elementor' ),
 				'type' => Controls_Manager::SELECT,
 				'options' => [
-					'star_fontawesome' => 'Font Awesome',
-					'star_unicode' => 'Unicode',
+					'fa' => 'Font Awesome Star',
+					'eicons' => 'Eicon Star',
 				],
-				'default' => 'star_fontawesome',
+				'default' => 'eicons',
```
rami-elementor, line 130: You are changing `options` and `default`... it doesn't break backwards compatibility for existing sites? Why not simple leave the existing option values?

or leave the original options and use a dictionary:

```
'selectors_dictionary' => [
	'star_fontawesome' => 'fa',
	'star_unicode' => 'eicons',
],

maork-elementor: I don't think that we need to keep supporting this kind of old hacks.
The old code was used the HTML ASCII codes to create the icons on the client side. 
I rewrite the widget, so he could work with Elementor Icon Manager. By using this class we can use the real icons both for E-icons & Font awesome which is already loaded in the user site. 
To support backward computability, you can look here : 

https://github.com/elementor/elementor/pull/20151/files#diff-71dc6c5ec721d83f0f0d6900a6915796ff1faa2abc0c30775c4c94d41da24aa8R363

If the user had font awesome, so we are good to go, but if he selects before the Unicode, then he will get the E-icons instead. As the new default.

PR #20193, opened 2022-10-30
Tweak: Small refactoring in image widget

rodolphebertozzo:
I moved the table that proposes the list of icons to put it directly below the range resolution operator, as on the other widgets when it comes to units


rodolphebertozzo:
@KingYes

rami-elementor (commented):
Seems like nothing really changed.

PR #20239, opened 2022-11-02
Internal: PHPUnit fails - Image HTML was changed [ED-ED-8575]

maork-elementor:


tests/phpunit/elementor/includes/controls/groups/test-image-size.php
```diff
@@ -30,9 +30,16 @@ public function test_print_attachment_image_html() {
 
 		$html = ob_get_clean();
 
+		$wp_attachment = wp_get_attachment_image( $attachment_id, 'large', false, [ 'class' => 'attachment-large size-large' ] );
+		
+		// Wordpress 6.1 add new attributes to the image tag, so we need to remove them to support all versions.
```
matipojo, line 35:
```suggestion
		// Wordpress 6.1 removed the `decoding` attribute, so we need to remove it  to support all versions.
```

PR #20304, opened 2022-11-07
Tweak: Add "League Spartan" font

rodolphebertozzo:
Tweak: Add "League Spartan" font to font to font awesome


rami-elementor:
Closing. It seems like we already have this font included.

PR #20326, opened 2022-11-08
Internal: Elements Regression - Screenshot also in the frontend [ED-8667]

Nevoss:


matipojo:
Continuing here: https://github.com/elementor/elementor/pull/20820

PR #20353, opened 2022-11-09
Tweak: Remove the Folder icon when using a nested tab [ed-8314]

hein-obox:


tests/playwright/sanity/modules/tabs-v2/tabs.test.js
```diff
@@ -0,0 +1,34 @@
+const { test, expect } = require( '@playwright/test' );
+const WpAdminPage = require( '../../../pages/wp-admin-page' );
+
+test.describe( 'Nested Tabs tests', () => {
+	test( 'Count the number of icons inside the Add Section element', async ( { page }, testInfo ) => {
+		// Arrange.
+		const wpAdmin = new WpAdminPage( page, testInfo );
+		await wpAdmin.setExperiments( {
+			container: true,
+            'nested-elements': true,
```
davseve, line 10: This object is not aligned well.

davseve (changes requested):

tests/playwright/sanity/modules/tabs-v2/tabs.test.js
```diff
@@ -0,0 +1,34 @@
...
+		} );
+
+		const editor = await wpAdmin.useElementorCleanPost(),
+			container = await editor.addElement( { elType: 'container' }, 'document' );
+
+		// Add widgets.
+		await editor.addWidget( 'tabs-v2', container );
+		await editor.getPreviewFrame().waitForSelector( '.elementor-tabs-content-wrapper .e-con.elementor-active' );
```
davseve, line 18: Should we delete the page?

davseve (changes requested):

tests/playwright/sanity/modules/tabs-v2/tabs.test.js
hein-obox: @davseve I don't think, but can you ask the question in the Playwright slack channel.

davseve: According to what was said in the Playwright channel, we should return the site to the same state we received it.

davseve (changes requested):

davseve (approved):

Issue #6778, opened 2019-01-17
✔️ 🚀 Feature Request: Background image size selection


ShadowXVII:
## Prerequisites

- [x] I have searched for similar features requests in both open and closed tickets and cannot find a duplicate.
- [x] The feature is still missing in the latest stable version of Elementor ( Elementor Pro. )

**What problem is your feature request going to solve? Please describe.**
Opening a new ticket for the reconsideration of #3722, as advised by Elementor support.

This feature will prevent clients (especially mobile or limited bandwidth connections) from having to download a 5mb background image, when a 50kb one would be more appropriate.

**Describe the solution you'd like**
Native, and responsive background image source selection based on size, dimensions.

Wordpress does this by standard for normal blog images -- Elementor backgrounds should be smart enough to do this also.

Other elementor widgets already do this, i.e. Posts widget lets you pick the size of the image to use as the featured image header.

Ideally, there would be:
- The ability to select a specific image size, rather than having it just use the "full" version.
- A responsive option that fills in srcset for responsive scaling (as WordPress standard does).
- Not just a single in-place src replacement -- otherwise caching proxies would pick up and save a specific size.

**Describe alternatives you've considered**
Third-party optimisation tools, but there are compatibility issues with Elementor.

**Additional context**
Elementor already has several performance-conscious features, I'm unsure why this particular scenario hasn't been addressed yet.

Multiple background images that was released in recent versions of Elementor is only going to multiply this issue.

black-eye:
+1 
Oh yes. Pls add an option to select image size (of the same image) for different media queries. Now we can chose only a different image.

And also pls add an option to remove bg image completely (on specific media query) and leave only the color.

olegyurist:
+

theorib:
+1 on this, so important!

parfilov:
+1 I wrote about this in 2.4 beta release post.

michael3371:
+1 this would be a game changer in mobile speed opt.

thenathantt:
+1 not having this feature is delaying load time across the board for me.

gabiudrescu:
+1

sermalefico:
+1

Micemade:
+1

blueglassdigital:
+1

Isnt its quite a large demand on this now? will you eventually add this feature? :)
It's a pain to crop images manually...

Also, why did the privious feature request topic is set as Closed, while feature still not there... :(

jazir5:
+1 

[...]

PR #20361, opened 2022-11-10
Tweak: Background image size enables responsive choice [ED-2108]

maork-elementor:


fabianenzensperger:
Hi all, is this Tweak for this request? #6778

DavidGMiles:
Does this finally solve the 4 year old ticket?

fabianenzensperger:
> Does this finally solve the 4 year old ticket?

Better than never! 🤗🙈

DavidGMiles:
> > Does this finally solve the 4 year old ticket?
> 
> Better than never! 🤗🙈

Assuming the answer is yes! If not aggggggggh!

nicholaszein:
Hey @fabianenzensperger and @DavidGMiles!

Yes, that's exactly what it is! 😉

DavidGMiles:
Thank you 🙏 RegardsDavid Milese: ***@***.***: http://natureslens.co.ukOn 10 Nov 2022, at 17:20, Nicholas Zein ***@***.***> wrote:﻿
Hey @fabianenzensperger and @DavidGMiles!
Yes, that's exactly what it is! 😉

includes/controls/groups/background.php
```diff
@@ -258,6 +258,25 @@ public function init_fields() {
...
+		$dimentions_options = [];
+		foreach ( get_intermediate_image_sizes() as $size ) {
+			$dimentions_options[ $size ] = $size;
+		}
+		$dimentions_options['full'] = 'full';
+
+		$fields['dimension'] = [
+			'label' => esc_html_x( 'Dimension', 'Background Control', 'elementor' ),
+			'type' => Controls_Manager::SELECT,
+			'options' => $dimentions_options,
```
matipojo, line 270: Translation?

matipojo (changes requested):
🔥
Please cover with tests

assets/dev/js/editor/utils/controls-css-parser.js
```diff
@@ -37,6 +37,33 @@ ControlsCSSParser = elementorModules.ViewModule.extend( {
 		} );
 	},
 
+	updateBackgroundImagesSizes( control, values ) {
```
StyleShit, line 40: the name is too generic, doesn't really say what the function does.

assets/dev/js/editor/utils/controls-css-parser.js
```diff
@@ -37,6 +37,33 @@ ControlsCSSParser = elementorModules.ViewModule.extend( {
...
+		if ( control.name.includes( 'background_image' ) ) {
+			const imageSize = values.background_dimension || false;
```
StyleShit, line 42: why do you need to make it false? it's already falsy at that point

assets/dev/js/editor/utils/controls-css-parser.js
```diff
@@ -37,6 +37,33 @@ ControlsCSSParser = elementorModules.ViewModule.extend( {
...
+			if ( imageSize ) {
+				let responsiveSize = 'background_dimension';
```
StyleShit, line 44: same as above, basing a whole feature on control name isn't a good idea.

assets/dev/js/editor/utils/controls-css-parser.js
```diff
@@ -37,6 +37,33 @@ ControlsCSSParser = elementorModules.ViewModule.extend( {
...
+				const breakpoints = elementorFrontend.config.responsive.breakpoints;
+				Object.entries( breakpoints ).forEach( ( [ breakpointName ] ) => {
+					if ( control.name.includes( breakpointName ) ) {
+						responsiveSize += '_' + breakpointName;
+					}
+				} );
```
StyleShit, line 50: I think you can get the breakpoint that's associated with the control without iterating over all of them like this. 
Maybe consult with @udidol

assets/dev/js/editor/utils/controls-css-parser.js
```diff
@@ -37,6 +37,33 @@ ControlsCSSParser = elementorModules.ViewModule.extend( {
...
+				const sizeValue = values[ responsiveSize ];
```
StyleShit, line 51: maybe it could just be something like this?
```suggestion
				const sizeValue = values[ sizeControlName + activeBreakpoint ];
```

assets/dev/js/editor/utils/controls-css-parser.js
```diff
@@ -37,6 +37,33 @@ ControlsCSSParser = elementorModules.ViewModule.extend( {
...
+				const originalImage = values[ control.name ];
+				if ( sizeValue && originalImage ) {
+					const image = elementor.imagesManager.getImageUrl( {
```
StyleShit, line 54:
```suggestion
					const imageURL = elementor.imagesManager.getImageUrl( {
```

core/files/css/base.php
```diff
@@ -666,6 +666,31 @@ protected function parse_content() {
 		return $this->get_stylesheet()->__toString();
 	}
 
+
+	private function update_background_images_sizes( $control, $values ) {
+		if ( strpos( $control['name'], 'background_image' ) !== false ) {
```
StyleShit, line 671: Same as above regarding the control name

core/files/css/base.php
```diff
@@ -666,6 +666,31 @@ protected function parse_content() {
...
+			$image_size = $values['background_dimension'] ?? null;
+			if ( ! empty( $image_size ) ) {
+				$responsive_size = 'background_dimension';
+				$breakpoints = Plugin::$instance->breakpoints->get_breakpoints();
+				foreach ( $breakpoints as $breakpoint_key => $breakpoint ) {
+					if ( strpos( $control['name'], $breakpoint_key ) !== false ) {
```
StyleShit, line 677: Should end with the `_{ breakpoint_name }`, not contain

includes/controls/groups/background.php
```diff
@@ -258,6 +258,19 @@ public function init_fields() {
 			],
 		];
 
+		$fields['dimension'] = [
```
StyleShit, line 261:
```suggestion
		$fields['dimensions'] = [
```

includes/controls/groups/background.php
```diff
@@ -258,6 +258,19 @@ public function init_fields() {
...
+			'label' => esc_html_x( 'Dimension', 'Background Control', 'elementor' ),
```
StyleShit, line 262:
```suggestion
			'label' => esc_html_x( 'Dimensions', 'Background Control', 'elementor' ),
```

includes/controls/groups/background.php
```diff
@@ -258,6 +258,19 @@ public function init_fields() {
...
+			'type' => Controls_Manager::SELECT,
+			'options' => $this->get_images_sizes(),
+			'responsive' => true,
+			'default' => 'full',
+			'condition' => [
+				'background' => [ 'classic' ],
+				'image[url]!' => '',
+			],
+			'separator' => 'after',
```
StyleShit, line 271: does it need to have a separator?

includes/controls/groups/background.php
```diff
@@ -814,4 +827,20 @@ protected function get_default_options() {
...
 	}
+
+	protected function get_images_sizes() {
+		$sizes = get_intermediate_image_sizes();
+		$sizes[] = 'full';
+		$result = [];
+		foreach ( $sizes as $size ) {
+			$width = get_option( "{$size}_size_w" );
+			$height = get_option( "{$size}_size_h" );
+			if ( ! $width && ! $height ) {
```
StyleShit, line 838:
```suggestion
			if ( ! $width || ! $height ) {
```

assets/dev/js/editor/utils/controls-css-parser.js
```diff
@@ -37,6 +37,33 @@ ControlsCSSParser = elementorModules.ViewModule.extend( {
 		} );
```
StyleShit, line 37: does it work with the lazyload experiment?

StyleShit (changes requested):

core/files/css/base.php
```diff
@@ -666,6 +666,30 @@ protected function parse_content() {
 		return $this->get_stylesheet()->__toString();
 	}
 
+
+	private function update_background_images_sizes( $control, $values ) {
+		if ( strpos( $control['name'], 'background_image' ) !== false ) {
```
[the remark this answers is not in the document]
maork-elementor: done

assets/dev/js/editor/utils/controls-css-parser.js
```diff
@@ -37,6 +37,33 @@ ControlsCSSParser = elementorModules.ViewModule.extend( {
...
 	},
 
+	updateBackgroundImagesSizes( control, values ) {
+		if ( control.name.includes( 'background_image' ) ) {
```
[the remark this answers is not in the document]
maork-elementor: This is the file corresponding to core/files/css/base.php of the editor fronted. 
Also talk about it with Mati. This is the location of the logic should be.

assets/dev/js/editor/utils/controls-css-parser.js
maork-elementor: Removed.

assets/dev/js/editor/utils/controls-css-parser.js
maork-elementor: I need to support all types of breakpoints. In the future also with custom breakpoints.

includes/controls/groups/background.php
maork-elementor: done

includes/controls/groups/background.php
maork-elementor: done

includes/controls/groups/background.php
maork-elementor: Yes

includes/controls/groups/background.php
maork-elementor: This is not the logic I need.

assets/dev/js/editor/utils/controls-css-parser.js
maork-elementor: Yes

includes/controls/groups/background.php
```diff
@@ -258,6 +258,19 @@ public function init_fields() {
 			],
 		];
 
+		$fields['dimensions'] = [
+			'label' => esc_html_x( 'Dimensions', 'Background Control', 'elementor' ),
+			'type' => Controls_Manager::SELECT,
+			'options' => $this->get_images_sizes(),
```
udidol, line 264:
```suggestion
			'options' => $this->get_image_size_options(),
```

includes/controls/groups/background.php
```diff
@@ -814,4 +827,20 @@ protected function get_default_options() {
 			'popover' => false,
 		];
 	}
+
+	protected function get_images_sizes() {
```
udidol, line 831:
```suggestion
	protected function get_image_size_options() {
```

includes/controls/groups/background.php
```diff
@@ -814,4 +827,20 @@ protected function get_default_options() {
...
+		$sizes = get_intermediate_image_sizes();
+		$sizes[] = 'full';
+		$result = [];
+		foreach ( $sizes as $size ) {
+			$width = get_option( "{$size}_size_w" );
+			$height = get_option( "{$size}_size_h" );
+			if ( ! $width && ! $height ) {
+				$result[ $size ] = $size;
+			} else {
+				$result[ $size ] = $size . ' (' . $width . 'x' . $height . ')';
```
udidol, line 841: The option labels need to be translated

assets/dev/js/editor/utils/controls-css-parser.js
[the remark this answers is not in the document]
udidol: The Controls CSS Parser is almost entirely generic (except for some Globals handling).
I would not put specific control handling here.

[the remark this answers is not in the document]
udidol: To add to that - you cannot use a hard-coded prefix here, since each Background Group Control has a different prefix.
Take the Flip Box widget for example, it has multiple Background Group Controls with different prefixes (`background_a`, `background_b`).

You will need to use a different control model property to identify it as a background control.

You might need to add a property (e.g. `'is_background_image'`) to the `'image'` field in `background.php` which you could then use in the CSS parser to identify it.

assets/dev/js/editor/utils/controls-css-parser.js
udidol: You don't need to iterate over these breakpoints to get the responsive control name.

Each responsive control has a `responsive` property that you can use to get the prefix. Specifically, we already have an existing method that does exactly what you need:
`elementor.conditions.getResponsiveControlDeviceSuffix( control.responsive )`

assets/dev/js/editor/utils/controls-css-parser.js
```diff
@@ -37,6 +37,34 @@ ControlsCSSParser = elementorModules.ViewModule.extend( {
...
+				const breakpoints = elementorFrontend.config.responsive.breakpoints;
+				Object.entries( breakpoints ).forEach( ( [ breakpointName ] ) => {
+					if ( control.name.includes( breakpointName ) ) {
+						responsiveSize += '_' + breakpointName;
+					}
+				} );
+				const sizeValue = values[ responsiveSize ];
+				const originalImage = values[ control.name ];
+				if ( sizeValue && originalImage ) {
+					const imageURL = elementor.imagesManager.getImageUrl( {
```
udidol, line 55: This is great, I didn't know this existed.

assets/dev/js/editor/utils/controls-css-parser.js
```diff
@@ -37,6 +37,34 @@ ControlsCSSParser = elementorModules.ViewModule.extend( {
 		} );
 	},
 
+	updateBackgroundImagesDimension( control, values ) {
+		const backgroundDimensionsPrefix = 'background_dimensions';
+		if ( control.name.includes( 'background_image' ) ) {
```
udidol, line 42: Also, instead of a large `if` statement, to make code more readable and avoid nesting, I would just `return values` if the control isn't a background image control

assets/dev/js/editor/utils/controls-css-parser.js
```diff
@@ -37,6 +37,34 @@ ControlsCSSParser = elementorModules.ViewModule.extend( {
...
+			const imageSize = values[ backgroundDimensionsPrefix ];
+			if ( imageSize ) {
```
udidol, line 44: Same here about the large `if` statement

core/files/css/base.php
```diff
@@ -681,6 +707,9 @@ protected function parse_content() {
 	 * @param array $replacements Replacements.
 	 */
 	protected function add_control_style_rules( array $control, array $values, array $controls, array $placeholders, array $replacements ) {
+
```
udidol, line 710: Please remove empty line

core/files/css/base.php
```diff
@@ -666,6 +666,32 @@ protected function parse_content() {
 		return $this->get_stylesheet()->__toString();
 	}
 
+
+	private function update_background_images_dimension( $control, $values ) {
+		$background_dimensions_prefix = 'background_dimensions';
+		if ( false !== strpos( $control['name'], 'background_image' ) ) {
```
udidol, line 672: All of my comments from this method's JS counterpart apply here too

udidol (changes requested):

includes/controls/groups/background.php
maork-elementor: done

includes/controls/groups/background.php
maork-elementor: done

core/files/css/base.php
maork-elementor: done

includes/controls/groups/background.php
maork-elementor: The options on the images sizes are not translated also.

includes/controls/groups/background.php
```diff
@@ -258,6 +258,19 @@ public function init_fields() {
 			],
 		];
 
+		$fields['dimensions'] = [
+			'label' => esc_html_x( 'Dimensions', 'Background Control', 'elementor' ),
+			'type' => Controls_Manager::SELECT,
+			'options' => $this->get_images_sizes_options(),
```
udidol, line 264:
```suggestion
			'options' => $this->get_image_sizes_options(),
```

udidol (changes requested):

includes/controls/groups/background.php
```diff
@@ -814,4 +827,20 @@ protected function get_default_options() {
 			'popover' => false,
 		];
 	}
+
+	protected function get_images_sizes_options() {
```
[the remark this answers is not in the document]
maork-elementor: done

yotamselementor:
@maork-elementor Did you look into using the `srcset` attribute?

yotamselementor:
@maork-elementor What about tests?

includes/controls/groups/background.php
yotamselementor: Are you sure?
Seems like this:
`$result[ $size ] = $size . ' (' . $width . 'x' . $height . ')';`
will result in something like `thumbnail(150xnull)`

assets/dev/js/editor/utils/controls-css-parser.js
yotamselementor: Where is the removal?

includes/controls/groups/background.php
yotamselementor: Where is the change?

includes/controls/groups/background.php
yotamselementor: Where is the change?

includes/controls/groups/background.php
yotamselementor: Is the `$size` string translated?

includes/controls/groups/background.php
maork-elementor: Yas its also like that with the image size, i looked it up.

yotamselementor: What does it look like when one of them is empty?

includes/controls/groups/background.php
maork-elementor: Not relevant anymore, the logic has changed.

core/files/css/base.php
maork-elementor: Not relevant anymore, the logic has changed.

assets/dev/js/editor/utils/controls-css-parser.js
maork-elementor: Not relevant anymore, the logic has changed.

assets/dev/js/editor/utils/controls-css-parser.js
maork-elementor: Not relevant anymore, the logic has changed.

includes/controls/groups/background.php
maork-elementor: Not relevant anymore, the logic has changed.

assets/dev/js/editor/utils/controls-css-parser.js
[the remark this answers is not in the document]
maork-elementor: Not relevant anymore, the logic has changed.

includes/controls/groups/background.php
maork-elementor: Not relevant anymore, the logic has changed.

includes/controls/groups/background.php
maork-elementor: Show the Dimensions name insted of showing 

xxxx 450x0

Nevoss (changes requested):
So there is no backend side for this feature? (Like an image widget that read the size on the backend and replaces the URL?)
And what about tests? seems like a unit test here will be hard to write. most of the code is related to the editor and not a backend or commands

includes/controls/groups/background.php
```diff
@@ -262,6 +262,7 @@ public function init_fields() {
 			'label' => esc_html_x( 'Position', 'Background Control', 'elementor' ),
 			'type' => Controls_Manager::SELECT,
 			'default' => '',
+			'separator' => 'before',
```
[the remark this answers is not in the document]
maork-elementor: Part of the task design. (You can take a look at Figma design)

assets/dev/js/editor/controls/media.js
```diff
@@ -13,6 +13,8 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
 		ui.frameOpeners = '.elementor-control-preview-area';
 		ui.removeButton = '.elementor-control-media__remove';
 		ui.fileName = '.elementor-control-media__file__content__info__name';
+		ui.mediaImageDimensions = '#media-display-dimensions';
+		ui.mediaDimensionsWrapper = '.media-dimensions';
```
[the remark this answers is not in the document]
maork-elementor: Done

assets/dev/js/editor/controls/media.js
```diff
@@ -243,6 +249,50 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
...
+
+	async onMediaImageDimensions() {
+		const selectedDimensions = this.ui.mediaImageDimensions.val();
+		let imageURL = await elementor.imagesManager.getImageUrl( {
+			id: this.getControlValue( 'id' ),
+			size: selectedDimensions,
+		} );
+
+		// In case the image is not set, try to get the image from WP media API.
+		if ( ! imageURL ) {
```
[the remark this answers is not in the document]
maork-elementor: elementor.imagesManager.getImageUrl fails in some cases, mainly when the first image request is made.

assets/dev/js/editor/controls/media.js
```diff
@@ -243,6 +249,50 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
...
+		if ( imageURL ) {
+			this.setValue( {
+				url: imageURL,
+				id: this.getControlValue( 'id' ),
+				alt: this.getControlValue( 'alt' ),
+				dimensions: selectedDimensions,
+				source: 'library',
+			} );
+		}
+		this.renderWithChildren();
```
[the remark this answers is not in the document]
maork-elementor: You are correct !

includes/controls/media.php
```diff
@@ -47,6 +47,7 @@ public function get_default_value() {
 		return [
 			'url' => '',
 			'id' => '',
+			'dimensions' => 'full',
```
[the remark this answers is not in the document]
maork-elementor: I just got new design update from Aviv to rename this control & the existing size control. 
`“Dimensions” to “Image Size”
“Size” to “Display Size”`

Done

includes/controls/media.php
```diff
@@ -271,9 +272,42 @@ public function content_template() {
...
+		<div class="elementor-control-content media-dimensions">
+			<div class="elementor-control-field elementor-control elementor-control-type-select">
+				<label for="elementor-control-default" class="elementor-control-title"><?php echo esc_html__( 'Dimensions', 'elementor' ); ?></label>
+				<div class="elementor-control-input-wrapper">
+				<select data-setting="dimensions" id="media-display-dimensions" class="elementor-control elementor-control-type-select elementor-control-input-wrapper select">
+				<?php
+				$sizes = $this->get_image_sizes_options();
+				foreach ( $sizes as $key => $value ) {
+					// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+					echo "<option value='" . esc_attr( $key ) . "' {{ data.controlValue.dimensions == '$key' ? 'selected' : '' }}>$value</option>";
```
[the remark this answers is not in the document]
maork-elementor: Looks great, thanks.

assets/dev/js/editor/controls/media.js
```diff
@@ -13,6 +13,8 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
 		ui.frameOpeners = '.elementor-control-preview-area';
 		ui.removeButton = '.elementor-control-media__remove';
 		ui.fileName = '.elementor-control-media__file__content__info__name';
+		ui.mediaImageSize = '.media-display-size';
+		ui.mediaSizeWrapper = '.media-size';
```
Nevoss, line 17: Use the above convention.

assets/dev/js/editor/controls/media.js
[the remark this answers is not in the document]
Nevoss: Check for existing solutions for loading the image before

Nevoss (changes requested):

assets/dev/js/editor/controls/media.js
maork-elementor: Done

includes/controls/groups/background.php
maork-elementor: Yes

assets/dev/js/editor/controls/media.js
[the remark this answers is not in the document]
maork-elementor: Done

assets/dev/js/editor/controls/media.js
```diff
@@ -50,10 +53,12 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
 
 		if ( [ 'image', 'svg' ].includes( mediaType ) ) {
 			this.ui.mediaImage.css( 'background-image', url ? 'url(' + url + ')' : '' );
-
```
StyleShit, line 53: revert the lie removal (to have less diff)

assets/dev/js/editor/controls/media.js
```diff
@@ -50,10 +53,12 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
...
 			if ( ! value && url ) {
 				this.ui.mediaImage.css( 'opacity', 0.5 );
 			}
+			this.ui.mediaImage.attr( 'data-id', this.getControlValue( 'id' ) || this.getControlPlaceholder()?.id );
+			this.ui.mediaImage.attr( 'data-alt', this.getControlValue( 'alt' ) || this.getControlPlaceholder()?.alt || '' );
+			this.ui.mediaImage.attr( 'data-size', this.getControlValue( 'size' ) || this.getControlPlaceholder()?.size || 'full' );
```
StyleShit, line 61: explain and/or move to a meaningful function

assets/dev/scss/editor/_global.scss
```diff
@@ -166,3 +165,31 @@
 i.eicon-tilted {
 	transform: rotate(90deg);
 }
+
+.elementor-control .elementor-control-content.elementor-control-media-size {
```
StyleShit, line 169: should it be here or in some `controls/_media.scss` file?

assets/dev/scss/editor/_global.scss
```diff
@@ -166,3 +165,31 @@
...
+	margin-top: 15px;
+}
+
+.elementor-control-media-size .elementor-control-field {
+	padding: 0px;
```
StyleShit, line 174:
```suggestion
	padding: 0;
```

assets/dev/scss/editor/_global.scss
```diff
@@ -166,3 +165,31 @@
...
+}
+
+.elementor-control-media-size .elementor-control-responsive-switchers {
+	display: none;
```
StyleShit, line 178: why?

assets/dev/scss/editor/_global.scss
```diff
@@ -166,3 +165,31 @@
...
+}
+
+.elementor-control-media-display-size.select {
+	padding-bottom: 0px;
```
StyleShit, line 182:
```suggestion
	padding-bottom: 0;
```

assets/dev/scss/editor/_global.scss
```diff
@@ -166,3 +165,31 @@
...
+	width: 72%;
```
StyleShit, line 183: what's this magic number?

assets/dev/scss/editor/_global.scss
```diff
@@ -166,3 +165,31 @@
...
+	float: right;
```
StyleShit, line 184: float?
can't it be solved with some `display` property?

assets/dev/scss/editor/_global.scss
```diff
@@ -166,3 +165,31 @@
...
+}
+
+.media-size-input-wrapper {
+	position: unset;
```
StyleShit, line 188: why?

assets/dev/scss/editor/_global.scss
```diff
@@ -166,3 +165,31 @@
...
+    float: right;
```
StyleShit, line 189: same as above regarding the float

assets/dev/scss/editor/_global.scss
```diff
@@ -166,3 +165,31 @@
...
+    width: 90%;
```
StyleShit, line 190:
```suggestion
.media-size-input-wrapper {
	position: unset;
	float: right;
	width: 90%;
```

assets/dev/scss/editor/_global.scss
```diff
@@ -166,3 +165,31 @@
...
+}
+
+.elementor-control-media-size .elementor-control-title {
+	width: 32%;
```
StyleShit, line 194: same as above regarding the magic number

includes/controls/media.php
```diff
@@ -47,6 +47,7 @@ public function get_default_value() {
 		return [
 			'url' => '',
 			'id' => '',
+			'size' => 'full',
```
StyleShit, line 50: do we want it to be saved in the DB by default?
maybe is should be calculated as default value on render?

includes/controls/media.php
```diff
@@ -271,9 +272,43 @@ public function content_template() {
...
+		$sizes = get_intermediate_image_sizes();
+		$sizes[] = 'full';
+		$result = [];
+		foreach ( $sizes as $size ) {
+			$width = get_option( "{$size}_size_w" );
+			$height = get_option( "{$size}_size_h" );
+			if ( ! $width && ! $height ) {
+				$result[ $size ] = ucfirst( $size );
+			} else {
+				$result[ $size ] = ucfirst( esc_html__( $size, 'elementor' ) ) . " ($width x $height)"; // phpcs:ignore WordPress.WP.I18n.NonSingularStringLiteralText
```
StyleShit, line 306: Is it translated properly?
coz using a non-string in this method usually doesn't work (expect for some edge cases)

includes/controls/media.php
```diff
@@ -271,9 +272,43 @@ public function content_template() {
...
 	}
 
+	protected function get_image_sizes_options() {
+		$sizes = get_intermediate_image_sizes();
+		$sizes[] = 'full';
+		$result = [];
+		foreach ( $sizes as $size ) {
+			$width = get_option( "{$size}_size_w" );
+			$height = get_option( "{$size}_size_h" );
+			if ( ! $width && ! $height ) {
```
StyleShit, line 303:
```suggestion
			if ( ! $width || ! $height ) {
```

assets/dev/js/editor/controls/media.js
```diff
@@ -243,6 +251,54 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
...
+				if ( imageURL ) {
+					stateOptions.url = imageURL;
+					this.setValue( stateOptions );
+				}
+			} );
+		}
+	},
+
+	toggleSizeControl() {
+		const previewImageExist = this.$el.find( '.elementor-control-media__preview' )[ 0 ]?.style?.backgroundImage;
```
StyleShit, line 290: use `$el.find( ... ).css()` or `getComputedStyle()` (also below)

assets/dev/js/editor/controls/media.js
```diff
@@ -243,6 +251,54 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
...
+		const backgroundControl = this.$el[ 0 ]?.className?.match( /background_image/g );
```
StyleShit, line 291: Maybe use `classList.contains()`?

assets/dev/js/editor/controls/media.js
```diff
@@ -243,6 +251,54 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
...
+		if ( backgroundControl && previewImageExist ) {
+			const inheritedBgImage = '0.5' === this.$el.find( '.elementor-control-media__preview' )[ 0 ].style.opacity;
```
StyleShit, line 293: I wouldn't trust this hard-coded value

assets/dev/js/editor/controls/media.js
```diff
@@ -243,6 +251,54 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
...
+			if ( inheritedBgImage ) {
+				this.ui.mediaImageSize.css( 'color', '#a0a5aa' );
+			}
+			this.ui.mediaSizeWrapper.css( 'display', 'flex' );
```
StyleShit, line 297: don't hardcode CSS values in JS, use classes or something

assets/dev/js/editor/controls/media.js
```diff
@@ -243,6 +251,54 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
 	onBeforeDestroy() {
 		this.$el.remove();
 	},
+
+	async onMediaImageSize() {
+		this.ui.mediaImageSize.css( 'color', 'inherit' );
```
StyleShit, line 256: explain

assets/dev/js/editor/controls/media.js
```diff
@@ -243,6 +251,54 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
...
+		const previewPlaceHolder = this.$el.find( '.elementor-control-media__preview' );
```
StyleShit, line 257: You should have this element under `this.ui.mediaImage`

assets/dev/js/editor/controls/media.js
```diff
@@ -243,6 +251,54 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
...
+		const selectedImageId = {
+			id: this?.getControlValue( 'id' ) || previewPlaceHolder[ 0 ].getAttribute( 'data-id' ),
+			alt: this?.getControlValue( 'alt' ) || previewPlaceHolder[ 0 ].getAttribute( 'alt' ),
```
StyleShit, line 260:
```suggestion
			id: this.getControlValue( 'id' ) || previewPlaceHolder[ 0 ].getAttribute( 'data-id' ),
			alt: this.getControlValue( 'alt' ) || previewPlaceHolder[ 0 ].getAttribute( 'alt' ),
```

assets/dev/js/editor/controls/media.js
```diff
@@ -243,6 +251,54 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
 	onBeforeDestroy() {
 		this.$el.remove();
 	},
+
+	async onMediaImageSize() {
+		this.ui.mediaImageSize.css( 'color', 'inherit' );
+		const previewPlaceHolder = this.$el.find( '.elementor-control-media__preview' );
+		const selectedImageId = {
```
StyleShit, line 258:
```suggestion
		const selectedImage = {
```

assets/dev/js/editor/controls/media.js
```diff
@@ -243,6 +251,54 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
...
+			id: selectedImageId.id,
+			alt: selectedImageId.alt,
+			dimensions: selectedSize,
+			source: 'library',
+		};
+		if ( imageURL ) {
+			stateOptions.url = imageURL;
+			this.setValue( stateOptions );
+		} else {
+			elementor.channels.editor.once( 'imagesManager:detailsReceived', ( data ) => {
```
StyleShit, line 279: You might have a race condition here, where you miss this event since it already finished before adding this listener. 
Maybe always add this event before calling `getImageUrl()`?
Or maybe always throw the same event when the value is returned from cache, and you'll be able to use the same mechanism for both?

Something like this:
```JS
elementor.channels.editor.once( 'imagesManager:detailsReceived', ( data ) => {
    this.setValue( { /* ... */ } );
} );

elementor.imagesManager.getImageUrl( ... );
```

Anyway, looks like the images manager will throw this event only if the panel is open (and when using "custom" size):
https://github.com/elementor/elementor/blob/39ffa61c3da41fc4ae87e97ba0a33ee3d6f51b8d/assets/dev/js/editor/utils/images-manager.js#L58-L65

So you can have another bug there. worth checking

StyleShit (changes requested):

assets/dev/js/editor/controls/media.js
maork-elementor: Done

assets/dev/js/editor/controls/media.js
maork-elementor: Done

assets/dev/scss/editor/_global.scss
maork-elementor: There is no such SCSS file that directly related to media control.

assets/dev/scss/editor/_global.scss
maork-elementor: Done

assets/dev/scss/editor/_global.scss
maork-elementor: There is no need to show twice the breakpoint switcher in the same control. 
I extend the media control, not added a new one.

assets/dev/js/editor/controls/media.js
maork-elementor: Remove the gray color text color when user changes the breakpoint view. 
The default value is with gray color, when the user change the size the gray color is removed.
As the design mentions.

assets/dev/js/editor/controls/media.js
maork-elementor: Done

assets/dev/js/editor/controls/media.js
maork-elementor: Done

assets/dev/js/editor/controls/media.js
maork-elementor: Done

assets/dev/js/editor/controls/media.js
maork-elementor: The first run triggers only the hook.
Then the function return the value and without the hook. 
The hook runs only on the first call when `getImageUrl` is actually request the image data.

assets/dev/scss/editor/_global.scss
maork-elementor: The panel has x Width, the select need to be 72% of his wrapper width in order to be aligned with all the other inputs.

assets/dev/scss/editor/_global.scss
maork-elementor: No, I gave it a try first.

assets/dev/scss/editor/_global.scss
maork-elementor: UI issue from the wrapper CSS

assets/dev/scss/editor/_global.scss
maork-elementor: UI issue from the wrapper CSS

assets/dev/scss/editor/_global.scss
maork-elementor: Done

assets/dev/scss/editor/_global.scss
maork-elementor: The panel has x Width, the select need to be 32% of his wrapper width in order to be aligned with all the other inputs.

includes/controls/media.php
maork-elementor: Full should be the default when the user is not selecting size. The original image size.

includes/controls/media.php
maork-elementor: Do we have a solution for this issue?

includes/controls/media.php
maork-elementor: This is not the logic I need.

assets/dev/js/editor/controls/media.js
maork-elementor: Done

assets/dev/js/editor/controls/media.js
maork-elementor: `classList` is no good for me. I need to test if there is any class which contain sub string of 

`xxxxxx-background_image`

assets/dev/js/editor/controls/media.js
maork-elementor: There is no other way to achieve this info, I also saw that there are more places that use the same logic. 
Do you have a better solution for that?

assets/dev/js/editor/controls/media.js
maork-elementor: There is already a use of this kind of functionality in the same file, and also in other widgets. Sample from the same file : 

`this.ui.mediaImage.css( 'background-image', url ? 'url(' + url + ')' : '' );
			if ( ! value && url ) {
				this.ui.mediaImage.css( 'opacity', 0.5 );
}`

assets/dev/scss/editor/_global.scss
StyleShit: can you fix it with flex or something?

assets/dev/scss/editor/_global.scss
StyleShit: Are you sure?
There must be a way

assets/dev/scss/editor/_global.scss
StyleShit: yeah, but explain the issue in the code. Plus, which issue can a position cause?

includes/controls/media.php
StyleShit: If you won't have one of them. you'll have a bug

assets/dev/js/editor/controls/media.js
StyleShit: background image is different than color / display.
the opacity in the original code is also bad

assets/dev/js/editor/controls/media.js
```diff
@@ -243,6 +252,54 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
...
+				if ( imageURL ) {
+					stateOptions.url = imageURL;
+					this.setValue( stateOptions );
+				}
+			} );
+		}
+	},
+
+	toggleSizeControl() {
+		const previewImageExist = this.$el.find( '.elementor-control-media__preview' )[ 0 ]?.style?.backgroundImage;
```
StyleShit, line 291: you have this under `this.ui.mediaImage`

includes/controls/media.php
StyleShit: yes, but the user might have unused data in the DB. We need to ensure that this value isn't saved when not needed

StyleShit (changes requested):

assets/dev/scss/editor/_global.scss
maork-elementor: In the near future, we will need to support several such breakpoint support buttons in one control, so I prefer not to make changes to this library.

assets/dev/scss/editor/_global.scss
maork-elementor: Done

assets/dev/scss/editor/_global.scss
maork-elementor: Done, i set it to `0`

assets/dev/scss/editor/_global.scss
maork-elementor: No, because he won't be aligned with all other inputs below him. (I test it)

assets/dev/scss/editor/_global.scss
maork-elementor: `position` removed. It was unnecessary after all.

includes/controls/media.php
maork-elementor: Removed

assets/dev/js/editor/controls/media.js
maork-elementor: This is the only indicator that I found.

assets/dev/js/editor/controls/media.js
maork-elementor: Only check for property existing & use class regarding the text color of the select input.

assets/dev/js/editor/controls/media.js
maork-elementor: I replaced this with class.

assets/dev/js/editor/controls/media.js
```diff
@@ -42,18 +45,21 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
...
+		this.ui.mediaImage.attr( 'data-size', this.getControlValue( 'size' ) || this.getControlPlaceholder()?.size || 'full' );
+	},
 	applySavedValue() {
 		const value = this.getControlValue( 'url' ),
 			url = value || this.getControlPlaceholder()?.url,
 			mediaType = this.getMediaType();
-
 		if ( [ 'image', 'svg' ].includes( mediaType ) ) {
 			this.ui.mediaImage.css( 'background-image', url ? 'url(' + url + ')' : '' );
-
```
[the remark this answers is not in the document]
maork-elementor: Only the upper line can be restored. 
The line below is reformatting on save from the ESLint. And the line is auto removed.

assets/dev/js/editor/controls/media.js
maork-elementor: Using `this.ui.mediaImage`

assets/dev/js/editor/controls/media.js
maork-elementor: Opacity value removed. Just checking for existent.

includes/controls/media.php
maork-elementor: Done

assets/dev/js/editor/controls/media.js
maork-elementor: Change the selectors

assets/dev/js/editor/controls/media.js
```diff
@@ -42,18 +45,21 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
 
 		return ( 'svg' === mediaType ) ? 'image/svg+xml' : mediaType;
 	},
-
+	setFallbackValues() {
+		this.ui.mediaImage.attr( 'data-id', this.getControlValue( 'id' ) || this.getControlPlaceholder()?.id );
+		this.ui.mediaImage.attr( 'data-alt', this.getControlValue( 'alt' ) || this.getControlPlaceholder()?.alt || '' );
+		this.ui.mediaImage.attr( 'data-size', this.getControlValue( 'size' ) || this.getControlPlaceholder()?.size || 'full' );
```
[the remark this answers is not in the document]
maork-elementor: Every time a break point is changed, the instance is reset

assets/dev/scss/editor/_global.scss
maork-elementor: Worked on the second e try, done.

assets/dev/js/editor/controls/media.js
maork-elementor: 1. It is not matter of fast request, The imagesManager has a cache mechanism that will always trigger the `imagesManager:detailsReceived` hook on the first request. And later on from cache object.
As you can see here :

https://github.com/elementor/elementor/blob/945ede74b8cfcff4478f8f5d0d752534e18930fa/assets/dev/js/editor/utils/images-manager.js#L150

 I also consulted with Mati and Nevo about this matter, this is the solution we found in the configuration that the imagesManager API is currently working on.

2. the images size currently not does not support custom size, So I guess we'll have to deal with the moment once we implement support for custom size.

includes/controls/media.php
maork-elementor: Fixed, as soon as you select a desired size and save, only this is saved in the db, that is, a controller that does not have the selection will not populate the database with the default value (`full`).

assets/dev/js/editor/controls/media.js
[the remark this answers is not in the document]
maork-elementor: Removed use getSetting - setSetting in browser memory

assets/dev/js/editor/controls/media.js
```diff
@@ -222,27 +254,79 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
...
+
+	async onMediaImageSize() {
+		this.ui.mediaImageSize.removeClass( 'e-select-placeholder' );
+		const selectedImage = {
+			id: this.getEditSettings( 'mediaId' ),
+			alt: this.getEditSettings( 'mediaAlt' ),
+		};
+
+		const selectedSize = this.ui.mediaImageSize.val();
+		let imageURL = await elementor.imagesManager.getImageUrl( {
```
StyleShit, line 290:
```suggestion
		let imageURL = elementor.imagesManager.getImageUrl( {
```

(you can probably also remove the `async` from the function signature)

assets/dev/js/editor/controls/media.js
```diff
@@ -222,27 +254,79 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
...
 	onBeforeDestroy() {
 		this.$el.remove();
 	},
+
+	async onMediaImageSize() {
+		this.ui.mediaImageSize.removeClass( 'e-select-placeholder' );
+		const selectedImage = {
+			id: this.getEditSettings( 'mediaId' ),
+			alt: this.getEditSettings( 'mediaAlt' ),
+		};
```
StyleShit, line 287: Maybe add a method `this.getPrevState()`?

assets/dev/js/editor/controls/media.js
```diff
@@ -222,27 +254,79 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
...
+					this.setValue( stateOptions );
+				}
+			} );
+		}
+	},
+
+	toggleSizeControl() {
+		const sizesSupport = this.model.get( 'sizes_supported' );
+		const hasImage = this.getEditSettings( 'image_url' );
+		const opacity = this.getEditSettings( 'image_opacity' );
```
StyleShit, line 320: `hasImage` & `opacity` don't really say a lot. maybe we should try finding better names (same as the `setEditSettings()` we need to change above).

And regarding the "sizes support", maybe let's ask @matipojo if there's any naming convention for that

assets/dev/js/editor/utils/images-manager.js
```diff
@@ -48,6 +48,7 @@ ImagesManager = function() {
 		} );
 	};
 
+	// Ir url is exist, return it from cache, otherwise, get it from server and trigger event.
 	self.getImageUrl = function( image ) {
```
StyleShit, line 52: maybe add a callback parameter, like @matipojo said?

assets/dev/js/editor/utils/images-manager.js
```diff
@@ -127,6 +128,7 @@ ImagesManager = function() {
 			} );
 		}
 
+		// On the request, the hook 'imagesManager:detailsReceived' is triggered with the response (image Id)
```
StyleShit, line 131:
```suggestion
		// On success, an 'imagesManager:detailsReceived' event is fired with the response.
```

assets/dev/scss/editor/panel/controls/_media.scss
```diff
@@ -245,3 +245,34 @@
...
+	margin-top: 15px;
+}
+
+.elementor-control-media-size {
+
+	.elementor-control-field {
+		padding: 0;
+	}
+
+	// Control won't support more than one breakpoints
```
StyleShit, line 259: explain it more (about the bug and everything we've talked about)

assets/dev/scss/editor/panel/controls/_media.scss
```diff
@@ -245,3 +245,34 @@
...
+	.media-size-input-wrapper {
+		margin-left: auto;
+	}
+
+}
+
+.elementor-control-media-display-size.select {
+
+	padding-bottom: 0;
+	margin-left: auto;
```
StyleShit, line 273:
```suggestion
	margin-inline-start: auto;
```

assets/dev/scss/editor/panel/controls/_media.scss
```diff
@@ -245,3 +245,34 @@
 		}
 	}
 }
+
+.elementor-control .elementor-control-content.elementor-control-media-size {
```
StyleShit, line 249: try to put this CSS inside the existing one (use SASS concatenation things `&-size`, etc.)

includes/controls/media.php
```diff
@@ -271,9 +271,50 @@ public function content_template() {
...
+				<label for="elementor-control-default" class="elementor-control-title"><?php echo esc_html__( 'Image Size', 'elementor' ); ?></label>
+				<div class="elementor-control-input-wrapper media-size-input-wrapper elementor-control-unit-5">
+				<select data-setting="size" class="elementor-control elementor-control-type-select elementor-control-input-wrapper elementor-control-media-display-size select">
+				<?php
+				$sizes = $this->get_image_sizes_options();
+				foreach ( $sizes as $key => $value ) {
+					?>
+						<option value="<?php echo esc_attr( $key ); ?>" 
+						{{ data.controlValue.size && data.controlValue.size == '<?php echo esc_attr( $key ); ?>' ? 'selected' : '' }}
+						{{ ! data.controlValue.size && '<?php echo esc_attr( $key ); ?>' == 'full' ? 'selected' : '' }}
```
StyleShit, line 288: maybe extract this logic to a function above the render?

includes/controls/media.php
```diff
@@ -271,9 +271,50 @@ public function content_template() {
...
+		$sizes[] = 'full';
+		$result = [];
+		foreach ( $sizes as $size ) {
+			$width = get_option( "{$size}_size_w" );
+			$height = get_option( "{$size}_size_h" );
+			if ( $width && $height ) {
+				$result[ $size ] = ucfirst( $size ) . " ($width x $height)";
+			} else {
+				$result[ $size ] = ucfirst( $size );
+			}
```
StyleShit, line 313:
```suggestion
			$result[ $size ] = ucfirst( $size );
			
			if ( $width && $height ) {
				$result[ $size ] .= " ($width x $height)";
			}
```

Maybe?

StyleShit (changes requested):

assets/dev/js/editor/controls/media.js
```diff
@@ -43,6 +46,21 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
...
 	},
 
+	updatePrevState( options ) {
+		if ( ! options.id ) {
+			return;
+		}
+		this.setEditSetting( 'mediaId', options.id );
+		this.setEditSetting( 'mediaAlt', options.alt );
+		this.setEditSetting( 'mediaSize', options.size );
+	},
```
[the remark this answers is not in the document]
maork-elementor: Comment added

assets/dev/js/editor/controls/media.js
```diff
@@ -43,6 +46,21 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
...
+
+	resetPrevState() {
+		this.setEditSetting( 'mediaId', null );
+		this.setEditSetting( 'mediaAlt', null );
+		this.setEditSetting( 'mediaSize', null );
+	},
```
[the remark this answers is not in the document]
maork-elementor: https://github.com/elementor/elementor/blob/cf465842dd6e009cd50847f4ed2b8e289b875a10/assets/dev/js/editor/controls/base-data.js#L175

Which problem it will resolve?

assets/dev/js/editor/controls/media.js
```diff
@@ -51,9 +69,20 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
 		if ( [ 'image', 'svg' ].includes( mediaType ) ) {
 			this.ui.mediaImage.css( 'background-image', url ? 'url(' + url + ')' : '' );
 
+			if ( value ) {
+				this.setEditSetting( 'image_url', true );
+			}
+
 			if ( ! value && url ) {
 				this.ui.mediaImage.css( 'opacity', 0.5 );
+				this.setEditSetting( 'image_opacity', true );
```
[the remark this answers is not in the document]
maork-elementor: Done

assets/dev/js/editor/controls/media.js
```diff
@@ -43,6 +46,21 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
 		return ( 'svg' === mediaType ) ? 'image/svg+xml' : mediaType;
 	},
 
+	updatePrevState( options ) {
```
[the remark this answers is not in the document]
maork-elementor: Done

assets/dev/js/editor/controls/media.js
```diff
@@ -222,27 +254,79 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
...
+			this.toggleSizeControl();
+		}
 		this.trigger( 'after:select' );
 	},
 
 	onBeforeDestroy() {
 		this.$el.remove();
 	},
+
+	async onMediaImageSize() {
```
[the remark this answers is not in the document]
maork-elementor: Done

assets/dev/js/editor/controls/media.js
maork-elementor: Removed

assets/dev/js/editor/controls/media.js
maork-elementor: Why to create a proxy function for `getEditSettings`? 
I don't think that this is a good idea.

assets/dev/js/editor/controls/media.js
maork-elementor: Vars are renamed from previews review.

assets/dev/js/editor/controls/media.js
```diff
@@ -222,27 +254,79 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
...
+		const hasImage = this.getEditSettings( 'image_url' );
+		const opacity = this.getEditSettings( 'image_opacity' );
+
+		if ( sizesSupport && hasImage ) {
+			if ( opacity ) {
+				this.ui.mediaImageSize.addClass( 'e-select-placeholder' );
+			}
+		} else {
+			this.ui.mediaSizeWrapper.hide();
+		}
```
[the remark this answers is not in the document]
maork-elementor: If there is no opacity than the select text color will keep be black.
which mean that the breakpoint dont have default value from previews break point

assets/dev/js/editor/utils/images-manager.js
```diff
@@ -48,6 +48,7 @@ ImagesManager = function() {
 		} );
 	};
 
+	// Ir url is exist, return it from cache, otherwise, get it from server and trigger event.
```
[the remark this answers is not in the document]
maork-elementor: Done

assets/dev/js/editor/utils/images-manager.js
maork-elementor: I am not changing the function with the current task scope.

assets/dev/js/editor/utils/images-manager.js
maork-elementor: Done

assets/dev/scss/editor/panel/controls/_media.scss
maork-elementor: Done

assets/dev/scss/editor/panel/controls/_media.scss
maork-elementor: Done

includes/controls/media.php
maork-elementor: Changed

includes/controls/media.php
maork-elementor: What for? It won't be reused.

assets/dev/scss/editor/panel/controls/_media.scss
maork-elementor: Changed.

assets/dev/js/editor/controls/media.js
```diff
@@ -51,9 +70,20 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
 		if ( [ 'image', 'svg' ].includes( mediaType ) ) {
 			this.ui.mediaImage.css( 'background-image', url ? 'url(' + url + ')' : '' );
 
+			if ( value ) {
+				this.setEditSetting( 'preview_image', true );
```
StyleShit, line 74:
```suggestion
				this.setEditSetting( 'has_image', true );
```

assets/dev/js/editor/controls/media.js
```diff
@@ -51,9 +70,20 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
...
+			}
+
 			if ( ! value && url ) {
 				this.ui.mediaImage.css( 'opacity', 0.5 );
+				this.setEditSetting( 'preview_image_opacity', true );
```
StyleShit, line 79:
```suggestion
				this.setEditSetting( 'has_placeholder_image', true );
```

assets/dev/scss/editor/panel/controls/_media.scss
```diff
@@ -238,10 +241,44 @@
...
+
+.elementor-control-media-display-size.select {
+
+	padding-bottom: 0;
+	margin-left: auto;
+
+	.e-select-placeholder {
+		display: flex;
+	}
+}
```
StyleShit, line 284: add empty line

assets/dev/scss/editor/panel/controls/_media.scss
```diff
@@ -238,10 +241,44 @@
...
+
+.elementor-control-media-size {
+
+	.elementor-control-field {
+		padding: 0;
+	}
+
+	// Multipal switcher are supported in the same control.
```
StyleShit, line 265: explain

includes/controls/groups/background.php
```diff
@@ -256,12 +256,14 @@ public function init_fields() {
 			'condition' => [
 				'background' => [ 'classic' ],
 			],
+			'sizes_supported' => true,
```
StyleShit, line 259: add to the control itself as default

StyleShit (changes requested):

includes/controls/media.php
maork-elementor: Done

assets/dev/js/editor/controls/media.js
maork-elementor: Done

assets/dev/js/editor/controls/media.js
maork-elementor: Done

assets/dev/scss/editor/panel/controls/_media.scss
maork-elementor: Done

assets/dev/scss/editor/panel/controls/_media.scss
maork-elementor: Done

assets/dev/js/editor/controls/media.js
```diff
@@ -222,27 +255,78 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
 		}
 
 		if ( attachment.url ) {
-			this.setValue( {
+			const updatedState = {
```
KingYes, line 258: Why?

assets/dev/js/editor/controls/media.js
```diff
@@ -222,27 +255,78 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
...
+			url: null,
+			id: selectedImage.id,
+			alt: selectedImage.alt,
+			dimensions: selectedSize,
+			source: 'library',
+		};
+
+		if ( imageURL ) {
+			stateOptions.url = imageURL;
+			this.setValue( stateOptions );
```
KingYes, line 306: You can exit from the function and remove the `else` statement.

assets/dev/js/editor/controls/media.js
```diff
@@ -222,27 +255,78 @@ ControlMediaItemView = ControlMultipleBaseItemView.extend( {
...
+		}
 		this.trigger( 'after:select' );
 	},
 
 	onBeforeDestroy() {
 		this.$el.remove();
 	},
+
+	onMediaImageSizeChange() {
+		this.ui.mediaImageSize.removeClass( 'e-select-placeholder' );
```
KingYes, line 284: Why you don't have one place to handle the UI like this?

includes/controls/media.php
```diff
@@ -271,9 +271,49 @@ public function content_template() {
...
+				?>
+				</select>
+				</div>
+			</div>
+		</div>
+	<# } #>
 		<?php
 	}
 
+	protected function get_image_sizes_options() {
```
KingYes, line 302: @StyleShit Do you have idea how to write function like this with better logic?

KingYes (changes requested):

assets/dev/js/editor/controls/media.js
maork-elementor: I save the control state on every change, So when the user switch from one breakpoint to another, I can get the placeholders from the preview's breakpoint. And show them as defaults.

assets/dev/js/editor/controls/media.js
maork-elementor: Screen Shot 2022-12-06 at 13 44 17

I also need the else to listen to `imagesManager:detailsReceived` hook, only on the first Ajax request to get the image data. You can also read my discussion with @StyleShit 

https://github.com/elementor/elementor/pull/20361#discussion_r1033167730

includes/controls/media.php
```diff
@@ -197,6 +197,16 @@ public function content_template() {
...
+
+			isSeletedSizeOption = ( key ) => {
+				if ( data.controlValue.size && data.controlValue.size == key ) {
+					return true;
+				}
+				if ( ! data.controlValue.size && 'full' == key ) {
+					return true;
+				}
+				return false;
+			}
```
StyleShit, line 209:
```suggestion
			isSeletedSizeOption = ( key ) => {
				if ( ! data.controlValue.size ) {
					return 'full' === key;
				}
				
				return data.controlValue.size === key;
			}
```

includes/controls/media.php
```diff
@@ -197,6 +197,16 @@ public function content_template() {
 						return '<?php esc_html_e( 'Choose File', 'elementor' ); ?>';
 				}
 			}
+
+			isSeletedSizeOption = ( key ) => {
```
StyleShit, line 201:
```suggestion
			isSelectedSizeOption = ( key ) => {
```

StyleShit, line 201:
```suggestion
			const isSeletedSizeOption = ( key ) => {
```

StyleShit (changes requested):

includes/controls/media.php
maork-elementor: done

includes/controls/media.php
maork-elementor: done

assets/dev/js/editor/controls/media.js
maork-elementor: Done

PR #20366, opened 2022-11-10
Tweak: Updated the content in nested tabs [ED-8546]

hein-obox:


modules/tabs-v2/widgets/tabs-v2.php
```diff
@@ -158,10 +158,11 @@ protected function register_controls() {
 				],
 			],
 			'title_field' => '{{{ tab_title }}}',
+			'button_text' => 'Add Tab',
 		] );
 
-		$this->add_responsive_control( 'tabs_position', [
-			'label' => esc_html__( 'Position', 'elementor' ),
+		$this->add_responsive_control( 'tabs_direction', [
```
davseve, line 164: I noticed that the order of the Direction buttons has changed between Aviv's video to what appears in this branch.
[Avivs video](https://nimb.ws/wc6Ure):
[Current branch](https://nimb.ws/pscJWC)

hein-obox: @davseve 
Can you please double check on your side if it works correctly?

davseve: It works as it should, only the order changed.

hein-obox: I have checked it. It seems to work correctly.
I think we continue to PR Review 2.

davseve (approved):

PR #20385, opened 2022-11-13
Fix: Lazy load is not working on widgets when using motion effects [ED-8709]

StyleShit:


modules/lazyload/assets/js/frontend.js
```diff
@@ -8,7 +8,7 @@ document.addEventListener( 'DOMContentLoaded', function() {
 
                 const lazyloadSelector = lazyloadBackground.getAttribute( dataAttribute );
                 if ( lazyloadSelector ) {
-                    lazyloadBackground = lazyloadSelector;
+                    lazyloadBackground = element.querySelector( lazyloadSelector );
```
Nevoss, line 11: Are we sure all those selectors are inside the element?

Nevoss (changes requested):

modules/lazyload/assets/js/frontend.js
StyleShit: Well, it should be, because then how it's the background of the element?
Anyway, currently it looks like the only place where the selector is used is inside the element.

Nevoss (approved):

PR #20416, opened 2022-11-15
Fix: Icon change position when clicking tabs location [ED-8316]

davseve:


modules/tabs-v2/assets/scss/frontend.scss
```diff
@@ -46,6 +46,7 @@
 	--tabs-v2-title-border-radius: initial;
 	--tabs-v2-title-transition: 0.3s;
 
+  	--tabs-v2-icon-vertical-align: center;
```
hein-obox, line 49: Please use the flex box methods for the naming of css variables, in this case possibly `--tabs-v2-icon-align-items`

modules/tabs-v2/assets/scss/frontend.scss
```diff
@@ -138,6 +140,7 @@
 
 				&,
 				&-active {
+				  	align-items: var( --tabs-v2-icon-vertical-align );
```
hein-obox, line 143: Possibly we only the `align-items` here.

modules/tabs-v2/assets/scss/frontend.scss
```diff
@@ -127,6 +128,7 @@
 
 			.elementor-tab-icon {
 				display: flex;
+			  	align-items: var( --tabs-v2-icon-vertical-align );
```
hein-obox, line 131: Can we delete this line?

hein-obox:
Please write a playwright test for this work. 

Have you double checked your work with the icon styling options?

image

PR #20421, opened 2022-11-16
Fix: Post type is overridden on autosave for library documents (#19941) [ED-8591]

### LINKED ISSUE #19941: 🐞 Bug Report: Post type override on autosave (ED-8591)
### Prerequisites

### Description

The plugin overrides a document type on autosave.

It tries to set the document to a custom type when elementor pro is disabled, and the "error 404" type when elementor pro is enabled.

**Possible solution**
Make sure we don't get a cached version of the document if someone tries to get it in some WP hook.

Bypass cache here:
https://github.com/elementor/elementor/blob/master/core/base/document.php#L528

`$document = Plugin::$instance->documents->get( $autosave_id );`
->
`$document = Plugin::$instance->documents->get( $autosave_id, false );`

### Steps to reproduce

-1-
https://github.com/elementor/elementor/blob/master/core/base/document.php#L516
This call triggers the "save_post" action.

-2-
In one of the functions associated with the "save_post" action, I want to get a document by the passed ID in this way.
`\Elementor\Plugin::$instance->documents->get($post_id);`

The problem is that the plugin will cache the requested document with the wrong type because we haven't copied the meta tags yet.

-3-
The plugin then copies the meta tags here
https://github.com/elementor/elementor/blob/master/core/base/document.php#L526

-4-
Here we will get a cached document with the wrong type (cached version from step #2)
https://github.com/elementor/elementor/blob/master/core/base/document.php#L528

-5-
And as a final step, we save the wrong type here and override the metadata from step #3.
https://github.com/elementor/elementor/blob/master/core/base/document.php#L529

### Isolating the problem

- [X] This bug happens with a Blank WordPress theme active ([Hello theme](https://wordpress.org/themes/hello-elementor/)).
- [X] I can reproduce this bug consistently using the steps above.

### System Info

    == Active Plugins ==
	Classic Editor
		Version: 1.5
		Author: WordPress Contributors

	Elementor
		Version: 3.7.6
		Author: Elementor.com

	Gravity Forms
		Version: 2.6.3.4
		Author: Gravity Forms

StyleShit:
The bug was caused due to this part:
https://github.com/elementor/elementor/blob/e8d4d693c3d59ee77aa69d97f397e0aeb342995e/core/documents-manager.php#L150-L154
Each library document has overridden the CPT for all of the others document types, so the last one was the default.

Closes #19941


modules/library/documents/library-document.php
```diff
@@ -44,7 +44,6 @@ public static function get_properties() {
 		$properties['admin_tab_group'] = 'library';
 		$properties['show_in_library'] = true;
 		$properties['register_type'] = true;
```
matipojo, line 46:
```suggestion
		$properties['register_type'] = true;
		$properties['cpt'] = [ Source_Local::CPT ];
		$properties['cpt'] = [ Source_Local::CPT ];
```

modules/library/documents/library-document.php
```diff
@@ -45,6 +45,7 @@ public static function get_properties() {
...
 		$properties['cpt'] = [ Source_Local::CPT ];
+		$properties['cpt'] = [ Source_Local::CPT ];
```
matipojo, line 48:
```suggestion
```

matipojo (changes requested):
Please fix the checks errors

core/documents-manager.php
```diff
@@ -719,4 +702,42 @@ public static function get_create_new_post_url( $post_type = 'page', $template_t
...
+		$template_type = $this->get_template_type_by_id( $post_id );
+
+		if ( $template_type && isset( $this->types[ $template_type ] ) ) {
+			return $template_type;
+		}
+
+		// Auto-save post type should be inherited from the original post.
+		if ( wp_is_post_autosave( $post_id ) ) {
+			$post_id = wp_get_post_parent_id( $post_id );
+		}
```
matipojo, line 717: Maybe move to the top and delete the related code on `get_template_type_by_id`

matipojo (changes requested):

core/documents-manager.php
StyleShit: Done

PR #20511, opened 2022-11-24
Tweak: Added Undo option in "Save as Default" experiment [ED-8785]

StyleShit:


modules/kit-elements-defaults/assets/js/editor/commands/create.js
```diff
@@ -10,16 +10,30 @@ export default class Create extends $e.modules.editor.CommandContainerBase {
...
 
 			elementor.notifications.showToast( {
 				message: __( 'Default settings changed.', 'elementor' ),
+				buttons: [ {
+					name: 'undo',
+					text: __( 'Undo', 'elementor' ),
+					callback() {
+						$e.run( 'kit-elements-defaults/restore', {
+							type,
+							defaults: previousDefaults,
```
Nevoss, line 33:
```suggestion
							settings: previousDefaults,
```

maybe?

Nevoss (dismissed):

modules/kit-elements-defaults/assets/js/editor/commands/create.js
StyleShit: Done

Nevoss (approved):

maork-elementor (approved):

ofmarconi:
It's common for a site to have more than one button style, consider adding save more than one style.

The logic of the models could be used: You can rename and choose through a modal which styles to apply.

Then this solution would be ultimate.

Thanks!

PR #20561, opened 2022-11-29
Fix: Lazy Load experiment caused the background not working some cases [ED-8898]

maork-elementor:


modules/lazyload/assets/scss/frontend.scss
```diff
@@ -1,3 +1,8 @@
-body.e-lazyload .lazyloaded, body.e-lazyload .lazyloaded .elementor-background-overlay {
+.lazyloaded:not( .elementor-motion-effects-element-type-background ),
```
rotemee, line 1: Please add comments that explain the selectors' targets.

includes/elements/container.php
```diff
@@ -690,6 +690,13 @@ protected function register_background_overlay_controls() {
 							$background_overlay_selector => '--background-overlay: \'\';',
 						],
 					],
+					'image' => [
```
rotemee, line 693: @maork-elementor The background doesn't work in a container (see take for more details).

modules/lazyload/module.php
```diff
@@ -54,6 +54,13 @@ private function update_element_attributes( Element_Base $element ) {
 
 		if ( $keys ) {
 			$background_image_url = Utils::get_array_value_by_keys( $settings, $keys );
+			if ( ! $background_image_url ) {
```
rotemee, line 57: Why do you have a separated condition when the $background_image_url does not exist?
why not if/else?

modules/lazyload/module.php
```diff
@@ -54,6 +54,13 @@ private function update_element_attributes( Element_Base $element ) {
...
+				// If the background image is not set, check also for overlay image.
+				$keys = Utils::get_array_value_by_keys( $control_data, [ 'background_lazyload', 'overlay_keys' ] );
+				if ( $keys ) {
+					$background_image_url = Utils::get_array_value_by_keys( $settings, $keys );
+				}
+			}
```
rotemee, line 63: Add a space below the condition and consider adding spaces in general for readability (to keep on a consistent code style in the project).

modules/lazyload/assets/scss/frontend.scss
maork-elementor: done

includes/elements/container.php
maork-elementor: @rotemee 
I just test is with container experiment ON : 

Only background image - worked.
Background image & overlay image - worked. 

Screen Shot 2022-12-01 at 9 03 56

This is the CSS for the container overlay : 

Screen Shot 2022-12-01 at 9 04 57

What do I need to do to reproduce the problem?

modules/lazyload/module.php
maork-elementor: My PHPCS is not allowing me to add space below the condition.
The empty lines are auto removed when I save the file.
I'm using our config file `phpcs.xml`

includes/elements/container.php
rotemee: @maork-elementor:
"Create a container, and set a background overlay → the background does not load."
Try only a background overlay (without a regular background).

maork-elementor: You are right, Fixed.

modules/lazyload/module.php
maork-elementor: PHPCS error is generated by the line break in both my local and the actual PR test.

rotemee (dismissed):

includes/elements/column.php
```diff
@@ -298,7 +298,8 @@ protected function register_controls() {
 						'background_lazyload' => [
 							'active' => true,
 							'keys' => [ 'background_image', 'url' ],
-							'selector' => '.elementor-column-wrap',
+							'overlay_keys' => [ 'background_overlay_image', 'url' ],
```
Nevoss, line 301: the overlay is not a different control? lets talk about it

Nevoss (changes requested):

includes/elements/column.php
maork-elementor: Keys are removed, using new logic to support multiple background controls in the same element.

Nevoss (approved):

PR #20573, opened 2022-11-30
Fix: Nested Tabs styling issues on tabs section [ED-8841]

davseve:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #


modules/nested-tabs/widgets/nested-tabs.php
```diff
@@ -425,6 +425,9 @@ protected function register_controls() {
 				'fields_options' => [
 					'color' => [
```
hein-obox, line 426: Please make sure to update the background gradient selector as well.

modules/nested-tabs/assets/scss/frontend.scss
```diff
@@ -34,7 +34,8 @@
 	--n-tabs-title-background-color: #{$editor-background-light};
 	--n-tabs-title-color-active: #{$white};
 	--n-tabs-title-background-color-active: var( --e-global-color-accent, var( --n-tabs-color-accent-fallback ) );
-	--n-tabs-title-width: initial;
+	--n-tabs-title-background-color-hover: var( --e-global-color-accent, var( --n-tabs-color-accent-fallback ) );
```
hein-obox, line 37: Safe to delete?

davseve: I will delete them once I'll see that all the backgrounds work well.
Good comment :)

modules/nested-tabs/widgets/nested-tabs.php
```diff
@@ -423,9 +423,6 @@ protected function register_controls() {
 				'exclude' => [ 'image' ],
 				'selector' => "{$nested_tabs_heading_selector_class} > .e-n-tab-title",
 				'fields_options' => [
-					'background' => [
-						'default' => 'classic',
-					],
```
hein-obox: Why is this removed?

modules/nested-tabs/widgets/nested-tabs.php
```diff
@@ -85,7 +85,7 @@ protected function register_controls() {
 		$end = is_rtl() ? 'left' : 'right';
 		$tooltip_start = is_rtl() ? esc_html__( 'Right', 'elementor' ) : esc_html__( 'Left', 'elementor' );
 		$tooltip_end = is_rtl() ? esc_html__( 'Left', 'elementor' ) : esc_html__( 'Right', 'elementor' );
-		$nested_tabs_heading_selector_class = '{{WRAPPER}} > .elementor-widget-container > .e-n-tabs > .e-n-tabs-heading';
+		$nested_tabs_heading_selector_class = ':is( {{WRAPPER}} > .elementor-widget-container > .e-n-tabs > .e-n-tabs-heading, {{WRAPPER}} > .elementor-widget-container > .e-n-tabs > .e-n-tabs-content )';
```
hein-obox, line 88: Could you add a `>` charachter after the ` .e-n-tabs-content`? If we do that, we don't have to add them at the selectors below.

modules/nested-tabs/widgets/nested-tabs.php
```diff
@@ -421,10 +421,13 @@ protected function register_controls() {
 				'name' => 'tabs_title_background_color',
 				'types' => [ 'classic', 'gradient' ],
 				'exclude' => [ 'image' ],
-				'selector' => "{$nested_tabs_heading_selector_class} .e-n-tab-title",
+				'selector' => "{$nested_tabs_heading_selector_class} > .e-n-tab-title",
```
hein-obox, line 424: Possibly delete and restore to original version. See comment above.

modules/nested-tabs/widgets/nested-tabs.php
```diff
@@ -972,7 +987,7 @@ protected function register_controls() {
 			Group_Control_Box_Shadow::get_type(),
 			[
 				'name' => 'box_shadow_box_shadow',
-				'selector' => "{$nested_tabs_content_selector_class}",
+				'selector' => '{{WRAPPER}} .e-n-tabs-content',
```
hein-obox, line 990: This is incorrect. Now the styling will be applied to child nested tab widgets as well.

modules/nested-tabs/widgets/nested-tabs.php
```diff
@@ -421,10 +421,13 @@ protected function register_controls() {
 				'name' => 'tabs_title_background_color',
 				'types' => [ 'classic', 'gradient' ],
 				'exclude' => [ 'image' ],
-				'selector' => "{$nested_tabs_heading_selector_class} .e-n-tab-title",
+				'selector' => "{$nested_tabs_heading_selector_class} > .e-n-tab-title",
 				'fields_options' => [
 					'color' => [
 						'label' => esc_html__( 'Background Color', 'elementor' ),
+						'selectors' => [
+							'{{WRAPPER}}' => '--n-tabs-title-background-color: {{VALUE}}',
```
hein-obox, line 429: I am curious about the logic.

We use a css variable here, but for the hover and active state we use a php selector.

What is the logic?

modules/nested-tabs/assets/scss/frontend.scss
```diff
@@ -229,6 +229,17 @@
 	}
 }
 
+// TODO: Remove once PHP global color support is fixed.
+.elementor .elementor-element.elementor-widget-n-tabs > .elementor-widget-container > .e-n-tabs > .e-n-tabs-heading .e-n-tab-title {
+	&.e-active {
+		background: var( --n-tabs-title-background-color-active );
```
hein-obox, line 235: Could it work if we use `background-color` here and then add `background-image: initial;`?

modules/nested-tabs/assets/scss/frontend.scss
```diff
@@ -159,7 +159,7 @@
 			&.e-active {
 				&:not(:hover) {
 					color: var( --n-tabs-title-color-active );
-					background-color: var( --n-tabs-title-background-color-active );
+					background: var( --n-tabs-title-background-color-active );
```
hein-obox, line 162: Shouldn't we remove these and move them to the new css selector with the higher specificity?

modules/nested-tabs/assets/scss/frontend.scss
```diff
@@ -191,7 +191,7 @@
 
 			&:hover {
 				color: var( --n-tabs-title-color-hover );
-				background-color: var( --n-tabs-title-background-color-hover );
+				background: var( --n-tabs-title-background-color-hover );
```
hein-obox, line 194: Shouldn't we remove these and move them to the new css selector with the higher specificity?

tests/playwright/sanity/modules/nested-tabs/nested-tabs.test.js
```diff
@@ -435,6 +435,86 @@ test.describe( 'Nested Tabs tests', () => {
 
 		await cleanup( wpAdmin );
 	} );
+
+	test( 'Check that active tab receives custom hover color when hovering above it', async ( { page }, testInfo ) => {
```
hein-obox, line 439: Suggestion: 'Check if the hover effect works on active tabs`

tests/playwright/sanity/modules/nested-tabs/nested-tabs.test.js
```diff
@@ -435,6 +435,86 @@ test.describe( 'Nested Tabs tests', () => {
...
+
+		await activeTab.hover();
+
+		// Check that active tab receives the hover defined background color.
+		await expect( activeTab ).toHaveCSS( 'background-color', backgroundHoverStyle );
+
+		await cleanup( wpAdmin );
+	} );
+
+	test( 'Check that while hovering an active tab hover styling is shown', async ( { page }, testInfo ) => {
```
hein-obox, line 475: Suggestion: 'Check if the hover effect works on active tabs`

modules/nested-tabs/assets/scss/frontend.scss
davseve: I think that the CSS property should be the same in both cases.
In the control, we are targeting the `background` CSS property in the PHP.
If my explanation is not clear I will elaborate on the explanation

modules/nested-tabs/assets/scss/frontend.scss
davseve: I think that the CSS property should be the same in both cases.
In the control, we are targeting the background CSS property in the PHP.
If my explanation is not clear I will elaborate on the explanation.

modules/nested-tabs/widgets/nested-tabs.php
davseve: That was my fist try :)
It didn't work, so I added the direct sign to after each and every use of the `nested_tabs_heading_selector_class` variable.

modules/nested-tabs/widgets/nested-tabs.php
davseve: The background color should be weaker (less explicit) than the `active` and `hover` states but stronger (more explicit) that the normal background color.
In this way, I achieve exactly that :)

modules/nested-tabs/widgets/nested-tabs.php
davseve: Probably missed it in the merging from 3.10.0.
Fixed

tests/playwright/sanity/modules/nested-tabs/nested-tabs.test.js
davseve: Great suggestion.
Done.

tests/playwright/sanity/modules/nested-tabs/nested-tabs.test.js
```diff
@@ -491,7 +571,9 @@ async function cleanup( wpAdmin ) {
 
 async function setTabItemColor( page, editor, panelClass, tabState, colorPickerClass, color ) {
 	await editor.activatePanelTab( 'style' );
-	await page.locator( `.elementor-control-${ panelClass }` ).click();
+	if ( 'tabs' !== panelClass ) {
```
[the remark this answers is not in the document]
davseve: `panelClass` is the class of the panel that I want to open in order to create changes in its controls,
When I choose the `style` tab, the tabs is already open, and therefore, clicking the tabs panel will close it. 
That is why I excluded the tabs panel.
See the image bellow.
image

modules/nested-tabs/widgets/nested-tabs.php
davseve: Please see comment above

modules/nested-tabs/assets/scss/frontend.scss
davseve: Changed to your suggestion, 
Looks like it works OK.

modules/nested-tabs/assets/scss/frontend.scss
hein-obox: Can't we remove this and move it to line 233 as well?

modules/nested-tabs/assets/scss/frontend.scss
hein-obox: Can't we remove this and move it? Don't we have this in line 241 already?

modules/nested-tabs/assets/scss/frontend.scss
```diff
@@ -229,6 +229,21 @@
 	}
 }
 
+// TODO: Remove once PHP global color support is fixed.
+:is( .elementor .elementor-element.elementor-widget-n-tabs > .elementor-widget-container > .e-n-tabs > .e-n-tabs-heading .e-n-tab-title,
+  .elementor .elementor-element.elementor-widget-n-tabs > .elementor-widget-container > .e-n-tabs > .e-n-tabs-content .e-n-tab-title ){
+
```
hein-obox, line 235: Can we add the normal styling here as well?

modules/nested-tabs/widgets/nested-tabs.php
hein-obox: Can we try it again? I believe it should work.

modules/nested-tabs/widgets/nested-tabs.php
hein-obox: We use a css variable here, but for the hover and active state we use a php selector.

Couldn't we use the same approach everywhere?

modules/nested-tabs/widgets/nested-tabs.php
davseve: I added this by mistake, AFAIR it doesn't appear in the original branch

modules/nested-tabs/widgets/nested-tabs.php
davseve: I am afraid we can't, if you remember, yesterday, we tried using the PHP approach in normal background color but it didn't work.
The use of the CSS variable is because we don't have a global variable in the light gray color.
This way the tab's normal background color is the BG fallback and if a color is selected that we assign the variable that appears in the PHP

modules/nested-tabs/assets/scss/frontend.scss
davseve: We don't need to. 
This solution solves the malfunction of the PHP global colors, in the normal state we are not using global colors.

tests/playwright/sanity/modules/nested-tabs/nested-tabs.test.js
davseve: Great suggestion.
Done.

modules/nested-tabs/assets/scss/frontend.scss
```diff
@@ -229,6 +229,21 @@
 	}
 }
 
+// TODO: Remove once PHP global color support is fixed.
```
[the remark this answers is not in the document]
davseve: Removed

modules/nested-tabs/widgets/nested-tabs.php
davseve: We tried together :)
We learned that the specificity sign can't the last element of an `:is` function.
Please resolve.

modules/nested-tabs/assets/scss/frontend.scss
davseve: We have decided to leave it because we are using it to fill the lack of PHP global colors.

hein-obox (dismissed):

TzviRabinovitch (dismissed):

modules/nested-tabs/widgets/nested-tabs.php
```diff
@@ -619,7 +634,7 @@ protected function register_controls() {
 			'global' => [
 				'default' => Global_Typography::TYPOGRAPHY_ACCENT,
 			],
-			'selector' => "{$nested_tabs_heading_selector_class} .e-n-tab-title-text",
+			'selector' => "{$nested_tabs_heading_selector_class} > .e-n-tab-title-text",
```
hein-obox, line 637: `'selector' => "{$nested_tabs_heading_selector_class} > .e-n-tab-title > .e-n-tab-title-text",`

hein-obox (dismissed):

TzviRabinovitch (dismissed):

hein-obox (dismissed):

TzviRabinovitch (approved):

PR #20641, opened 2022-12-07
Fix: Nested Tabs - Icons aren't visible in mobile view [ed-8980]

hein-obox:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #


tests/playwright/sanity/modules/nested-tabs/nested-tabs.test.js
```diff
@@ -475,17 +514,19 @@ const clickTab = async ( context, tabPosition ) => {
 	await context.locator( `.elementor-widget-n-tabs .e-n-tab-title >> nth=${ tabPosition } ` ).first().click();
 };
 
-async function setup( wpAdmin ) {
+async function setup( wpAdmin, svgActive = false ) {
```
davseve, line 517: I don't think this is a good implementation, this to specific and covers only one use case.

Can you try this instead?
The suggestion bellow is more generic.
This way we can add any experiment we want and we are not limited to SVG icons expiriment. 

```
async function setup( wpAdmin, customExperiment = '' ) {
    let experiments = {
        container: 'active',
        'nested-elements': 'active',
    };

    experiments = { ...experiments, ...customExperiment };
    await wpAdmin.setExperiments( experiments );
}

async function cleanup( wpAdmin, customExperiment = '' ) {
    let experiments = {
        container: 'inactive',
        'nested-elements': 'inactive',
    };

    experiments = { ...experiments, ...customExperiment };
    await wpAdmin.setExperiments( experiments );
}
```

The implementation should look like:


```
        await setup( wpAdmin, { e_font_icon_svg: 'active' } );
        await cleanup( wpAdmin, { e_font_icon_svg: 'inactive' } );
```

tests/playwright/sanity/modules/nested-tabs/nested-tabs.test.js
```diff
@@ -435,6 +419,61 @@ test.describe( 'Nested Tabs tests', () => {
...
+		// Act.
+		// Add Icons.
+		await setIconsToTabs( page, TabsIcons );
+
+		// Open front end.
+		await editor.publishAndViewPage();
+		await page.waitForSelector( '.elementor-widget-n-tabs' );
+
+		// Assert
+		await page.setViewportSize( { width: 400, height: 480 } );
```
davseve, line 443: instead of duplicating the dimensions object, I would suggest adding a dimensions object.
This way you'll have one source of truth.

Something like:

```
const viewportSize = {
    desktop: { width: 1920, height: 1080 },
    mobile: { width: 400, height: 480 },
};
```

The implementation should be as follows: 

```
 await page.setViewportSize( viewportSize.mobile );
  await page.setViewportSize( viewportSize.desktop );
```

tests/playwright/sanity/modules/nested-tabs/nested-tabs.test.js
```diff
@@ -435,6 +419,61 @@ test.describe( 'Nested Tabs tests', () => {
...
+		// Act.
+		// Add Icons.
+		await setIconsToTabs( page, TabsIcons );
+
+		// Open front end.
+		await editor.publishAndViewPage();
+		await page.waitForSelector( '.elementor-widget-n-tabs' );
+
+		// Assert
+		await page.pause();
```
davseve, line 470: Why do we need the pause for?
Is there a script running in the background that we want to stop?

davseve (changes requested):

tests/playwright/sanity/modules/nested-tabs/nested-tabs.test.js
hein-obox: Implemented

tests/playwright/sanity/modules/nested-tabs/nested-tabs.test.js
hein-obox: Implemented

davseve (dismissed):

TzviRabinovitch (dismissed):

TzviRabinovitch (approved):

PR #20727, opened 2022-12-13
Tweak: Add caption space option to image gallery and image carousel widgets [ED-10064]

rodolphebertozzo:


rodolphebertozzo:
@arielk

includes/widgets/image-gallery.php
```diff
@@ -342,6 +342,32 @@ protected function register_controls() {
 				'selector' => '{{WRAPPER}} .gallery-item .gallery-caption',
 			]
 		);
+		
+
```
rodolphebertozzo, line 346:
```suggestion
```

rodolphebertozzo:
@arielk

rodolphebertozzo:
@rotemee

includes/widgets/image-gallery.php
```diff
@@ -343,6 +343,31 @@ protected function register_controls() {
...
+						'min' => 0,
+						'max' => 100,
+					],
+					'%' => [
+						'min' => 0,
+						'max' => 100,
+					],
+				],
+				'selectors' => [
+					'{{WRAPPER}} gallery-item .gallery-caption' => 'margin-top: {{SIZE}}{{UNIT}};',
```
kobizz, line 366:
```suggestion
					'{{WRAPPER}} .gallery-item .gallery-caption' => 'margin-top: {{SIZE}}{{UNIT}};',
```

kobizz (dismissed):

includes/widgets/image-gallery.php
rodolphebertozzo: Done.

rodolphebertozzo:
@TzviRabinovitch

rodolphebertozzo:
@KingYes

rodolphebertozzo:
@arielk

rodolphebertozzo:
@rami-elementor

rodolphebertozzo:
@arielk

rodolphebertozzo:
@arielk @rami-elementor Status ?

rodolphebertozzo:
@rami-elementor

rami-elementor:
I personally like this PR, it adds consistency to Elementor widgets using image captions.

It's in my TODO list to test this patch.

doronwolf:
Hi @rodolphebertozzo 
A. There is a bug in the display of the light box's icons on the top right side.. which may or may not be related to the changes in this build, but on the other hand I didn't run into it in other builds either.. so maybe it is related? (see the first screenshot)
Screenshot1

B. Even when choosing the 'hide' option for the caption, the spacing slider is still available 
and also the text shadow also continues to appear (see the second screenshot)
Screenshot2

C. Before this addition the default distance of the caption from the images was 0.75rem but after the change, the default distance is different (smaller, and until you don't move the slider, it is probably only affected by the line height of the text) and although it is a seemingly small difference, it is a bit of a problem for backward compatibility

  ...Beyond those issues, I actually think it's a great feature and I'm sure that those who do use captions will be happy with this addition so please review those issues and once they are fixed we would be happy to approve it :)

rodolphebertozzo:
Hi, I've corrected the underlined aspects by adding conditions and default values (0.75 rem = 12px) so that there won't be any backward compatibility problems with sites that already display captions. Adding this control without a default value surely overrides what is agreed in the SASS file at this level.
As for the lightbox, it's very strange. Any ideas?

rodolphebertozzo:
@doronwolf Hi, I've corrected the underlined aspects by adding conditions and default values (0.75 rem = 12px) so that there won't be any backward compatibility problems with sites that already display captions. Adding this control without a default value surely overrides what is agreed in the SASS file at this level.
As for the lightbox, it's very strange. Any ideas?

doronwolf:
Hi @rodolphebertozzo thank you very much for making the necessary changes 🙏 looks very good!
about the lightbox - I agree, it seems not related to your code
I need to do a few more tests on my end but I believe they will go smoothly and then we all approve and merge this PR
will try to update here once I have news

rodolphebertozzo:
> Hi @rodolphebertozzo thank you very much for making the necessary changes 🙏 looks very good! about the lightbox - I agree, it seems not related to your code I need to do a few more tests on my end but I believe they will go smoothly and then we all approve and merge this PR will try to update here once I have news

Hello,
I've just checked the SASS files, it seems that there are no default style rules for the caption, so it's the PHP file that manages this and prints some style rules in the DOM for this default control. Yes, I've just restored all the files as they were, and the problem is not with my PR

doronwolf:
Hi @rodolphebertozzo I also checked again... there is a default style rules for the caption - but it comes from the theme (see the attached screenshot) in my case it's the 'hello' theme
Screenshot 2023-07-12 at 9 46 48 but since we can't know which theme each user chooses to use, and other themes may have different default (or no default) then I agree with you👍 and thanks for the catch

I hope this is the only change you reverted? because initially when choosing the 'hide' option for the caption, the spacing slider, and the text-shadow were still available, and that was a fix you shouldn't revert

rodolphebertozzo:
> Salut@rodolphebertozzoJ'ai également vérifié à nouveau... il y a des règles de style par défaut pour la légende - mais cela vient du thème (voir la capture d'écran ci-jointe) dans mon cas c'est le thème 'hello' mais puisque nous ne pouvons pas savoir quel thème chaque utilisateur Capture d'écran 2023-07-12 à 9 46 48choisit à utiliser, et d'autres thèmes peuvent avoir des valeurs par défaut différentes (ou aucune valeur par défaut), alors je suis d'accord avec vous👍et merci pour la prise
> 
> J'espère que c'est le seul changement que vous avez annulé? parce qu'au départ, lors du choix de l'option 'masquer' pour la légende, le curseur d'espacement et l'ombre du texte étaient toujours disponibles, et c'était une solution que vous ne devriez pas revenir

For text shadow and spacing, I've added the condition that the control is only displayed if the caption is active.

includes/widgets/image-carousel.php
```diff
@@ -826,6 +826,32 @@ protected function register_controls() {
...
+						'min' => 0,
+						'max' => 100,
+					],
+					'%' => [
+						'min' => 0,
+						'max' => 100,
+					],
+				],
+				'selectors' => [
+					'{{WRAPPER}} .elementor-image-carousel-caption' => 'margin-top: {{SIZE}}{{UNIT}};',
```
rami-elementor, line 850:
```suggestion
					'{{WRAPPER}} .elementor-image-carousel-caption' => 'margin-block-start: {{SIZE}}{{UNIT}};',
```

includes/widgets/image-gallery.php
```diff
@@ -346,6 +346,38 @@ protected function register_controls() {
...
+						'min' => 0,
+						'max' => 100,
+					],
+					'%' => [
+						'min' => 0,
+						'max' => 100,
+					],
+				],
+				'selectors' => [
+					'{{WRAPPER}} .gallery-item .gallery-caption' => 'margin-top: {{SIZE}}{{UNIT}};',
```
rami-elementor, line 376:
```suggestion
					'{{WRAPPER}} .gallery-item .gallery-caption' => 'margin-block-start: {{SIZE}}{{UNIT}};',
```

includes/widgets/image-gallery.php
```diff
@@ -346,6 +346,38 @@ protected function register_controls() {
...
+				],
+				'range' => [
+					'px' => [
+						'min' => 0,
+						'max' => 100,
+					],
+					'%' => [
+						'min' => 0,
+						'max' => 100,
+					],
```
rami-elementor, line 373:
```suggestion
```

You don't need to set this as this is already set by default for `%` units.

includes/widgets/image-carousel.php
```diff
@@ -826,6 +826,32 @@ protected function register_controls() {
...
+				],
+				'range' => [
+					'px' => [
+						'min' => 0,
+						'max' => 100,
+					],
+					'%' => [
+						'min' => 0,
+						'max' => 100,
+					],
```
rami-elementor, line 847:
```suggestion
```

You don't need to set this as this is already set by default for `%` units.

includes/widgets/image-carousel.php
```diff
@@ -826,6 +826,32 @@ protected function register_controls() {
...
 		);
 
+		$this->add_responsive_control(
+			'caption_space',
+			[
+				'label' => esc_html__( 'Spacing', 'elementor' ),
+				'type' => Controls_Manager::SLIDER,
+				'size_units' => [ 'px', '%', 'em', 'rem', 'custom' ],
+				'default' => [
+					'unit' => 'px',
```
rami-elementor, line 836:
```suggestion
```

Remove, the default unit is `px`, you don't need to set this again.

includes/widgets/image-gallery.php
```diff
@@ -346,6 +346,38 @@ protected function register_controls() {
...
+		);
+
+		$this->add_responsive_control(
+			'caption_space',
+			[
+				'label' => esc_html__( 'Spacing', 'elementor' ),
+				'type' => Controls_Manager::SLIDER,
+				'size_units' => [ 'px', '%', 'em', 'rem', 'custom' ],
+				'default' => [
+					'unit' => 'px',
```
rami-elementor, line 362:
```suggestion
```

Remove, the default unit is `px`, you don't need to set this again.

rami-elementor (changes requested):

includes/widgets/image-carousel.php
```diff
@@ -833,6 +833,27 @@ protected function register_controls() {
...
 
+		$this->add_responsive_control(
+			'caption_space',
+			[
+				'label' => esc_html__( 'Spacing', 'elementor' ),
+				'type' => Controls_Manager::SLIDER,
+				'size_units' => [ 'px', '%', 'em', 'rem', 'custom' ],
+				'default' => [
+					'size' => 12,
+				],
```
rami-elementor, line 844:
```suggestion
```

Remove the default value. This way we keep backwards compatibility.

includes/widgets/image-gallery.php
```diff
@@ -359,6 +359,33 @@ protected function register_controls() {
...
+
+		$this->add_responsive_control(
+			'caption_space',
+			[
+				'label' => esc_html__( 'Spacing', 'elementor' ),
+				'type' => Controls_Manager::SLIDER,
+				'size_units' => [ 'px', '%', 'em', 'rem', 'custom' ],
+				'default' => [
+					'size' => 12,
+				],
```
rami-elementor, line 376:
```suggestion
```

Remove the default value. This way we keep backwards compatibility.

includes/widgets/image-carousel.php
```diff
@@ -833,6 +833,27 @@ protected function register_controls() {
...
+				'range' => [
+					'px' => [
+						'min' => 0,
+						'max' => 100,
+					],
+				],
```
rami-elementor, line 850:
```suggestion
```

Remove this section, you are redefining default range values.

includes/widgets/image-gallery.php
```diff
@@ -359,6 +359,33 @@ protected function register_controls() {
...
+				'range' => [
+					'px' => [
+						'min' => 0,
+						'max' => 100,
+					],
+				],
```
rami-elementor, line 382:
```suggestion
```

Remove this section, you are redefining default range values.

rami-elementor (changes requested):

rami-elementor (dismissed):

rodolphebertozzo:
Hello @rami-elementor,
Yes, I've just been made aware of your changes. At the time I started working on this improvement, I wasn't fully aware of how Elementor works with units, and even less of the fact that they are set in a separate file, and that specifying maximum, minimum or default unit values is only done if you want to get away from the operating logic in a specific case.

rami-elementor:
@rodolphebertozzo 

Thank you for your valuable contribution! I approved this PR. Moving to @arielk to merge.

arielk (dismissed):

rami-elementor (approved):

arielk (approved):

arielk (approved):

PR #20742, opened 2022-12-14
Fix: Lazy Load Background Images not loading for Template widget in the editor [ED-8994]

maork-elementor:


Nevoss (dismissed):

modules/lazyload/module.php
```diff
@@ -90,6 +90,11 @@ private function apply_dominant_color_background( $control, $value, $selector )
 	public function __construct() {
 		parent::__construct();
 
+		// Disable lazyload in admin area, works for all roles ( True if inside WordPress administration interface - Editor, Admin, etc.)
+		if ( is_admin() ) {
```
StyleShit, line 94: check if the lazy load works in the preview, it should be active, like the frontend.

also, try to use something like this:
https://github.com/elementor/elementor/blob/fa41b8219d4241c7e3885b473e287d0e52bd5fe7/includes/base/widget-base.php#L1190-L1194

StyleShit (changes requested):

modules/lazyload/module.php
maork-elementor: `$is_active` it True both in fronted & editor mode.

StyleShit (dismissed):

StyleShit (approved):

PR #20791, opened 2022-12-19
Tweak: Nested Elements - Prevent copying the tab container [ED-9128]

Nevoss:


matipojo (changes requested):
Please cover with tests

Nevoss:
It seems like this is a specific and small case for e2e tests, and I just tried to do a unit test and it didn't go nice.

Issue #20669, opened 2022-12-08
✔️ 🐞 Bug report: Gallery widget broken after updating Elementor from 3.8.2 to 3.9.0 (ED-9020)


puregraphx:
### Prerequisites

- [X] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [X] The issue still exists against the latest stable version of Elementor.

### Description

After updating to 3.9.0 the Gallery Widget is still visible in the DOM, but every div.e-gallery-image is 33554400x0 pixels.
--item-width and --item-height on a.e-gallery-item are set to NaN%;

### Steps to reproduce

Have a working Gallery Widget in 3.8.2.
Update Elementor Pro to 3.9.0.
Visit the page again

### Isolating the problem

- [X] This bug happens with only Elementor plugin active (and Elementor Pro).
- [X] This bug happens with a Blank WordPress theme active ([Hello theme](https://wordpress.org/themes/hello-elementor/)).
- [X] I can reproduce this bug consistently using the steps above.

### System Info

				
== Server Environment ==
	Operating System: Linux
	Software: Apache/2.4.54 (Debian)
	MySQL version: mariadb.org binary distribution v10.4.20
	PHP Version: 7.4.30
	PHP Memory Limit: 256M
	PHP Max Input Vars: 2500
	PHP Max Post Size: 10M
	GD Installed: Yes
	ZIP Installed: Yes
	Write Permissions: All right
	Elementor Library: Connected

== WordPress Environment ==
	Version: 6.1.1
	Site URL: https://new.decoratiehuren.be
	Home URL: https://new.decoratiehuren.be
	WP Multisite: No
	Max Upload Size: 10 MB
	Memory limit: 2048M
	Max Memory limit: 2048M
	Permalink Structure: /%postname%/
	Language: en-US
	Timezone: Europe/Brussels
	Admin Email: info@puregraphx.be
	Debug Mode: Inactive

== Theme ==
	Name: Astra Child
	Version: 1.0.0
	Author: Brainstorm Force
	Child Theme: Yes
	Parent Theme Name: Astra
	Parent Theme Version: 3.9.4
	Parent Theme Author: Brainstorm Force

== User ==
	Role: administrator
	WP Profile lang: nl_NL
	User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/108.0.0.0 Safari/537.36

== Active Plugins ==
[...]

PR #20814, opened 2022-12-21
Fix: Pro Gallery widget won't load unless set to 100% (#20669) [ED-9020]

louiswol94:


assets/dev/scss/frontend/_container.scss
```diff
@@ -6,6 +6,11 @@
 	--flex-basis: auto;
 	--flex-grow: 0;
 	--flex-shrink: 1;
+	--container-widget-width: 100%;
+	--container-widget-height: initial;
+	--container-widget-flex-grow: 0;
+	--container-widget-align-self: initial;
+
```
louiswol94, line 13: Without specifying these values in the SCSS, child containers inherited these styles from parent containers.

assets/dev/scss/frontend/widgets/spacer.scss
```diff
@@ -5,14 +5,11 @@
 	}
 }
 
-.e-con {
-	--container-widget-width: 100%;
-}
-
```
louiswol94, line 11: This affected other widgets, so I moved it down.

assets/dev/scss/frontend/widgets/spacer.scss
```diff
@@ -5,14 +5,11 @@
...
 .e-con,
 .e-con-inner {
 
 	& > .elementor-widget-spacer {
+		--container-widget-width: 100%;
 		width: var( --container-widget-width, var( --spacer-size ) );
 		--align-self: var( --container-widget-align-self, initial );
 		--flex-shrink: 0;
```
hein-obox, line 15: One small comment.
For the readibility of the code it might nice to first display all css variables and only afterwards the normal css selector.

hein-obox:
I have tried to rerun the playwright test.
I would be great if we can use the existing playwright tests, but fix them where necessary.

assets/dev/scss/frontend/widgets/spacer.scss
louiswol94: @hein-obox Done.

assets/dev/scss/frontend/widgets/spacer.scss
udidol: Good call

udidol (changes requested):
This PR is checked out of `release/3.9.0`,
please check the code out of `develop` and PR into `develop`

udidol (approved):

PR #20879, opened 2022-12-29
Internal: Editor V2 - CI

Nevoss:


.github/workflows/packages.yml
```diff
@@ -0,0 +1,82 @@
...
+        uses: actions/setup-node@v2
+        with:
+          node-version: 18.x
+      - name: Cache node modules
+        uses: actions/cache@v3
+        env:
+          cache-name: cache-packages-node-modules
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('packages/package-lock.json') }}
```
StyleShit, line 46:
```suggestion
          key: ${{ runner.os }}-packages-${{ env.cache-name }}-${{ hashFiles('packages/package-lock.json') }}
```

Maybe? Also others

.github/workflows/packages.yml
```diff
@@ -0,0 +1,82 @@
...
+          restore-keys: |
+            ${{ runner.os }}-build-${{ env.cache-name }}-
+            ${{ runner.os }}-build-
+            ${{ runner.os }}-
+      - name: Install Dependencies
+        run: npm run packages:install
```
StyleShit, line 52: not sure how it works in Github CI, but maybe we can `npm i` once and then run the tasks? because seems like you'll get a cache-miss twice for the current YAML file

StyleShit (changes requested):

StyleShit (approved):

PR #20907, opened 2023-01-03
Fix: Added WC filter to prevent duplicate pages in WC kits [ED-9140]

ZeyadMahagna:


app/modules/import-export/module.php
```diff
@@ -383,6 +383,18 @@ private function register_actions() {
 		$page_id = Tools::PAGE_ID;
 
 		add_action( "elementor/admin/after_create_settings/{$page_id}", [ $this, 'register_settings_tab' ] );
+
+		add_filter( 'woocommerce_create_pages', [ $this, 'empty_pages' ], 10, 0 );
+	}
+
+	/**
+	 * empty_pages
```
ManorHazazE, line 391: No need to mention the function.

ghost (changes requested):

app/modules/import-export/module.php
ZeyadMahagna: Do you mean I should remove the whole function doc or just the name?

ghost (approved):

nivnoimanelementor (approved):

PR #20959, opened 2023-01-10
Internal: Remove 3.5.0 deprecations in Finder Categories_Manager & Controls_Manager [ED-9325]

StyleShit:


matipojo (changes requested):
Please disable the hard deprecation

PR #20981, opened 2023-01-11
Internal: Experiments - Adding non-existent experiment dependency crashes the plugin [ED-9261]

Nevoss:


core/experiments/manager.php
```diff
@@ -115,7 +115,12 @@ public function add_feature( array $options ) {
 					throw new Exceptions\Dependency_Exception( 'Depending on a hidden experiment is not allowed.' );
 				}
 
-				if ( ! class_exists( $dependency ) && ! empty( $feature ) ) {
+				$feature_not_exists = ! class_exists( $dependency ) && empty( $feature );
+				$feature_without_class = ! class_exists( $dependency ) && ! empty( $feature );
+
+				if ( $feature_not_exists ) {
+					$experimental_data['dependencies'][ $key ] = new Non_Existing_Dependency( $dependency );
+				} elseif ( $feature_without_class ) {
```
StyleShit, line 123:
```PHP
if ( $feature ) {
    return Wrap();
} elseif ( class_exists( $dependency ) ) {
    return $dependency::instance();
} else {
    return Non_Existing();
}
```

Maybe?

core/experiments/manager.php
```diff
@@ -594,6 +599,15 @@ private function render_feature_dependency( $feature ) {
 		<?php
 	}
 
+	private function has_non_existing_dependency( $feature ) {
+		$non_existing_dep = ( new Collection( $feature['dependencies'] ?? [] ) )
+			->find( function ( $dependency ) {
+				return $dependency instanceof Non_Existing_Dependency;
+			} );
```
StyleShit, line 606: do you still need the `Non_Existing_Dependency` instance if you're hiding the experiment entirely? Can't you just return null?

StyleShit (changes requested):

core/experiments/manager.php
Nevoss: It feels weird to me checking if there is a null in the array and this is an indicator that experiment is not exists.

core/experiments/manager.php
Nevoss: Done

core/experiments/manager.php
```diff
@@ -785,6 +790,18 @@ private function should_show_hidden() {
...
 	}
 
+	private function create_dependency_class( $dependency_name, $dependency_args ) {
+		if ( class_exists( $dependency_name ) ) {
+			return $dependency_name::instance();
+		}
+
+		if ( ! empty( $dependency_args ) ) {
+			return new Wrap_Core_Dependency( $dependency_args );
+		}
```
StyleShit, line 800: What the priority should be? maybe we first want the Wrap and then the instance?
IDK

StyleShit (changes requested):

StyleShit (approved):

tests/phpunit/elementor/core/experiments/test-manager.php
```diff
@@ -84,6 +85,27 @@ public function test_add_feature__ensure_wrap_core_dependency() {
...
+		$this->add_test_feature( $test_feature_data );
+
+		// Assert.
+		$feature = $this->experiments->get_features( 'test_feature' );
+
+		$this->assertCount( 1, $feature['dependencies'] );
+		$this->assertInstanceOf( Non_Existing_Dependency::class, $feature['dependencies'][0] );
+		$this->assertEquals( 'feature-that-not-exists', $feature['dependencies'][0]->get_name() );
+	}
+
```
matipojo, line 108: Can you assert that it will not appear in the experiments list?

matipojo (changes requested):

matipojo (approved):

PR #21013, opened 2023-01-14
Tweak: Add `max-height` control to containers and sections [ED-9575]

m1ga:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

* adding `max-height` value to the container element

## Description

* adding `max-height` value to the container element

## Test instructions

* create a container
* add a value the the new `max-height` property (below min-height)
* switch overflow mode to `auto` or `hidden`

## Quality assurance

Fixes #

https://user-images.githubusercontent.com/4334997/212478029-e39967bb-0a89-4a47-915c-5f8bb66a4d81.mp4


rodolphebertozzo (dismissed):

includes/elements/container.php
```diff
@@ -443,6 +443,28 @@ protected function register_container_layout_controls() {
...
+				'size_units' => [ 'px', 'vh', 'custom' ],
+				'range' => [
+					'px' => [
+						'min' => 0,
+						'max' => 1440,
+					],
+					'vh' => [
+						'min' => 0,
+						'max' => 100,
+					],
```
rami-elementor, line 460: Can be removed as this is has already defined in `includes\controls\base-units.php`

includes/elements/container.php
```diff
@@ -443,6 +443,28 @@ protected function register_container_layout_controls() {
...
+		$this->add_responsive_control(
+			'max_height',
+			[
+				'label' => esc_html__( 'Max Height', 'elementor' ),
+				'type' => Controls_Manager::SLIDER,
+				'size_units' => [ 'px', 'vh', 'custom' ],
+				'range' => [
+					'px' => [
+						'min' => 0,
+						'max' => 1440,
```
rami-elementor, line 455: Why limiting the `max-height` to `1440`? What if I need a larger container?

includes/elements/container.php
m1ga: I just copied it from existing `'Min Height'`. The values are there too :smile: 
I'll remove them :+1:

rami-elementor:
For consistency, you should add the `max-height` control to sections too.

includes/elements/container.php
m1ga: Shall I use a high value or is there an unlimited option?

m1ga: I've raised the value to 10000. Otherwise the slide is very hard to control. User can always enter a higher value if needed.

includes/elements/container.php
```diff
@@ -443,6 +439,24 @@ protected function register_container_layout_controls() {
 			]
 		);
 
+		$this->add_responsive_control(
+			'max_height',
+			[
+				'label' => esc_html__( 'Max Height', 'elementor' ),
+				'type' => Controls_Manager::SLIDER,
+				'size_units' => [ 'px', 'vh', 'custom' ],
```
rami-elementor, line 447:
```suggestion
				'size_units' => [ 'px', 'vh', 'vw', 'custom' ],
```

rami-elementor: I've noticed that sections use three unit types (`px`, `vh`, `vw`) while containers use two (`px`, `vh`). Let's fix that too.

rodolphebertozzo: Is it really relevant to add the vw for the height, it seems to me that this unit is only suitable for the width?

m1ga: I don't mind having it there. If it's not there there will be one user that will ask for it :smile:

rodolphebertozzo: Obviously, this is not disturbing in itself, other builders have all the units everywhere to offer something complete, because the code structure is different. Elementor wants to offer more units, just as the base code requires units to be added to containers and elements on a case by case basis for each of them, so we might as well think about this detail.

m1ga:
any news if this can be included?

m1ga:
Still no feedback here?

m1ga:
@yotamselementor I think it would be easier to redo the PR as there are so many merge conflicts now.
The only question is: is there any interest in this PR at all since I've made that over a year ago and nothing happened.

PR #21029, opened 2023-01-15
Internal: Editor V2 - Adapter for invoking commands [ED-9292]

StyleShit:


packages/packages/v1-adapters/src/dispatchers/dispatchers.ts
```diff
@@ -0,0 +1,38 @@
+import { ExtendedWindow } from './types';
+import { isJQueryDeferred, promisifyJQueryDeferred } from './utils';
+
```
Nevoss, line 3: maybe?

```ts
declare global {
	interface Window {
		$e?: {
			run?: ( command: string, args?: object ) => unknown;
			route?: ( route: string ) => void;
		},
	}
}
```

packages/packages/v1-adapters/src/dispatchers/dispatchers.ts
```diff
@@ -0,0 +1,38 @@
...
+export function runCommand( command: string, args?: object ) : Promise<unknown> {
+	const extendedWindow = window as unknown as ExtendedWindow;
+
+	if ( ! extendedWindow.$e?.run ) {
```
Nevoss, line 7:
```suggestion
	if ( ! window.$e?.run ) {
```

Nevoss (changes requested):

packages/packages/v1-adapters/src/dispatchers/dispatchers.ts
StyleShit: it will make this type globally available to every package that imports this adapter, and we don't want that.

packages/packages/v1-adapters/src/dispatchers/dispatchers.ts
StyleShit: NO

Nevoss (dismissed):

packages/packages/v1-adapters/src/dispatchers/dispatchers.ts
```diff
@@ -0,0 +1,38 @@
...
+
+	if ( isJQueryDeferred( result ) ) {
+		return promisifyJQueryDeferred( result );
+	}
+
+	return Promise.resolve( result );
+}
+
+export function openRoute( route: string ) : Promise<void> {
+	const extendedWindow = window as unknown as ExtendedWindow;
```
rotemee, line 25: why twice? why not at the top of the module?

packages/packages/v1-adapters/src/dispatchers/dispatchers.ts
```diff
@@ -0,0 +1,38 @@
...
+	}
+
+	if ( isJQueryDeferred( result ) ) {
+		return promisifyJQueryDeferred( result );
+	}
+
+	return Promise.resolve( result );
+}
+
+export function openRoute( route: string ) : Promise<void> {
```
rotemee, line 24: if the function always returns a Promise, why not:
```
export function openRoute( route: string ) {
	return new Promise( ... );
}
```

packages/packages/v1-adapters/src/dispatchers/__tests__/index.test.ts
```diff
@@ -0,0 +1,115 @@
+import { openRoute, runCommand } from '../';
+
+describe( '@elementor/v1-adapters/dispatchers', () => {
+	beforeEach( () => {
+		( window as any ).$e = {
```
rotemee, line 5: Consider defining one reusable variable.

packages/packages/v1-adapters/src/dispatchers/__tests__/index.test.ts
```diff
@@ -0,0 +1,115 @@
...
+
+	it( 'should run a V1 command that returns Promise', () => {
+		// Arrange.
+		const command = 'editor/documents/open',
+			args = { test: 'arg' };
+
+		( window as any ).$e.run.mockReturnValue( Promise.resolve( 'result' ) );
+
+		// Act.
+		const actual = runCommand( command, args );
```
rotemee, line 23:
```suggestion
		const result = runCommand( command, args );
```

packages/packages/v1-adapters/src/dispatchers/__tests__/index.test.ts
```diff
@@ -0,0 +1,115 @@
...
+
+		// Assert.
+		expect( ( window as any ).$e.run ).toHaveBeenCalledWith( command, args );
+		expect( actual ).toEqual( Promise.resolve( 'result' ) );
+	} );
+
+	it( 'should run a V1 command that returns a regular value', () => {
```
rotemee, line 45:
```suggestion
	it( 'should run a V1 command that returns a plain value', () => {
```

rotemee (changes requested):

packages/packages/v1-adapters/src/dispatchers/dispatchers.ts
StyleShit: IDK, maybe `window` isn't fully initialized yet when the module loads, some references might change, didn't want to take that risk 🤷

packages/packages/v1-adapters/src/dispatchers/dispatchers.ts
StyleShit: Tried to write it with this approach. but having multiple conditions feels cleaner and more readable IMO

packages/packages/v1-adapters/src/dispatchers/__tests__/index.test.ts
StyleShit: for `window as any`?
same as above, maybe it'll have references issues

packages/packages/v1-adapters/src/dispatchers/__tests__/index.test.ts
StyleShit: Yup, that's better... 
copilot did this 😅

packages/packages/v1-adapters/src/dispatchers/__tests__/index.test.ts
StyleShit: Done

rotemee (dismissed):

rotemee (approved):

PR #21047, opened 2023-01-17
Internal: Generate RC tag GitHub action [ED-9375]

maork-elementor:


matipojo (changes requested):

.github/workflows/build-plugin/action.yml
```diff
@@ -22,7 +22,13 @@ runs:
                 echo "PACKAGE_VERSION is empty string or undefined"
                 exit 1
             fi
-            bash ${{ inputs.UPDATE_VERSIONS_SCRIPT_PATH }}
+
+            if [ -f "${{ inputs.UPDATE_VERSIONS_SCRIPT_PATH }}" ];
+              then
+               bash ${{ inputs.UPDATE_VERSIONS_SCRIPT_PATH }}
+            fi
```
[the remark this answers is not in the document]
maork-elementor: When Cloud publish begins, it will run from the tag.
By updating the tag's version first, then running the cloud publish action, the version will increase twice.

If we take the current version x.5 as an example
This will increase the tag version by one, resulting in x.6.
As soon as the publish cloud runs, it will increase it again, but this time from x.6 to x.7.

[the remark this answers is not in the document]
matipojo: Try to make it explicit (e.g., `create_new_version`=true|false), not as guessing the logic based on if a file is existing or not.

matipojo (changes requested):

.github/workflows/create-tag.yml
```diff
@@ -0,0 +1,78 @@
+name: Create Tag
+# Required secrets for this workflow: MAINTAIN_USERNAME, MAINTAIN_EMAIL, SLACK_TOKEN, SLACK_TAG_CHANNELS, MAINTAIN_TOKEN
+
+on:
+  workflow_dispatch:
+    inputs:
+      type:
+        type: choice
+        description: 'Release type'
```
matipojo, line 9:
```suggestion
        description: 'Tag Type'
```

.github/workflows/create-tag.yml
```diff
@@ -0,0 +1,78 @@
+name: Create Tag
+# Required secrets for this workflow: MAINTAIN_USERNAME, MAINTAIN_EMAIL, SLACK_TOKEN, SLACK_TAG_CHANNELS, MAINTAIN_TOKEN
```
matipojo, line 2:
```suggestion
# Required secrets for this workflow: MAINTAIN_USERNAME, MAINTAIN_EMAIL, SLACK_TOKEN, MAINTAIN_TOKEN, and a var: SLACK_TAG_CHANNELS.
```

matipojo (approved):

package.json
```diff
@@ -5,6 +5,7 @@
     "last_beta_version": "3.10.0-beta3",
     "last_cloud_version": "3.11.0-cloud1",
     "last_dev_version": "3.9.0-dev2",
+    "last_rc_version": "3.11.0-rc5",
```
matipojo, line 8:
```suggestion
    "last_rc_version": "3.11.0-rc1",
```

PR #21061, opened 2023-01-18
Internal: Add I/E infra for supporting future features [ED-9431]

ghost:


app/modules/import-export/processes/import.php
```diff
@@ -698,7 +700,9 @@ private function add_import_session_option( $start_time ) {
 
 		$import_sessions[ $this->session_id ] = [
 			'session_id' => $this->session_id,
-			'kit_name' => $this->manifest['name'],
+			'kit_title' => $this->manifest['title'] ?? '',
+			'kit_name' => $this->manifest['name'] ?? '',
+			'kit_thumbnail' => $this->manifest['thumbnail'] ?? '',
 			'kit_source' => $this->settings_referrer,
 			'user_id' => get_current_user_id(),
```
ZeyadMahagna, line 707: Can you combine this array with the one at line 386? maybe a common function?

app/modules/import-export/processes/revert.php
```diff
@@ -25,8 +25,8 @@ class Revert {
 	 * @throws \Exception
 	 */
 	public function __construct() {
-		$this->import_sessions = $this->get_import_sessions();
-		$this->revert_sessions = $this->get_revert_sessions();
+		$this->import_sessions = self::get_import_sessions();
+		$this->revert_sessions = self::get_revert_sessions();
```
ZeyadMahagna, line 29: If there is no specific reason to use 'self' here I prefer to use 'static' to allow easier extension to this class in the future

app/modules/import-export/utils.php
```diff
@@ -80,4 +80,20 @@ public static function get_registered_cpt_names() {
 
 		return array_keys( $post_types );
 	}
+
+	/**
+	 * Transform a string name to title format.
+	 *
+	 * @param $name
+	 * @return string
```
ZeyadMahagna, line 88:
```
	 * @param $name
	 *
	 * @return string
```

app/modules/import-export/utils.php
```diff
@@ -80,4 +80,20 @@ public static function get_registered_cpt_names() {
...
+	 */
+	public static function transform_name_to_title( $name ): string {
+		if ( empty( $name ) ) {
+			return '';
+		}
+
+		$title = str_replace( [ '-', '_' ], ' ', $name );
+
+		return ucwords( $title );
```
ZeyadMahagna, line 97: I would personally get rid of the title variable here as it doesn't really provide much extra context and used only once but up to you

ZeyadMahagna (changes requested):
Nice work! A few points:

app/modules/import-export/utils.php
ManorHazazE: In my opinion, it makes it more readable.

app/modules/import-export/utils.php
ManorHazazE: Done

app/modules/import-export/processes/revert.php
ManorHazazE: Actually the opposite, if there is no reason to use static you should use self, just like if there is no reason to use "protected" function you should use "private" unless you think someone should use it.
No one should change those functions they only get the data from options, this will probably last forever.

app/modules/import-export/processes/import.php
ManorHazazE: I changed the logic a little bit and merged them.

ZeyadMahagna (approved):
Good job!

PR #21064, opened 2023-01-18
Tweak: Remove Kit alert - flow [ED-9421]

ZeyadMahagna:


app/modules/kit-library/assets/js/components/kit-already-applied-dialog.js
```diff
@@ -0,0 +1,26 @@
...
+		<Dialog
+			title={ __( 'You\'ve already applied a Kit.', 'elementor' ) }
+			text={ <>
+				{ __( 'Applying two Kits on the same website will mix global styles and colors and hurt your site\'s performance.', 'elementor' ) }
+				<br /><br />
+				{ __( 'Remove the existing Kit before applying a new one.', 'elementor' ) }
+			</> }
+			approveButtonText={ __( 'Remove existing Kit', 'elementor' ) }
+			approveButtonColor="primary"
+			approveButtonOnClick={ () => location.href = elementorAppConfig.admin_url + 'admin.php?page=elementor-tools&referrer_kit=' + props.kitId + '#tab-import-export-kit' }
```
ManorHazazE, line 14: I don't think hardcoded link is the right way to do it.
Maybe think of a different solution? Maybe externalize the tab link from the import/export tools tab.

app/modules/kit-library/assets/js/components/kit-dialog.js
```diff
@@ -0,0 +1,31 @@
+import { useState } from 'react';
+
+import KitAlreadyAppliedDialog from './kit-already-applied-dialog';
+import ApplyKitDialog from './apply-kit-dialog';
+
+export default function KitDialog( props ) {
+	const [ applyAnyway, setApplyAnyway ] = useState( false );
+
+	if ( elementorAppConfig[ 'import-export' ].importSessions.length && ! applyAnyway ) {
```
ManorHazazE, line 9: Maybe for better readability.

```suggestion
const kitAlreadyApplied = elementorAppConfig[ 'import-export' ].importSessions.length;
	if (  kitAlreadyApplied && ! applyAnyway ) {
```

assets/dev/scss/global/_dialog.scss
```diff
@@ -140,6 +140,13 @@
 	}
 }
 
+#e-kit-deleted-dialog {
```
ManorHazazE, line 143: Maybe? So you know it's related to the revert?
```suggestion
#e-revert-kit-deleted-dialog {
```

ghost (changes requested):

app/modules/import-export/assets/js/admin.js
```diff
@@ -1,19 +1,189 @@
...
+	/**
+	 * MaybeShowKitReferrerDialog
+	 */
+	maybeShowKitReferrerDialog() {
+		const isKitReferred = sessionStorage.getItem( this.isKitReferredKey );
+		if ( ! isKitReferred ) {
+			return;
+		}
+
+		if ( 'false' === isKitReferred ) {
```
ranshoshani34, line 91:
```suggestion
		if ( 'no' === isKitReferred ) {
```

app/modules/import-export/assets/js/admin.js
```diff
@@ -1,19 +1,189 @@
...
+	onRevertConfirm( revertBtn, kitNameToRemove ) {
+		const referrerKit = new URLSearchParams( revertBtn.href ).get( 'referrer_kit' );
+
+		this.setSessionStorage( Boolean( referrerKit ), referrerKit, kitNameToRemove );
+
+		location.href = revertBtn.href;
+	}
+
+	setSessionStorage( referred, referrerKit, kitNameToRemove ) {
+		sessionStorage.setItem( this.isKitReferredKey, referred.toString() );
```
ranshoshani34, line 75:
```suggestion
		sessionStorage.setItem( this.isKitReferredKey, referred ? 'yes' : 'no' );
```

app/modules/import-export/assets/js/admin.js
```diff
@@ -1,19 +1,189 @@
...
+		elementorCommon.dialogsManager.createWidget( 'confirm', {
+			id: 'e-revert-kit-deleted-dialog',
+			headerMessage: sessionStorage.getItem( this.kitNameToRemoveKey ) + __( ' was successfully deleted', 'elementor' ),
+			message: options.message,
+			strings: {
+				confirm: options.strings.confirm,
+				cancel: options.strings.cancel,
+			},
+			...( options.onConfirm ) && { onConfirm: options.onConfirm },
+			...( options.onCancel ) && { onCancel: options.onCancel },
```
ranshoshani34, line 136:
```suggestion
			onConfirm: options.onConfirm ,
			onCancel: options.onCancel ,
```

You dont need the complex syntax. If the method doesnt exist it will be undefined in both ways.

app/modules/import-export/assets/js/admin.js
```diff
@@ -1,19 +1,189 @@
...
+
+	/**
+	 * Retrieving the last imported kit from the elementorAppConfig global
+	 *
+	 * @return {string}
+	 */
+	getKitToRemoveName() {
+		const lastKit = elementorAppConfig[ 'import-export' ].lastImportedSession;
+
+		if ( lastKit.kit_title ) {
```
ranshoshani34, line 161:
```suggestion
		if ( lastKit?.kit_title ) {
```

app/modules/import-export/assets/js/admin.js
```diff
@@ -1,19 +1,189 @@
...
+			return lastKit.kit_title;
+		}
+
+		if ( lastKit.kit_name ) {
```
ranshoshani34, line 165:
```suggestion
		if ( lastKit?.kit_name ) {
```

app/modules/import-export/assets/js/admin.js
```diff
@@ -1,19 +1,189 @@
...
+
+	/**
+	 * ConvertNameToTitle
+	 *
+	 * @param {string} name
+	 *
+	 * @return {string}
+	 */
+	convertNameToTitle( name ) {
+		const words = name.split( /[-_]+/ );
```
ranshoshani34, line 180:
```suggestion
		const words = name.split( /[-_]+/ ).map( word => word.charAt(0).toUpperCase() + word.substring(1));
```

ranshoshani34 (changes requested):

ranshoshani34 (dismissed):
Great job!

ranshoshani34 (changes requested):

ranshoshani34 (dismissed):

app/modules/import-export/assets/js/admin.js
```diff
@@ -1,19 +1,245 @@
...
+	/**
+	 * ConvertNameToTitle
+	 *
+	 * @param {string} name
+	 *
+	 * @return {string}
+	 */
+	convertNameToTitle( name ) {
+		const words = name.split( /[-_]+/ ).map( ( word ) => word[ 0 ].toUpperCase() + word.substring( 1 ) );
+		return words.join( ' ' );
```
StyleShit, line 239:
```suggestion
		return name
			.split( /[-_]+/ )
			.map( ( word ) => word[ 0 ].toUpperCase() + word.substring( 1 ) )
			.join( ' ' );
```


Or maybe shorter:

```suggestion
		return name.replace(
			/(?:\b|[-_])(\w)/g,
			( match, w ) => ' ' + w.toUpperCase()
		).trim();
```

app/modules/import-export/module.php
```diff
@@ -89,13 +89,11 @@ public function __construct() {
...
 		( new Usage() )->register();
+
+		$this->revert = new Revert();
 	}
 
 	public function get_init_settings() {
-		if ( ! Plugin::$instance->app->is_current() ) {
-			return [];
-		}
-
```
StyleShit, line 98: why?

app/modules/import-export/module.php
```diff
@@ -208,8 +205,10 @@ private function render_import_export_tab_content() {
 
 			<?php
 			if ( $should_show_revert_section ) {
+				$admin_post_url = admin_url( 'admin-post.php?action=elementor_revert_kit' );
+				$nonced_admin_post_url = wp_nonce_url( $admin_post_url, 'elementor_revert_kit' );
 				$link_attributes = [
-					'href' => wp_nonce_url( admin_url( 'admin-post.php?action=elementor_revert_kit' ), 'elementor_revert_kit' ),
+					'href' => $this->maybe_add_referrer_param( $nonced_admin_post_url ),
```
StyleShit, line 211: just extract all of it into a method that builds the URL.
When reading it, the developer doesn't care how the URL is gonna look like

app/modules/import-export/module.php
```diff
@@ -601,16 +615,24 @@ private function get_config_data() {
...
 			'recentlyEditedElementorPageUrl' => $this->get_recently_edited_elementor_page_url(),
+			'tools_url' => Tools::get_url(),
 			'importSessions' => Revert::get_import_sessions(),
 		];
+
+		$last_import_session = $this->revert->get_last_import_session();
+		if ( ! empty( $last_import_session ) ) {
+			$config_data['lastImportedSession'] = $last_import_session;
+		}
+
```
StyleShit, line 634: maybe just always return it? 
so on the client side you won't have to check each time for the key existence

app/modules/kit-library/assets/js/components/kit-already-applied-dialog.js
```diff
@@ -0,0 +1,27 @@
...
+		<Dialog
+			title={ __( 'You\'ve already applied a Kit.', 'elementor' ) }
+			text={ <>
+				{ __( 'Applying two Kits on the same website will mix global styles and colors and hurt your site\'s performance.', 'elementor' ) }
+				<br /><br />
+				{ __( 'Remove the existing Kit before applying a new one.', 'elementor' ) }
+			</> }
+			approveButtonText={ __( 'Remove existing Kit', 'elementor' ) }
+			approveButtonColor="primary"
+			approveButtonOnClick={ () => location.href = elementorToolsUrl + '&referrer_kit=' + props.id + '#tab-import-export-kit' }
```
StyleShit, line 15: don't build a URL like that. Use the `URL` object

app/modules/kit-library/assets/js/components/kit-dialog.js
```diff
@@ -0,0 +1,32 @@
...
+	if ( kitAlreadyApplied && ! applyAnyway ) {
+		return (
+			<KitAlreadyAppliedDialog
+				id={ props.id }
+				dismissButtonOnClick={ () => setApplyAnyway( true ) }
+				onClose={ props.onClose }
+			/>
+		);
+	}
+	return (
```
StyleShit, line 19: add some spaces between the things

StyleShit (changes requested):

app/modules/import-export/assets/js/admin.js
```diff
@@ -1,19 +1,245 @@
-const revertButton = document.getElementById( 'elementor-import-export__revert_kit' );
+/**
+ * Class Admin
+ */
+class Admin {
```
StyleShit, line 4: tests?

StyleShit (changes requested):

app/modules/import-export/assets/js/admin.js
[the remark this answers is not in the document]
ZeyadMahagna: It doesn't need to, it's just for organization because the file got too bloated

app/modules/import-export/assets/js/admin.js
```diff
@@ -1,19 +1,245 @@
...
+		const revertKitHeight = revertButton.parentElement.offsetHeight;
+
+		return (
+			document.body.clientHeight -
+			adminBarHeight -
+			revertKitHeight -
+			document.getElementById( 'wpfooter' ).offsetHeight -
+			15 // Extra margin at the top
+		) + 'px';
+	}
```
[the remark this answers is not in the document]
ZeyadMahagna: This is only applied when we are coming from the Kit Library (referred).
We are checking if the WP Admin Bar exists and subtracting its offset height if it does, I don't know how to do this in CSS.
I don't think this looks weird

app/modules/import-export/assets/js/admin.js
```diff
@@ -1,19 +1,245 @@
...
+
+	/**
+	 * Scroll to the bottom of the page
+	 */
+	scrollToBottom() {
+		setTimeout( () => {
+			window.scrollTo( 0, document.body.scrollHeight );
+		} );
```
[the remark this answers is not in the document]
ZeyadMahagna: In order to enqueue the scroll function after the execution stack, otherwise the scrolling does not work

app/modules/import-export/assets/js/admin.js
```diff
@@ -1,19 +1,245 @@
...
+	 *
+	 * @return {string}
+	 */
+	convertNameToTitle( name ) {
+		const words = name.split( /[-_]+/ ).map( ( word ) => word[ 0 ].toUpperCase() + word.substring( 1 ) );
+		return words.join( ' ' );
+	}
 }
+
+window.addEventListener( 'load', () => {
```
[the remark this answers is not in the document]
ZeyadMahagna: need this to be on 'load' here as DOMContentLoaded is not late enough

app/modules/import-export/module.php
ZeyadMahagna: Because we need the 'lastImportedSession' in the admin

app/modules/import-export/assets/js/admin.js
```diff
@@ -1,19 +1,245 @@
...
+		}
+		sessionStorage.removeItem( this.kitNameToRemoveKey );
+	}
+
+	/**
+	 * Retrieving the last imported kit from the elementorAppConfig global
+	 *
+	 * @return {string}
+	 */
+	getKitToRemoveName() {
```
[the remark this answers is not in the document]
ZeyadMahagna: Current Kit might imply the Kit we are coming from

app/modules/import-export/assets/js/admin.js
```diff
@@ -1,19 +1,245 @@
...
+		} );
+		this.cleanUp( true );
+	}
+
+	/**
+	 * CreateKitDeletedWidget
+	 *
+	 * @param {Object} options
+	 */
+	createKitDeletedWidget( options ) {
```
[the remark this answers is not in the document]
ZeyadMahagna: Why? the two 'delete' widgets it's used for both share the same 'id' and 'headerMessage' field. Why duplicate this?

app/modules/import-export/assets/js/admin.js
```diff
@@ -1,19 +1,245 @@
...
+		elementorCommon.dialogsManager.createWidget( 'confirm', {
+			id: 'e-revert-kit-deleted-dialog',
+			headerMessage: sessionStorage.getItem( this.kitNameToRemoveKey ) + __( ' was successfully deleted', 'elementor' ),
```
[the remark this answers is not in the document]
ZeyadMahagna: Is there a sprintf for browser JS? I can't find it in MDN or any other docs

app/modules/import-export/assets/js/admin.js
```diff
@@ -1,19 +1,245 @@
...
+				cancel: __( 'Cancel', 'elementor' ),
 			},
-			onConfirm() {
-				location.href = revertButton.href;
+			onConfirm: this.onRevertConfirm.bind( this, event.target, kitNameToRemove ),
+		} ).show();
+	}
+
+	onRevertConfirm( revertBtn, kitNameToRemove ) {
+		const referrerKit = new URLSearchParams( revertBtn.href ).get( 'referrer_kit' );
```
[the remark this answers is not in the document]
ZeyadMahagna: It's only passed once, here

app/modules/import-export/assets/js/admin.js
```diff
@@ -1,19 +1,245 @@
...
+		location.href = revertBtn.href;
+	}
+
+	setSessionStorage( referred, referrerKit, kitNameToRemove ) {
+		sessionStorage.setItem( this.isKitReferredKey, referred ? 'yes' : 'no' );
+		if ( referred ) {
+			sessionStorage.setItem( this.kitReferrerKey, referrerKit );
+		}
+		sessionStorage.setItem( this.kitNameToRemoveKey, kitNameToRemove );
+	}
```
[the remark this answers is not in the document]
ZeyadMahagna: About the third question, I used the `referred` because there are 3 possibilities when showing the referrer kit dialog:
1. Don't show the dialog (User is on the page but has not removed a kit)
2. show not referred dialog
3. show referred dialog

Perhaps I could use `referrerKit` in the same way and set it to an empty string when not referred but a kit has been reverted, but this seems more semantic

app/modules/import-export/assets/js/admin.js
```diff
@@ -1,19 +1,245 @@
...
+	 */
+	constructor() {
+		this.configureRevertBtn();
+		this.maybeShowKitReferrerDialog();
+	}
+
+	/**
+	 * ConfigureRevertBtn
+	 */
+	configureRevertBtn() {
```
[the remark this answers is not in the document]
StyleShit: I would split it to `getElements()` and `bindEvents()`, which is our convention for widgets, etc.

app/modules/import-export/assets/js/admin.js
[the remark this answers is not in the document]
StyleShit: There is no any class on the body or something?

app/modules/import-export/assets/js/admin.js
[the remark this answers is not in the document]
StyleShit: what do you have in the stack that causes this?
if it's the margin calculation, let's try to remove it and then we'll be able to remove the timeout

app/modules/import-export/assets/js/admin.js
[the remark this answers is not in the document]
StyleShit: yeah, that's still weird to pass it if you have a class. if you don't use the features of a class, I don't see a reason to use a class

app/modules/import-export/assets/js/admin.js
```diff
@@ -1,19 +1,214 @@
...
+	saveToCache( referrerKitId, kitNameToRemove ) {
+		sessionStorage.setItem( this.KIT_DATA_KEY, JSON.stringify( { referrerKitId, kitNameToRemove } ) );
+	}
+
+	getDataFromCache() {
+		if ( this.cachedKitData ) {
+			return this.cachedKitData;
+		}
+
+		this.cachedKitData = JSON.parse( sessionStorage.getItem( this.KIT_DATA_KEY ) );
```
StyleShit, line 202: it might throw when the data isn't JSON, maybe add a try catch here

app/modules/import-export/assets/js/admin.js
```diff
@@ -1,19 +1,245 @@
...
+	revertBtnOnClick( event ) {
 		event.preventDefault();
 
+		const kitNameToRemove = this.getKitToRemoveName();
+
 		elementorCommon.dialogsManager.createWidget( 'confirm', {
-			headerMessage: __( 'Sure you want to make these changes?', 'elementor' ),
-			message: __( 'Removing assets or changing your site settings can drastically change the look of your website.', 'elementor' ),
+			headerMessage: __( 'Are you sure?', 'elementor' ),
+			message: __( 'Removing ', 'elementor' ) + kitNameToRemove + __( ' will permanently delete changes made to the Kit\'s content and site settings', 'elementor' ),
```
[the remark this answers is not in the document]
StyleShit: ?

app/modules/import-export/assets/js/admin.js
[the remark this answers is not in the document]
StyleShit: you can make a simple one:
https://github.com/elementor/elementor/blob/88cf8c7f0101a23f58e022ee696d656d84fe0bb7/app/modules/kit-library/assets/js/components/apply-kit-dialog.js#L23

app/modules/import-export/assets/js/admin.js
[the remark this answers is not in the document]
StyleShit: so share only those, just put them in variables and consume when creating the widgets

app/modules/import-export/assets/js/admin.js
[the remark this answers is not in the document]
StyleShit: so maybe `getActiveKit()`?

app/modules/import-export/module.php
StyleShit: yeah, but won't it expose all of the IE data even when not needed?

app/modules/import-export/module.php
```diff
@@ -230,6 +228,27 @@ private function render_import_export_tab_content() {
...
+	 *
+	 * @return string
+	 */
+	private function maybe_add_referrer_param( string $href ): string {
+		$param_name = 'referrer_kit';
+		if ( empty( $_GET[ $param_name ] ) ) {
+			return $href;
+		}
+		return add_query_arg( $param_name, sanitize_key( $_GET[ $param_name ] ), $href );
+	}
```
StyleShit, line 250: add some spaces, let the code breathe 😄

app/modules/import-export/assets/js/admin.js
```diff
@@ -1,19 +1,214 @@
...
+
+		if ( 0 === referrerKitId.length ) {
+			this.createKitDeletedWidget( {
+				message: __( 'Try a different Kit or build your site from scratch.', 'elementor' ),
+				strings: {
+					confirm: __( 'OK', 'elementor' ),
+					cancel: __( 'Kit Library', 'elementor' ),
+				},
+				onCancel: () => {
+					location.href = elementorAppConfig.base_url + '#/kit-library';
```
StyleShit, line 118: shouldn't the URLs come from the backend?

app/modules/kit-library/assets/js/components/kit-already-applied-dialog.js
```diff
@@ -0,0 +1,36 @@
+import { Dialog } from '@elementor/app-ui';
+
+export default function KitAlreadyAppliedDialog( props ) {
+	const elementorToolsUrl = elementorAppConfig[ 'import-export' ].tools_url;
```
StyleShit, line 4: put it inside `getRemoveKitUrl()`

app/modules/kit-library/assets/js/components/kit-dialog.js
```diff
@@ -0,0 +1,34 @@
+import { useState } from 'react';
+
+import KitAlreadyAppliedDialog from './kit-already-applied-dialog';
+import ApplyKitDialog from './apply-kit-dialog';
+
+export default function KitDialog( props ) {
+	const [ applyAnyway, setApplyAnyway ] = useState( false );
+
+	const kitAlreadyApplied = elementorAppConfig[ 'import-export' ].lastImportedSession.session_id;
```
StyleShit, line 9:
```suggestion
	const kitAlreadyApplied = !! elementorAppConfig[ 'import-export' ].lastImportedSession.session_id;
```

tests/jest/unit/app/modules/import-export/assets/js/admin.test.js
```diff
@@ -0,0 +1,30 @@
+describe( 'Admin Test', () => {
```
StyleShit, line 1: change to something more specific

tests/jest/unit/app/modules/import-export/assets/js/admin.test.js
```diff
@@ -0,0 +1,30 @@
...
+	it( 'should check if revert button handler is added', function () {
```
StyleShit, line 2: eslint whole file

StyleShit, line 2: the test doesn't really check anything useful

StyleShit (changes requested):

app/modules/import-export/assets/js/admin.js
ZeyadMahagna: the base is coming from the backend, only `kit-library` no sense sending this from the backend

StyleShit: IDK, seems like there are many URLs here:

https://github.com/elementor/elementor/blob/9fcaebbf68f755acf89d743e0b2234aac8cc4d5f/app/app.php#L87-L99

app/modules/import-export/assets/js/admin.js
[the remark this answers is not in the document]
ZeyadMahagna: What's wrong with simple concatenation?

app/modules/import-export/module.php
ZeyadMahagna: it's exposed on the admin side now too, and it's needed there now

app/modules/import-export/assets/js/admin.js
```diff
@@ -1,19 +1,245 @@
...
+		if ( referred ) {
+			sessionStorage.setItem( this.kitReferrerKey, referrerKit );
+		}
+		sessionStorage.setItem( this.kitNameToRemoveKey, kitNameToRemove );
+	}
+
+	/**
+	 * MaybeShowKitReferrerDialog
+	 */
+	maybeShowKitReferrerDialog() {
```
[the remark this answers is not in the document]
ZeyadMahagna: What functions do you want me to split it to?

app/modules/import-export/assets/js/admin.js
```diff
@@ -1,19 +1,231 @@
...
+		const revertKitHeight = this.revertBtn.parentElement.offsetHeight;
+
+		return (
+			document.body.clientHeight -
+			adminBarHeight -
+			revertKitHeight -
+			document.getElementById( 'wpfooter' ).offsetHeight -
+			15 // Extra margin at the top
+		) + 'px';
+	}
```
StyleShit, line 73: OK, after understanding the feature better:

1. I can't merge it like this, we need to tell the product that this UI isn't gonna happen. I talked with @nivnoimanelementor about it.
2. It doesn't work like the Figma anyway, you scroll the whole body instead of the inner wrapper.
3. The best I can suggest is adding some highlight effect to the button somehow (add class on page load?), you can suggest it to the product.

app/modules/import-export/assets/js/admin.js
```diff
@@ -1,19 +1,231 @@
...
+			window.scrollTo( 0, document.body.scrollHeight );
+		} );
+	}
+
+	/**
+	 * RevertBtnOnClick
+	 *
+	 * @param {Event} event
+	 */
+	revertBtnOnClick( event ) {
```
StyleShit, line 89:
```suggestion
	onRevertButtonClick( event ) {
```

app/modules/import-export/assets/js/admin.js
```diff
@@ -1,19 +1,231 @@
...
 		event.preventDefault();
 
 		elementorCommon.dialogsManager.createWidget( 'confirm', {
-			headerMessage: __( 'Sure you want to make these changes?', 'elementor' ),
-			message: __( 'Removing assets or changing your site settings can drastically change the look of your website.', 'elementor' ),
+			headerMessage: __( 'Are you sure?', 'elementor' ),
+			message: __( 'Removing ', 'elementor' ) + this.activeKitName + __( ' will permanently delete changes made to the Kit\'s content and site settings', 'elementor' ),
```
StyleShit, line 94: sprintf, here and others

app/modules/import-export/assets/js/admin.js
```diff
@@ -1,19 +1,231 @@
...
+		this.saveToCache( referrerKitId ?? '' );
+
+		location.href = this.revertBtn.href;
+	}
+
+	maybeShowReferrerKitDialog() {
+		const { referrerKitId } = this.getDataFromCache();
+		if ( undefined === referrerKitId ) {
+			return;
+		}
```
StyleShit, line 115: do you really need the undefined?

app/modules/kit-library/assets/js/components/kit-dialog.js
```diff
@@ -0,0 +1,35 @@
+import { useState } from 'react';
+
+import KitAlreadyAppliedDialog from './kit-already-applied-dialog';
```
StyleShit, line 3:
```suggestion
import { useState } from 'react';
import KitAlreadyAppliedDialog from './kit-already-applied-dialog';
```

tests/jest/unit/app/modules/import-export/assets/js/admin.test.js
```diff
@@ -0,0 +1,121 @@
+describe( 'Import Export Admin Test', () => {
+	window.elementorAppConfig = {
+		'import-export': {
+			lastImportedSession: {
+				kit_title: 'Kit Title',
+			},
+		},
+	};
+
+	require( 'elementor/app/modules/import-export/assets/js/admin.js' );
```
StyleShit, line 10: use regular import

tests/jest/unit/app/modules/import-export/assets/js/admin.test.js
```diff
@@ -0,0 +1,121 @@
+describe( 'Import Export Admin Test', () => {
+	window.elementorAppConfig = {
+		'import-export': {
+			lastImportedSession: {
+				kit_title: 'Kit Title',
+			},
+		},
+	};
```
StyleShit, line 8: put it in beforeEach, and remove in afterEach

tests/jest/unit/app/modules/import-export/assets/js/admin.test.js
```diff
@@ -0,0 +1,121 @@
...
+		// Arrange
+		window.elementorCommon = {
+			dialogsManager: {
+				createWidget: jest.fn( () => {
+					return {
+						show: jest.fn( () => {} ),
+					};
+				} ),
+			},
+		};
```
StyleShit, line 22: if you mock it in every test, remove the top one (and cleanup after each test)

tests/jest/unit/app/modules/import-export/assets/js/admin.test.js
```diff
@@ -0,0 +1,121 @@
...
+
+		// Act
+		window.dispatchEvent( new Event( 'load' ) );
```
StyleShit, line 25: I think it'll be easier to unit test the Admin class only, so you can just:
```JS
// Act
( new Admin() ).init()
```
or something

tests/jest/unit/app/modules/import-export/assets/js/admin.test.js
```diff
@@ -0,0 +1,121 @@
...
+		'import-export': {
+			lastImportedSession: {
+				kit_title: 'Kit Title',
+			},
+		},
+	};
+
+	require( 'elementor/app/modules/import-export/assets/js/admin.js' );
+
+	it( 'should check if "kit deleted" dialog is not shown without cache', () => {
```
StyleShit, line 12: would probably be better to use non-techincal test names, form the user POV, instead of writing "cache"

tests/jest/unit/app/modules/import-export/assets/js/admin.test.js
```diff
@@ -0,0 +1,121 @@
...
+		// Arrange
+		window.elementorCommon = {
+			dialogsManager: {
+				createWidget: jest.fn( () => {
+					return {
+						show: jest.fn( () => {} ),
+					};
```
StyleShit, line 19: you don't need to mock "show" if you don't use it

tests/jest/unit/app/modules/import-export/assets/js/admin.test.js
```diff
@@ -0,0 +1,121 @@
...
+		};
+
+		// Act
+		window.dispatchEvent( new Event( 'load' ) );
+
+		// No Kit referrer dialog shown without refferer in cache
+		expect( window.elementorCommon.dialogsManager.createWidget ).not.toHaveBeenCalled();
+	} );
+
+	it( 'should check if the referred "kit deleted" dialog is shown', () => {
```
StyleShit, line 31: when it should be shown?
A nice convention I found goes something like this:
`it( should check if ??? is ??? when ??? )`
`it( should do ??? when ??? )`

Maybe you can try using it in your tests

tests/jest/unit/app/modules/import-export/assets/js/admin.test.js
```diff
@@ -0,0 +1,121 @@
...
+				createWidget: jest.fn( ( type, options ) => {
+					// Auto confirm
+					sessionStorage.setItem( 'elementor-kit-data', JSON.stringify( {
+						referrerKitId: '123',
+						activeKitName: 'Active Kit Name',
+					} ) );
+
+					// To expect
+					window.elementorCommon.lastWidgetHeaderMessage = options.headerMessage;
+					window.elementorCommon.lastWidgetMessage = options.message;
```
StyleShit, line 44: you can use `.toHaveBeenCalledWith()`

tests/jest/unit/app/modules/import-export/assets/js/admin.test.js
```diff
@@ -0,0 +1,121 @@
...
+					return {
+						show: jest.fn( () => {} ),
+					};
+				} ),
+			},
+		};
+		window.document.body.innerHTML = `
+			<button id="elementor-import-export__revert_kit">
+			</button>
+		`;
```
StyleShit, line 54: cleanup after test

tests/jest/unit/app/modules/import-export/assets/js/admin.test.js
```diff
@@ -0,0 +1,121 @@
...
+		expect( window.elementorCommon.lastWidgetHeaderMessage ).toEqual( 'Are you sure?' );
+		expect( window.elementorCommon.lastWidgetMessage ).toEqual( 'Removing Kit Title will permanently delete changes made to the Kit\'s content and site settings' );
+
+		// Simulate reload
+		window.dispatchEvent( new Event( 'load' ) );
+
+		// Kit deleted dialog
+		expect( window.elementorCommon.dialogsManager.createWidget ).toHaveBeenCalledTimes( 2 );
+		expect( window.elementorCommon.lastWidgetHeaderMessage ).toEqual( 'Active Kit Name was successfully deleted' );
+		expect( window.elementorCommon.lastWidgetMessage ).toEqual( 'You\'re ready to apply a new Kit!' );
```
StyleShit, line 73: you can use `toHaveBeenNthCalledWith()`

tests/jest/unit/app/modules/import-export/assets/js/admin.test.js
```diff
@@ -0,0 +1,121 @@
...
+				createWidget: jest.fn( ( type, options ) => {
+					// Auto confirm
+					sessionStorage.setItem( 'elementor-kit-data', JSON.stringify( {
+						referrerKitId: '', // Non referred
+						activeKitName: 'Active Kit Name',
+					} ) );
+
+					// To expect
+					window.elementorCommon.lastWidgetHeaderMessage = options.headerMessage;
+					window.elementorCommon.lastWidgetMessage = options.message;
```
StyleShit, line 89: same as above

StyleShit (changes requested):

app/modules/import-export/assets/js/admin.js
ZeyadMahagna: yes this is one of three cases, the case where we don't show a dialog.
We can't check for a falsey value because that the second case is for an empty string (false value) which shows the non-referred dialog

tests/jest/unit/app/modules/import-export/assets/js/admin.test.js
ZeyadMahagna: the admin file does not export anything and it doesn't need to

tests/jest/unit/app/modules/import-export/assets/js/admin.test.js
ZeyadMahagna: we mock it in a different way everytime..

tests/jest/unit/app/modules/import-export/assets/js/admin.test.js
ZeyadMahagna: I am not exporting the class, and if the whole class is meant to run on load like in our case, essentially the same thing as doing `new Admin()` I don't see what is wrong with this

tests/jest/unit/app/modules/import-export/assets/js/admin.test.js
ZeyadMahagna: It is called in the class

tests/jest/unit/app/modules/import-export/assets/js/admin.test.js
ZeyadMahagna: I tried this. In this case the second `options` parameter is an object and not possible to check equality with it

app/modules/import-export/assets/js/admin.js
nivnoimanelementor: This is Ok by me. I approve it 👍

nivnoimanelementor (approved):

PR #21072, opened 2023-01-19
Fix: Links cannot be crawled on Toggle and Accordion Widget

Ludwig44:
Google Page Speed Insight shows SEO errors and says it can't crawl links because of Toggle and Accordion widgets.

So here is an adjustment of the code to switch to span.


Ludwig44:
👍

PR #21098, opened 2023-01-22
Tweak: Added Lazy Load support for background images in Slides widget (Core) [ED-9118]

maork-elementor:


modules/lazyload/module.php
```diff
@@ -42,34 +42,62 @@ private function update_element_attributes( Element_Base $element ) {
 		$settings = $element->get_settings_for_display();
 		$controls = $element->get_controls();
 		$lazyload_attribute_name = 'data-e-bg-lazyload';
-
+		$attributes = [];
 		$controls_with_background_image = array_filter( $controls, function( $control ) {
 			return Utils::get_array_value_by_keys( $control, [ 'background_lazyload', 'active' ] );
 		} );
 
+		$repeater = array_filter( $controls, function( $control ) {
```
Nevoss, line 50: Extract this code into functions, what is the input and the output of the function?

modules/lazyload/module.php
```diff
@@ -156,9 +184,31 @@ public function __construct() {
...
+					return lazyloadBackgrounds;
+				};
+
 				const lazyloadRunObserver = () => {
 					const dataAttribute = 'data-e-bg-lazyload';
-					const lazyloadBackgrounds = document.querySelectorAll( `[${ dataAttribute }]:not(.lazyloaded)` );
+					const reapterAttribute = 'data-e-bg-repeater';
+
+					let lazyloadBackgrounds = document.querySelectorAll( `[${ dataAttribute }]:not(.lazyloaded, [${reapterAttribute}])` );
+					lazyloadBackgrounds = getReapetersBackgroundLazyload( lazyloadBackgrounds, reapterAttribute, dataAttribute );
```
Nevoss, line 210: What is the different between repeaters and selectors?

Nevoss (changes requested):

modules/lazyload/module.php
maork-elementor: Repeaters are the item of repeater control.
The selector is CSS rule which apply to repeater control.

modules/lazyload/module.php
maork-elementor: controls -> repeater

Done

modules/lazyload/module.php
```diff
@@ -156,9 +192,31 @@ public function __construct() {
...
```
Nevoss, line 218: try make it simpler by only observe selectors without the need to know if it is repeater or not

Nevoss (changes requested):

modules/lazyload/module.php
maork-elementor: I pushed a POC version, please take a look if this is simpler enough. I used like you offered to use simple CSS selectors without knowing which is what on the client, Also in the backend we have now arrayed the selectors for each wrapper.

modules/lazyload/module.php
```diff
@@ -38,38 +38,118 @@ private function enqueue_styles() {
...
 
+		$repeaters_selectors = $this->get_repeaters_selectors( $controls, $settings );
+
 		foreach ( $controls_with_background_image as $control_name => $control_data ) {
+
+			$selectors = [];
+
+			if ( $repeaters_selectors ) {
+				$selectors = array_merge( $repeaters_selectors, $selectors );
+			}
```
Nevoss, line 103: Why it is inside the foreach loop

modules/lazyload/module.php
```diff
@@ -38,38 +38,118 @@ private function enqueue_styles() {
...
+
+					if ( $background_image_url && ! $wrapper_bg_selector ) {
+						$selectors[] = $wrapper_selector;
+					}
+
+					$attributes[ $lazyload_attribute_name ] = implode( ',', $selectors );
+
+					$element->add_render_attribute( '_wrapper',
+						$attributes
+					);
```
Nevoss, line 132: Move outside of the loop

Nevoss (changes requested):

modules/lazyload/module.php
maork-elementor: It is inside the loop because of it's running for each control inside the element. For each control has his own `$selectors` & `$repeaters_selectors`. If I move it outside the for each loop, the lazyload will not work. Did i miss anything?

modules/lazyload/module.php
```diff
@@ -38,38 +38,125 @@ private function enqueue_styles() {
...
+		$repeater_background_image_url = Utils::get_array_value_by_keys( $repeater_settings, $repeater_keys );
+
+		if ( $repeater_background_image_url ) {
+			return $repeater_selector;
+		}
+
+		return false;
+	}
+
 	private function update_element_attributes( Element_Base $element ) {
```
Nevoss, line 85: Maybe consider doing it like this:
```php
$settings = $element->get_settings_for_display();
$controls = $element->get_controls();

$selectors = [];

foreach ( $controls as $control_name => $control_data ) {
	if (
		$is_repeater // Here should be the check if the control is repeater.
	) {
		// You dont need to run over all the controls again, because you have the control data already.
		$selectors = array_merge(
			$this->get_repeater_selector( $control_data ),
			$selectors // Merge all teh selectors to one array.
		);

		// Because it is a repeater, you already handle it and you should not continue the code.
		continue;
	}

	// Here should be the code the normal control and get its selectors.
	// In this method if there is no selector specified, it will return the wrapper selector.
	$selectors = array_merge(
		$this->get_control_selector( $control_data ),
		$selectors // Merge all teh selectors to one array.
	);
}

// Here it should make the array_unique and trim and merge theme into one string (which is the one selector)
$final_selector = $this->merge_selectors( $selectors );
if ( ! empty( $final_selector ) ) {
	$element->add_render_attribute(
		'_wrapper',
		[ 'data-e-bg-lazyload' => $final_selector ]
	);
}
```

This is only a guideline,  I think there will be more changes you will have to do.

In my opinion it will make the code more readable, let me know what do you think?

Nevoss (changes requested):

modules/lazyload/module.php
```diff
@@ -38,38 +38,146 @@ private function enqueue_styles() {
...
+				// If the control (wrapper) has a background image, add the wrapper selector to the control selectors.
+				if ( $background_image_url ) {
+					$control_selectors[] = $wrapper_selector;
+				}
+
+				// Trim spaces and remove duplicates.
+				$control_selectors = array_unique( array_map( 'trim', $control_selectors ) );
+
+				// Build the lazyload attribute selectors value. e.g. ".elementor-element-123 .selector1, .elementor-element-123 .selector2".
+				$attributes[ $lazyload_attribute_name ] = implode( ',', $control_selectors );
```
Nevoss, line 154: Can it move outside of the loop?

modules/lazyload/module.php
```diff
@@ -38,38 +38,146 @@ private function enqueue_styles() {
...
-					$bg_selector = Utils::get_array_value_by_keys( $control_data, [ 'background_lazyload', 'selector' ] ) ?? '';
-					$element->add_render_attribute( '_wrapper', [
-						$lazyload_attribute_name => $bg_selector,
-					] );
+			$control_selectors = [];
+
+			// Marge repeaters selectors to the wrapper control selectors if exist.
+			if ( $repeaters_selectors ) {
+				$control_selectors = array_merge( $repeaters_selectors, $control_selectors );
+			}
```
Nevoss, line 127: Can it move outside of the loop

Nevoss (changes requested):

modules/lazyload/module.php
maork-elementor: Moving the $selectors = []; statement outside the loop would cause the $selectors array to retain the values from the previous iteration. This would result in incorrect values being added to the $control_selectors array for each iteration.

In the current implementation, the $selectors array is initialized as an empty array at the start of each iteration of the loop. This ensures that the $control_selectors array only contains selectors from the current iteration and not from previous iterations.

If you move the $selectors = []; statement outside the loop, then the $control_selectors array would accumulate selectors from all iterations of the loop. This would result in incorrect values being added to the $attributes array for the lazyload attribute.

modules/lazyload/module.php
maork-elementor: The same answer as the issue below.

modules/lazyload/module.php
maork-elementor: Update - We sit together and refactor the code.

PR #21105, opened 2023-01-23
Tweak: Add thumbnail to most recently installed kit [ED-9422]

ZeyadMahagna:


ranshoshani34 (dismissed):

app/modules/import-export/processes/import.php
```diff
@@ -413,6 +417,31 @@ public function init_import_session() {
...
+
+		$kits = json_decode( wp_remote_retrieve_body( $response ) , true );
+		foreach ( $kits as $kit ) {
+			if ( $this->manifest['title'] === $kit['title'] ) {
+				return $kit['thumbnail'];
+			}
+		}
+
+		return '';
+	}
```
ranshoshani34, line 443: Can't you get the individual kit instead of all kits?

ranshoshani34 (changes requested):

app/modules/import-export/processes/import.php
ZeyadMahagna: We don't have the API Kit ID in the manifest

app/modules/import-export/processes/import.php
```diff
@@ -18,7 +18,11 @@
 use Elementor\App\Modules\ImportExport\Runners\Import\Wp_Content;
 use Elementor\App\Modules\ImportExport\Module;
 
+use Elementor\App\Modules\KitLibrary\Connect\Kit_Library as Kit_Library_Api;
```
StyleShit, line 21:
```suggestion
use Elementor\App\Modules\ImportExport\Module;
use Elementor\App\Modules\KitLibrary\Connect\Kit_Library as Kit_Library_Api;
```

app/modules/import-export/processes/import.php
```diff
@@ -18,7 +18,11 @@
...
+use Elementor\Data\V2\Base\Exceptions\WP_Error_Exception;
+
 class Import {
+
```
StyleShit, line 25: remove, to have less unrelated diff

app/modules/import-export/module.php
```diff
@@ -171,16 +171,16 @@ private function render_import_export_tab_content() {
 			if ( ! empty( $penultimate_imported_kit ) ) {
 				$revert_text = sprintf(
 					esc_html__( 'Remove all the content and site settings that came with "%1$s" on %2$s %3$s and revert to the site setting that came with "%4$s" on %5$s.', 'elementor' ),
-					! empty( $last_imported_kit['kit_name'] ) ? $last_imported_kit['kit_name'] : esc_html__( 'imported kit', 'elementor' ),
+					! empty( $last_imported_kit['kit_title'] ) ? $last_imported_kit['kit_title'] : esc_html__( 'imported kit', 'elementor' ),
```
StyleShit, line 174: what's the difference between kit name and title?

StyleShit (changes requested):

app/modules/import-export/module.php
StyleShit, line 174:
```suggestion
					$last_imported_kit['kit_title'] ?? esc_html__( 'imported kit', 'elementor' ),
```

also others

app/modules/import-export/processes/import.php
```diff
@@ -418,6 +430,32 @@ public function init_import_session( $save_instance_data = false ) {
...
+		if ( ! empty( $this->manifest['thumbnail'] ) ) {
+			return $this->manifest['thumbnail'];
+		}
+
+		if ( empty( $this->api_id ) ) {
+			return '';
+		}
+
+		$api = new Kit_Library_Api();
+		$kit = $api->get_by_id( $this->api_id );
```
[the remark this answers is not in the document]
ZeyadMahagna: This is a fallback, most manifests do not have a thumbnail link currently

app/modules/import-export/processes/import.php
```diff
@@ -418,6 +428,32 @@ public function init_import_session( $save_instance_data = false ) {
 		update_option( Module::OPTION_KEY_ELEMENTOR_IMPORT_SESSIONS, $import_sessions, false );
 	}
 
+	/**
+	 * Get the manifest thumbnail, goes to the home page thumbnail if main doesn't exist
```
StyleShit, line 432:
```suggestion
	 * Get the kit thumbnail, goes to the home page thumbnail if main doesn't exist,
```

app/modules/import-export/processes/import.php
```diff
@@ -36,6 +38,13 @@ class Import {
 	 */
 	private $session_id;
 
+	/**
+	 * The Kit BA-API ID.
+	 *
+	 * @var string
+	 */
```
StyleShit, line 45:
```suggestion
	/**
	 * The Kit ID.
	 *
	 * @var string
	 */
```

app/modules/import-export/processes/import.php
```diff
@@ -387,7 +397,7 @@ public function init_import_session( $save_instance_data = false ) {
 			'session_id' => $this->session_id,
 			'kit_title' => $this->manifest['title'] ?? '',
 			'kit_name' => $this->manifest['name'] ?? '',
-			'kit_thumbnail' => $this->manifest['thumbnail'] ?? '',
+			'kit_thumbnail' => $this->get_kit_thumbnail(),
```
StyleShit, line 400: test the parent function will all of the edge cases of the thumbnail

StyleShit (changes requested):

tests/phpunit/elementor/app/import-export/processes/test-import.php
```diff
@@ -475,6 +475,49 @@ public function test_get_manifest__kit_library() {
 		}
 	}
 
+	public function test_init_import_session__kit_thumbnail() {
+		// Arrange
+		$import_settings = [
+			'referrer' => 'kit-library',
+		];
+		$module = new Module();
```
StyleShit, line 483: why do you run the whole module? can't you do it like the other tests?
something like: `( new Import() )->run()`

tests/phpunit/elementor/app/import-export/processes/test-import.php
```diff
@@ -475,6 +475,49 @@ public function test_get_manifest__kit_library() {
 		}
 	}
 
+	public function test_init_import_session__kit_thumbnail() {
```
StyleShit, line 478: maybe something like this? to use the same convention as the previous tests:

```suggestion
	public function test_run__get_kit_thumbnail_from_manifest() {
```

```suggestion
	public function test_run__get_kit_thumbnail_from_api() {
```

tests/phpunit/elementor/app/import-export/processes/test-import.php
```diff
@@ -475,6 +475,49 @@ public function test_get_manifest__kit_library() {
...
+		$module->import_kit( __DIR__ . '/../mock/kit-library-manifest-with-thumbnail.zip', $import_settings );
+		$import_sessions = get_option( Module::OPTION_KEY_ELEMENTOR_IMPORT_SESSIONS );
+		$this->assertEquals( 'manifest-thumbnail', $import_sessions[ $module->import->get_session_id() ]['kit_thumbnail'] );
+
+		// Case 2: no thumbnail & no kit id
+		$module->import_kit( __DIR__ . '/../mock/kit-library-manifest-only.zip', $import_settings );
+		$import_sessions = get_option( Module::OPTION_KEY_ELEMENTOR_IMPORT_SESSIONS );
+		$this->assertEmpty( $import_sessions[ $module->import->get_session_id() ]['kit_thumbnail'] );
+
+		// Case 3: no thumbnail -> request to API
```
StyleShit, line 495: test also a failing API request

tests/phpunit/elementor/app/import-export/processes/test-import.php
```diff
@@ -475,6 +475,49 @@ public function test_get_manifest__kit_library() {
...
+		$import_settings['id'] = '123';
+		$this->mock_get_remote_kit_by_id( [
+			'thumbnail' => 'api-thumbnail',
+		] );
```
StyleShit, line 499: remove the filter after the test

tests/phpunit/elementor/app/import-export/processes/test-import.php
```diff
@@ -475,6 +475,49 @@ public function test_get_manifest__kit_library() {
...
+				'headers' => [],
+				'response' => [
+					'code' => 200,
+					'message' => 'OK',
+				],
+				'cookies' => [],
+				'filename' => '',
+				'body' => wp_json_encode( $kit ),
+			];
+		}, 10, 3 );
```
StyleShit, line 518:
```suggestion
		} );
```

do you need that?

app/modules/import-export/processes/import.php
```diff
@@ -418,6 +428,32 @@ public function init_import_session( $save_instance_data = false ) {
...
+		if ( empty( $this->kit_id ) ) {
+			return '';
+		}
+
+		$api = new Kit_Library_Api();
+		$kit = $api->get_by_id( $this->kit_id );
+
+		if ( is_wp_error( $kit ) ) {
+			throw new WP_Error_Exception( $kit );
+		}
```
StyleShit, line 452: maybe instead of throwing an error, return an empty string? same as line 444

app/modules/import-export/module.php
StyleShit, line 174: maybe add a fallback to the kit name? (probably need to check with the product)

`kit_title ?? kit_name ?? "imported kit"`

and then extract to some reusable method (`extract_kit_title()` or something?)

app/modules/import-export/module.php
```diff
@@ -230,6 +231,35 @@ private function render_import_export_tab_content() {
...
+		<div class="tab-import-export-kit__kit-item-row">
+			<article class="tab-import-export-kit__kit-item">
+				<header>
+					<h3>
+						<?php echo esc_html( $last_imported_kit['kit_title'] ); ?>
+					</h3>
+				</header>
+				<img
+					src="<?php echo esc_url( $last_imported_kit['kit_thumbnail'] ); ?>"
+					alt="<?php echo esc_attr( $last_imported_kit['kit_title'] ); ?>"
```
StyleShit, line 256: and then you can use here the method I suggested above

app/modules/kit-library/assets/js/components/apply-kit-dialog.js
```diff
@@ -7,7 +7,8 @@ export default function ApplyKitDialog( props ) {
 
 	const startImportProcess = useCallback( ( applyAll = false ) => {
 		let url = '/import/process' +
-			`?file_url=${ encodeURIComponent( props.downloadLink ) }` +
+			`?id=${ props.id }` +
+			`&file_url=${ encodeURIComponent( props.downloadLink ) }` +
 			`&nonce=${ props.nonce }&referrer=kit-library`;
```
StyleShit, line 12: At this stage of the concatenation, it might be the time to start using the `URL()` object IMO, it's too big and the next dev can easily break it

app/modules/kit-library/assets/js/components/item-header.js
```diff
@@ -156,6 +156,7 @@ export default function ItemHeader( props ) {
 			}
 			{
 				downloadLinkData && <ApplyKitDialog
+					id={ props.model.id }
```
StyleShit, line 159: where did you add the `id` prop to the model?

app/modules/kit-library/connect/kit-library.php
```diff
@@ -20,6 +20,10 @@ public function get_all( $args = [] ) {
 		return $this->http_request( 'GET', 'kits/plugin-version/' . ELEMENTOR_VERSION, $args );
 	}
 
+	public function get_by_id( $id ) {
+		return $this->http_request( 'GET', 'kits/' . $id );
+	}
```
StyleShit, line 25: IDK if the rest of the class is tested, but maybe add a test 🤷

StyleShit (changes requested):

app/modules/kit-library/assets/js/components/item-header.js
ZeyadMahagna: The model already has an id

app/modules/kit-library/assets/js/components/apply-kit-dialog.js
ZeyadMahagna: This is a relative URL used after with the `navigate` not a full one, can't use the URL class here

tests/phpunit/elementor/app/import-export/processes/test-import.php
ZeyadMahagna: The `import_kit` method wraps the `run` and makes all the neccessary arrangments that need to be made before calling `run` like ensuring writing permissions and registering default runners which we would need to do anyway if we were to call run directly in the test

app/modules/import-export/module.php
ZeyadMahagna: These were the specifications of the task. The name is dashed and they don't want it to be used here and this decision should not be part of this PR anyway

app/modules/import-export/processes/import.php
```diff
@@ -17,6 +17,8 @@
 use Elementor\App\Modules\ImportExport\Runners\Import\Templates;
 use Elementor\App\Modules\ImportExport\Runners\Import\Wp_Content;
 use Elementor\App\Modules\ImportExport\Module;
+use Elementor\App\Modules\KitLibrary\Connect\Kit_Library as Kit_Library_Api;
+use Elementor\Data\V2\Base\Exceptions\WP_Error_Exception;
```
StyleShit, line 21:
```suggestion
```

app/modules/kit-library/assets/js/components/apply-kit-dialog.js
StyleShit: So you can use this:
https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

tests/phpunit/elementor/app/import-export/processes/test-import.php
```diff
@@ -475,6 +475,82 @@ public function test_get_manifest__kit_library() {
...
+		];
+		$module = new Module();
+
+		$module->import_kit( __DIR__ . '/../mock/kit-library-manifest-with-thumbnail.zip', $import_settings );
+
+		$import_sessions = get_option( Module::OPTION_KEY_ELEMENTOR_IMPORT_SESSIONS );
+		$this->assertEquals( 'manifest-thumbnail', $import_sessions[ $module->import->get_session_id() ]['kit_thumbnail'] );
+	}
+
+	public function init_import_session__get_empty_kit_thumbnail() {
```
StyleShit, line 490:
```suggestion
	public function test_init_import_session__get_empty_kit_thumbnail() {
```

also others

tests/phpunit/elementor/app/import-export/processes/test-import.php
StyleShit: I know it wraps the import, but how is this necessary in this test?
every other test in this file doesn't use the module
The behavior of the module should be tested inside the module test, not inside this unit test of the import process.

It should also simplify the tests (e.g. you'll be able to remove this: `get_option( Module::OPTION_KEY_ELEMENTOR_IMPORT_SESSIONS )`, which is an implementation detail of the module)

tests/phpunit/elementor/app/import-export/processes/test-import.php
```diff
@@ -475,6 +475,82 @@ public function test_get_manifest__kit_library() {
...
+		$import_settings = [
+			'referrer' => 'kit-library',
+		];
+		$module = new Module();
+
+		$module->import_kit( __DIR__ . '/../mock/kit-library-manifest-only.zip', $import_settings );
+
+		$import_sessions = get_option( Module::OPTION_KEY_ELEMENTOR_IMPORT_SESSIONS );
+		$this->assertEmpty( $import_sessions[ $module->import->get_session_id() ]['kit_thumbnail'] );
```
StyleShit, line 499: maybe assert an empty string, to make sure there is always a string
(`assertEmpty` will pass for nullish values)

tests/phpunit/elementor/app/import-export/processes/test-import.php
```diff
@@ -475,6 +475,82 @@ public function test_get_manifest__kit_library() {
...
+			'filename' => '',
+			'body' => wp_json_encode( [
+				'thumbnail' => 'api-thumbnail'
+			] ),
+		];
+	}
+
+	private function unmock_get_remote() {
+		remove_filter( 'pre_http_request', [ $this, 'get_kit' ] );
+	}
```
StyleShit, line 552: you can simplify it to:

```PHP
public function mock_get_kit() {
	$filter = function() {
		return [
			'headers' => [],
			'response' => [
				'code' => 200,
				'message' => 'OK',
			],
			'cookies' => [],
			'filename' => '',
			'body' => wp_json_encode( [
				'thumbnail' => 'api-thumbnail'
			] ),
		];
	};

	add_filter( 'pre_http_request',  $filter );
	
	return function() use( $filter ) {
		remove_filter( 'pre_http_request', $filter );
	}
}
```

and then:

```PHP
function test_asd() {
	$cleanup = $this->mock_get_kit();

	// ...

	$cleanup();
}
```

tests/phpunit/elementor/app/import-export/processes/test-import.php
```diff
@@ -475,6 +475,82 @@ public function test_get_manifest__kit_library() {
...
+			'referrer' => 'kit-library',
+			'id' => '123',
+		];
+		$module = new Module();
+
+		$module->import_kit( __DIR__ . '/../mock/kit-library-manifest-only.zip', $import_settings );
+
+		$import_sessions = get_option( Module::OPTION_KEY_ELEMENTOR_IMPORT_SESSIONS );
+		$this->assertEmpty( $import_sessions[ $module->import->get_session_id() ]['kit_thumbnail'] );
+	}
```
StyleShit, line 529: you don't mock the API request, which (probably?) means that the test will make an actual request. you need to mock it

tests/phpunit/elementor/app/import-export/processes/test-import.php
StyleShit: where are the changes?

tests/phpunit/elementor/app/import-export/processes/test-import.php
```diff
@@ -475,6 +475,95 @@ public function test_get_manifest__kit_library() {
...
+		$import = new Import( __DIR__ . '/../mock/kit-library-manifest-only.zip', $import_settings );
+		$import->init_import_session();
+
+		$import_sessions = get_option( 'elementor_import_sessions' );
+		$this->assertEquals( '', $import_sessions[ $import->get_session_id() ]['kit_thumbnail'] );
+
+		$cleanup();
+	}
+
+	public function mock_get_kit() {
```
StyleShit, line 530:
```suggestion
	private function mock_get_kit() {
```

tests/phpunit/elementor/app/import-export/processes/test-import.php
```diff
@@ -475,6 +475,95 @@ public function test_get_manifest__kit_library() {
...
+		};
+
+		add_filter( 'pre_http_request',  $filter );
+
+		return function() use( $filter ) {
+			remove_filter( 'pre_http_request', $filter );
+		};
+	}
+
+	public function mock_get_kit_wp_error() {
```
StyleShit, line 553:
```suggestion
	private function mock_get_kit_wp_error() {
```

StyleShit (changes requested):

StyleShit (dismissed):

StyleShit (approved):

PR #21118, opened 2023-01-24
Internal: Add a vertical alignment control to icons in the Icon List widget [ED-9265]

lukecarbis:


includes/widgets/icon-list.php
```diff
@@ -459,7 +459,7 @@ protected function register_controls() {
 		$this->add_responsive_control(
 			'icon_self_align',
 			[
-				'label' => esc_html__( 'Alignment', 'elementor' ),
+				'label' => esc_html__( 'Horizontal Alignment', 'elementor' ),
```
TzviRabinovitch, line 462: duplicate name ?

includes/widgets/icon-list.php
```diff
@@ -487,6 +487,62 @@ protected function register_controls() {
...
+					'unit' => 'px',
+					'size' => 0,
+				],
+				'range' => [
+					'px' => [
+						'min' => -15,
+						'max' => 15,
+					],
+				],
+				'selectors' => [
```
TzviRabinovitch, line 540: condition ?
If the user chose ‘center’ (or left the default ‘center’ option on) the slider should not be available

includes/widgets/icon-list.php
```diff
@@ -487,6 +487,62 @@ protected function register_controls() {
...
+					'middle' => [
+						'title' => esc_html__( 'Center', 'elementor' ),
+						'icon' => 'eicon-v-align-middle',
+					],
+					'bottom' => [
+						'title' => esc_html__( 'End', 'elementor' ),
+						'icon' => 'eicon-v-align-bottom',
+					],
+				],
+				'default' => '',
```
TzviRabinovitch, line 509: default = center ?

TzviRabinovitch (changes requested):

includes/widgets/icon-list.php
lukecarbis: The default is intentionally blank, which matches all other alignment controls.

The result is four possible values:

- `'top'`: This styles the control alignment explicitly as `flex-start`
- `'bottom'`: This styles the control alignment explicitly as `flex-end`
- `'center'`: This styles the control alignment explicitly as `center`
- `''`: This applies no explicit styling, allowing the CSS to fall back to `center`

I believe that the other alignment controls are intentionally implemented this way so that themes can override the default list styling, but also accept the set alignment property when it's explicitly set.

includes/widgets/icon-list.php
lukecarbis: This is implemented on line 526—528, however I will remove this as per the updated requirements.

lukecarbis: Done in 2eddc0bd0.

assets/dev/scss/frontend/widgets/icon-list.scss
```diff
@@ -68,6 +70,7 @@
 
 	.elementor-icon-list-icon {
 		display: flex;
+		position: relative;
```
louiswol94, line 73: Why?

includes/widgets/icon-list.php
louiswol94: @TzviRabinovitch @lukecarbis What about tests?

louiswol94 (changes requested):

assets/dev/scss/frontend/widgets/icon-list.scss
lukecarbis: This is required for the `top` property to work when we adjust the vertical offset position. Using margin / negative margin doesn't work with `flex-start` or `flex-end` positions.

includes/widgets/icon-list.php
```diff
@@ -487,6 +487,64 @@ protected function register_controls() {
...
+						'max' => 15,
+					],
+					'em' => [
+						'min' => -1,
+						'max' => 1,
+						'step' => 0.1,
+					],
+				],
+				'selectors' => [
+					'{{WRAPPER}} .elementor-icon-list-icon' => 'top: {{SIZE}}{{UNIT}};',
```
louiswol94, line 543: @lukecarbis Can you add a CSS variable for this as well?

Something like:

`'{{WRAPPER}}' => '--icon-vertical-position: {{SIZE}}{{UNIT}};',`

```
.elementor-icon-list-icon {
    top: var( --icon-vertical-position, 0px );
}
```

louiswol94 (approved):

Issue #19993, opened 2022-10-05
🐞 Bug Report: Button widget adds extra space for icon even if no icon is selected - [ED-9569]


Michaelgimii:
### Prerequisites

- [X] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [X] The issue still exists against the latest stable version of Elementor.

### Description

If you add a button widget, Inspect in the editor will show the added code. tags related to the icon while no icon is selected.

![a](https://user-images.githubusercontent.com/88034902/194134317-744ef69c-fdd9-49af-a37f-4c3b93904886.png)

Also, if you pay attention to the button widget in the editor, it is clear that the space on the left side of the button is more than the right side.

![b](https://user-images.githubusercontent.com/88034902/194135335-2b25c2ee-0d8d-4cbd-b059-b36dc9eba23f.png)

In fact, the reason for this is the presence of the same additional code that can be seen in the browser's inspect.

Finally, I had a question about how to code Elementor...

Is there a specific reason why Elementor generates 4 lines of code for a button widget, when it could have been a very clean code?

Please see the code below. Elementor has done it in 4 lines.

![code](https://user-images.githubusercontent.com/88034902/194137817-a74f41ed-38ce-4427-984c-2ca74deed11e.png)

### Steps to reproduce

-

### Isolating the problem

- [X] This bug happens with only Elementor plugin active (and Elementor Pro).
- [X] This bug happens with a Blank WordPress theme active ([Hello theme](https://wordpress.org/themes/hello-elementor/)).
- [X] I can reproduce this bug consistently using the steps above.

### System Info

-

nicholaszein:
Hello @Michaelgimii,

Thank you for reaching out! 🙏 

#### About the rendering of the additional HTML tags for the button icon:
I've tested your issue with both Elementor v3.7 and v3.8-Beta, with and without the _Inline SVG Icons_ experiment turned on, and I could not reproduce it. This might be happening to you because of either or a combination of these issues:

* You added an SVG icon, but the file has since been deleted;
* Cache.

Most likely you're having a cache issue. When the Icon is not selected, the `.elementor-button-icon` is not rendered in the final HTML.

#### About your second question:
In short, yes. Having inner tags allows Elementor _(and you)_ to manipulate the elements inside the button widget with more precision and flexibility. Especially when you add inline SVGs.

I hope this clarifies your questions.

Kind regards

PR #21185, opened 2023-01-29
Fix: Button Widget - Inline editing and icon space [ED-9569]

matipojo:
Fix: https://github.com/elementor/elementor/issues/19993


Joliamendi:
An option to control the icon size in the button widget would be great. This is a missing feature. Please don't forget it.
(I apologize for mentioning it here)

StyleShit (approved):

PR #21191, opened 2023-01-30
Internal: Add support for disabling lazyload at the document level [ED-8993]

maork-elementor:


core/files/css/base.php
```diff
@@ -327,7 +327,7 @@ public function add_control_rules( array $control, array $controls_stack, callab
 
 		$stylesheet = $this->get_stylesheet();
```
rami-elementor, line 329:
```suggestion

		/**
		 * CSS file control.
		 *
		 * Filters the control used to to generate the CSS file.
		 *
		 * @since 3.12.0
		 *
		 * @param array $control The controls.
		 * @param array $value The value.
		 * @param Elementor\Core\Files\Base $this CSS file instance.
		 */
```

yotamselementor (approved):

PR #21213, opened 2023-02-01
Tweak: Update obscure error messages [ED-9591]

rami-elementor:


arielk (commented):
@rami-elementor This is an internal error message for developers, I dont think we need to translate it

rami-elementor:
@arielk We are already translating the message.

rami-elementor:
In any case, we can remove the i18n function, but the text change will remain.

PR #21381, opened 2023-02-17
Fix: Content styling controls are not targeting the right container in Tabs widget [ED-9752]

StuartElementor:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #


modules/nested-tabs/assets/scss/frontend.scss
```diff
@@ -229,25 +229,33 @@
 		}
 	}
 
-	.e-con,
 	.e-n-tabs-content {
-		border: 1px none var( --n-tabs-border-color );
-		border-width: var( --n-tabs-border-width );
+
+		:where( & > .e-con ) {
```
hein-obox, line 234: Why are we using `:where()`. Is that for specificity? Is that necessary for the Nested tabs? 

I would consider just using `> e.con` instead. 

What are your thoughts?

modules/nested-tabs/assets/scss/frontend.scss
```diff
@@ -229,25 +229,33 @@
...
+			border-radius: var(--border-radius);
+			padding-right: var(--padding-right);
+			padding-left: var(--padding-left);
```
hein-obox, line 237: We had a discussion with Udi about this whether padding-left/right sould be on `e-con` or `e-con-inner` level. 

I am curious to hear your opinion. 

I am not 100% sure what would be best.

modules/nested-tabs/widgets/nested-tabs.php
```diff
@@ -81,7 +81,7 @@ protected function register_controls() {
 		$tooltip_start = is_rtl() ? esc_html__( 'Right', 'elementor' ) : esc_html__( 'Left', 'elementor' );
 		$tooltip_end = is_rtl() ? esc_html__( 'Left', 'elementor' ) : esc_html__( 'Right', 'elementor' );
 		$nested_tabs_heading_selector_class = ':is( {{WRAPPER}} > .elementor-widget-container > .e-n-tabs > .e-n-tabs-heading, {{WRAPPER}} > .elementor-widget-container > .e-n-tabs > .e-n-tabs-content )';
-		$nested_tabs_content_selector_class = '{{WRAPPER}} > .elementor-widget-container > .e-n-tabs > .e-n-tabs-content';
+		$nested_tabs_content_selector_class = ':where( {{WRAPPER}} > .elementor-widget-container > .e-n-tabs > .e-n-tabs-content ) > .e-con';
```
hein-obox, line 85: Do we need `where()` for this? 
Just curious.

modules/nested-tabs/widgets/nested-tabs.php
```diff
@@ -949,7 +949,7 @@ protected function register_controls() {
 				'name' => 'box_background_color',
 				'types' => [ 'classic', 'gradient' ],
 				'exclude' => [ 'image' ],
-				'selector' => "{$nested_tabs_content_selector_class}",
+				'selector' => $nested_tabs_content_selector_class,
```
hein-obox, line 952: I know we have used this `{$..}` throughout the whole php file. I am thinking can you check if we can remove it everywhere, for confirmity sake?

modules/nested-tabs/assets/scss/frontend.scss
StuartElementor: using the `:where()` in all the places where we set styling for the `.e-con` are necessary so that when someone chooses to style the drop down Container itself, that the long chain of specific selectors in the Widget here don't override the selector of the Container.

and the same goes for the `'selector' => '...'` in the Controls.

This was the solve for the problem in Mega Menu where the Widget Container controls were overriding any custom dropdown Container controls. and we decided to apply the same logic to the Tabs.

modules/nested-tabs/widgets/nested-tabs.php
StuartElementor: It does seem redundant in this case, where there is nothing else being processed inside the double quotes.
I will do as you said and change it in the whole file.

modules/nested-tabs/assets/scss/frontend.scss
StuartElementor: This got to me thinking. I don't actually have to set any of these `.e-con` styles in our css.
This all gets handled by the Container css. All I need to do is stick to setting the vars and the CSS from the controls.

I've removed all that related css. And I will need to open a task on Monday to do the same the Mega Menu.
It's unnecessarily redundant code.

modules/nested-tabs/widgets/nested-tabs.php
StuartElementor: I've checked and there are no more unnecessary occurrences :)

hein-obox (approved):

PR #21393, opened 2023-02-19
Internal: Editor V2 - Breakpoints selector auto scale [ED-9829]

StyleShit:


assets/dev/js/editor/editor-base.js
```diff
@@ -648,6 +648,12 @@ export default class EditorBase extends Marionette.Application {
 			.trigger( 'resize' );
```
Nevoss, line 648: In the PHP staff we are still loading old code in the v2 config

Nevoss (changes requested):

PR #21443, opened 2023-02-23
Fix: Containers are still editable in preview mode [ED-8696]

ArmanPurtoian:


TzviRabinovitch (dismissed):

assets/dev/scss/editor/preview/_container.scss
```diff
@@ -1,4 +1,14 @@
 // Container in preview
+.elementor-editor-preview {
```
udidol, line 2: This will stop the editing handles from doing anything, but they should not exist in preview mode.
The resize behavior is created using a Marionette `Behavior` class extension attached to the Container `View` instance.

See here: https://github.com/elementor/elementor/blob/55d0e27ecb6bf49c823f77c1e005d8ba392a0bac/assets/dev/js/editor/elements/views/container.js#L62

See the behavior class here:
https://github.com/elementor/elementor/blob/master/assets/dev/js/editor/elements/views/behaviors/widget-resizeable.js

Read more about Marionette Behaviors here:
https://marionettejs.com/docs/v2.4.5/marionette.behavior.html

If I had to guess, the problem is here:
https://github.com/elementor/elementor/blob/55d0e27ecb6bf49c823f77c1e005d8ba392a0bac/assets/dev/js/editor/elements/views/behaviors/widget-resizeable.js#L58

I would ask Evyatar (@StyleShit) if he remembers why he did it.
The fix should be somewhere there.

udidol (changes requested):

louiswol94 (changes requested):
@ArmanPurtoian Can you add a PlayWright test?

louiswol94 (approved):

udidol (approved):

PR #21444, opened 2023-02-23
Internal: Editor V2 - Documents UI package [ED-9866]

StyleShit:


core/editor/config-providers/editor-v2-config-provider.php
```diff
@@ -21,6 +21,7 @@ class Editor_V2_Config_Provider implements Config_Provider_Interface {
 	];
 
 	const UTIL_PACKAGES = [
+		'documents',
```
Nevoss, line 24: I think this is still an extension, it cannot live outside of the editor like `store` or `locations`

Nevoss (changes requested):

Nevoss (dismissed):

rotemee (approved):

PR #21468, opened 2023-02-25
Tweak: Delete `aspect-ratio` SASS [ED-9875]

rami-elementor:


arielk:
@rami-elementor please replace the current usage in the lightbox with the new method before we remove this one

PR #21521, opened 2023-03-02
Tweak: Modify cli commands to support cloud site creation [ED-9953]

lenalElementor:


modules/wp-cli/library.php
```diff
@@ -4,6 +4,7 @@
 use Elementor\Api;
 use Elementor\Plugin;
 use Elementor\TemplateLibrary\Source_Local;
+use ElementorPro\Modules\ThemeBuilder\Classes\Conditions_Cache;
```
nivnoimanelementor, line 7: you can't use elementor pro here..
you inside elementor core ( they don't know each other )

nivnoimanelementor (changes requested):
please check the review

modules/wp-cli/library.php
lenalElementor: Moving this cli command to pro

modules/wp-cli/library.php
```diff
@@ -66,6 +66,10 @@ public function sync( $args, $assoc_args ) {
 	/**
 	 * Import template files to the Library.
 	 *
+	 *  [--returnType]
```
matipojo, line 69: Please add an example after line 75

modules/wp-cli/library.php
```diff
@@ -93,9 +99,17 @@ public function import( $args ) {
...
 
-		\WP_CLI::success( count( $imported_items ) . ' item(s) has been imported.' );
-	}
+		foreach ( $imported_items as $item ) {
+			$imported_items_ids[] = $item['template_id'];
+		}
+		$imported_items_ids = implode( ',', $imported_items_ids );
 
+		if ( 'ids' === $return_type ) {
+			\WP_CLI::line( $imported_items_ids );
```
matipojo, line 108:
```suggestion
			\WP_CLI::success( $imported_items_ids );
```

modules/wp-cli/library.php
```diff
@@ -77,12 +81,14 @@ public function sync( $args, $assoc_args ) {
...
 	 */
-	public function import( $args ) {
+	public function import( $args, $assoc_args ) {
 		if ( empty( $args[0] ) ) {
 			\WP_CLI::error( 'Please set file path.' );
 		}
 
 		$file = $args[0];
+		$imported_items_ids = [];
+		$return_type = \WP_CLI\Utils\get_flag_value( $assoc_args, 'returnType', 'info' );
```
matipojo, line 91: Can you get it directly from `assoc_args`?

matipojo (changes requested):

modules/wp-cli/library.php
lenalElementor: I can do it with `$assoc_args['returnType']`, but it was the quickest way the get the value and set a default value if not exist.

modules/wp-cli/library.php
lenalElementor: I'm sending the ids here, with the "Success: " in the response string it will be difficult to fetch those ids.

modules/wp-cli/library.php
```diff
@@ -66,36 +66,71 @@ public function sync( $args, $assoc_args ) {
...
 
 		/** @var Source_Local $source */
 		$source = Plugin::$instance->templates_manager->get_source( 'local' );
 
+		if ( 'url' === $file_protocol ) {
+			if ( false === filter_var( $file, FILTER_VALIDATE_URL ) ) {
+				\WP_CLI::error( "Invalid file URL" );
+			}
+
+			$file_path = download_url( $file );
```
nivnoimanelementor, line 110: Please remove this tmp file at the end

nivnoimanelementor (changes requested):
Please check request changes

modules/wp-cli/library.php
```diff
@@ -66,36 +66,76 @@ public function sync( $args, $assoc_args ) {
 	/**
 	 * Import template files to the Library.
 	 *
+	 *  [--returnType]
+	 *      Forms of output. Possible values are 'ids', 'info'.
+	 *      if this parameter won't be specified, the import info will be output.
+	 *
+	 *  [--fileProtocol]
+	 *      Forms of file source. Possible values are 'path', 'url'.
+	 *      if this parameter won't be specified, path will be the default.
```
matipojo, line 75: Why not check if it starts with `http`?

matipojo (changes requested):

modules/wp-cli/library.php
lenalElementor: We want to keep the old logic, and separate it from the new one (the url).

nivnoimanelementor (dismissed):

core/experiments/wp-cli.php
```diff
@@ -105,6 +105,27 @@ public function deactivate( $args, $assoc_args ) {
...
+	 * Experiment Status
+	 *
+	 * ## EXAMPLES
+	 *
+	 * 1. wp elementor experiments status container
+	 *      - This will return the status of Container experiment. (active/inactive)
+	 *
+	 * @param array $args
+	 */
+	public function status( $args ) {
```
matipojo, line 118: Please cover with tests

matipojo (changes requested):

core/experiments/wp-cli.php
nivnoimanelementor: The core team need to update the feature manager before. ( approved by Yotam )

nivnoimanelementor (dismissed):

nivnoimanelementor (approved):

PR #21568, opened 2023-03-06
Tweak: Suggest better behavior for Nested Tabs opening on the accordion layout [ED-9927]

hein-obox:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #


modules/nested-tabs/assets/scss/frontend.scss
```diff
@@ -56,9 +52,6 @@
 		flex-direction: var( --n-tabs-direction );
 		gap: var( --n-tabs-gap );
 		text-align: $start;
-		height: var( --n-tabs-height );
-		overflow-x: var( --n-tabs-overflow-x );
-		overflow-y: var( --n-tabs-overflow-y );
```
hein-obox, line 61: Remove unnecessary scrolling.
Originally we had implemented a `max-height` control for this widget, but we removed it during the development process.

modules/nested-tabs/assets/scss/frontend.scss
```diff
@@ -243,7 +236,9 @@
 		}
 
 		> .e-con:not(.e-active) {
-			display: none;
```
StuartElementor, line 246: Without the display none, see what I'm getting
image

- Set Border on the Content Containers
- Set Margin on the Content Containers

Test using your fix, with the `display: none;` that might solve the porblem.

StuartElementor: Also to think about:
If you set padding to 0, and margin to 0, and there is a transition set on the Content Container, then when the `.e-active` is removed, it will animate from 0 to 20px for the set transition time. So we could even take of the `padding: 0` if the display none works.

StuartElementor: Also,
Seeing as this is not an issue on Desktop, maybe we should nest it in the `.e-collapse`, so this fix only effect responsive.

hein-obox: > Without the display none, see what I'm getting image
> 
> * Set Border on the Content Containers
> * Set Margin on the Content Containers
> 
> Test using your fix, with the `display: none;` that might solve the porblem.

@StuartElementor 

`display: none` is causing problems. 

I can fix the border and margin issues.

hein-obox: > 

Will do.

hein-obox: @StuartElementor 

I have assigned my styling to the accordion version only. 
And I have removed all styling from not active containers.

modules/nested-tabs/widgets/nested-tabs.php
```diff
@@ -81,7 +81,7 @@ protected function register_controls() {
 		$tooltip_start = is_rtl() ? esc_html__( 'Right', 'elementor' ) : esc_html__( 'Left', 'elementor' );
 		$tooltip_end = is_rtl() ? esc_html__( 'Left', 'elementor' ) : esc_html__( 'Right', 'elementor' );
 		$nested_tabs_heading_selector_class = ':is( {{WRAPPER}} > .elementor-widget-container > .e-n-tabs > .e-n-tabs-heading, {{WRAPPER}} > .elementor-widget-container > .e-n-tabs > .e-n-tabs-content )';
-		$nested_tabs_content_selector_class = ':where( {{WRAPPER}} > .elementor-widget-container > .e-n-tabs > .e-n-tabs-content ) > .e-con';
+		$nested_tabs_content_selector_class = ':where( {{WRAPPER}} > .elementor-widget-container > .e-n-tabs > .e-n-tabs-content ) > .e-con.e-active';
```
StuartElementor, line 84: I think that using the `.e-con.e-active` as the selector to target for the many styling controls will have more adverse effect than helpful effects.
If a css `transition: 1s;` (for example) is added for any reason (by a theme, by a future control, by a user, by a plugin that adds transision features to the container widget, etc), then then what you will immediately see  when the tab is shown you will also for example see the `border-color` transition from the default `transparent, black, white` (whatever it is) to your chosen color, and the same for any styled properties `border-width`, `padding`, even `background-color`, and etc. So I this selector should stay as it is.

You and I have spoekn about this in a huddle, I've noted my feedback here, and you already have another JS option, which I like.

We can invlove Udi in the AM and chose the best option.

Nice work.

PR #21574, opened 2023-03-07
Internal: Editor V2 - Add `import/no-cycle` eslint rule [ED-9997]

StyleShit:


packages/.eslintrc.js
```diff
@@ -36,6 +36,9 @@ module.exports = {
 			},
 		],
 
+		// Prevent circular dependencies.
```
Nevoss, line 39:
```suggestion
		// Prevent circular dependencies. (if the lint is slow consider removing it)
```

Nevoss (changes requested):

packages/packages/documents/src/types.ts
```diff
@@ -1,4 +1,4 @@
-import { createSlice } from './store';
+import type { createSlice } from './store';
```
Nevoss, line 1: ?

Nevoss (changes requested):

packages/packages/documents/src/types.ts
StyleShit: Changed

Nevoss (approved):

PR #21603, opened 2023-03-09
Fix: Lazy Load - Disable 'Dominant Color' on PNG and GIF images [ED-10001]

maork-elementor:


modules/lazyload/module.php
```diff
@@ -64,21 +65,46 @@ private function update_element_attributes( Element_Base $element ) {
...
+							$control = $this->apply_dominant_color_background( $control, $value, $selector );
+						}
 					}
 				}
 			}
 		}
 		return $control;
 	}
 
+	private function is_dominant_color_enabled() {
```
Nevoss, line 87: It feels like the code should not be inside the lazy load module, maybe it should not be in Elementor at all and we should provide a hook to the `PERFORMANCE_LAB` team, and they will add this code.

Nevoss (changes requested):

modules/lazyload/module.php
maork-elementor: PERFORMANCE_LAB does not know about our CSS or background images, we have our own post.css for each page or post. Our controls and CSS are not parsed by them.

We use an internal CSS mechanism to create the background image & all the page css. Previously, we had a problem with this plugin and our CSS, so I created a module for it.

https://github.com/elementor/elementor/blob/ef19421b7637f93d427e1691427e8e87304015a7/modules/performance-lab/module.php

StyleShit (changes requested):

modules/lazyload/module.php
```diff
@@ -64,21 +65,46 @@ private function update_element_attributes( Element_Base $element ) {
...
+		$performance_lab_settings = (array) get_option( self::PERFORMANCE_LAB_OPTION_NAME, [] );
```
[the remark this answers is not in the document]
maork-elementor: Querying the database for the relevant option is the only way to find out if the option is enabled. As performance_lab is also doing this by themselves.

There is no need to check if the plugin is activated, or if any event files exist by trying to use some namespace. This will only indicate that the plugin exists in general, but not the value of the option.

Could you think of a way to get the option value without using the DB?

modules/lazyload/module.php
```diff
@@ -64,21 +65,46 @@ private function update_element_attributes( Element_Base $element ) {
...
+		return Utils::get_array_value_by_keys( $performance_lab_settings, [ 'images/dominant-color', 'enabled' ] );
+	}
+
+	private function image_extension_validation( $value ) {
+		// Disable 'Dominant Color' on PNG and GIF images
+		$url = $value['url'];
+		$ignored_extensions = [ 'png', 'gif' ];
```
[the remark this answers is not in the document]
maork-elementor: 1. PNG, GIF, TIFF, TGA, BMP, and SVG are all formats with transparency support.
    Would you recommend adding all of them? I added the most popular formats with transparency support.  
2. Dominant Color Support - Set as background color, the user will see the Dominant Color of the background image 
    until the image loads to the browser. In response to your question, the user will see a white background by default. 
    
Performance labs did not change anything regarding this issue; it was the behavior from the beginning of the experiment. I discovered this bug while working on another task recently.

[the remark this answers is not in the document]
maork-elementor: According to PERFORMANCE_LAB, each attachment is set with a transparency metadata. Currently, I use it to determine whether an image has transparency.

modules/lazyload/module.php
```diff
@@ -64,26 +64,45 @@ private function update_element_attributes( Element_Base $element ) {
...
 		$metadata = wp_get_attachment_metadata( $value['id'] );
-		$dominant_color = Utils::get_array_value_by_keys( $metadata, [ 'dominant_color' ] );
-		if ( $dominant_color ) {
+
+		// Performance Lab adds these metadata
+		$has_transparency = $metadata['has_transparency'] ?? false;
+		$dominant_color = $metadata['dominant_color'] ?? false;
+
+		if ( $dominant_color && ! $has_transparency ) {
 			$control['selectors'][ $selector ] .= "background-color: #{$dominant_color};";
```
StyleShit, line 103:
```suggestion
			$control['selectors'][ $selector ] .= "background-color: #{esc_attr( $dominant_color )};";
```

modules/lazyload/module.php
```diff
@@ -64,26 +64,45 @@ private function update_element_attributes( Element_Base $element ) {
...
+	}
+
 	private function apply_dominant_color_background( $control, $value, $selector ) {
 		$metadata = wp_get_attachment_metadata( $value['id'] );
-		$dominant_color = Utils::get_array_value_by_keys( $metadata, [ 'dominant_color' ] );
-		if ( $dominant_color ) {
+
+		// Performance Lab adds these metadata
+		$has_transparency = $metadata['has_transparency'] ?? false;
+		$dominant_color = $metadata['dominant_color'] ?? false;
```
StyleShit, line 100: validate with doron

StyleShit (changes requested):

modules/lazyload/module.php
maork-elementor: Done

modules/lazyload/module.php
```diff
@@ -64,26 +64,45 @@ private function update_element_attributes( Element_Base $element ) {
...
 	}
 
+	private function is_dominant_color_enabled() {
+		if ( function_exists( 'perflab_get_active_modules' ) && function_exists( 'perflab_is_valid_module' ) ) {
+			$active_and_valid_modules = array_filter( perflab_get_active_modules(), 'perflab_is_valid_module' );
+			$has_dominant_color_module = in_array( 'images/dominant-color', $active_and_valid_modules );
+			return $has_dominant_color_module;
+		}
+		return false;
+	}
```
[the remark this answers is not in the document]
maork-elementor: Done

modules/lazyload/module.php
```diff
@@ -64,26 +64,45 @@ private function update_element_attributes( Element_Base $element ) {
...
 		return $control;
 	}
 
+	private function is_dominant_color_enabled() {
+		if ( ! function_exists( 'perflab_get_active_modules' ) ) {
+			return false;
+		}
+
+		$active_modules = perflab_get_active_modules();
+		return in_array( 'images/dominant-color', $active_modules, true );
```
StyleShit, line 92:
```suggestion
		$active_modules = perflab_get_active_modules();

		return in_array( 'images/dominant-color', $active_modules, true );
```

StyleShit (approved):

modules/lazyload/module.php
```diff
@@ -64,26 +64,46 @@ private function update_element_attributes( Element_Base $element ) {
...
-		$dominant_color = Utils::get_array_value_by_keys( $metadata, [ 'dominant_color' ] );
-		if ( $dominant_color ) {
-			$control['selectors'][ $selector ] .= "background-color: #{$dominant_color};";
+
+		// Performance Lab adds these metadata
+		$has_transparency = $metadata['has_transparency'] ?? false;
+		$dominant_color = esc_attr( $metadata['dominant_color'] ?? false );
+
+		if ( $dominant_color && ! $has_transparency ) {
+			$control['selectors'][ $selector ] .= "background-color: #{$dominant_color} ;";
```
StyleShit, line 104:
```suggestion
			$control['selectors'][ $selector ] .= "background-color: #{$dominant_color};";
```

StyleShit (changes requested):

StyleShit (dismissed):

modules/lazyload/module.php
```diff
@@ -64,26 +64,46 @@ private function update_element_attributes( Element_Base $element ) {
...
+		if ( ! function_exists( 'perflab_get_active_modules' ) ) {
+			return false;
+		}
+
+		$active_modules = perflab_get_active_modules();
+
+		return in_array( 'images/dominant-color', $active_modules, true );
+	}
+
 	private function apply_dominant_color_background( $control, $value, $selector ) {
```
Nevoss, line 96:
```suggestion
	private function get_dominant_color( $attachment_id ) {
```

modules/lazyload/module.php
```diff
@@ -64,26 +64,46 @@ private function update_element_attributes( Element_Base $element ) {
...
 		$metadata = wp_get_attachment_metadata( $value['id'] );
-		$dominant_color = Utils::get_array_value_by_keys( $metadata, [ 'dominant_color' ] );
-		if ( $dominant_color ) {
+
+		// Performance Lab adds these metadata
+		$has_transparency = $metadata['has_transparency'] ?? false;
+		$dominant_color = esc_attr( $metadata['dominant_color'] ?? false );
+
+		if ( $dominant_color && ! $has_transparency ) {
 			$control['selectors'][ $selector ] .= "background-color: #{$dominant_color};";
```
Nevoss, line 104:
```suggestion
			retrun $dominant_color;
```

modules/lazyload/module.php
```diff
@@ -64,26 +64,46 @@ private function update_element_attributes( Element_Base $element ) {
...
 		}
+
 		return $control;
```
Nevoss, line 107:
```suggestion
		return null;
```

modules/lazyload/module.php
```diff
@@ -64,26 +64,46 @@ private function update_element_attributes( Element_Base $element ) {
...
 			foreach ( $control['selectors'] as $selector => $css_property ) {
 				if ( 0 === strpos( $css_property, 'background-image' ) ) {
 					if ( ! empty( $value['url'] ) ) {
 						$css_property  = str_replace( 'url("{{URL}}")', 'var(--e-bg-lazyload-loaded)', $css_property );
 						$control['selectors'][ $selector ] = $css_property . '--e-bg-lazyload: url("' . $value['url'] . '");';
-						$control = $this->apply_dominant_color_background( $control, $value, $selector );
+
+						if ( $is_dominant_color_enabled ) {
+							$control = $this->apply_dominant_color_background( $control, $value, $selector );
+						}
```
Nevoss, line 78:
```suggestion
						$color = $is_dominant_color_enabled ? $this->get_dominant_color( $value['id'] ) : null;
						 
						if ( $color ) {
							$control['selectors'][ $selector ] .= "background-color: #{$color};";
						}
```

Nevoss (changes requested):

modules/lazyload/module.php
maork-elementor: Done

modules/lazyload/module.php
```diff
@@ -64,26 +64,46 @@ private function update_element_attributes( Element_Base $element ) {
...
+			return false;
+		}
+
+		$active_modules = perflab_get_active_modules();
+
+		return in_array( 'images/dominant-color', $active_modules, true );
+	}
+
 	private function apply_dominant_color_background( $control, $value, $selector ) {
 		$metadata = wp_get_attachment_metadata( $value['id'] );
```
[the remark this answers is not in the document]
maork-elementor: Done

modules/lazyload/module.php
maork-elementor: Done

modules/lazyload/module.php
maork-elementor: Done

modules/lazyload/module.php
maork-elementor: Done

Nevoss (dismissed):

Nevoss (approved):

PR #21616, opened 2023-03-10
Fix: Latest Chrome release breaks sticky effects [ED-10063]

louiswol94:


udidol (commented):
Test?
You can load a frontend/editor page and evaluate the device mode. Before the fix it should be `widescreen`, after the fix it should be `desktop`.

To verify try running the test on `develop` then on your branch. Develop should fail and yours should pass.

louiswol94:
@udidol Test added.

udidol (approved):

tests/playwright/sanity/device-mode.test.js
```diff
@@ -0,0 +1,22 @@
...
+		await editor.addWidget( 'heading', container );
+		await editor.publishAndViewPage();
+		await page.waitForSelector( '.e-con' );
+
+		const deviceMode = await page.evaluate( () => {
+			return elementorFrontend.getCurrentDeviceMode();
+		} );
+
+		const isInDesktopMode = 'desktop' === deviceMode;
+
```
hein-obox, line 19: I believe it is good to test all device modes, especially for all the additional breakpoints. 

On frontend, see my code for the nav menu for reference: https://github.com/elementor/elementor-pro/blob/99e6e591e1a19e843152bd2181d2a3b900d0c095/tests/playwright/sanity/navMenu.test.js#L193. 

If you would like to test it in the editor, you can check the Nested Tabs tests.

hein-obox (approved):

PR #21646, opened 2023-03-13
Internal: Fix onboarding dark mode issues

arielk:


core/common/app.php
```diff
@@ -245,6 +245,11 @@ private function enqueue_theme_ui_styles() {
 			}
 		}
 
+		if ( in_array( current_action(), [ 'admin_enqueue_scripts' ], true ) ) {
```
KingYes, line 248:
```suggestion
		if ( 'admin_enqueue_scripts' === current_action() ) {
```

KingYes (approved):

PR #21692, opened 2023-03-19
Internal: Editor V2 - Change tab name to "Widgets" [ED-10070]

StyleShit:


rotemee (dismissed):

core/editor/assets/scss/editor-v2-overrides.scss
```diff
@@ -28,3 +28,7 @@ body.elementor-navigator-docked #elementor-navigator {
 	border-bottom: var( --e-a-border );
 	height: 48px;
 }
+
+#elementor-panel-elements-navigation .elementor-panel-navigation-tab {
+	text-transform: none;
+}
```
Nevoss, line 34: I think it should be a change in the normal design and not only in V2

Nevoss (changes requested):

Nevoss (approved):

PR #21851, opened 2023-03-30
Fix: Receiving Error 500 when downgraded version [ED-10164]

louiswol94:
Error message: Uncaught Error: Class 'Elementor\Control_Gaps' not found


includes/managers/controls.php
```diff
@@ -436,7 +436,17 @@ public static function get_controls_names() {
 	private function register_controls() {
 		$this->controls = [];
 
-		foreach ( self::get_controls_names() as $control_id ) {
+		$controls_names = self::get_controls_names();
+
+		if ( ! Plugin::instance()->experiments->is_feature_active( 'container_grid' ) ) {
```
TzviRabinovitch, line 441: maybe extract the logic to an external function 
should be something like :
if (! Plugin::instance()->experiments->is_feature_active( 'container_grid' )){
 unsetGapsIfNotUsed()
}

TzviRabinovitch (changes requested):

TzviRabinovitch (approved):

PR #21871, opened 2023-03-31
Internal: Create a wizard for creating a container flexbox/grid - Part 2 [ED-10130]

ArmanPurtoian:


TzviRabinovitch (dismissed):

assets/dev/js/editor/views/add-section/base.js
```diff
@@ -185,6 +187,34 @@ import environment from 'elementor-common/utils/environment';
 		};
 	}
 
+	onGridPresetSelected( event ) {
+		this.closeSelectPresets();
+
+		const selectedStructure = event.currentTarget.dataset.preset;
+		const chunks = selectedStructure.split( '-' );
+		const rows = chunks[ 0 ];
+		const columns = chunks[ 1 ];
```
louiswol94, line 196: @ArmanPurtoian Is it possible to use `elementor.presetsFactory.getParsedStructure()` to parse the structure?

assets/dev/js/editor/views/add-section/base.js
```diff
@@ -185,6 +187,34 @@ import environment from 'elementor-common/utils/environment';
 		};
 	}
 
+	onGridPresetSelected( event ) {
+		this.closeSelectPresets();
+
+		const selectedStructure = event.currentTarget.dataset.preset;
```
louiswol94, line 193: Does the flex container also use `dataset.preset`?

assets/dev/js/editor/utils/container-helper.js
```diff
@@ -53,6 +54,25 @@ export class ContainerHelper {
 		} );
 	}
 
+	static createGridContainer( settings = {}, target, options = {} ) {
```
louiswol94, line 57: @ArmanPurtoian I don't think we need a new function for this? It looks identical to `createContainer()`. You just need to pass the correct `settings` to `createContainer()`, no?

louiswol94 (changes requested):

assets/dev/js/editor/utils/presets-factory.js
```diff
@@ -47,6 +47,11 @@ presetsFactory = {
 		return this.getPresets( parsedStructure.columnsCount, parsedStructure.presetIndex );
 	},
 
+	// Grid preset looks like 1-2 ( 1 rows, 2 columns )
+	getParsedGridStructure( selectedStructure ) {
+		return selectedStructure.split( '-' );
+	},
```
louiswol94, line 53: @ArmanPurtoian 

Please add this line - same as the other function in the beginning of your function:
`selectedStructure += ''; // Make sure this is a string`
What do you think?

Then can you also rather return an object? Because to use `rowsAndCols[ 1 ]` below is a bit confusing. `parsedStructure.rows` and `parsedStructure.columns` would be better?

louiswol94 (changes requested):

assets/dev/js/editor/views/add-section/base.js
louiswol94: @ArmanPurtoian ?

louiswol94 (dismissed):

TzviRabinovitch (approved):

PR #21910, opened 2023-04-03
Fix: Hover active color is not correct in Tabs widget [ED-10305]

davseve:


tests/playwright/sanity/modules/nested-tabs/nested-tabs.test.js
```diff
@@ -433,26 +433,32 @@ test.describe( 'Nested Tabs tests @nested-tabs', () => {
 		await editor.addWidget( 'nested-tabs', container );
 		await editor.getPreviewFrame().waitForSelector( '.e-n-tab-title.e-normal.e-active' );
 
+		await page.pause();
```
hein-obox, line 436: Remove.

modules/nested-tabs/assets/scss/frontend.scss
```diff
@@ -253,7 +253,7 @@
 :is( .elementor .elementor-element.elementor-widget-n-tabs > .elementor-widget-container > .e-n-tabs > .e-n-tabs-heading .e-n-tab-title,
 	.elementor .elementor-element.elementor-widget-n-tabs > .elementor-widget-container > .e-n-tabs > .e-n-tabs-content .e-n-tab-title ) {
 
-	&.e-normal:hover {
+	&:hover {
```
hein-obox, line 256: Have you tried this? 

With doing this, we are targetting the `e-normal` directly, without changing the specifity. 

And we are remove the hover effect from the tabs `.e-n-tabs-content`.

```.elementor .elementor-element.elementor-widget-n-tabs > .elementor-widget-container > .e-n-tabs > .e-n-tabs-heading .e-normal:hover {
		background-color: var( --n-tabs-title-background-color-hover );
		background-image: initial;
	}
```

davseve: I tried that snippet, it makes no change.
We can take it offline and investigate it together.
Removing the `.e-normal` class decreases the specificity and allows the `:not(:hover)` rule to be assigned to the active tab.

tests/playwright/sanity/modules/nested-tabs/nested-tabs.test.js
davseve: Removed.
My mistake

modules/nested-tabs/widgets/nested-tabs.php
```diff
@@ -416,7 +416,7 @@ protected function register_controls() {
 				'name' => 'tabs_title_background_color',
 				'types' => [ 'classic', 'gradient' ],
 				'exclude' => [ 'image' ],
-				'selector' => "{$nested_tabs_heading_selector_class} > .e-n-tab-title:not( .e-active ):not( :hover )",
+				'selector' => "{$nested_tabs_heading_selector_class} > .e-n-tab-title:not( .e-active )",
```
hein-obox, line 419: Does this work correctly for both mobile and desktop? I assume we do need the `:not( :hover )` for the desktop version.

hein-obox:
@davseve Can you please add a mindmap, to make sure that we cover all scenarios.

modules/nested-tabs/widgets/nested-tabs.php
davseve: Correct, 
The hover worked but something else broke.
I fixed it now.

hein-obox (approved):

TzviRabinovitch (approved):

PR #22002, opened 2023-04-11
Internal: FR unit should be presented in all devices (Grid) [ED-10363]

louiswol94:


davseve (approved):

hein-obox:
@louiswol94 One question: can you please indicated on the mindmap if you have checked the inheritance for all devices? I think we need to make sure that the inheritance works e.g. when desktop is custom, laptop is fr, tablet-extra is custom etc.

ArmanPurtoian (approved):

PR #22029, opened 2023-04-13
Internal: Admin can change DB attributes from "Replace URL" - this option should be blocked [ED-9500]

maork-elementor:


includes/utils.php
```diff
@@ -193,12 +193,16 @@ public static function replace_urls( $from, $to ) {
...
-		// @codingStandardsIgnoreEnd
+			$wpdb->prepare(
+				"UPDATE {$wpdb->postmeta} " .
+				'SET `meta_value` = REPLACE(`meta_value`, %s, %s) ' .
+				"WHERE `meta_key` = '_elementor_data' AND `meta_value` LIKE %s ;",
+				str_replace( '/', '\\/', $from ),
+				str_replace( '/', '\\/', $to ),
+				'[%' // meta_value LIKE '[%' are json formatted
+			)
+		);
```
Nevoss, line 205:
```suggestion
			$query = "UPDATE {$wpdb->postmeta} SET `meta_value` = REPLACE(`meta_value`, %s, %s) WHERE `meta_key` = '_elementor_data' AND `meta_value` LIKE '[%';"

			$wpdb->prepare( $query, $from, $to );
		);
```

I think it's more readable.
\+ I think there is no need for `str_replace` to the URLs if you are using `prepare` BUT please validate me.

Nevoss (changes requested):

tests/phpunit/elementor/includes/test-utils.php
```diff
@@ -83,6 +83,22 @@ public function test_should_throw_error_because_urls_are_invalid() {
...
+		global $wpdb;
+
+		//Arrange
+		$old_home_url = 'http://example.org';
+		$new_home_url = 'http://example2.org';
+		$expected_query = "UPDATE wptests_postmeta SET `meta_value` = REPLACE(`meta_value`, 'http:\\\/\\\/example.org', 'http:\\\/\\\/example2.org') WHERE `meta_key` = '_elementor_data' AND `meta_value` LIKE '[%';";
+		$get_query = true;
+
+		//Act
+		$last_query = Utils::replace_urls( $old_home_url, $new_home_url, $get_query );
```
Nevoss, line 97:
```suggestion
		Utils::replace_urls( $old_home_url, $new_home_url, $get_query );
		
		$last_query =  $wpdb->last_query;
```

use that instead of changing the method

includes/utils.php
```diff
@@ -192,13 +192,24 @@ public static function replace_urls( $from, $to ) {
...
+				"WHERE `meta_key` = '_elementor_data' AND `meta_value` LIKE %s;",
+				$escaped_from,
+				$escaped_to,
+				$meta_value_like
+			)
+		);
+
+		if ( $get_query ) {
+			return $wpdb->last_query;
+		}
```
Nevoss, line 212: remove

includes/utils.php
```diff
@@ -170,7 +170,7 @@ public static function get_pro_link( $link ) {
 	 * @return string
 	 * @throws \Exception
 	 */
-	public static function replace_urls( $from, $to ) {
+	public static function replace_urls( $from, $to, $get_query = false ) {
```
Nevoss, line 173:
```suggestion
	public static function replace_urls( $from, $to ) {
```

Nevoss (changes requested):

tests/phpunit/elementor/includes/test-utils.php
maork-elementor: This is not working this is not the last query. 
The function is triggering more queries. so i lost the actual query that i need. this is why i add this code in this specific location.

Nevoss (approved):

Nevoss:
Replaced by #22124

PR #22031, opened 2023-04-13
Fix: Progress Bar widget wrong ARIA text value [HTS-433]

matipojo:
If the widget inner text is empty, there shouldn't be an ARIA text value attribute


Nevoss (approved):

arielk (approved):

macedvisioned:
@rami-elementor Hi Rami, I have tested the Progress bar published in 3.13 update. 
I used a plain WordPress website www.visioned.work with only a hello child theme + some basic plugin's not lay-out related.
Only one issue in the title 
AxeTestProgressbarWidget
Missing title role attribute.
In the link I sent you before some documented details:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/progressbar_role

Further 2 issues in skiplink and header navigation, but that's another story.
If you want want more information let's talk online by Teams or Zoom.

PR #22145, opened 2023-04-21
Tweak: Change Nested Tabs behavior [ED-10465]

ArmanPurtoian:


hein-obox:
@ArmanPurtoian 
Please test all possible scenarios and add to the mindmap.

Direction
Justify content
Tab Alignment
Tab width
Device (desktop + mobile)
Also add icons + icon position

I have noticed that there is an issue with Direction : Start / End (left / right) and Justify content: justified on mobile.

Flex grow causes problems in combination with 'flex-grow'. 

I think in general to spend a bit more time for testing and to map all scenarios on the mindmap.

hein-obox (approved):

louiswol94 (approved):

PR #22191, opened 2023-04-25
Tweak: Added responsive control to Auto Flow in Grid Container [ED-10501]

ArmanPurtoian:


hein-obox:
@ArmanPurtoian Can you please check why the existing `Assert correct icons for Justify Items and Align Items` test is failing? I think there more selectors now with the same selector, because we have added the responsive selectors. Please make sure that that test only targets the desktop selector.

hein-obox:
@ArmanPurtoian Can you please manually test the inheritance as well? Meaning a setting that we change on desktop, should be applied to tablet, unless we change the setting on tablet.

hein-obox (approved):

louiswol94 (approved):

PR #22195, opened 2023-04-25
Fix: [Core] Nested Tabs / Remove Duplicated CSS ID [ED-10533]

hein-obox:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #


modules/nested-tabs/assets/js/frontend/handlers/nested-tabs.js
```diff
@@ -157,7 +158,7 @@ export default class NestedTabs extends Base {
 		};
 	}
 
-	onHideTabContent( $activeContent ) {}
```
hein-obox, line 160: I have removed this unused variable.

ArmanPurtoian (dismissed):

modules/nested-tabs/assets/js/frontend/utils.js
```diff
@@ -0,0 +1,7 @@
+export function isMenuInDropdownMode( elementSettings ) {
```
louiswol94, line 1: @hein-obox This is duplicated code. You can rather move the mega-menu file somewhere more general and import it in both places.

louiswol94 (changes requested):

modules/nested-tabs/assets/js/frontend/utils.js
hein-obox: @louiswol94 

In my version there was a slight difference between the versions, but I have made them identical now.

Proposal: I can remove the version inside the MM in the pro version of this task.
In that way the base function is available in the Core plugin inside the Nested Tabs widget, and is inherited inside the MM widget in the pro version.

modules/nested-tabs/assets/js/frontend/handlers/nested-tabs.js
```diff
@@ -368,6 +374,38 @@ export default class NestedTabs extends Base {
 		}
 	}
 
+	setResponsiveTabTitleId( resizeEvent = false ) {
```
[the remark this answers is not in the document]
hein-obox: Aviv asked David and me to decide on the implementation. 

We think it is best if we have the ID available on all devices, without and prefixes or suffices.

modules/nested-tabs/assets/js/frontend/utils.js
louiswol94: @hein-obox Sorry I forgot this is in core. But okay cool, thanks.

modules/nested-tabs/assets/js/frontend/handlers/nested-tabs.js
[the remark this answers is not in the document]
louiswol94: @hein-obox I understand. I am not sure I like this approach though :-)

And also. if we go with it, should you not unbind the event?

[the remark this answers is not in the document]
hein-obox: Unbinding added.

Approach: I feel like no solution will be really nice until we remove the duplicated mobile tab titles.

What is your suggestion?

PR #22201, opened 2023-04-26
Fix: Lightbox image captions are not aligned to the center [ED-10401]

Barcov:


louiswol94 (dismissed):

tests/playwright/sanity/includes/widgets/image.test.js
```diff
@@ -0,0 +1,32 @@
...
+	const editor = await wpAdmin.useElementorCleanPost();
+
+	// Act.
+	const frame = page.frameLocator( '#elementor-preview-iframe' );
+	await editor.addWidget( 'image' );
+
+	await page.locator( '.elementor-control-media__preview' ).click();
+	await page.getByRole( 'tab', { name: 'Media Library' } ).click();
+	await page.setInputFiles( 'input[type="file"]', './tests/playwright/resources/elementor1.png' );
+	await page.locator( '#attachment-details-title' ).fill( 'Elementor Logo (title)' );
```
hein-obox, line 16: Do we need to add alt text and caption?
I don't think that we are asserting these.

hein-obox:
@Barcov 
Could this test be added inside this file tests/playwright/sanity/image.test.js? I don't think we have have to create a new test file for this.

And could we try to implement the `test.step()` method, see: tests/playwright/sanity/container-grid.test.js.

tests/playwright/sanity/includes/widgets/image.test.js
Barcov: @hein-obox  By default Lightbox has a Title and Description but caption and Alt tags can be assigned to either of these in SiteSettings => Lightbox. I've tested Alt, Caption by hand. I'll remove them from the test.

Barcov:
> @Barcov Could this test be added inside this file tests/playwright/sanity/image.test.js? I don't think we have have to create a new test file for this.
> 
> And could we try to implement the `test.step()` method, see: tests/playwright/sanity/container-grid.test.js.

Well spotted. I've combined the two image tests. I've moved sanity/image.test.js into sanity/includes/widgets/image.test.js to match the path convention used with other tests like modules. I.e. the test has the same path in playwright tests as in the project. I.e. sanity/modules and includes/widgets/ < base widgets

Also included your suggestion for using test.step()

hein-obox (approved):

PR #22204, opened 2023-04-26
Internal: Adding more template tests: accordion, icon_box, icon_list ...   [ED-10538]

dmaziar:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ x] Other... Please describe:  added playwright template tests.

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #


tests/elements-regression/tests/elements-regression.test.js
```diff
@@ -77,9 +85,5 @@ test.describe( 'Elementor regression tests with templates for CORE', () => {
 			await expect( page.locator( EditorSelectors.container ) ).toHaveScreenshot( `${ widgetType }_published.png`, { maxDiffPixels: 100 }, { timeout: 10000 } );
 		} );
 	}
-
-	test.afterAll( async ( { request } ) => {
```
yotamselementor, line 81: Why don't we want to clean up after ourselves?

yotamselementor (changes requested):

tests/elements-regression/tests/elements-regression.test.js
dmaziar: sometimes for some weird reasons, afterAll hook deletes all the images before the test with images is passed. This causes flakiness. Some images are not loaded, because they are deleted.
There should be no impact for ci because the docker image will be deleted.

dmaziar: https://playwright.dev/docs/api/class-test#test-before-all. So it does before all, after all for each worker, which means we are adding media and deleting 3 times in parallel which of course brings flakiness. 
The solution is to move the upload and delete of media to global-setup.js

yotamselementor (approved):

PR #22249, opened 2023-05-01
Tweak: Update `Tested up to: 6.2` [ED-10569]

ramiy:
Branch 3.12

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -2,7 +2,7 @@
 Contributors: elemntor
 Tags: page builder, editor, landing page, drag-and-drop, elementor, visual editor, wysiwyg, design, maintenance mode, coming soon, under construction, website builder, landing page builder, front-end builder
 Requires at least: 5.9
-Tested up to: 6.1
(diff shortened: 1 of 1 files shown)


yotamselementor (changes requested):
I'm not sure this is OK - we did NOT fully test it with 6.2 AFAIK
In fact, our Playwright tests fail when we try to use 6.2, see here: https://github.com/elementor/elementor/pull/22217

arielk (approved):

ramiy:
@yotamselementor I was told that all the PR tested with the "Latest" WordPress version. Which was 6.2 when 3.12 features merged.

yotamselementor:
> @yotamselementor I was told that all the PR tested with the "Latest" WordPress version. Which was 6.2 when 3.12 features merged.

Yes & No - we ask WP-Env to run the latest WP version, but, for some reason, it runs WP 6.1.1
I think this is a problem fixed in future versions of WP-Env

PR #22254, opened 2023-05-01
Fix: The drag and resize option breaks Grid layout [ED-10542]

davseve:


TzviRabinovitch (dismissed):

louiswol94 (changes requested):
@davseve Can we have the resize handles not initialised in the first place? Instead of just hiding it with CSS afterwards.

See this PR for reference:
https://github.com/elementor/elementor/pull/21443/files
https://github.com/elementor/elementor/pull/21443/files#diff-805cb9891b59a77e9bccf7fb5e1d4d5086888f0398abadcd585916a2992ba36dR54

assets/dev/js/editor/container/container.js
```diff
@@ -521,6 +521,10 @@ export default class Container extends ArgsObject {
 		return elementor.userCan( 'design' ) && this.isEditable();
 	}
 
+	isGridContainer() {
+		return 'grid' === this.parent.args.model.attributes.settings.attributes.container_type;
```
louiswol94, line 525:
```suggestion
		return 'grid' === this.settings.get( 'container_type' );
```
@davseve Please check if this works?

louiswol94 (changes requested):

assets/dev/js/editor/container/container.js
davseve: I think that we would like to check the parent state and not the container state itself.
Let's take the following examples:
**Parent**: grid. **Child**: widget - in this case we wat the grid to be resizable but we don't want its children to be resizable.
**Grandparent**: Grid. **Parent**: Flex. **Child**: Widget. - in this case, we would like the Grandparent to be resizable, the flex should not be resizable but the widget can be resizable. 

So AFAIU we need to check the parent value and if the parent is of type `grid`, than its direct child will not be resizable.

louiswol94: Discussed on Slack to use `this.parent.settings.get()`

louiswol94 (approved):

PR #22329, opened 2023-05-04
Internal: POC - support correlation between widget and parent container [ED-10076]

louiswol94:


assets/dev/js/editor/controls/base.js
```diff
@@ -126,6 +126,13 @@ ControlBaseView = Marionette.CompositeView.extend( {
 
 		var isVisible = elementor.helpers.isActiveControl( this.model, settings.attributes, settings.controls );
 
+		const parentContainerType = this.model.get( 'parent_container_type' );
+
+		if ( isVisible && parentContainerType ) {
+			isVisible = this.container.settings.get( 'container_type' ) === parentContainerType ||
```
davseve, line 132: In which scenario will we need to check the item itself?
AFAIU we need to check only the parent item

davseve, line 132: I am not sure about it but should the `ControlBaseView` be aware of the `this.container`?

davseve (commented):
Great work!!
1. I think that we should add `POC` to the PR title.
2. I know this is only a POC but I think that we should handle this condition in the editor PHP side as well.
The control now is hidden but its values are still assigned to the element.
Another issue the that we will need to support BC problems that will come up.
In the following example, we see that the control doesn't exist abuot the value is still rendered.
I would expect that the control value won't be rendered.
See the attached video.  https://www.loom.com/share/62a0cfb172654715bd3a60a1ca07485f

assets/dev/js/editor/controls/base.js
louiswol94: @davseve The container itself is also a "widget". So we need to update the flex based controls in the container widget itself's Advanced tab also.

louiswol94:
> Great work!!
> 
> 1. I think that we should add `POC` to the PR title.
> 2. I know this is only a POC but I think that we should handle this condition in the editor PHP side as well.
>    The control now is hidden but its values are still assigned to the element.
>    Another issue the that we will need to support BC problems that will come up.
>    In the following example, we see that the control doesn't exist abuot the value is still rendered.
>    I would expect that the control value won't be rendered.
>    See the attached video.  https://www.loom.com/share/62a0cfb172654715bd3a60a1ca07485f

@davseve 
1) Done
2) Agreed. If this POC is approved we can continue with that 😃

PR #22346, opened 2023-05-07
Internal: General code structure improvements and adjustments [ED-10651]

matipojo:


modules/safe-mode/module.php
```diff
@@ -64,7 +64,14 @@ public function on_update_safe_mode( $value ) {
 		return $value;
 	}
 
-	public function ajax_enable_safe_mode( $data ) {
+	/**
+	 * @throws \Exception
+	 */
+	public function ajax_enable_safe_mode($data ) {
```
matipojo, line 70:
```suggestion
	public function ajax_enable_safe_mode( $data ) {
```

PR #22350, opened 2023-05-08
Internal: Support for stage-based SVN publishing on pre-release [ED-10630]

maork-elementor:


.github/scripts/publish-to-wordpress-org-staging.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
```
matipojo, line 1: Extract to an action shared with WordPress SVN or just drop it.

matipojo (changes requested):

PR #22373, opened 2023-05-09
Internal: Editor V2 - Allow loading a document without reloading the page [ED-10379]

StyleShit:


assets/dev/js/editor/components/documents/commands/open.js
```diff
@@ -17,12 +17,20 @@ export class Open extends $e.modules.CommandBase {
 			elementor.$previewContents.find( `.elementor-${ id }` ).addClass( 'loading' );
 		}
 
+		if ( setAsInitial ) {
+			// Set the new initial document id to be sent with all requests.
```
Nevoss, line 21: Why? Please explain that the next request uses the initial document id to return the preview URL that's why it has to be here and not when the document is loaded

assets/dev/js/editor/editor-base.js
```diff
@@ -977,7 +977,7 @@ export default class EditorBase extends Marionette.Application {
 		// TODO: Should be command?
 		jQuery( '#elementor-preview-loading' ).show();
 
-		this.$preview[ 0 ].contentWindow.location.reload( true );
+		this.$preview[ 0 ].src = this.config.initial_document.urls.preview;
```
Nevoss, line 980: Please check the case when you are editing loop item that the initial document was some page and then switch the canvas view

Nevoss (changes requested):

assets/dev/js/editor/components/documents/commands/open.js
StyleShit: Good point, done

assets/dev/js/editor/editor-base.js
StyleShit: NO

Nevoss (approved):

PR #22401, opened 2023-05-10
Tweak: Added a horizontal scrolling option to the Nested Tabs settings [ED-9311]

ArmanPurtoian:


modules/nested-tabs/assets/js/frontend/handlers/nested-tabs.js
```diff
@@ -479,4 +518,112 @@ export default class NestedTabs extends Base {
...
+		const slider = this.elements.$headingContainer[ 0 ];
+
+		return slider.clientWidth < this.getChildrenWidth( slider.children ) && 'enable' === this.getHorizontalScrollSetting();
+	}
+
+	getChildrenWidth( children ) {
+		let totalWidth = 0;
+		const parentContainer = children[ 0 ].parentNode;
+		const computedStyles = getComputedStyle( parentContainer );
+		const gap = parseFloat( computedStyles.gap ) || 0; // Get the gap value or default to 0 if it's not specified
```
davseve, line 544:
```suggestion
		const parentContainer = children[ 0 ].parentNode,
		       computedStyles = getComputedStyle( parentContainer ),
		       gap = parseFloat( computedStyles.gap ) || 0; // Get the gap value or default to 0 if it's not specified
```

modules/nested-tabs/assets/js/frontend/handlers/nested-tabs.js
```diff
@@ -479,4 +518,112 @@ export default class NestedTabs extends Base {
...
+		return totalWidth;
+	}
+
+	setHorizontalScrollAlignment( event = {} ) {
+		if ( ! this.elements ) {
+			return;
+		}
+
+		const slider = this.elements.$headingContainer[ 0 ];
+		const tabsDirection = this.getTabsDirection();
```
davseve, line 559:
```suggestion
		const slider = this.elements.$headingContainer[ 0 ],
		     tabsDirection = this.getTabsDirection();
```

modules/nested-tabs/assets/js/frontend/handlers/nested-tabs.js
```diff
@@ -479,4 +518,112 @@ export default class NestedTabs extends Base {
...
+
+		event.preventDefault();
+
+		const previousPositionX = parseFloat( slider.dataset.pageX ),
+			mouseMoveX = event.pageX - previousPositionX,
+			maximumScrollValue = 5;
+
+		let toScrollDistanceX = 0;
+
+		if ( 20 < mouseMoveX ) {
```
davseve, line 619: Why 20 is a fixed value?
I suggest placing it in a variable so we will know the intention of this value.

modules/nested-tabs/assets/js/frontend/handlers/nested-tabs.js
```diff
@@ -479,4 +518,112 @@ export default class NestedTabs extends Base {
...
+			toScrollDistanceX = maximumScrollValue;
+		} else if ( -20 > mouseMoveX ) {
+			toScrollDistanceX = -1 * maximumScrollValue;
+		} else {
+			toScrollDistanceX = mouseMoveX / 10;
```
davseve, line 624: Same as above

davseve (changes requested):

davseve (dismissed):

modules/nested-tabs/assets/js/frontend/handlers/nested-tabs.js
```diff
@@ -278,13 +304,26 @@ export default class NestedTabs extends Base {
...
 	}
 
 	onEditSettingsChange( propertyName, value ) {
 		if ( 'activeItemIndex' === propertyName ) {
 			this.changeActiveTab( value, false );
 		}
 	}
+	onElementChange( propertyName ) {
+		const propsToListen = [
+			'horizontal_scroll',
```
louiswol94, line 318: @ArmanPurtoian This will not work for responsive controls like `horizontal_scroll`.
rather use something like `0 === propertyName.indexOf( 'horizontal_scroll' )`
Because this needs to check for horizontal_scroll_tablet, horizontal_scroll_mobile etc as well, right?

modules/nested-tabs/assets/js/frontend/handlers/nested-tabs.js
```diff
@@ -479,4 +518,114 @@ export default class NestedTabs extends Base {
...
+		return totalWidth;
+	}
+
+	setHorizontalScrollAlignment( event = {} ) {
+		if ( ! this.elements ) {
+			return;
+		}
+
+		const slider = this.elements.$headingContainer[ 0 ],
+			tabsDirection = this.getTabsDirection();
```
louiswol94, line 560: `tabsDirection` can be inside the if statement to avoid an unnecessary call to this function? You only need `slider` in both cases.

louiswol94 (changes requested):

modules/nested-tabs/assets/js/frontend/handlers/nested-tabs.js
```diff
@@ -479,4 +518,115 @@ export default class NestedTabs extends Base {
...
+		} else {
+			slider.classList.remove( 'e-scroll' );
+			slider.dataset.pageX = '';
+		}
+	}
+
+	isHorizontalScroll() {
+		const slider = this.elements.$headingContainer[ 0 ];
+
+		return slider.clientWidth < this.getChildrenWidth( slider.children ) && 'enable' === this.getHorizontalScrollSetting();
```
hein-obox, line 537: Would is be possible to use `slider.scrollWidth` instead of `this.getChildrenWidth()`? `slider.scrollWidth` will give you the full width of all scrollable elements together.

modules/nested-tabs/assets/js/frontend/handlers/nested-tabs.js
```diff
@@ -479,4 +518,115 @@ export default class NestedTabs extends Base {
...
+	}
+
+	initialScrollPosition( slider, tabsDirection ) {
+		switch ( tabsDirection ) {
+			// To cover defaults and center
+			case undefined:
+			case '':
+			case 'center':
+				slider.style.setProperty( '--n-tabs-heading-justify-content', 'start' );
+				slider.scrollLeft = ( this.getChildrenWidth( slider.children ) - slider.offsetWidth ) / 2;
```
hein-obox, line 584: Why do we subtract `slider.offsetWidth`?

modules/nested-tabs/assets/scss/frontend.scss
```diff
@@ -59,9 +61,33 @@
...
+				cursor: -webkit-grabbing;
+
+				&-active {
+					position: relative;
+
+					&::before {
+						content: '';
+						position: absolute;
+						inset-block: 0;
+						inset-inline: -1000vw;
```
louiswol94, line 86: What is this about?
`inset-inline: -1000vw;`

modules/nested-tabs/assets/js/frontend/handlers/nested-tabs.js
```diff
@@ -278,13 +304,26 @@ export default class NestedTabs extends Base {
...
+		if ( this.checkSliderPropsToWatch( propertyName ) ) {
+			this.setHorizontalScrollAlignment();
+		}
+	}
+
+	checkSliderPropsToWatch( propertyName ) {
+		return 0 === propertyName.indexOf( 'horizontal_scroll' ) ||
+			0 === propertyName.indexOf( 'tabs_justify_horizontal' ) ||
+			0 === propertyName.indexOf( 'tabs_title_space_between' );
+	}
```
hein-obox, line 326: I am not sure if I understand what happens here.
I assume we only use the horizontal scroll function when the `propertyName` horizontal_scroll` has been activated.

Why do we check `tabs_justify_horizontal` and `tabs_title_space_between` as well?

About `propertyName`: I would assume that always to return a string, but you seem to treat it like an array. Are you sure about that? I am just curious.

modules/nested-tabs/assets/js/frontend/handlers/nested-tabs.js
```diff
@@ -479,4 +518,115 @@ export default class NestedTabs extends Base {
...
+		}
+	}
+
+	isHorizontalScroll() {
+		const slider = this.elements.$headingContainer[ 0 ];
+
+		return slider.clientWidth < this.getChildrenWidth( slider.children ) && 'enable' === this.getHorizontalScrollSetting();
+	}
+
+	getChildrenWidth( children ) {
```
hein-obox, line 540: I am not sure if we need this function. I hope that we can replace it with the JS property `Element.scrollWidth`.

modules/nested-tabs/assets/js/frontend/handlers/nested-tabs.js
```diff
@@ -479,4 +518,115 @@ export default class NestedTabs extends Base {
 
 		return !! $tabTitleContainerElement && isTabTitleActive ? true : false;
 	}
+
+	changeScrollStatus( event ) {
```
hein-obox, line 522: I am thinking possibly we could add a comment to the codepen that we used for this, for reference for future maintenance.

modules/nested-tabs/assets/scss/frontend.scss
louiswol94: CC @hein-obox

modules/nested-tabs/assets/js/frontend/handlers/nested-tabs.js
hein-obox: @ArmanPurtoian @louiswol94 
100%

modules/nested-tabs/assets/js/frontend/handlers/nested-tabs.js
hein-obox: Done already?

modules/nested-tabs/assets/js/frontend/handlers/nested-tabs.js
hein-obox: @ArmanPurtoian This part isn't clear to me.
Doesn't `slider.style.setProperty( '--n-tabs-heading-justify-content', 'start' );` overwrite center styling?
I am curious about this aspect.

modules/nested-tabs/assets/scss/frontend.scss
hein-obox: @ArmanPurtoian @louiswol94 I had added the `.e-scroll-active` to make sure that we couldn't hover over elements while dragging them to the left or right hand side.
I found that a bit annoying that the elements would receive a hover effect, while moving them. 

What's your take or approach about that?

hein-obox: @louiswol94 

I had added the `.e-scroll-active` class, to be able to stop hover effects from happening when moving the elements from left to right.

I used an 'inset-inline: -1000vw` to make sure that this element would cover all tab elements.

hein-obox (dismissed):

louiswol94 (dismissed):

louiswol94 (approved):

PR #22465, opened 2023-05-16
Fix: Templates responsive setting won't update unless 'Regenerate CSS' is used [ED-10594]

maork-elementor:


includes/managers/controls.php
```diff
@@ -862,6 +875,28 @@ public function remove_control_from_stack( $stack_id, $control_id ) {
...
+	/**
+	 * Should clear stack cache.
+	 * Whether the CSS requires to clear the controls stack cache.
+	 * @since 3.13.0
+	 * @access protected
+	 * @return bool True if the CSS requires to clear the controls stack cache, False otherwise.
+	 */
+	public function stack_cache_has_been_cleared() {
+		return $this->has_been_cleared;
+	}
```
udidol, line 887: I would change the name of the private variable and the method to accurately describe what it does.
The Controls Manager shouldn't be aware of CSS, it should exist as a separate entity.

Also would need to be changed in the Pro PR.

```suggestion
	/**
	 * Has Stacks Cache Been Cleared.
	 *
	 * @since 3.13.0
	 * @access protected
	 * @return bool True if the CSS requires to clear the controls stack cache, False otherwise.
	 */
	public function has_stacks_cache_been_cleared() {
		return $this->has_stacks_cache_been_cleared;
	}
```

includes/managers/controls.php
```diff
@@ -862,6 +875,28 @@ public function remove_control_from_stack( $stack_id, $control_id ) {
...
+
+	/**
+	 * Clear stack.
+	 * This method clears the stack.
+	 * @since 3.13.0
+	 * @access public
+	 */
+	public function clear_stack_cache() {
+		$this->stacks = [];
+		$this->has_been_cleared = true;
```
udidol, line 897:
```suggestion
		$this->has_stacks_cache_been_cleared = true;
```

includes/managers/controls.php
```diff
@@ -290,6 +290,19 @@ class Controls_Manager {
...
+	 *
+	 * Holds whether the controls manager has been clear stack.
+	 *
+	 * @since 3.13.0
+	 * @access private
+	 * @static
+	 *
+	 * @var array
+	 */
+	public $has_been_cleared = false;
```
udidol, line 304:
```suggestion
	public $has_stacks_cache_been_cleared = false;
```

includes/managers/controls.php
```diff
@@ -290,6 +290,19 @@ class Controls_Manager {
 	 */
 	private static $tabs;
 
+	/**
+	 * Has been cleared.
+	 *
+	 * Holds whether the controls manager has been clear stack.
```
udidol, line 296:
```suggestion
	 * Has stacks cache been cleared.
	 *
	 * Boolean flag used to determine whether the controls manager stack cache has been cleared once during the current runtime.
```

includes/managers/controls.php
udidol: What about the use case where controls are already duplicated when CSS is generated?

udidol (changes requested):

core/files/css/template.php
```diff
@@ -0,0 +1,47 @@
...
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+
+class Template extends Post_CSS {
+
+	protected function parse_content() {
+		$stack_cache_has_been_cleared = Plugin::$instance->controls_manager->stack_cache_has_been_cleared();
+		$has_document = Plugin::$instance->documents->get_current();
```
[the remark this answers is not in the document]
maork-elementor: Removed

includes/managers/controls.php
maork-elementor: Done

includes/managers/controls.php
maork-elementor: Done

includes/managers/controls.php
maork-elementor: Done

tests/phpunit/elementor/includes/managers/test-controls.php
```diff
@@ -54,4 +57,61 @@ public function test_delete_stack() {
...
+			'settings' => [],
+			'id' => '1',
+		], $stacks_mock );
+
+		// Act
+		Plugin::$instance->controls_manager->open_stack( $widget );
+
+		// Assert
+		$stacks = Plugin::$instance->controls_manager->get_stacks();
+		$this->assertNotEmpty( $stacks );
```
udidol, line 98: Elementor is instantiated for the test run, so theoretically there could be unrelated stacks in the cache. Did you check that your mock widget's stack is the only one? If it isn't the only stack this could be a false negative

maork-elementor: I checked if the current `stacks` are actually our stacks. 
They seems to be OK, and not false positive. 

I also found some issue with the assigned the control_data to the actual widget.
`add_control_to_stack` that was fixed.

udidol (dismissed):

includes/managers/controls.php
```diff
@@ -290,6 +290,19 @@ class Controls_Manager {
...
+	 *
+	 * Boolean flag used to determine whether the controls manager stack cache has been cleared once during the current runtime.
+	 *
+	 * @since 3.13.0
+	 * @access private
+	 * @static
+	 *
+	 * @var array
+	 */
+	public $has_stacks_cache_been_cleared = false;
```
matipojo, line 304: Private or public?

includes/managers/controls.php
```diff
@@ -862,6 +875,27 @@ public function remove_control_from_stack( $stack_id, $control_id ) {
 		return true;
 	}
 
+	/**
+	 * Has Stacks Cache Been Cleared.
+	 * @since 3.13.0
+	 * @access protected
+	 * @return bool True if the CSS requires to clear the controls stack cache, False otherwise.
+	 */
+	public function has_stacks_cache_been_cleared() {
```
matipojo, line 884: Protected or public?

matipojo (changes requested):

includes/managers/controls.php
maork-elementor: Private, updated.

includes/managers/controls.php
maork-elementor: This function is public, like all the other methods of this Manager.

matipojo: Please fix the PHPDoc

maork-elementor: PHPDoc fixed

matipojo (approved):

PR #22485, opened 2023-05-18
Fix: Background video is not working in Nested containers [ED-10744]

davseve:


assets/dev/js/editor/elements/views/base.js
```diff
@@ -553,7 +554,9 @@ BaseElementView = BaseContainer.extend( {
 	renderCustomElementID() {
 		const customElementID = this.getEditModel().get( 'settings' ).get( '_element_id' );
 
-		this.$el.attr( 'id', customElementID );
+		if ( customElementID ) {
+			this.$el.attr( 'id', customElementID );
```
davseve, line 558: Not related to this task, this condition will prevent the creation of redundant empty `id` attribute tags in the elements.

louiswol94 (approved):

PR #22504, opened 2023-05-21
Fix: Arbitrary File Write via Archive Extraction (Zip Slip) [ED-10576]

maork-elementor:


core/files/file-types/zip.php
```diff
@@ -128,6 +128,11 @@ private function get_allowed_files( $zip, $allowed_file_types ) {
 			$filename = $zip->getNameIndex( $i );
 			$extension = pathinfo( $filename, PATHINFO_EXTENSION );
 
+			// Skip files with transversal paths.
+			if ( strpos( $filename, '..' ) !== false ) {
```
matipojo, line 132: Please check also other places like https://github.com/elementor/elementor/blob/3921cf2e16a364a75908b731a243930335517962/core/files/uploads-manager.php#L78

matipojo (changes requested):

core/files/file-types/zip.php
maork-elementor: This class is usign the 
`use Elementor\Core\Files\File_Types\Zip;`

there for he is using the validation from `$this->get_allowed_files` inside the `extract` method from `Elementor\Core\Files\File_Types\Zip`.

Did i miss something?

matipojo (approved):

PR #22550, opened 2023-05-26
Fix: Unable to save Global Widget [ED-10855]

louiswol94:


tests/phpunit/elementor/includes/sources/test-local.php
```diff
@@ -78,22 +75,6 @@ public function test_save_item__contributors_cannot_publish_templates() {
 		$this->assertEquals( 'pending', $document->get_post()->post_status );
 	}
 
-	public function test_save_item__contributors_cannot_create_pages() {
```
louiswol94, line 81: This can maybe be changed to assert the post status instead of completely deleting.

StyleShit:
resolved by #22564

PR #22552, opened 2023-05-26
Fix: Unable to save Global Widget [ED-10855]

louiswol94:


includes/template-library/sources/local.php
```diff
@@ -574,7 +574,7 @@ public function save_item( $template_data ) {
 	protected function is_valid_template_type( $type ) {
 		$allowed_document_types = Plugin::$instance->documents->get_document_types( [
 			'show_in_library' => true,
-		] );
+		] ) + [ 'widget' => 'ElementorPro\Modules\GlobalWidget\Documents\Widget' ];
```
louiswol94, line 577: I realise Core is not supposed to be aware of this Pro Document, but it is just a sample illustrating that adding another item to this array is a possible solution, maybe via a filter hook.

StyleShit:
resolved by #22564

PR #22586, opened 2023-05-29
Tweak: Unified the appearance of 'stretch' and 'center' buttons in Tabs widgets [ED-10876]

rami-elementor:


tests/playwright/sanity/modules/nested-tabs/nested-tabs.test.js
```diff
@@ -50,7 +50,7 @@ test.describe( 'Nested Tabs tests @nested-tabs', () => {
 
 		// Act.
 		// Set tabs direction to 'stretch'.
-		await page.locator( '.elementor-control-tabs_justify_horizontal .elementor-control-input-wrapper .eicon-h-align-stretch' ).click();
+		await page.locator( '.elementor-control-tabs_justify_horizontal .elementor-control-input-wrapper .eicon-flex eicon-align-stretch-h' ).click();
```
hein-obox, line 53: I think we need to remove the '.eicon-flex' everywhere and add a dot to the next selector.

rami-elementor: Yes, you are right. Good catch!

arielk (approved):

PR #22594, opened 2023-05-29
New: [NA] [Style] [Accordion] Created a Accordion section [ED-10521]

TzviRabinovitch:


modules/nested-accordion/assets/scss/frontend.scss
```diff
@@ -4,15 +4,28 @@
 	--n-accordion-border-width: 1px;
 	--n-accordion-border-color: #{$gray-light};
 	--n-accordion-border-style: solid;
+  	--n-accordion-item-title-space-between: 0px;
+  	--n-accordion-item-title-distance-from-content: 0px;
+  	--n-accordion-padding-top:0px;
```
hein-obox, line 9: Add space between.

modules/nested-accordion/assets/scss/frontend.scss
```diff
@@ -4,15 +4,28 @@
...
+
+		  &:not(:last-child) {
+				margin-bottom: var(--n-accordion-item-title-space-between);
+		  	}
+
 			> .e-con {
 				border-width: var(--n-accordion-border-width);
 				border-color: var(--n-accordion-border-color);
 				border-style: var(--n-accordion-border-style);
+			  	margin-top: var(--n-accordion-item-title-distance-from-content);
```
hein-obox, line 28: This might not be the best approach, as the .e-con element has margin settings that could overwrite this value. I think it might be safer to use gap styling on the wrapper.

modules/nested-accordion/assets/scss/frontend.scss
```diff
@@ -4,15 +4,28 @@
...
+  	--n-accordion-padding-left:0px;
 
 	.e-n-accordion {
 
 		&-item {
 
+		  padding: var(--n-accordion-padding-top) var(--n-accordion-padding-right) var(--n-accordion-padding-bottom) var(--n-accordion-padding-left);
+
+		  &:not(:last-child) {
+				margin-bottom: var(--n-accordion-item-title-space-between);
```
hein-obox, line 21: This might not be the best approach, as the .e-con element has margin settings that could overwrite this value. It me safer to add padding/margin/gap on the wrapper element...

modules/nested-accordion/widgets/nested-accordion.php
```diff
@@ -44,6 +46,148 @@ public function show_in_panel(): bool {
...
+					'max' => 200,
+				],
+			],
+			'default' => [
+				'size' => 0,
+			],
+			'selectors' => [
+				'{{WRAPPER}}' => '--n-accordion-item-title-distance-from-content: {{SIZE}}{{UNIT}}',
+			],
+			'frontend_available' => true,
```
hein-obox, line 151: Do we need this inside JS? If so, possibly we can add it when we implement it?

modules/nested-accordion/widgets/nested-accordion.php
```diff
@@ -44,6 +46,148 @@ public function show_in_panel(): bool {
...
+		] );
+
+		$this->start_controls_tabs( 'accordion__border_and_background' );
+
+		foreach ( array( 'normal', 'hover', 'active' ) as &$state ) {
```
hein-obox, line 156: I like this :)

modules/nested-accordion/widgets/nested-accordion.php
```diff
@@ -44,6 +46,148 @@ public function show_in_panel(): bool {
...
+		$this->end_controls_tabs();
+
+		$this->add_responsive_control(
+			'accordion_border_radius',
+			[
+				'label' => esc_html__( 'Border Radius', 'elementor' ),
+				'type' => Controls_Manager::DIMENSIONS,
+				'size_units' => [ 'px', '%', 'em', 'rem', 'vw', 'custom' ],
+				'selectors' => [
+					'{{WRAPPER}} .e-n-accordion-item'  => 'border-radius: {{TOP}}{{UNIT}} {{RIGHT}}{{UNIT}} {{BOTTOM}}{{UNIT}} {{LEFT}}{{UNIT}};',
```
hein-obox, line 169: Could we use a css variable instead? Cf. the Nested Carousel widget.

modules/nested-accordion/widgets/nested-accordion.php
```diff
@@ -44,6 +46,148 @@ public function show_in_panel(): bool {
...
+		);
+
+		$this->add_responsive_control(
+			'accordion_padding',
+			[
+				'label' => esc_html__( 'Padding', 'elementor' ),
+				'type' => Controls_Manager::DIMENSIONS,
+				'size_units' => [ 'px', '%', 'em', 'rem', 'custom' ],
+				'selectors' => [
+					'{{WRAPPER}} .e-n-accordion-item' => '--n-accordion-padding-top: {{TOP}}{{UNIT}}; --n-accordion-padding-right: {{RIGHT}}{{UNIT}}; --n-accordion-padding-bottom: {{BOTTOM}}{{UNIT}}; --n-accordion-padding-left: {{LEFT}}{{UNIT}};',
```
hein-obox, line 182: Do we need the full selector '{{WRAPPER}} .e-n-accordion-item'?
'{{WRAPPER}}' might be sufficient?  

Not sure. But it would make the code even a bit cleaner.

tests/playwright/pages/editor-page.js
```diff
@@ -715,4 +715,14 @@ module.exports = class EditorPage extends BasePage {
...
+
+	async setColorControlValue( color, colorControlId ) {
+		const controlSelector = '.elementor-control-' + colorControlId;
+
+		await this.page.locator( controlSelector + ' .pcr-button' ).click();
+		await this.page.locator( '.pcr-app.visible .pcr-interaction input.pcr-result' ).fill( color );
+		await this.page.locator( controlSelector ).click();
+	}
+
+
```
hein-obox, line 727: Remove empty line?

tests/playwright/sanity/modules/nested-accordion/nested-accordion.test.js
```diff
@@ -1,4 +1,4 @@
-import { test, expect } from '@playwright/test';
+import { expect, test } from '@playwright/test';
```
hein-obox, line 1: Unneccesary change?

tests/playwright/sanity/modules/nested-accordion/nested-accordion.test.js
```diff
@@ -150,5 +150,101 @@ test.describe( 'Nested Accordion @nested-accordion', () => {
...
+			await test.step( 'test space between Items and content', async () => {
+				// Act
+				await editor.setSliderControlValue( 'accordion_item_title_space_between', '15' );
+				await editor.setSliderControlValue( 'accordion_item_title_distance_from_content', '5' );
+
+				// Assert.
+				await expect( nestedAccordionItemTitle.first() ).toHaveCSS( 'margin-bottom', '15px' );
+				await expect( nestedAccordionItemContent.first() ).toHaveCSS( 'margin-top', '5px' );
+			} );
+			await test.step( 'test normal background color and border style', async () => {
```
hein-obox, line 178: Maybe: Capital letter for 'Test normal'?

tests/playwright/sanity/modules/nested-accordion/nested-accordion.test.js
```diff
@@ -150,5 +150,101 @@ test.describe( 'Nested Accordion @nested-accordion', () => {
...
+				await setBorderAndBackground( editor, 'hover', '#00ff00', 'dashed', '#0000ff' );
+				nestedAccordionItemTitle.first().hover();
+
+				// Assert
+				await expect( nestedAccordionItemTitle.first() ).toHaveCSS( 'background-color', 'rgb(0, 255, 0)' );
+				await expect( nestedAccordionItemTitle.first() ).toHaveCSS( 'border-style', 'dashed' );
+				await expect( nestedAccordionItemTitle.first() ).toHaveCSS( 'border-color', 'rgb(0, 0, 255)' );
+			} );
+
+			await test.step( 'test active background color and border style', async () => {
```
hein-obox, line 198: Same as above.

tests/playwright/sanity/modules/nested-accordion/nested-accordion.test.js
```diff
@@ -150,5 +150,101 @@ test.describe( 'Nested Accordion @nested-accordion', () => {
...
+				await setBorderAndBackground( editor, 'active', '#0000ff', 'dotted', '#ff0000' );
+				nestedAccordionItemTitle.first().click();
+
+				// Assert
+				await expect( nestedAccordionItemTitle.first() ).toHaveCSS( 'background-color', 'rgb(0, 0, 255)' );
+				await expect( nestedAccordionItemTitle.first() ).toHaveCSS( 'border-style', 'dotted' );
+				await expect( nestedAccordionItemTitle.first() ).toHaveCSS( 'border-color', 'rgb(255, 0, 0)' );
+			} );
+
+			await test.step( 'test border radius ', async () => {
```
hein-obox, line 209: Capital letter/

modules/nested-accordion/widgets/nested-accordion.php
TzviRabinovitch: nice catch

modules/nested-accordion/assets/scss/frontend.scss
TzviRabinovitch: @davseve 
i need your input about naming css vars

hein-obox: @davseve I wonder if we should use specific widget variables here or that it is better to use the container variables like `--border-width`. 
I know you have done more work on the relationship between nested widget container styling & container styling.

Possibly should we apply the low specificity aspect here as well?

davseve: 1.  Regarding the variable name, this is the exact same name we used in the mega menu, so I think that the var name is OK, unless we have a better suggestion.
2.  Regarding the border, we can use distributed CSS rules for each property, The problem is different, as Hein pointed out, we want the container's style to have a higher specificity than the accordion container style.
 We can't strengthen the specificity of the container, so we need to weaken accordion container specificity.
 In order to achieve that, we would like to assign the CSS styling in the selector in the PHP control.
 The selector should be defined with the `:where()` CSS function.

> The difference between :where() and [:is()](https://developer.mozilla.org/en-US/docs/Web/CSS/:is) is that :where() always has 0 [specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity), whereas :is() takes on the specificity of the most specific selector in its arguments. ([https://developer.mozilla.org/en-US/docs/Web/CSS/:where](https://developer.mozilla.org/en-US/docs/Web/CSS/:where))

Like what we did in the nested carousel:
`$low_specificity_slider_container_selector = ':where( {{WRAPPER}} .swiper-slide ) > .e-con';
`

```
$this->add_group_control(
			Group_Control_Border::get_type(),
			[
				'name' => 'content_border',
				'selector' => $low_specificity_slider_container_selector,
				'fields_options' => [
					'color' => [
						'label' => esc_html__( 'Border Color', 'elementor-pro' ),
					],
					'width' => [
						'label' => esc_html__( 'Border Width', 'elementor-pro' ),
					],
				],
			]
		);`
`
```

Eventualy,

modules/nested-accordion/assets/scss/frontend.scss
```diff
@@ -4,11 +4,22 @@
 	--n-accordion-border-width: 1px;
 	--n-accordion-border-color: #{$gray-light};
 	--n-accordion-border-style: solid;
+  	--n-accordion-item-title-space-between: 0px;
+  	--n-accordion-item-title-distance-from-content: 0px;
+  	--n-accordion-padding:0px;
```
hein-obox, line 9: Add space between `:0px;`.

modules/nested-accordion/assets/scss/frontend.scss
```diff
@@ -4,11 +4,22 @@
...
+  	--n-accordion-border-radius :0px;
```
hein-obox, line 10: `--n-accordion-border-radius: 0px;`

modules/nested-accordion/assets/scss/frontend.scss
```diff
@@ -4,11 +4,22 @@
...
 
 	.e-n-accordion {
 
 		&-item {
 
+		  padding: var(--n-accordion-padding) ;
+		  border-radius:  var(--n-accordion-border-radius);
```
hein-obox, line 17: `border-radius: var(--n-accordion-border-radius);`

modules/nested-accordion/assets/scss/frontend.scss
```diff
@@ -30,6 +41,10 @@
 				svg {
 					fill: var(--n-accordion-icon-color);
 				}
+
+			}
+			&[open] .e-n-accordion-item-title {
+				margin-bottom: var(--n-accordion-item-title-distance-from-content);
```
hein-obox, line 47: Add default value at the top of the file.

modules/nested-accordion/assets/scss/frontend.scss
```diff
@@ -4,11 +4,22 @@
...
 
 	.e-n-accordion {
 
 		&-item {
 
+		  padding: var(--n-accordion-padding) ;
+		  border-radius:  var(--n-accordion-border-radius);
+
+		  &:not(:last-child) {
+				margin-bottom: var(--n-accordion-item-title-space-between);
```
hein-obox, line 20: Add default value at the top of the file.

modules/nested-accordion/widgets/nested-accordion.php
```diff
@@ -44,6 +46,148 @@ public function show_in_panel(): bool {
...
+
+		$this->end_controls_section();
+	}
+
+	/**
+	 * @string $state
+	 */
+	private function add_border_and_radius_style( $state ) {
+
+		$selector = '{{WRAPPER}}  .e-n-accordion-item';
```
hein-obox, line 136: Remove `.e-n-accordion-item` from selector.

modules/nested-accordion/assets/scss/frontend.scss
```diff
@@ -4,11 +4,22 @@
...
+		  	}
+
 			> .e-con {
```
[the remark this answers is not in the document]
hein-obox: The logic is: 
We should be able to overwrite the '.e-con' styling made in this PR( e.g. border styling) with the specific containers of the nested elements. Makes sense?
I believe David is right.
If you are not convinced, can you please check David?

modules/nested-accordion/widgets/nested-accordion.php
hein-obox: Sorry, I had missed that. 

I do believe we need some kind of low specificity selector here.

modules/nested-accordion/assets/scss/frontend.scss
```diff
@@ -4,11 +4,22 @@
...
 				border-width: var(--n-accordion-border-width);
 				border-color: var(--n-accordion-border-color);
```
hein-obox, line 25: What is this styling used for? Are these not defined by the php controls? Is this pure as default values?

TzviRabinovitch: yes
it will be overridden in the content styling PR

modules/nested-accordion/assets/scss/frontend.scss
hein-obox: I am still seeing the old version.

hein-obox (dismissed):

hein-obox (approved):

PR #22605, opened 2023-05-30
Internal: Changelog generator [ED-10792]

maork-elementor:


matipojo (changes requested):

.github/scripts/changelog-generator-v2.js
```diff
@@ -0,0 +1,92 @@
...
+		owner,
+		repo,
+		tag: headTag
+	})
+	return new Date(release.data.created_at)
+}
+
+// Fetch all pull requests with pagination
+async function fetchAllPullRequests (releaseDate) {
+	const pullRequests = await octokit.paginate(octokit.rest.pulls.list, {
```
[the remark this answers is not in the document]
maork-elementor: Done : )

.github/scripts/changelog-generator-v2.js
```diff
@@ -0,0 +1,88 @@
...
+// Fetch the last 100 commits
+async function fetchAllCommits (releaseDate) {
+	const { data: commits } = await octokit.rest.repos.listCommits({
+		owner,
+		repo,
+		per_page: 100
+	})
+
+	return commits.filter(commit => {
+		const commitDate = new Date(commit.commit.committer.date)
```
matipojo, line 32: Why do you need the date? you need the diff between the baseBranch and the headTag

matipojo (changes requested):

.github/scripts/changelog-generator-v2.js
maork-elementor: Squash merging will affect the results of compareCommits.

In a squash merge scenario, these two methods are likely to yield different results because the "compareCommits" will consider the single squash commit while the "Filtering commits by date" won't see the individual commits that were included in the squash if they were made before the date you're filtering from.

matipojo (approved):

PR #22644, opened 2023-06-01
New: [NA] [Style] [Content] Changed Title HTML default value and styling [ED-10884]

Barcov:


Barcov:
See https://github.com/elementor/elementor/pull/22651

PR #22688, opened 2023-06-06
Fix: Cloud Hosting version control mechanism [ED-10790]

maork-elementor:


includes/settings/tools.php
```diff
@@ -181,8 +186,17 @@ public function __construct() {
...
+	public function get_rollback_versions() {
+
 		$rollback_versions = get_transient( 'elementor_rollback_versions_' . ELEMENTOR_VERSION );
+
+		$plugin_slug = basename( ELEMENTOR__FILE__, '.php' );
+		$is_cloud = strpos( ELEMENTOR_VERSION, '-cloud' ) !== false;
+
+		if ( $is_cloud ) {
+			$rollback_versions = apply_filters( 'elementor/settings/tools/rollback/versions', $rollback_versions, $plugin_slug );
+		}
```
Nevoss, line 198: Why all this check is needed? why not only put the filter?

includes/settings/tools.php
```diff
@@ -128,21 +128,26 @@ public function post_elementor_rollback() {
 
 		$rollback_versions = $this->get_rollback_versions();
 		$version = Utils::get_super_global_value( $_GET, 'version' );
+		$is_cloud = strpos( ELEMENTOR_VERSION, '-cloud' ) !== false;
 
-		if ( empty( $version ) || ! in_array( $version, $rollback_versions, true ) ) {
+		if ( ! $is_cloud && empty( $version ) || ! in_array( $version, $rollback_versions, true ) ) {
```
Nevoss, line 133: Why needed?

Nevoss (changes requested):

includes/settings/tools.php
maork-elementor: There is no point in running code that does not need to be run. 

Is there a reason to run this hook outside the cloud? Cloud websites are the only ones that use this. When we don't need hook, why take the risk of activating him?

includes/settings/tools.php
maork-elementor: It should be the default behavior, if the plugin version is not cloud, just act as usual

PR #22429, opened 2023-05-11
Fix: Changed http: to https: in some URL

DAnn2012:
## PR Checklist

- [x] The commit message follows our guidelines:  https://github.com/elementor/elementor/blob/master/.github/CONTRIBUTING.md

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## Summary

This PR can be summarized in the following changelog entry:

* Changed http: to https: in some URL

## Description
An explanation of what is done in this PR

* Changed http: to https: in some URL (site go.elementor.com and placeholder new-url.com)

## Test instructions
This PR can be tested by following these steps:

*

## Quality assurance

- [ ] I have tested this code to the best of my abilities
- [ ] I have added unittests to verify the code works as intended
- [ ] Docs have been added / updated (for bug fixes / features)

Fixes #

PR #22713, opened 2023-06-08
Fix: Use default placeholder set in URL control

DAnn2012:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [x] Code style update (formatting, local variables)
- [ ] Build related changes
- [ ] CI related changes

## Summary

* Replace your-link.com in placeholders with site URL for more security

## Description

* Replace your-link.com in placeholders with site URL for more security (See also this comment https://github.com/elementor/elementor/pull/22429#discussion_r1211443638)

## Test instructions

*

## Quality assurance

Fixes #


rami-elementor:
Why not `https://elementor.com` ? or even better, the current site domain ?

DAnn2012:
Hi @Barcov and @yotamselementor  ,

please could you kindly take a look at my PR?

Thank you.

Barcov:
@rami-elementor excellent suggestion. 

@DAnn2012 Thanks for the work you've done on this PR. You've pointed out something I've not considered and this lead us to the fix Rami suggests. I think by using the actual Site URL as the placeholder the worst inexperienced user can do is redirect to their home page. 

In the case where we use the site URL http would also be valid if there is no SSL certificate for the domain. 

I've included code suggestions on each of the fixes to make it easier to change inline or you can submit the changes via your branch. I'm happy either way.

Thanks for contributing to Elementor to help improve it.

Barcov:
> 'placeholder' => esc_html__( 'https://your-link.com', 'elementor' ),

@DAnn2012  I suggest you replace it with `'placeholder' => esc_html__( site_url(), 'elementor' ),`

I've tested this replacement and it works well. 

[esc_html__()](https://developer.wordpress.org/reference/functions/esc_html/) has an important security role here to sanitise URLs.  
The 'elementor' part in the placeholder is a [textdomain that's part of how controls are constructed to avoid naming conflicts](https://developers.elementor.com/docs/widgets/widget-controls/).  The concept of textdomains is similar to Php Namespaces.

DAnn2012:
@Barcov Thank you for the explanation.

But I would like to ask you, isn't `esc_html(` enough as this is not a translatable string to include in the .POT / .PO / .MO files and maybe it could create problems with the strings on translate.wordpress.org?

I've seen that in these other Elementor files only `esc_html(` is used when it comes to non-translatable strings:

https://github.com/search?q=repo%3Aelementor%2Felementor%20esc_html(&type=code

Don't you think `'placeholder' => esc_html( site_url() ),` would be just as good?

Thank you.

rami-elementor:
I suggest using:

```php
'placeholder' => get_home_url(),
```

The `get_site_url()` return the location of WP installation, if WP installed in sub-folder you won't see the correct URL. That is why I prefer to use `get_home_url()`.

Another suggestion is to delete all the `placeholder` attributes from all these controls and add a placeholder to the [URL control](https://developers.elementor.com/docs/editor-controls/control-url/), this way we won't need to set a placeholder for each URL control, it will be defined globally.

DAnn2012:
@rami-elementor As you suggested, I did the replacement with `get_home_url()`.

Thanks.

rami-elementor:
After talking with our product team, it seems like you raised an interesting point, but from a different perspective.

The **[URL Control](https://developers.elementor.com/docs/editor-controls/control-url/)** has a default placeholder "**Paste URL or type**", but in many widgets we replace the original placeholder with "https://your-link.com". Doing so we hide the text which mentions the fact that you can type to search for posts/pages.

We believe that hiding the type option is bad for UX, therefore we should remove all the custom placeholders from the URL controls in all the widgets.

Please remove all the placeholders.

DAnn2012:
@rami-elementor As you asked, I removed all placeholders.

Thank you.

rami-elementor:
@DAnn2012 I can't merge the PR, please sign again on the CLA.

rami-elementor (approved):

DAnn2012:
@rami-elementor 

I just checked, and for the Elementor repository it is active since May 11, 2023.

What should I do?

PR #22725, opened 2023-06-09
Fix: Video background added [...] only works on the first tab in Tabs widget [ED-10932]

hein-obox:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #


modules/nested-tabs/assets/js/frontend/handlers/nested-tabs.js
```diff
@@ -194,6 +195,8 @@ export default class NestedTabs extends Base {
 			() => this.onShowTabContent( $requestedContent ),
 		);
 		$requestedContent.removeAttr( 'hidden' );
+
+		this.handleElementHandlers();
 	}
 
 	onShowTabContent( $requestedContent ) {
```
louiswol94, line 202: @hein-obox I found this issue can be fixed with a one liner 😄 

In `nested-tabs.js` add `elementorFrontend.elements.$window.trigger( 'resize' );` into `onShowTabContent()`

```
onShowTabContent( $requestedContent ) {
        elementorFrontend.elements.$window.trigger( 'elementor-pro/motion-fx/recalc' );
        elementorFrontend.elements.$window.trigger( 'elementor/nested-tabs/activate', $requestedContent );
        elementorFrontend.elements.$window.trigger( 'resize' );
}
```

I got the inspiration from here:
https://github.com/elementor/elementor-pro/blob/main/modules/video-playlist/assets/js/frontend/base-tabs.js#L137

Please check it and let me know. Thanks!

louiswol94 (changes requested):

modules/nested-tabs/assets/js/frontend/handlers/nested-tabs.js
hein-obox: @louiswol94 

Very interesting!

I am hesitant to implement it though, for 2 reasons: 
1. This would trigger all resize events on the page. I fear that will trigger many unneccessary events.
2. The background video handler by default will resize all videos when it gets triggered. I have made our change event call conditional. In my code only one specific video will be resized.

I have removed the run event handler function (as per your example) and trigger the background resize from the nested tabs immediately.

louiswol94: @hein-obox Not sure I agree. If this is the solution in the base tabs, then it should be safe for us to use, it was just lost because we overwrote the function.

louiswol94: @hein-obox I see we did not extend it yes ;-) But the fact that this is used in the base tabs means that it should be a viable solution. Please implement it 🙏 It will make this PR a lot less complex. Thanks.

hein-obox: @louiswol94 Implemented. Advantage: we are 100% that this solution works, as this solution has been used for a long time already.

louiswol94 (dismissed):

modules/nested-tabs/assets/js/frontend/handlers/nested-tabs.js
```diff
@@ -199,6 +199,7 @@ export default class NestedTabs extends Base {
...
 		elementorFrontend.elements.$window.trigger( 'elementor-pro/motion-fx/recalc' );
 		elementorFrontend.elements.$window.trigger( 'elementor/nested-tabs/activate', $requestedContent );
+		elementorFrontend.elements.$window.trigger( 'resize' );
```
davseve, line 202: How resizing the window is related to the video?

tests/playwright/sanity/modules/nested-tabs/nested-tabs.test.js
```diff
@@ -1109,8 +1113,35 @@ test.describe( 'Nested Tabs tests @nested-tabs', () => {
 		await page.locator( '.eicon-video-camera' ).first().click();
 		await page.locator( '.elementor-control-background_video_link input' ).fill( videoUrl );
 
+		await editor.selectElement( contentContainerTwoId );
+		await editor.activatePanelTab( 'style' );
+		await page.locator( '.eicon-video-camera' ).first().click();
+		await page.locator( '.elementor-control-background_video_link input' ).fill( videoUrl );
```
davseve, line 1119:
```
await page.locator( '.eicon-video-camera' ).first().click();
await page.locator( '.elementor-control-background_video_link input' ).fill( videoUrl );
```
Suggestion.
I see that the lines above are used in several places. 
We can extract it to a local function.

tests/playwright/sanity/modules/nested-tabs/nested-tabs.test.js
```diff
@@ -1109,8 +1113,35 @@ test.describe( 'Nested Tabs tests @nested-tabs', () => {
...
+
+		await editor.selectElement( contentContainerThreeId );
+		await editor.activatePanelTab( 'style' );
+		await page.locator( '.eicon-video-camera' ).first().click();
+		await page.locator( '.elementor-control-background_video_link input' ).fill( videoUrl );
+
 		await expect( contentContainerOne ).toHaveAttribute( 'data-model-cid', firstTabContainerModelCId );
 		await expect( videoContainer ).toHaveCount( 1 );
+
+		await page.waitForTimeout( 3000 );
```
davseve, line 1129: `waitForTimeout` is not the best practice but sometimes there are no other options.  
What `3000ms` means?
If we are using it in several places we can set it as a variable.

modules/nested-tabs/assets/js/frontend/handlers/nested-tabs.js
louiswol94: @davseve

1) You will see `base-tabs.js` also runs `elementorFrontend.elements.$window.trigger( 'resize' );` inside `activateTab()`. We decided to match this functionality in nested-tabs. Certain handlers listen for this event.

2) `background-video.js` listens for the `resize` event to calculate the video size. The problem was that the video width/height was on 0px. So we needed to get `this.changeVideoSize` to run. https://github.com/elementor/elementor/blob/main/assets/dev/js/frontend/handlers/background-video.js#L255

tests/playwright/sanity/modules/nested-tabs/nested-tabs.test.js
hein-obox: @davseve Ideally I would wait for the iframe width value to be updated. Do you have an idea how we could wait for that value to be updated inside PlayWright?

davseve: I don't have a good suggestion, maybe using a `mutationObserver` but it will be an overkill

tests/playwright/sanity/modules/nested-tabs/nested-tabs.test.js
hein-obox: Done

modules/nested-tabs/assets/js/frontend/handlers/nested-tabs.js
hein-obox: See Louis' comment.

davseve (approved):

PR #22524, opened 2023-05-22
Tweak: Use CSS logical properties in Elementor Editor [ED-10827]

rami-elementor:
Related: #22068

PR #22735, opened 2023-06-11
Tweak: Use CSS logical properties in Elementor App [ED-11020]

rami-elementor:
Related:

- https://github.com/elementor/elementor/pull/22524
- https://github.com/elementor/elementor/pull/22068


app/assets/js/app.scss
```diff
@@ -25,17 +25,18 @@
...
 		flex-shrink: 0;
 		font-size: $app-header-font-size;
-		border-bottom: var(--e-a-border);
+		border-block-end: var(--e-a-border);
 		position: relative;
 		z-index: $app-header-z-index;
 		height: $app-header-height;
-		padding: $app-header-padding-y $spp-header-padding-x;
+		padding-block: $app-header-padding-block;
+		padding-inline: $app-header-padding-inline;
```
rami-elementor, line 39: Had to split this after a manual test of the screen...

app/assets/js/app.scss
```diff
@@ -25,17 +25,18 @@
 		background-color: var(--app-lightbox-background-color);
 		z-index: $app-lightbox-z-index;
 		bottom: 0;
```
arielk, line 27: What about bottom?

app/assets/js/ui/card/card.scss
```diff
@@ -58,7 +58,7 @@ $root: $eps-prefix + card;
 		object-position: top;
 		position: absolute;
 		top: 0;
-		left: 0;
```
arielk, line 61: Please review it in the front end

rami-elementor: Tested. Works as expected.

This change is for consistency:
- All `left: 0;` properties replaced with `@include start(0);`
- All `right: 0;` properties replaced with `@include end(0);`

We do the same in other places. In the App I found a few places where LRT properties have been hardcoded.

app/assets/js/app.scss
rami-elementor: I wanted to start refactoring all the `top`, `right`, `bottom` and `left` properties to `inset-*`. But this PR is becoming huge (76 files), so I separated the work to several tasks.

@arielk If your prefer to do this in a single PR, I can add the changes to this PR. Your call.

app/assets/js/ui/card/card.scss
rami-elementor: @arielk 

In this specific case, it doesn't matter if we use `left` or `right` because we have `width: 100%;` and `position: absolute;`. In both cases it will cover 100% no matter where is the starting point.

arielk (approved):

arielk (approved):

PR #22804, opened 2023-06-18
Tweak: [NA] Change default style values [ED-11051]

TzviRabinovitch:


modules/nested-accordion/widgets/nested-accordion.php
```diff
@@ -439,6 +439,23 @@ private function add_content_style_section() {
 			[
 				'name' => 'content_border',
 				'selector' => $low_specificity_accordion_item_selector,
+				'fields_options' => [
+					'color' => [
```
davseve, line 443: We usually set the style defaults in the CSS file itself. 
But I guess this is valid as well.

modules/nested-accordion/widgets/nested-accordion.php
```diff
@@ -439,6 +439,23 @@ private function add_content_style_section() {
...
+							'top' => 1,
+							'right' => 1,
+							'bottom' => 1,
+							'left' => 1,
+							'unit' => 'px',
+						],
+					],
+					'border' => [
+						'default' => 'solid',
+					],
```
davseve, line 457: We usually set the style defaults in the CSS file itself.
But I guess this is valid as well.

davseve (approved):

PR #22805, opened 2023-06-18
Internal: Nested Accordion animation duration control [ED-11011]

davseve:


modules/nested-accordion/assets/js/frontend/handlers/nested-accordion.js
```diff
@@ -88,7 +86,7 @@ export default class NestedAccordion extends Base {
 
 		animation = accordionItem.animate(
 			{ height: [ startHeight, endHeight ] },
-			{ duration: ANIMATION_DURATION },
+			{ duration: this.animationDuration() },
```
TzviRabinovitch, line 89: maybe getanimationDuration ?

modules/nested-accordion/widgets/nested-accordion.php
```diff
@@ -314,6 +314,26 @@ protected function register_controls() {
...
+			'n_accordion_animation_duration',
+			[
+				'label' => esc_html__( 'Animation Duration', 'elementor' ),
+				'type' => Controls_Manager::SLIDER,
+				'size_units' => [ 's', 'ms' ],
+				'default' => [
+					'unit' => 'ms',
+					'size' => 400,
+				],
+				'range' => [
```
TzviRabinovitch, line 327: do we have a range for seconds?

TzviRabinovitch: 0-3 ?

modules/nested-accordion/assets/js/frontend/handlers/nested-accordion.js
davseve: Fixed

modules/nested-accordion/widgets/nested-accordion.php
davseve: 0-3 is the second's default value.

TzviRabinovitch (approved):

modules/nested-accordion/widgets/nested-accordion.php
```diff
@@ -314,6 +314,20 @@ protected function register_controls() {
 			]
 		);
 
+		$this->add_control(
+			'n_accordion_animation_duration',
+			[
+				'label' => esc_html__( 'Animation Duration', 'elementor' ),
+				'type' => Controls_Manager::SLIDER,
+				'size_units' => [ 's', 'ms' ],
```
rodolphebertozzo, line 322:
```suggestion
				'size_units' => [ 's', 'ms', 'custom' ],
```

rodolphebertozzo (changes requested):

modules/nested-accordion/widgets/nested-accordion.php
davseve: @rodolphebertozzo 
Thank's for noticing.
Great suggestion, This was the original product manager's suggestion.
In this case, the toggle animation is controlled by the Javascript animate API.
The animate API receives data in milliseconds.
Just for curiosity, which other time units a user would like to use?
Minutes? Hours?
ודא

rami-elementor: we have only `s` and `ms`, other using are css functions that return calculations of `s` and `ms`.

davseve: @rami-elementor so should we add custom units to this control?

PR #22874, opened 2023-06-23
Internal: Fix: Icon is not presented as expected on front in Nested Accordion widget [ED-11148]

Barcov:


modules/nested-accordion/widgets/nested-accordion.php
```diff
@@ -703,6 +703,10 @@ private function is_active_icon_exist( $settings ):bool {
 	}
 
 	private function render_accordion_icons( $settings ) {
+		if ( '' === $settings['accordion_item_title_icon']['value'] ) {
+			return '';
```
louiswol94, line 707:
```suggestion
			return;
```
@Barcov Maybe?

modules/nested-accordion/widgets/nested-accordion.php
```diff
@@ -711,8 +715,8 @@ private function render_accordion_icons( $settings ) {
 		ob_start();
 		?>
 		<span class='e-n-accordion-item-title-icon'>
-			<span class='e-opened' ><?php echo wp_kses_post( $icon_active_html ); ?></span>
-			<span class='e-closed'><?php echo wp_kses_post( $icon_html ); ?></span>
+			<span class='e-opened' ><?php echo ( $icon_active_html ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?></span>
```
louiswol94, line 718:
```suggestion
			<span class='e-opened' ><?php echo $icon_active_html; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?></span>
```

modules/nested-accordion/widgets/nested-accordion.php
```diff
@@ -711,8 +715,8 @@ private function render_accordion_icons( $settings ) {
...
+			<span class='e-closed'><?php echo ( $icon_html ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?></span>
```
louiswol94, line 719:
```suggestion
			<span class='e-closed'><?php echo $icon_html; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?></span>
```

modules/nested-accordion/widgets/nested-accordion.php
```diff
@@ -754,7 +758,9 @@ protected function render() {
 				<details <?php echo wp_kses_post( $title_render_attributes ); ?>>
 					<summary class='e-n-accordion-item-title'>
 						<span class='e-n-accordion-item-title-header'><?php echo wp_kses_post( "<$title_html_tag class=\"e-n-accordion-item-title-text\"> $item_title </$title_html_tag>" ); ?></span>
-						<?php echo wp_kses_post( $icons_content ); ?>
+						<?php if ( '' !== $icons_content ) {
+							echo( $icons_content ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
louiswol94, line 762:
```suggestion
							echo $icons_content; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```

modules/nested-accordion/widgets/nested-accordion.php
```diff
@@ -754,7 +758,9 @@ protected function render() {
 				<details <?php echo wp_kses_post( $title_render_attributes ); ?>>
 					<summary class='e-n-accordion-item-title'>
 						<span class='e-n-accordion-item-title-header'><?php echo wp_kses_post( "<$title_html_tag class=\"e-n-accordion-item-title-text\"> $item_title </$title_html_tag>" ); ?></span>
-						<?php echo wp_kses_post( $icons_content ); ?>
+						<?php if ( '' !== $icons_content ) {
```
louiswol94, line 761:
```suggestion
						<?php if ( ! empty( $icons_content ) ) {
```
Maybe?

modules/nested-accordion/widgets/nested-accordion.php
```diff
@@ -703,6 +703,10 @@ private function is_active_icon_exist( $settings ):bool {
 	}
 
 	private function render_accordion_icons( $settings ) {
+		if ( '' === $settings['accordion_item_title_icon']['value'] ) {
```
louiswol94, line 706:
```suggestion
		if ( empty( $settings['accordion_item_title_icon']['value'] ) ) {
```

louiswol94: @Barcov You can maybe even put this `if()` in line 761. Then this function does not need an early return. And also if `$settings['accordion_item_title_icon']['value']` is not empty, then there will always be content returned, which is why you can replace the if() on line 761 with this if(). WDYT?

louiswol94 (changes requested):

modules/nested-accordion/widgets/nested-accordion.php
Barcov: @louiswol94 Yes I like that suggestion. I generally try to keep the html portion of render quite lean. But in this case the if on 706 does feel redundant.

louiswol94 (approved):

hein-obox (approved):

PR #22906, opened 2023-06-26
Fix: Responsive settings for Section Templates wont fully save [ED-11103]

maork-elementor:


core/files/css/base.php
```diff
@@ -643,6 +643,34 @@ protected function is_update_required() {
...
+	 *
+	 * @since 3.14.0
+	 *
+	 * @access private
+	 */
+	private function clear_stacks_on_additional_breakpoints_active() {
+		$is_edit_mode = Plugin::$instance->editor->is_edit_mode();
+		if ( ! $is_edit_mode && method_exists( Plugin::$instance->controls_manager, 'clear_stack_cache' ) ) {
+			$currnet_post_id = null;
+			if ( method_exists( $this, 'get_post_id' ) ) {
```
matipojo, line 662: Why do you need to check if `method_exists`?

includes/managers/controls.php
```diff
@@ -301,8 +301,11 @@ class Controls_Manager {
 	 *
 	 * @var array
 	 */
+	// TODO: Remove in the future. - When Pro will remove the template class use in location manager.
```
matipojo, line 304: It's a private property, and it's safe to remove it

matipojo (changes requested):

includes/managers/controls.php
maork-elementor: In pro version, there is a use in :

https://github.com/elementor/elementor-pro/blob/c35e2084b5884c3b92de9da3817655a0c10e2432/modules/theme-builder/classes/locations-manager.php

of 

https://github.com/elementor/elementor-pro/blob/6ab8d9dc94328c75ba80396c64992fe51dd63137/modules/theme-builder/files/css/template.php

with the next function 

https://github.com/elementor/elementor/blob/3994d6d3c801e2234f60a0f150630061fd54f20c/includes/managers/controls.php#L894

That has support only for one time flag value update per page render per document.
So I need to support both ways until we will remove the support from the Pro version.
Then the location manager in pro will inherit the new logic from base that support multipal page ID's support.

https://github.com/maork-elementor/elementor/blob/b36f98759cad724e413d10200400ae1bb2aabd89/includes/managers/controls.php#L906

core/files/css/base.php
maork-elementor: Because not in every iteration, we actually have `post_id`

matipojo: A parent class shouldn't know about his children, so this code should be moved to the child that always has the method

core/files/css/base.php
```diff
@@ -643,6 +643,34 @@ protected function is_update_required() {
...
+	 * the templates responsive setting will not be updated unless the 'Regenerate CSS' option is used.
+	 * Reproduce only when additional breakpoints are active.
+	 *
+	 * @since 3.14.0
+	 *
+	 * @access private
+	 */
+	private function clear_stacks_on_additional_breakpoints_active() {
+		$is_edit_mode = Plugin::$instance->editor->is_edit_mode();
+		if ( ! $is_edit_mode && method_exists( Plugin::$instance->controls_manager, 'clear_stack_cache' ) ) {
```
matipojo, line 660: Why do you need to check if `method_exists`?

matipojo (changes requested):

includes/managers/controls.php
matipojo: It can't be an external direct use of a private property.

core/files/css/base.php
maork-elementor: Fixed.

core/files/css/base.php
maork-elementor: Removed.

matipojo (changes requested):

includes/managers/controls.php
maork-elementor: In my understanding, we will continue to support each other in both directions?
I have removed the comment.

core/files/css/base.php
```diff
@@ -643,6 +643,42 @@ protected function is_update_required() {
 		return false;
 	}
 
+	/**
```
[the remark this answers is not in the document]
maork-elementor: Nice, looks great.

core/files/css/base.php
```diff
@@ -643,6 +643,28 @@ protected function is_update_required() {
...
+	 *
+	 * If a page contains templates, such as loop items, shortcodes, kits, or template widgets,
+	 * the templates responsive setting will not be updated unless the 'Regenerate CSS' option is used.
+	 * Reproduce only when additional breakpoints are active.
+	 *
+	 * @since 3.14.0
+	 *
+	 * @access private
+	 */
+	private function clear_stacks() {
```
matipojo, line 657: I think the full logic is not related to here, so please move the logic to the app/module layers.

You can work with hooks, for example, on `parse_content`, trigger an action: 
```
do_action( 'elementor/css_file/parse_content' );
```
Then from `additional_custom_breakpoints` experiment, run logic and then `clear_stack_cache`:
```
add_action( 'elementor/css_file/parse_content', function() { 
	if...
	Plugin::$instance->controls_manager->clear_stack_cache( $handle_id );
});
```

matipojo (changes requested):

core/files/css/base.php
```diff
@@ -652,6 +652,9 @@ protected function is_update_required() {
 	 * @access protected
 	 */
 	protected function parse_content() {
+
+		do_action( 'elementor/css_file/parse_content' );
```
matipojo, line 656:
```suggestion
		do_action( 'elementor/css_file/parse_content', $this );
```

core/breakpoints/manager.php
```diff
@@ -534,4 +534,36 @@ private static function get_stylesheet_templates() {
...
+			$post_id = $current_document->get_main_id();
+			return 'elementor-post-' . $post_id;
+		}
+
+		return false;
+	}
+
+	public function __construct() {
+		add_action( 'elementor/css_file/parse_content', function() {
+			$handle_id = $this->get_file_handle_id();
```
matipojo, line 560:
```suggestion
		add_action( 'elementor/css_file/parse_content', function( $css_file ) {
			$handle_id = $css_file->get_file_handle_id();
```

matipojo (changes requested):

core/breakpoints/manager.php
```diff
@@ -534,4 +534,36 @@ private static function get_stylesheet_templates() {
...
+	protected function get_file_handle_id() {
+		$current_document = Plugin::$instance->documents->get_current();
+		if ( $current_document ) {
+			$post_id = $current_document->get_main_id();
+			return 'elementor-post-' . $post_id;
+		}
+
+		return false;
+	}
+
```
matipojo, line 557: Remove

matipojo (changes requested):

core/files/css/base.php
```diff
@@ -652,6 +652,9 @@ protected function is_update_required() {
 	 * @access protected
 	 */
 	protected function parse_content() {
+
```
matipojo, line 655:
```suggestion
```

matipojo (changes requested):

core/breakpoints/manager.php
```diff
@@ -534,4 +534,36 @@ private static function get_stylesheet_templates() {
...
+	public function __construct() {
+		add_action( 'elementor/css_file/parse_content', function( $css_file ) {
+			$handle_id = $css_file->get_file_handle_id();
+			$is_edit_mode = Plugin::$instance->editor->is_edit_mode();
+			$additional_breakpoints_active = Plugin::$instance->experiments->is_feature_active( 'additional_custom_breakpoints' );
+
+			if ( ! $is_edit_mode && $handle_id && $additional_breakpoints_active ) {
+				Plugin::$instance->controls_manager->clear_stacks( $handle_id );
+			}
+		});
```
matipojo, line 567: PHPCS

maork-elementor: PHPCS saying everything is linted.

core/breakpoints/manager.php
maork-elementor: As a result of these changes, it looks like this :
Template widget and shortcode widget stop generating responsive controls, like in this initial state. 

I'm looking into this at the moment

includes/managers/controls.php
```diff
@@ -881,8 +881,12 @@ public function remove_control_from_stack( $stack_id, $control_id ) {
...
 	 * @return bool True if the CSS requires to clear the controls stack cache, False otherwise.
 	 */
-	public function has_stacks_cache_been_cleared() {
-		return $this->has_stacks_cache_been_cleared;
+	public function has_stacks_cache_been_cleared( $handle_id = 'default' ) {
+		if ( ! isset( $this->has_post_stacks_cache_been_cleared[ $handle_id ] ) ) {
+			return false;
+		}
+
+		return true;
```
matipojo, line 889:
```suggestion
		return isset( $this->has_post_stacks_cache_been_cleared[ $handle_id ] );
```

includes/managers/controls.php
```diff
@@ -1174,4 +1179,24 @@ public function add_custom_attributes_controls( Controls_Stack $controls_stack )
...
+	 * @since 3.14.0
+	 *
+	 * @access private
+	 */
+	public function clear_stacks( $handle_id ) {
+		if ( $handle_id ) {
+			$stack_cache_has_been_cleared = $this->has_stacks_cache_been_cleared( $handle_id );
+			if ( ! $stack_cache_has_been_cleared ) {
+				$this->clear_stack_cache( $handle_id );
+			}
```
matipojo, line 1199:
```suggestion
			if ( ! $this->has_stacks_cache_been_cleared( $handle_id ) ) {
				$this->clear_stack_cache( $handle_id );
			}
```

tests/phpunit/elementor/includes/managers/test-controls.php
```diff
@@ -105,6 +105,72 @@ public function test_clear_stack_cache() {
...
+		$this->assertNotEmpty( $stacks );
+
+		// Assert
+		$stack_cache_has_been_cleared = Plugin::$instance->controls_manager->has_stacks_cache_been_cleared( $handle_id );
+		$this->assertFalse( $stack_cache_has_been_cleared );
+
+		// Act
+		if ( ! $stack_cache_has_been_cleared ) {
+			Plugin::$instance->controls_manager->clear_stack_cache( $handle_id );
+		}
```
matipojo, line 164: It's already asserted as false.
```suggestion
		Plugin::$instance->controls_manager->clear_stack_cache( $handle_id );
```

matipojo (changes requested):

matipojo (approved):

PR #22949, opened 2023-07-02
Fix: Wrong video lightbox aspect ratio when multiple video widgets used on the same page [ED-10146]

rami-elementor:


bainternet (dismissed):

arielk (dismissed):

includes/widgets/video.php
```diff
@@ -873,6 +873,7 @@ protected function register_controls() {
 			[
 				'label' => esc_html__( 'Content Width', 'elementor' ),
 				'type' => Controls_Manager::SLIDER,
+				'description' => esc_html__( 'This control was deprecated.', 'elementor' ),
```
arielk, line 876: Please add an alert class to this description with alert type. In addition display this control for websites that already used this control only.

arielk: Add copy team to change the text

rami-elementor: Done.

arielk (approved):

PR #22952, opened 2023-07-02
Tweak: Remove deprecated schemes area [ED-9386]

KingYes:


core/upgrade/upgrades.php
```diff
@@ -945,14 +787,6 @@ public static function _v_3_4_8_fix_font_awesome_default_value_from_1_to_yes() {
 		}
 	}
 
-	public static function _v_3_5_0_remove_old_elementor_scheme() {
```
Nevoss, line 948: Why not keep it? if users comes from old version so the schema will be removed

includes/plugin.php
```diff
@@ -754,7 +739,6 @@ private function init_components() {
 			$this->wordpress_widgets_manager = new WordPress_Widgets_Manager();
 			$this->admin = new Admin();
 			$this->beta_testers = new Beta_Testers();
-			new Elementor_3_Re_Migrate_Globals();
```
Nevoss, line 757: Maybe we still need it?

Nevoss (changes requested):

includes/plugin.php
KingYes: It's used with a deprecation functions.

Nevoss (dismissed):

matipojo (changes requested):
Please keep the PHP public API, so, old plugins and themes that calling [`Elementor::instance()->schemes_manager`](https://github.com/search?type=code&q=-%3Eschemes_manager) will not throw a fatal error.

arielk (approved):

matipojo (approved):

PR #22979, opened 2023-07-06
Internal: Preparation for Nested Accordion [ED-11286]

TzviRabinovitch:


davseve (changes requested):
If the nested accordion exp doesn't exist anymore we should change the `setExperiment` function so it will not include the ansted accordion experiment.

tests/playwright/sanity/modules/nested-accordion/nested-accordion.test.js
```diff
@@ -5,27 +5,14 @@ import { borderStyle } from '../../../enums/border-styles';
...
-
-			await page.close();
-		} );
-
-		test.afterAll( async ( { browser }, testInfo ) => {
-			const context = await browser.newContext();
-			const page = await context.newPage();
-			const wpAdmin = new WpAdminPage( page, testInfo );
-			await wpAdmin.setExperiments( {
 				'nested-elements': 'inactive',
```
davseve, line 15: The nested elements experiment is active by default?

davseve (approved):

PR #23024, opened 2023-07-11
Tweak: Improved Ajax permissions functionality for better security enforcement [ED-11233]

StyleShit:


includes/api.php
```diff
@@ -277,6 +277,10 @@ public static function send_feedback( $feedback_key, $feedback_text ) {
 	public static function ajax_reset_api_data() {
 		check_ajax_referer( 'elementor_reset_library', '_nonce' );
 
+		if ( ! current_user_can( 'manage_options' ) ) {
+			wp_send_json_error( 'Permission denied' );
+		}
```
Nevoss, line 282: is only admin has access to the library?

Nevoss (changes requested):

includes/api.php
StyleShit: it's the reset library button under admin tools

Nevoss (approved):

PR #23025, opened 2023-07-11
Tweak: Improve render issue when changing title name in Accordion widget [ED-11315]

Barcov:


modules/nested-accordion/widgets/nested-accordion.php
```diff
@@ -271,7 +271,6 @@ protected function register_controls() {
 				],
 				'default' => 'div',
 				'separator' => 'before',
-				'render_type' => 'template',
```
louiswol94, line 274: @Barcov Not sure this is correct? This will cause the HTML tag to not change in the Editor. The data bindings should fix the issue. WDYT?

modules/nested-accordion/widgets/nested-accordion.php
```diff
@@ -767,20 +766,34 @@ protected function content_template() {
...
+			});
+
+			view.addRenderAttribute( itemTitleText, {
+				'class': ['e-n-accordion-item-title-text'],
+				'data-binding-type': 'repeater-item',
+				'data-binding-repeater-name': 'items',
+				'data-binding-setting': ['item_title'],
+				'data-binding-index': itemCount,
+			});
+
```
louiswol94, line 796: @Barcov Spacing/Indentation issue

modules/nested-accordion/widgets/nested-accordion.php
```diff
@@ -822,7 +835,7 @@ protected function content_template() {
 			</details>
 			<# } ); #>
 		<# } #>
-	</div>
+		</div>
```
louiswol94, line 838: @Barcov ?

louiswol94 (changes requested):
@Barcov Why did you have to update a screenshot?

modules/nested-accordion/widgets/nested-accordion.php
Barcov: @louiswol94 Databindings have fixed most of this PR. The version you tested had a no render_type(). But I did not see the heading tags update in Chromium as discussed. I've done some testing. Adding `render-type => template` back in has solved the tag update in the Dom when I'm in the editor. I've added it back in.

Barcov:
> @Barcov Why did you have to update a screenshot?

@louiswol94 attempt to fix issues fixed in the E2E PR so this one can be unblocked. Been waiting a while to merge this one.

louiswol94 (dismissed):

modules/nested-accordion/widgets/nested-accordion.php
```diff
@@ -802,48 +802,60 @@ protected function content_template() {
...
+				const itemCount = index + 1,
+					itemUid = elementUid + itemCount,
+					itemWrapperKey = itemUid,
+					itemTitleKey = 'item-' + itemUid;
+
+					view.addRenderAttribute( itemTitleKey, {
+						'class': ['e-n-accordion-item-title'],
+					});
+
+					view.addRenderAttribute( itemTitleText, {
```
hein-obox, line 825: I would move this to after `view.addRenderAttribute( itemTitleKey` to keep the original structure of this code.

modules/nested-accordion/widgets/nested-accordion.php
```diff
@@ -802,48 +802,60 @@ protected function content_template() {
 		?>
 		<div class="e-n-accordion" role="tablist" aria-orientation="vertical">
 			<# if ( settings['items'] ) {
-			const elementUid = view.getIDInt().toString().substr( 0, 3 ),
+			const elementUid = view.getIDInt().toString().substring( 0, 3 ),
```
hein-obox, line 805: Is this change on purpose?

> The difference between substring() and substr()
The two parameters of substr() are start and length , while for substring() , they are start and end . substr() 's start index will wrap to the end of the string if it is negative, while substring() will clamp it to 0 .

Barcov: @hein-obox Yes this was purposeful, `substr` was depricated in js and replaced by `substring`

modules/nested-accordion/widgets/nested-accordion.php
Barcov: @hein-obox Good idea

hein-obox (approved):

PR #23040, opened 2023-07-12
Fix: WC page settings on Elementor site settings won't save correctly [ED-11236]

ArmanPurtoian:


louiswol94 (changes requested):
@ArmanPurtoian In general, I don't think this **save command** is supposed to be aware of any **specific settings**. So we should not do the `startsWith( 'woocommerce' )`.

Then, we should always remove the default, otherwise it will create a huge payload.

My suggestion is after removing the defaults, check that every item in `container.settings.changed` is in the `settings` object that will be saved. This will make sure any settings that were changed will be in the payload.

`document.container.settings.changed`:
```
{
    "woocommerce_myaccount_page_id": "9"
}
```

WDYT?

louiswol94 (changes requested):

assets/dev/js/editor/document/save/commands/internal/save.js
```diff
@@ -153,6 +155,14 @@ export class Save extends $e.modules.CommandInternalBase {
 	onAfterAjax( document ) {
 		document.editor.isSaving = false;
 	}
+
+	checkIfPersistentSettingsAddedToPayload( settings, container ) {
```
louiswol94, line 159:
```suggestion
	maybeAddPersistentSettingsToPayload( settings, container ) {
```

assets/dev/js/editor/document/save/commands/internal/save.js
```diff
@@ -153,6 +155,14 @@ export class Save extends $e.modules.CommandInternalBase {
 	onAfterAjax( document ) {
 		document.editor.isSaving = false;
 	}
+
+	addPersistentSettingsToPayload( settings, container ) {
+		elementor?.config?.persistent_keys?.forEach( ( setting ) => {
+			if ( container.settings.defaults.hasOwnProperty( setting ) && ! settings.hasOwnProperty( setting ) ) {
+				settings[ setting ] = container.settings.defaults[ setting ];
```
louiswol94, line 162:
```suggestion
				settings[ setting ] = container?.settings?.defaults[ setting ];
```

louiswol94 (approved):

TzviRabinovitch (approved):

PR #23105, opened 2023-07-19
Internal: Build Faster Playwright tests [ED-11343]

Barcov:
Speed up experiment activation for Playwright tests.

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -103,7 +103,7 @@ jobs:
           npx wp-env run cli "wp --info"
       - name: Activate performance experiments
         run: |
-          npx wp-env run cli "wp elementor experiments activate e_font_icon_svg,e_lazyload,e_dom_optimization,e_optimized_assets_loading,e_optimized_css_loading,additional_custom_breakpoints"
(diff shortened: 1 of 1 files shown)


.github/workflows/playwright.yml
```diff
@@ -103,7 +103,7 @@ jobs:
           npx wp-env run cli "wp --info"
       - name: Activate performance experiments
         run: |
-          npx wp-env run cli "wp elementor experiments activate e_font_icon_svg,e_lazyload,e_dom_optimization,e_optimized_assets_loading,e_optimized_css_loading,additional_custom_breakpoints"
+          npx wp-env run cli "wp elementor experiments activate e_font_icon_svg,e_lazyload,e_dom_optimization,e_optimized_assets_loading,e_optimized_css_loading,additional_custom_breakpoints --skip-themes --skip-plugins=woocommerce"
```
Barcov, line 106: Skip iterating through hooks for themes and woocommerce hooks when activating experiments.

Barcov: Execution time before:
image

Barcov: Execution time with skip themes and plugins 2s 539ms
image

Barcov: Execution time with skip themes only 2s 382ms
image

Barcov: Execution time with skip Woocommerce only 
image

Barcov:
There seems to be no benefit in skipping plugins and themes when activating Elementor experiments using the CLI. Local tests showed more promise since my Dev environment has more active themes and plugins.  Executing those options in the CI context just adds more execution time.

PR #23374, opened 2023-08-10
Internal: Container Upgrade scripts - Gap Control [ED-11606]

davseve:


murzin-nikita (approved):

core/upgrade/upgrades.php
```diff
@@ -854,6 +854,7 @@ public static function _v_3_16_0_container_updates( $updater ) {
 
 			$data = self::maybe_convert_to_inner_containers( $data );
 			$data = self::maybe_convert_to_grid_container( $data );
+			$data = self::maybe_convert_flex_gap_control( $data );
```
louiswol94, line 857: @davseve Nice work!!

@TzviRabinovitch @davseve I just noticed that we are iterating through the same data set 3 times with these functions.

```
$data = self::maybe_convert_to_inner_containers( $data );
$data = self::maybe_convert_to_grid_container( $data );
$data = self::maybe_convert_flex_gap_control( $data );
```

Not really the best for performance :-) But we can talk about it, I will approve this task for now.

davseve: I agree.
We should improve the functionality here.

PR #23386, opened 2023-08-11
Fix: The display for width units of 'Full-width' [ED-11381]

ArmanPurtoian:


includes/elements/container.php
```diff
@@ -466,6 +466,24 @@ protected function register_container_layout_controls() {
 			'separator' => 'none',
 		];
 
+		$active_breakpoints = Plugin::$instance->breakpoints->get_active_breakpoints();
```
louiswol94, line 469: @ArmanPurtoian `$active_breakpoints = Plugin::$instance->breakpoints->get_active_breakpoints();` has already been done on line 407. Can you use that variable?

louiswol94 (changes requested):

louiswol94 (approved):

hein-obox (approved):

PR #23399, opened 2023-08-13
Internal: Improved version control mechanism [ED-10790]

ghost:


includes/settings/tools.php
```diff
@@ -212,6 +215,11 @@ private function get_rollback_versions() {
 				return [];
 			}
 
+			$plugin_information->versions = apply_filters(
```
Nevoss, line 218: If there something returns from the filter there is no need to call WordPress API up there `$plugin_information = plugins_api(`

Nevoss (changes requested):

includes/settings/tools.php
ManorHazazE: done

includes/settings/tools.php
```diff
@@ -197,32 +208,41 @@ public function __construct() {
...
 			$current_index = 0;
-			foreach ( $plugin_information->versions as $version => $download_link ) {
+			foreach ( $versions as $version ) {
 				if ( $max_versions <= $current_index ) {
 					break;
 				}
 
+				if ( substr( $version, 0, 1 ) == 'v' ) {
+					$version = substr( $version, 1 );
+				}
```
Nevoss, line 244: Why needed?

includes/settings/tools.php
```diff
@@ -197,32 +208,41 @@ public function __construct() {
...
+					]
+				);
+
+				if ( empty( $plugin_information->versions ) || ! is_array( $plugin_information->versions ) ) {
+					return [];
+				}
 
-			uksort( $plugin_information->versions, 'version_compare' );
-			$plugin_information->versions = array_reverse( $plugin_information->versions );
+				uksort( $plugin_information->versions, 'version_compare' );
```
Nevoss, line 230: Are you bringing the versions in the right order?

Nevoss (changes requested):

includes/settings/tools.php
ManorHazazE: removed

includes/settings/tools.php
ManorHazazE: YES

Nevoss (approved):

PR #23405, opened 2023-08-14
Internal: Container Upgrade scripts Patch [ED-11617]

davseve:


tests/phpunit/elementor/core/upgrade/test-upgrades.php
```diff
@@ -453,4 +478,50 @@ private function create_image() {
...
 	}
+
+	/**
+	 * @param $documents
+	 *
+	 * @return void
+	 */
+	private function assert_flex_gap_control_has_changed( $elementor_data ) {
+		$top_level_container = $elementor_data[0];
+		self::assertEquals( [
```
mykytamurzin, line 489: is there any specific purpose to access assertion function through `self::` in non-static method? 
Let's do it using `$this` like in `test_v_3_15_9_container_updates`

davseve: Correct
Fixed

murzin-nikita (approved):

louiswol94 (approved):

PR #23409, opened 2023-08-14
Internal: Remove event listeners in grid container handler [ED-11747]

StyleShit:


assets/dev/js/frontend/handlers/container/grid-container.js
```diff
@@ -1,4 +1,11 @@
 export default class GridContainer extends elementorModules.frontend.handlers.Base {
+	__construct( settings ) {
```
Nevoss, line 2: What about other handlers?

Nevoss (changes requested):

assets/dev/js/frontend/handlers/container/grid-container.js
StyleShit: From what I saw, they're fine

Nevoss (approved):

PR #23414, opened 2023-08-14
Fix: Admin Top Bar - Remove from non Elementor pages [ED-11695]

Nevoss:


modules/admin-top-bar/module.php
```diff
@@ -88,22 +88,43 @@ private function add_frontend_settings() {
...
 
+	private function should_be_shown() {
+		$current_screen = get_current_screen();
+
+		if ( ! $current_screen ) {
+			return false;
+		}
+
+		$is_elementor_page = strpos( $current_screen->id ?? '', 'elementor' ) !== false;
+		$is_elementor_post_type_page = strpos( $current_screen->post_type ?? '', 'elementor' ) !== false;
```
StyleShit, line 99: hmm, what's gonna happen with 3rd party plugins that uses "elementor" in their name?

modules/admin-top-bar/module.php
```diff
@@ -88,22 +88,43 @@ private function add_frontend_settings() {
...
+
+		return apply_filters(
+			'elementor/admin-top-bar/should-be-shown',
```
StyleShit, line 102: IDK why, but this hook name sound weird ("should be shown"). maybe let's talk about it and try to find a different name?

StyleShit (changes requested):

modules/admin-top-bar/module.php
Nevoss: Done

modules/admin-top-bar/module.php
Nevoss: Done

StyleShit (approved):

PR #23420, opened 2023-08-15
Internal: POC - Upgrade conditions - Beta

louiswol94:


core/upgrade/conditions.php
```diff
@@ -0,0 +1,13 @@
...
+namespace Elementor\Core\Upgrade;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+class Conditions {
+
+	public static function _v_3_15_9_container_updates__should_run() {
+		return ( 'yes' === get_option( 'elementor_container_gap_updated' ) ) ? false : true;
```
louiswol94, line 11: `elementor_container_gap_updated` flag is added when the reversal process took place. And this flag is deleted again in 3.16

core/base/db-upgrades-manager.php
```diff
@@ -212,12 +214,32 @@ public function get_upgrade_callbacks() {
 				continue;
 			}
 
+			if ( ! $this->should_run( $method_name ) ) {
+				continue;
+			}
```
louiswol94, line 219: Don't add the callback in the first place, so the user won't be prompted.

PR #23422, opened 2023-08-15
Internal: Gaps control update script improvements [ED-11753]

davseve:


core/upgrade/upgrades.php
```diff
@@ -934,62 +936,51 @@ public static function get_post_ids_by_element_type( $updater, string $element_t
...
 
-				foreach ( $element['elements'] as &$inner_element ) {
-					if ( 'container' === $inner_element['elType'] && ! $inner_element['isInner'] ) {
-						$inner_element['isInner'] = true;
-					}
-				}
-
-				return $element;
+				$element = self::convert_to_inner_container( $element );
+				$element = self::convert_to_grid_container( $element );
```
louiswol94, line 948:
```suggestion
				$element = self::maybe_convert_to_inner_container( $element );
				$element = self::maybe_convert_to_grid_container( $element );
```

louiswol94: WDYT?

louiswol94 (changes requested):

core/upgrade/upgrades.php
davseve: Suggestion accepted

louiswol94 (approved):

hein-obox:
@davseve Would it be helpful to add a CSS fallback variable?

hein-obox (approved):

davseve:
@hein-obox 
Adding a CSS fallback won't help as we didn’t change the CSS props.
We are still using the gap prop.
In addition, if we will add a fallback it will return the default value as it won’t know how to interpret the new gaps value control.
Even if we will support the CSS, the values will still not appear in the new gaps control.
We need to handle the template import mechanism so it will convert the old slider control to the new gaps control, similar to what we did in the upgrade script.

PR #23449, opened 2023-08-17
Internal: Fix one click release process [ED-11789] 

davseve:
add default value to `OVERRIDE_PACKAGE_VERSION` and return the one click conditions in `get-previous-release/action.yml`


.github/workflows/get-previous-release/action.yml
```diff
@@ -5,16 +5,23 @@ inputs:
...
     - shell: bash
       run: |
         # If the channel is either "cloud" or "beta", fetch all Git tags matching the channel pattern,
-        # excluding any "-rc" tags, and store them in the 'tags' variable. e.g. "refs/tags/v3.11.0-cloud1 , refs/tags/v3.11.0-cloud2"
+        # excluding any "-rc" and "-test" tags, and store them in the 'tags' variable. e.g. "refs/tags/v3.11.0-cloud1 , refs/tags/v3.11.0-cloud2"
         if [[ ${{ inputs.CHANNEL }} == "cloud" || ${{ inputs.CHANNEL }} == "beta" ]]; then
-           tags=$(git ls-remote --tags | grep ${{ inputs.CHANNEL }} | grep -v "\-rc" | awk '{print $2}')
+          if [[ "${{ inputs.POSTFIX }}" != "-test" ]]; then
+            tags=$(git ls-remote --tags | grep ${{ inputs.CHANNEL }} | grep -v "\-rc" | grep -v "\-test" | awk '{print $2}')
+          elif [[ "${{ inputs.POSTFIX }}" == "-test" ]]; then
```
TzviRabinovitch, line 22: do we need elseif ?
why not just else ?

TzviRabinovitch (changes requested):

.github/workflows/get-previous-release/action.yml
TzviRabinovitch, line 22:
```suggestion
          else
```

davseve: Done, 
I've changed the condition so the `if` will check the positive scenario.

PR #23467, opened 2023-08-18
Tweak: Make sure the Container Gap does not break BC [ED-11791]

louiswol94:


murzin-nikita (approved):

includes/controls/groups/flex-container.php
```diff
@@ -164,6 +164,10 @@ protected function init_fields() {
 				'{{SELECTOR}}' => '--gap: {{ROW}}{{UNIT}} {{COLUMN}}{{UNIT}}',
 			],
 			'responsive' => true,
+			'conversion_map' => [
+				'old_key' => 'size',
+				'new_key' => 'column',
```
davseve, line 169: The new key is set to `column`, but shouldn't we add `row` as well?

louiswol94: @davseve Only the `column` value should be mapped to the `size` value. Of the two values we said we'll use the column value as the one to use for the downgrade.

davseve (approved):

TzviRabinovitch (approved):

PR #23516, opened 2023-08-22
Internal: Accurately Convert values from Normal to Custom option in Mobile [ED-11687]

louiswol94:


assets/dev/js/frontend/handlers/container/grid-container.js
```diff
@@ -68,7 +68,10 @@ export default class GridContainer extends elementorModules.frontend.handlers.Ba
 		// Re-init empty view element after container layout change
 		const selectors = this.getSettings( 'selectors' );
 		this.elements.emptyView = this.findElement( selectors.emptyView )[ 0 ];
-		this.elements.emptyView.style.display = this.shouldRemoveEmptyView() ? 'none' : 'block';
+
```
louiswol94, line 71: Unrelated bug fix

assets/dev/js/editor/controls/slider.js
```diff
@@ -213,6 +220,20 @@ ControlSliderItemView = ControlBaseUnitsItemView.extend( {
...
+
+	onDeviceModeChange() {
+		const currentDeviceMode = elementor.channels.deviceMode.request( 'currentMode' ),
+			isMobile = 'mobile' === currentDeviceMode,
+			isMobileValue = this.model.get( 'name' ).includes( '_mobile' ),
+			hasDefault = this.model.get( 'default' ),
+			shouldRunConversion = isMobile && isMobileValue && hasDefault && this.isCustomUnit();
+
+		if ( shouldRunConversion ) {
+			setTimeout( () => {
```
louiswol94, line 232: Defer to wait for render to complete (before `maybeDoFractionToCustomConversions()` might trigger a re-render).

louiswol94:
I did not add a test as this is a follow up task, and the original task already has Jest tests.

assets/dev/js/editor/controls/slider.js
```diff
@@ -200,8 +200,15 @@ ControlSliderItemView = ControlBaseUnitsItemView.extend( {
 			return;
 		}
 
+		const currentSize = this.getSize(),
+			alreadyConverted = 'string' === typeof currentSize && currentSize.includes( 'fr' );
```
mykytamurzin, line 204: could this check replace `isFrToCustom` on 199 line ?

louiswol94: @mykytamurzin Nope it can't. If I remove `isFrToCustom`, then when the user switches from `px -> custom` it will also run the logic. We don't know what the previous unit was at this point in execution, which is why we are checking that the control should only have `fr`, `custom` for this logic to run.

murzin-nikita (approved):

TzviRabinovitch (approved):

PR #23546, opened 2023-08-24
Internal: Convert Gap control for kits and templates [ED-11782]

davseve:


assets/dev/js/editor/controls/gaps.js
```diff
@@ -41,10 +41,6 @@ ControlGapItemView = ControlDimensionsView.extend( {
 
 		this.setSettingsModel( values );
 	},
-
-	getControlPlaceholder() {
-		return '';
-	},
```
davseve, line 47: Show gaps control placeholders

includes/controls/gaps.php
```diff
@@ -60,4 +59,74 @@ protected function get_dimensions() {
...
+						</button>
+					</li>
+				</ul>
+			</div>
+		</div>
+		<# if ( data.description ) { #>
+			<div class="elementor-control-field-description">{{{ data.description }}}</div>
+		<# } #>
+		<?php
+	}
```
davseve, line 131: Related to the placeholder appearence

includes/controls/gaps.php
```diff
@@ -60,4 +59,74 @@ protected function get_dimensions() {
 			'row' => esc_html__( 'Row', 'elementor' ),
 		];
 	}
+
```
TzviRabinovitch, line 62: extract to different PR and create a story if needed

modules/control-replacer/module.php
```diff
@@ -0,0 +1,33 @@
+<?php
```
TzviRabinovitch, line 1: refactor to a single-responsibility command  pattern

suggestions :
modules/control-converters/flex-container-gaps-converter.php

with an execute function that does the conversion

TzviRabinovitch (changes requested):

modules/control-replacer/module.php
louiswol94: Het @TzviRabinovitch I think this is overkill. `on_import()` is already like a command that is being executed, it does not really make sense to then execute another command. The logic can just be inside `on_import()` directly. But because the same logic is needed in the Upgrade scripts for example we need to either make it in a trait or just a static function inside the container.php file. It's basically just a small piece of logic that needs to be shared.

TzviRabinovitch (approved):

PR #23577, opened 2023-08-27
Internal: Fix gaps control placeholder display to main [ED-11951]

davseve:


TzviRabinovitch (dismissed):

includes/controls/gaps.php
```diff
@@ -41,13 +41,12 @@ public function get_type() {
...
-				'row' => '',
-				'isLinked' => true,
-			]
-		);
+		return [
+			'column' => '',
+			'row' => '',
+			'isLinked' => true,
+			'unit' => 'px',
+		];
```
louiswol94, line 49: Thank you!

includes/controls/gaps.php
```diff
@@ -60,4 +59,75 @@ protected function get_dimensions() {
...
+							<input id="<?php $this->print_control_uid( $dimension_key ); ?>" type="text" data-setting="<?php
+							// PHPCS - the variable $dimension_key is a plain text.
+							echo $dimension_key; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+							?>" placeholder="<#
+									placeholder = view.getControlPlaceholder();
+									if ( _.isObject( placeholder ) && ! _.isUndefined( placeholder.<?php
+									// PHPCS - the variable $dimension_key is a plain text.
+									echo $dimension_key; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+									?> ) ) {
+										console.log( {placeholder} )
```
louiswol94, line 94:
```suggestion
```

includes/controls/gaps.php
```diff
@@ -60,4 +59,75 @@ protected function get_dimensions() {
...
+	 * Render gaps control output in the editor.
+	 *
+	 * Used to generate the control HTML in the editor using Underscore JS
+	 * template. The variables for the class are available using `data` JS
+	 * object.
+	 *
+	 * @since 3.16.0
+	 * @access public
+	 */
+	public function content_template() {
```
louiswol94, line 73: Hey @davseve 
 What was the fix? What changed in the `content_template()` that differs from the parent?

louiswol94 (changes requested):

includes/controls/gaps.php
davseve: I just saw it :)
Removed

includes/controls/gaps.php
davseve: We had to do it sometime :)

includes/controls/gaps.php
davseve: In the dimensions, we had a condition that tried to print the entire object so I tweaked the condition.

louiswol94: @davseve If that is the only change, can you just make a `protected function print_placeholder() {}`
Then in the `gaps.php` file you can just override this small function instead of all this code duplication?

louiswol94 (approved):

PR #23591, opened 2023-08-28
Internal: Convert the site settings Gap control for kits [11952]

davseve:


app/modules/import-export/utils.php
```diff
@@ -115,4 +115,13 @@ public static function get_import_sessions( $should_run_cleanup = false ) {
...
 	}
+
+	public static function update_space_between_widgets_values( $space_between_widgets ) {
+		$size = strval( $space_between_widgets['size'] );
+		$space_between_widgets['column'] = $size;
+		$space_between_widgets['row'] = $size;
+		$space_between_widgets['isLinked'] = true;
+
+		return $space_between_widgets;
+	}
```
louiswol94, line 126: @davseve You need a safeguard in place if `$space_between_widgets['size']` does not exist.

tests/phpunit/elementor/core/utils/import-export/test-site-settings.php
```diff
@@ -0,0 +1,30 @@
+<?php
```
louiswol94, line 1: @davseve I don't see a need for this test because I already covered this in my test. The only added value would be if you can actually Act the whole import process...

louiswol94 (changes requested):

app/modules/import-export/utils.php
davseve: Done

tests/phpunit/elementor/core/utils/import-export/test-site-settings.php
davseve: Test removed

core/kits/documents/tabs/settings-layout.php
```diff
@@ -103,7 +103,7 @@ protected function register_tab_controls() {
 			: esc_html__( 'Widgets Space', 'elementor' );
 
 		$this->add_control(
-			'space_between_widgets',
+			'Gaps',
```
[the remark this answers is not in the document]
davseve: Good save!!
my mistake

louiswol94 (dismissed):

louiswol94 (approved):

PR #23592, opened 2023-08-28
Internal: Convert the site settings Gap control for kits to beta [11952]

davseve:


app/modules/import-export/utils.php
```diff
@@ -115,4 +115,20 @@ public static function get_import_sessions( $should_run_cleanup = false ) {
 
 		return $import_sessions;
 	}
+
+	public static function update_space_between_widgets_values( $space_between_widgets ) {
+		$setting_not_exist = ! isset( $space_between_widgets['size'] );
```
TzviRabinovitch, line 120: maybe use positive 
settting_exist 
?

core/upgrade/upgrades.php
```diff
@@ -882,10 +883,8 @@ private static function maybe_add_gap_control_data( $option_name ) {
 			return;
 		}
 
-		$size = strval( $kit_data_array['space_between_widgets']['size'] );
-		$kit_data_array['space_between_widgets']['column'] = $size;
-		$kit_data_array['space_between_widgets']['row'] = $size;
-		$kit_data_array['space_between_widgets']['isLinked'] = true;
+		$kit_data_array['space_between_widgets'] = Utils::update_space_between_widgets_values( $kit_data_array['space_between_widgets'] );
+
```
TzviRabinovitch, line 887: do we need this empty line ?

TzviRabinovitch (changes requested):

app/modules/import-export/utils.php
davseve: OK, But we will have to negate it in the condition.

TzviRabinovitch: it wont work otherwise :)

davseve: We can't always be positive :)

TzviRabinovitch (dismissed):

TzviRabinovitch (dismissed):

app/modules/import-export/utils.php
louiswol94: LOL

louiswol94 (approved):

PR #23641, opened 2023-09-01
Tweak: Added a FAQ Schema support for the Nested Accordion widget [ED-11948]

Barcov:


modules/nested-accordion/widgets/nested-accordion.php
```diff
@@ -277,6 +277,31 @@ protected function register_controls() {
 
 			]
 		);
+
+		$this->add_control(
+			'faq_schema',
+			[
+				'label' => esc_html__( 'FAQ Schema', 'elementor' ),
+				'type' => \Elementor\Controls_Manager::SWITCHER,
```
louiswol94, line 285:
```suggestion
				'type' => Controls_Manager::SWITCHER,
```

modules/nested-accordion/widgets/nested-accordion.php
```diff
@@ -799,6 +828,28 @@ protected function render() {
 			<?php echo $items_title_html; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>
 		</div>
 		<?php
+
+		if ( $settings['faq_schema'] ) {
```
louiswol94, line 832:
```suggestion
		if ( isset( $settings['faq_schema'] ) && 'yes' === $settings['faq_schema'] ) {
```

modules/nested-accordion/widgets/nested-accordion.php
```diff
@@ -799,6 +828,28 @@ protected function render() {
...
+					'name' => wp_strip_all_tags( $name ),
+					'acceptedAnswer' => [
+						'@type' => 'Answer',
+						'text' => wp_strip_all_tags( $text ),
+					],
+				];
+			}
+		}
+		?>
+		<script type="application/ld+json"><?php echo wp_json_encode( $json ?? [] ); ?></script>
```
louiswol94, line 851:
```suggestion
		<script type="application/ld+json"><?php echo wp_json_encode( $json ); ?></script>
```
Maybe? Because the `$json` variable will never be empty, it is declared outside the loop.

louiswol94, line 851: The `<script>` tag should be inside the if block?

louiswol94 (changes requested):

modules/nested-accordion/widgets/nested-accordion.php
Barcov: @louiswol94 I also expected this initially but `$json` is not declared anywhere outside a loop. I've specified a fallback empty value which is functionally equivalent to adding an empty `$json` array outside of the loop.

Barcov: The issue was that the conditional above this would not execute if the faq schema control was in it's default state (off) resulting in `$json` having no value in some cases.

image

I've moved the script tag into the PHP conditional above which means it always have a value or never gets called.

Barcov: Agree

tests/playwright/sanity/modules/nested-accordion/nested-accordion-content.test.ts
```diff
@@ -397,4 +397,39 @@ test.describe( 'Nested Accordion Content Tests @nested-accordion', () => {
 			expect.soft( maxHeightAfterOpen ).not.toEqual( '0px' );
 		} );
 	} );
+
+	test.only('Nested Accordion FAQ Schema', async ( {page}, testInfo ) => {
```
hein-obox, line 401: Remove test.only()

modules/nested-accordion/widgets/nested-accordion.php
```diff
@@ -799,6 +829,28 @@ protected function render() {
...
+				'mainEntity' => [],
+			];
+
+			foreach ( $faq_schema as $name => $text ) {
+				$json['mainEntity'][] = [
+					'@type' => 'Question',
+					'name' => wp_strip_all_tags( $name ),
+					'acceptedAnswer' => [
+						'@type' => 'Answer',
+						'text' => wp_strip_all_tags( $text ),
```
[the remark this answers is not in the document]
louiswol94: @Barcov As `$allowed_html` is the second parameter for `wp_kses()`, I though we could keep the HTML tags that Google allows?

modules/nested-accordion/widgets/nested-accordion.php
```diff
@@ -799,6 +828,28 @@ protected function render() {
...
+					],
+				];
+			}
+			$structured_data = wp_json_encode( $json );
+			echo "<script type=\"application/ld+json\">$structured_data</script>";
+		}
+		?>
+		<?php
```
louiswol94, line 852:
```suggestion
```
Not needed.

tests/playwright/sanity/modules/nested-accordion/nested-accordion-content.test.ts
```diff
@@ -397,4 +397,39 @@ test.describe( 'Nested Accordion Content Tests @nested-accordion', () => {
...
+
+		await page.waitForLoadState( 'load' );
+
+		// Assert
+		const faqSchema = await page.evaluate( async () => {
+			let jsonSchema = await document.querySelectorAll( 'script[type="application/ld+json"]' );
+			return await JSON.parse( jsonSchema[ 0 ][ 'text' ] );
+		} );
+
+		await expect.soft( faqSchema[ '@type' ], 'Expecting Structured data Schema to be of type FAQPage but found' ).toBe('FAQPage' );
```
louiswol94, line 424:
```suggestion
		await expect.soft( faqSchema[ '@type' ], 'Expecting Structured data Schema to be of type FAQPage but found' ).toBe( 'FAQPage' );
```

louiswol94 (approved):

PR #23649, opened 2023-09-03
Internal: Add beta tag to readme.txt for all versions [ED-11893]

TzviRabinovitch:


.github/workflows/bump-channel-version/action.yml
```diff
@@ -44,14 +44,18 @@ runs:
 
             if [[ ${{ inputs.CHANNEL }} == "beta" ]]; then
                 BETA_VERSION=${{ inputs.CLEAN_PACKAGE_VERSION }}-${{ inputs.CHANNEL }}${NEW_CHANNEL_VERSION}
```
davseve, line 46: `BETA_VERSION` can be equal to `ELEMENTOR_CHANNEL_PACKAGE_VERSION`, already defined in line 26.

.github/workflows/bump-channel-version/action.yml
```diff
@@ -44,14 +44,18 @@ runs:
...
-                if grep -q "Beta tag:" ./readme.txt
-                  then
-                    echo "Replace existing beta tag in readme file"
-                    sed -i -E "s/Beta tag: .*/Beta tag: $BETA_VERSION/g" ./readme.txt
-                  else
-                    echo "Add beta tag to readme file"
-                    sed -i -E '/^Stable tag: .*/a\'$'\n'"Beta tag: $BETA_VERSION$(printf '\n\r')" ./readme.txt
-                fi
+            else
+              BETA_VERSION=$(git ls-remote --tags | awk '{print $2}' | sed 's/^refs\/tags\/v//' | sed 's/^refs\/tags\///' | sort -V | grep -E '^[0-9]+\.[0-9]+\.[0-9]-beta[1-9]$' | sort -V | tail -n 1)
```
davseve, line 48: Do we need `sort -V` twice?

davseve (changes requested):

.github/workflows/bump-channel-version/action.yml
TzviRabinovitch: if the channel is Beta than it should be the version number that is being releases
if not it should be an existing beta version

davseve (approved):

Issue #22991, opened 2023-07-06
✅ 🐞 Bug Report: 3 Warnings on new Theme Builder when Landing Page is disabled when using PHP 8.2 - [ED-11843]


banlieues-asbl:
### Prerequisites

- [X] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [x] The issue still exists against the latest stable version of Elementor.

### Description

When you disable Landing page (in Elementor features) and then go to the new theme builder page, when the theme builder page loads, 3 warning messages appear with Elementor Pro + PHP 8.2.6.

Warning: Attempt to read property "labels" on null in D:\xampp\htdocs\wordpress\themes\wp-content\plugins\elementor\app\modules\import-export\module.php on line 733

Warning: Attempt to read property "singular_name" on null in D:\xampp\htdocs\wordpress\themes\wp-content\plugins\elementor\app\modules\import-export\module.php on line 733

Warning: Attempt to read property "label" on null in D:\xampp\htdocs\wordpress\themes\wp-content\plugins\elementor\app\modules\import-export\module.php on line 734

### Steps to reproduce

1. Disable landing page
2. Go to New theme builder

### Isolating the problem

- [x] This bug happens with only Elementor plugin active (and Elementor Pro).
- [x] This bug happens with a Blank WordPress theme active ([Hello theme](https://wordpress.org/themes/hello-elementor/)).
- [x] I can reproduce this bug consistently following the steps above.

### System Info

Win 10 + PHP 8.2.6

nicholaszein:
Hello @banlieues-asbl , 

Thank you for submitting your report! 🙏 

### ❌ While reviewing your report, we noticed it was missing some key pieces of information:

- 🔬 **Isolate the problem:** Deactivate all of your plugins (except Elementor/Pro), and temporarily change your WordPress theme to [Hello theme](https://wordpress.org/themes/hello-elementor/). If this solves the problem, gradually activate your plugins and theme one by one, until you spot the conflicting plugin or theme.

- 🕵️ **Validate Your Issue:** If the issue can't be recreated consistently or reproduced on other sites, it's likely not an Elementor bug. In that case, your issue is likely caused by a conflict with something else.

- 💾 **System info:** Please include the System Info report as outlined in our [guidelines](https://github.com/elementor/elementor/issues/new?assignees=&labels=status%2Fawaiting_triage&template=bug.yml). 
[...]

PR #23650, opened 2023-09-03
Fix: Error messages appear in Theme Builder with PHP 8.X (#22991) [ED-11843]

StyleShit:


app/modules/import-export/utils.php
```diff
@@ -59,7 +59,7 @@ public static function map_old_new_term_ids( array $imported_data ) {
 	}
 
 	public static function get_elementor_post_types() {
-		return [ 'post', 'page', 'e-landing-page' ];
+		return get_post_types_by_support( 'elementor' );
```
ManorHazazE, line 62: It will also return "elementor_library". I think it's going to be a problem.

ghost (changes requested):

app/modules/import-export/utils.php
ManorHazazE: Have you checked it?

StyleShit: Fixed

ghost (approved):

Issue #23402, opened 2023-08-13
✅ 🐞 Bug Report: Testimonial Widget's Style tab has a white space on the top - [ED-11956]


iqbal-web:
### Prerequisites

- [X] I have searched for similar issues in both open and closed tickets and cannot find a duplicate.
- [X] The issue still exists against the latest stable version of Elementor.

### Description

here a screenshort https://prnt.sc/OcEdOD26OZAO
looks like there hide a section & more white space, same problem many widget

### Steps to reproduce

Got to elementor testimonial widget then go style section,

### Isolating the problem

- [ ] This bug happens with only Elementor plugin active (and Elementor Pro).
- [ ] This bug happens with a Blank WordPress theme active ([Hello theme](https://wordpress.org/themes/hello-elementor/)).
- [X] I can reproduce this bug consistently following the steps above.

### System Info
<details>
<summary>Click to reveal</summary>

````txt

### wp-core ###

version: 6.3
site_language: en_US
user_language: en_US
timezone: +00:00
permalink: /%postname%/
https_status: false
multisite: false
user_registration: 0
blog_public: 1
default_comment_status: open
environment_type: production
user_count: 1
dotorg_communication: true

### wp-paths-sizes ###

wordpress_path: F:\xampp\htdocs\xs-test
wordpress_size: 52.04 MB (54568229 bytes)
uploads_path: F:\xampp\htdocs\xs-test/wp-content/uploads
uploads_size: 93.40 KB (95646 bytes)
themes_path: F:\xampp\htdocs\xs-test/wp-content/themes
themes_size: 12.35 MB (12951623 bytes)
plugins_path: F:\xampp\htdocs\xs-test/wp-content/plugins
plugins_size: 67.76 MB (71050124 bytes)
database_size: 3.83 MB (4014080 bytes)
total_size: 136.07 MB (142679702 bytes)

### wp-active-theme ###

name: Twenty Twenty-Three (twentytwentythree)
version: 1.1 (latest version: 1.2)
author: the WordPress team
author_website: https://wordpress.org
parent_theme: none
theme_features: core-block-patterns, post-thumbnails, responsive-embeds, editor-styles, html5, automatic-feed-links, block-templates, widgets-block-editor
theme_path: F:\xampp\htdocs\xs-test/wp-content/themes/twentytwentythree
auto_update: Disabled

### wp-themes-inactive (2) ###

Twenty Twenty-One: version: 1.8, author: the WordPress team, Auto-updates disabled
[...]

PR #23658, opened 2023-09-04
Fix: UI Glitch in widget Style/Advanced tab (#23402) [ED-11956]

StyleShit:
We don't delete the popovers properly, so they still exist even when they shouldn't, and it breaks the CSS selectors


Michaelgimii:
Hi @StyleShit,

In fact, this much space is somehow at the end of all sections of the editor.

space

StyleShit:
@Michaelgimii 
The top one is the bug, and the bottom one is the intended behavior. see:
https://github.com/elementor/elementor/blob/main/assets/dev/scss/editor/panel/controls/_section.scss#L22-L24

Nevoss (approved):

PR #23673, opened 2023-09-05
Internal: [Rating] [Style] Create Icon section and Size control [ED-11943]

TzviRabinovitch:


includes/widgets/rating.php
```diff
@@ -30,6 +30,45 @@ public function get_keywords() {
 		return [ 'star', 'rating', 'review', 'score', 'scale' ];
 	}
 
+	/**
+	 * @return void
+	 */
+	public function add_style_tab() {
```
hein-obox, line 36: Is it possible to use a 'protected' function? That is the default function type that we usually use. Do you have a specific reason for using a 'public' type?

TzviRabinovitch: nice catch 
This probably should be private

louiswol94 (approved):

hein-obox (approved):

PR #23675, opened 2023-09-05
Internal: [Rating] [Style] Added Spacing control [ED-11944]

TzviRabinovitch:


louiswol94 (approved):

includes/widgets/rating.php
```diff
@@ -66,6 +66,30 @@ private function add_style_tab() {
...
+					'em' => [
+						'min' => 0,
+						'max' => 10,
+						'step' => 0.1,
+					],
+					'rem' => [
+						'min' => 0,
+						'max' => 10,
+						'step' => 0.1,
+					],
```
davseve, line 84: Should we add defaults for PX?

includes/widgets/rating.php
```diff
@@ -66,6 +66,30 @@ private function add_style_tab() {
...
+				],
+				'size_units' => [ 'px', 'em', 'rem', 'vw', 'custom' ],
+				'selectors' => [
+					'{{WRAPPER}}' => '--e-rating-gap: {{SIZE}}{{UNIT}}',
```
davseve, line 88: I am not sure about that.
AFAIK, as the implementation of `--e-rating-gap` is at the moment we will have vertical and horizontal gaps.
Shouldn't the gap be horizontal only?

davseve (changes requested):

davseve (approved):

PR #23728, opened 2023-09-11
Internal: Change Getting Started video and links to support Container release [ED-12069]

Barcov:


modules/shapes/widgets/text-path.php
```diff
@@ -659,7 +659,7 @@ protected function render() {
 		// Add Text Path attributes.
 		$this->add_render_attribute( 'text_path', [
 			'class' => 'e-text-path',
-			'data-text' => esc_attr( $settings['text'] ),
+			'data-text' => htmlentities( esc_attr( $settings['text'] ) ),
```
hein-obox, line 662: This seems to be unrelated. If so, please restore it to the original version.

Barcov: Thanks

hein-obox (approved):

PR #23733, opened 2023-09-11
Tweak: Move Swiper library experiment to the advanced tab [ED-10215]

davseve:


core/experiments/manager.php
```diff
@@ -405,18 +405,6 @@ private function add_default_features() {
...
-			'title' => esc_html__( 'Upgrade Swiper Library', 'elementor' ),
-			'description' => esc_html__( 'Prepare your website for future improvements to carousel features by upgrading the Swiper library integrated into your site from v5.36 to v8.45. This experiment includes markup changes so it might require updating custom code and cause compatibility issues with third party plugins.', 'elementor' ),
-			'release_status' => self::RELEASE_STATUS_STABLE,
-			'new_site' => [
-				'default_active' => true,
-				'minimum_installation_version' => '3.11.0',
-			],
-			'default' => self::STATE_INACTIVE,
-		] );
-
```
louiswol94, line 419: @TzviRabinovitch @doronwolf @avivu 
Maybe we should 'hide' this experiment for a while? Instead of removing it. In Pro we still use the `function is_feature_active()`, which will fail if the experiment is totally removed. Then after a few versions we can delete this.

modules/swiper/module.php
```diff
@@ -0,0 +1,100 @@
...
+	/**
+	 * Swiper version 8.4.5
+	 */
+	const SWIPER_VERSION_8_4_5 = '8.4.5';
+
+	public function get_name() {
+		return 'swiper';
+	}
+
+	public static function swiper_active_version() {
```
louiswol94, line 31:
```suggestion
	public static function get_active_version() {
```

modules/swiper/module.php
```diff
@@ -0,0 +1,100 @@
...
+		return get_option( self::SWIPER_SETTINGS_OPTION_NAME );
+	}
+
+	public static function swiper_css_class() {
```
louiswol94, line 35:
```suggestion
	public static function get_css_class() {
```

modules/swiper/module.php
```diff
@@ -0,0 +1,100 @@
...
+		return self::SWIPER_VERSION_8_4_5 === self::swiper_active_version() ? 'swiper' : 'swiper-container';
+	}
+
+	public static function swiper_assets_path() {
```
louiswol94, line 39:
```suggestion
	public static function get_assets_path() {
```

modules/swiper/module.php
```diff
@@ -0,0 +1,100 @@
...
+					),
+				],
+			]
+		);
+	}
+
+	/**
+	 * Convert Swiper experiment value to settings.
+	 */
+	public static function swiper_experiment_converter() {
```
louiswol94, line 81:
```suggestion
	public static function convert_experiment_to_setting() {
```

modules/swiper/module.php
```diff
@@ -0,0 +1,100 @@
...
+	/**
+	 * Register_Admin Settings
+	 *
+	 * adds Swiper update admin settings
+	 * @param Settings $settings
+	 */
+	public function register_admin_settings( Settings $settings ) {
+		$settings->add_field(
+			Settings::TAB_ADVANCED,
+			Settings::TAB_ADVANCED,
```
louiswol94, line 58: Twice? `Settings::TAB_ADVANCED`

modules/swiper/module.php
```diff
@@ -0,0 +1,100 @@
+<?php
```
louiswol94, line 1: Maybe it can be a bit cleaner if not all these functions are static functions?
WDYT?

core/experiments/manager.php
doronwolf: Hi @louiswol94 I understand the reason but wonder, can't the fact there are 2 places that affect the Swiper library version cause other issues/conflicts?

modules/swiper/module.php
davseve: Apparently, we need to supply 4 arguments to the `add_field` function.
In the first two arguments the word 'advanced' this word is taken from a constant in `settings.php`

core/experiments/manager.php
davseve: @doronwolf The experiment will not be visible to regular users. so regular users will see it only in the settings.

PR #23888, opened 2023-09-22
Fix: Links inside Loop Carousel are not working on initial load in Menu widget [ED-12102]

louiswol94:


assets/dev/js/frontend/handlers/base-carousel.js
```diff
@@ -133,6 +133,9 @@ export default class CarouselHandlerBase extends SwiperHandlerBase {
 				this.a11ySetPaginationTabindex();
 				this.handleElementHandlers();
 			},
+			init: () => {
+				this.a11ySetSlideAriaHidden( 'initialisation' );
```
louiswol94, line 137: @hein-obox Can I remove `this.a11ySetSlideAriaHidden()` in any other place in this file? (Now that I call it in Swiper's init).

- Does it need to stay in the handler's `onInit()` function?
- Does it need to stay in Swiper's `slideChangeTransitionEnd` function?
- Should I move `this.a11ySetWidgetAriaDetails()` and `this.a11ySetPaginationTabindex()` into this Swiper init as well rather?

hein-obox: Q1: I believe we can remove it from 'onInit()'.
Q2: We need to keep it in 'slideChangeTransitionEnd'.
Q3: Yes, I believe it is best to handle it in one place.

assets/dev/js/frontend/handlers/base-carousel.js
```diff
@@ -133,6 +133,9 @@ export default class CarouselHandlerBase extends SwiperHandlerBase {
 				this.a11ySetPaginationTabindex();
 				this.handleElementHandlers();
 			},
+			init: () => {
```
hein-obox, line 136: Could you please check if this issue is related as well: https://elementor.atlassian.net/browse/ED-11434?

murzin-nikita (approved):

PR #23931, opened 2023-10-02
Fix: Show the right type inside the no history message [ED-12342]

pkniazevych:
Replaced all `promptType` variable names with `historyType` as it makes more sense and saves consistency between history type constants and variable names.


modules/ai/assets/js/editor/components/prompt-history/parts/modal-upgrade.js
```diff
@@ -1,24 +1,34 @@
...
 const VARIANT_FULL = 'full';
 const VARIANT_SMALL = 'small';
 
 const messages = {
-	[ VARIANT_FULL ]: __( 'Want to see your image generation history for as far as the past 90 days?', 'elementor' ),
-	[ VARIANT_SMALL ]: __( 'Want to see your image generation history for the past 90 days?', 'elementor' ),
+	// Translators: %s: History type.
+	[ VARIANT_FULL ]: __( 'Want to see your %s generation history for as far as the past 90 days?', 'elementor' ),
+	// Translators: %s: History type.
+	[ VARIANT_SMALL ]: __( 'Want to see your %s generation history for the past 90 days?', 'elementor' ),
```
aviranLevi1, line 12: I suggest you to move the messages to consts/enum file

pkniazevych: They're actually needed only here, so I think keeping them as constants in the same file should work for us 🙂

aviranLevi1 (approved):

bainternet (approved):

bainternet (approved):

PR #24011, opened 2023-10-12
Internal: Enhanced functionality of the Dynamic Tag control [ED-12197]

ArmanPurtoian:


core/dynamic-tags/tag.php
```diff
@@ -51,7 +52,7 @@ public function get_content( array $options = [] ) {
 			endif;
 
 		} elseif ( ! Utils::is_empty( $settings, 'fallback' ) ) {
-			$value = $settings['fallback'];
+			$value = Str::sanitize_input_string_or_array( $settings['fallback'] );
```
StyleShit, line 55: why not `wp_kses_post` like others?

```suggestion
			$value = wp_kses_post( $settings['fallback'] );
```

tests/phpunit/elementor/core/utils/test-str.php
```diff
@@ -49,4 +49,166 @@ public function test_ends_with() {
...
+
+	public function test_basic_html() {
+		// Arrange
+		$input = "<b>Bold</b> <i>Italic</i>";
+
+		// Act
+		$output = Str::sanitize_input_string_or_array( $input );
+
+		// Assert
+		$this->assertEquals( "<b>Bold</b> <i>Italic</i>", $output );
```
louiswol94, line 201:
```suggestion
		$this->assertEquals( $input, $output );
```

tests/phpunit/elementor/core/utils/test-str.php
```diff
@@ -49,4 +49,166 @@ public function test_ends_with() {
...
+
+	public function test_html_entities() {
+		// Arrange
+		$input = "Hello &amp; World";
+
+		// Act
+		$output = Str::sanitize_input_string_or_array( $input );
+
+		// Assert
+		$this->assertEquals( "Hello &amp; World", $output );
```
louiswol94, line 190:
```suggestion
		$this->assertEquals( $input, $output );
```

louiswol94 (changes requested):

core/dynamic-tags/tag.php
louiswol94: @StyleShit `Str::sanitize_input_string_or_array()` will just be a wrapper for `wp_kses_post()` to handle scenarios where the fallback value might be an array.

louiswol94: @StyleShit Checking...

louiswol94: @StyleShit When the fallback is a file upload for example.

Barcov: @louiswol94 `wp_kses_post_deep()` looks equivalent to `sanitize_input_string_or_array()`. I like your approach because we can extend the function to handle future unexpected changes to ACF data types if needed as long as the function signature (naming in this case) does not make extension unexpected.

@StyleShit we reverted this change in a previous release because image dynamic tags sometimes returns an array. This was unexpected when I tested for the release.

louiswol94: @ArmanPurtoian Looks like `wp_kses_post_deep()` should work. Can you please implement? You'll also have to remove your unit tests from the `test-str.php` file and see if there's another relevant way to test this - like for example, creating a new tag and testing the output of `get_content()`.

Thanks @StyleShit

louiswol94: > @louiswol94 `wp_kses_post_deep()` looks equivalent to `sanitize_input_string_or_array()`. I like your approach because we can extend the function to handle future unexpected changes to ACF data types if needed as long as the function signature (naming in this case) does not make extension unexpected.
> 
> @StyleShit we reverted this change in a previous release because image dynamic tags sometimes returns an array. This was unexpected when I tested for the release.

Thanks Barco. I prefer to rather use built-in WP functions if we can 😄

louiswol94: @StyleShit @ArmanPurtoian @Barcov I am going to approve this PR. We don't have to test a WP function `wp_kses_post_deep()`, and if we want to further test the tag values we can rather expand Pro tests such as `test-acf-date-time.php`.

louiswol94 (approved):

core/dynamic-tags/tag.php
StyleShit: @louiswol94 

what about testing the tag itself?

PR #24066, opened 2023-10-22
Tweak: Reduce loading controls by remove UI controls and some arguments from frontend [ED-12500]

KingYes:


includes/base/controls-stack.php
```diff
@@ -393,6 +393,12 @@ public function add_control( $id, array $args, $options = [] ) {
 			'position' => null,
 		];
 
+		$should_optimize_controls = $this->should_optimize_controls();
+
+		if ( $should_optimize_controls && ! empty( $args['type'] ) && in_array( $args['type'], [ Controls_Manager::RAW_HTML, Controls_Manager::DIVIDER, Controls_Manager::HEADING, Controls_Manager::BUTTON ] ) ) {
+			return false;
```
louiswol94, line 399: Can we not add this code totally at the beginning of this method? `$default_options` is redundant if we return early here.

includes/base/controls-stack.php
```diff
@@ -393,6 +393,12 @@ public function add_control( $id, array $args, $options = [] ) {
 			'position' => null,
 		];
 
+		$should_optimize_controls = $this->should_optimize_controls();
+
+		if ( $should_optimize_controls && ! empty( $args['type'] ) && in_array( $args['type'], [ Controls_Manager::RAW_HTML, Controls_Manager::DIVIDER, Controls_Manager::HEADING, Controls_Manager::BUTTON ] ) ) {
```
louiswol94, line 398: `[ Controls_Manager::RAW_HTML, Controls_Manager::DIVIDER, Controls_Manager::HEADING, Controls_Manager::BUTTON ]`

Can we maybe name this array so it's easy to understand why these types have been added here?

louiswol94, line 398: `if ( $should_optimize_controls && ! empty( $args['type'] ) && in_array( $args['type'], [ Controls_Manager::RAW_HTML, Controls_Manager::DIVIDER, Controls_Manager::HEADING, Controls_Manager::BUTTON ] ) ) {`

Maybe we can extract this entire check into a helper function as well?

`if ( $this->control_can_be_skipped() ) {` <-- Not sure about the naming, but it would help improve readability.

includes/base/controls-stack.php
```diff
@@ -431,9 +437,41 @@ public function add_control( $id, array $args, $options = [] ) {
...
+
+		if ( null === $retval ) {
+			$retval = (
+				! is_admin()
+				&& ! \Elementor\Plugin::$instance->preview->is_preview_mode()
+			);
+		}
+
+		return $retval;
+	}
```
louiswol94, line 473: Is this basically just a check for the front end? Do we need to consider the following as well?

`( ! is_admin() || wp_doing_ajax() ) && ! wp_doing_cron()` <-- Not sure if these actions will invoke this file or not, but in general this is a good WordPress way to be sure we are on Frontend.

```
$retval = (
	! is_admin()
	&& ! \Elementor\Plugin::$instance->preview->is_preview_mode()
);
```
Maybe this logic can be moved to a helper like `\Elementor\Plugin::$instance->frontend->is_frontend()`? I also need to do FE checks in another place so it might be helpful to have one source.

louiswol94 (changes requested):
Good job! My comments here are just some minor "suggestions". You can take it or leave it ;-) Manual testing is definitely also needed.

includes/base/controls-stack.php
KingYes: This is the first draft of it. Right now, I make it on the front end only, and I'm not sure if we can expand it to other areas.

includes/base/controls-stack.php
```diff
@@ -431,9 +444,41 @@ public function add_control( $id, array $args, $options = [] ) {
...
+				$args['range'],
+				$args['render_type'],
+				$args['toggle'],
+				$args['ai'],
+				$args['label_on'],
+				$args['label_off'],
+				$args['labels'],
+				$args['handles']
+			);
+		}
```
StyleShit, line 464: might be easier (and safer?) to have a whitelist here of the data you actually need?
with the current approach, you'll have to maintain this list each time someone (and maybe 3rd party) adds a property

includes/base/controls-stack.php
```diff
@@ -431,9 +444,41 @@ public function add_control( $id, array $args, $options = [] ) {
...
+
 		return Plugin::$instance->controls_manager->add_control_to_stack( $this, $id, $args, $options );
 	}
 
+	private function should_optimize_controls() {
+		static $retval = null;
```
StyleShit, line 470: retval?

includes/base/controls-stack.php
```diff
@@ -431,9 +444,41 @@ public function add_control( $id, array $args, $options = [] ) {
...
+
+		if ( null === $retval ) {
+			$retval = (
+				! is_admin()
+				&& ! \Elementor\Plugin::$instance->preview->is_preview_mode()
```
StyleShit, line 475:
```suggestion
				&& ! Plugin::$instance->preview->is_preview_mode()
```

StyleShit (commented):
🔥

includes/base/controls-stack.php
louiswol94: @KingYes Again it's a blacklist vs whitelist 😉 hehe

KingYes: @StyleShit Why did you think it's not safe? As Elementor controls, I should know which controls we use on the frontend and which on the editor. No?

StyleShit: I mean "safe" in terms of future proof, maintainability, and performance - If you have a whitelist of controls, you know that those controls are the only ones that will be loaded to the memory, regardless of whatever we and/or 3rd parties have added. 

This way, you can know for sure that in any future cases, only the controls that you actually need are there.

Let's say we had this PR merged before the AI was introduced, you would have needed to come back here and add the "ai" key to the blacklist or you'll gradually hurt the performance without notice (and we both know that we'll forget to come back here).

I think it's also easier to understand and determine which controls are required and affect the rendering when reading the code as a whitelist.

louiswol94: @StyleShit It's a good point! But it might be tricky to construct a full whitelist without missing **anything** 🤔

louiswol94 (dismissed):

louiswol94 (approved):

PR #24092, opened 2023-10-25
Tweak: Get super global changed to handle multiple files [ED-2192]

Omerisra6:


StyleShit (dismissed):

includes/utils.php
```diff
@@ -822,14 +822,26 @@ public static function get_super_global_value( $super_global, $key ) {
 		}
 
 		if ( $_FILES === $super_global ) {
-			$super_global[ $key ]['name'] = sanitize_file_name( $super_global[ $key ]['name'] );
-
-			return $super_global[ $key ];
+			return isset( $super_global[ $key ]['name'] ) ?
```
mark-elementor, line 825: I think it makes a bit more sense to do:
```suggestion
			return is_multi_upload( $super_global[ $key ] ) ?
				self::sanitize_multi_upload( $super_global[ $key ] ) :
				self::sanitize_file_name( $super_global[ $key ] );
```

with a helper function like this:
```
private static function is_multi_upload( $array ) {
	return is_array( $array ) && count( array_filter( $array, 'is_array' ) ) === count( $array ) && count( array_filter( array_keys( $array ), 'is_numeric' ) ) === count( $array );
}
```

mark-elementor (dismissed):

StyleShit (dismissed):

blackcolt (approved):

PR #24141, opened 2023-10-29
Internal: Prepare AI Module for context menu [ED-12555]

matipojo:


modules/ai/assets/js/editor/ai-layout-behavior.js
```diff
@@ -25,93 +23,24 @@ export default class AiLayoutBehavior extends Marionette.Behavior {
...
-			template,
-		};
-	}
-
-	onSelect( template ) {
-		this.previewContainer.setContent( template );
-	}
-
 	onInsert( template ) {
 		this.hideDropArea();
```
StyleShit, line 37: isn't hiding the drop area also an editor integration?

modules/ai/assets/js/editor/utils/editor-integration.js
```diff
@@ -0,0 +1,99 @@
...
+	return {
+		colorScheme,
+		isRTL,
+	};
+};
+
+export const renderLayoutApp = ( options = {
+	at: null,
+	onInsert: null,
+} ) => {
```
StyleShit, line 36:
```suggestion
export const renderLayoutApp = ( {
	at: null,
	onInsert: null,
} ) => {
```

maybe?

modules/ai/assets/js/editor/utils/editor-integration.js
```diff
@@ -0,0 +1,99 @@
...
+			onGenerate={ () => previewContainer.reset() }
+			onData={ async ( template ) => {
+				const screenshot = await takeScreenshot( template );
+
+				return {
+					screenshot,
+					template,
+				};
+			} }
+			onSelect={ ( template ) => previewContainer.setContent( template ) }
```
StyleShit, line 72: any reason those are inlined?
don't you think it's cleaner to separate it and make this render function a little less ugly? 😅 

___
OK, after reading it again I'm assuming it's because you don't have a `previewContainer` state

StyleShit (changes requested):

modules/ai/assets/js/editor/ai-layout-behavior.js
matipojo: There is no "Drop area" from a context menu on an existing container

StyleShit (approved):

PR #24146, opened 2023-10-30
Tweak: Add note in background control that dynamic image is not supported in image size [ED-9824]

KingYes:


includes/controls/media.php
```diff
@@ -287,6 +287,8 @@ public function content_template() {
 						</select>
 					</div>
 				</div>
+
+				<div class="elementor-control-field-description"><?php echo esc_html__( 'Image size settings don’t apply to Dynamic Images.', 'elementor' ); ?></div>
```
rodolphebertozzo, line 291: Why don't you use a <p> tag to display the error message?

rodolphebertozzo (changes requested):

includes/controls/media.php
rami-elementor: @rodolphebertozzo

Currently, Elementor use a `<div>` tag to display field descriptions. So, for consistency, this control should use a `<div>` too.

However, you are correct, to improve semantics we should replace this to `<p>`. We will do that in a separate PR.

arielk (approved):

PR #24171, opened 2023-11-01
Internal: Create a build artifact on PR [ED-12617]

matipojo:


.github/workflows/build.yml
```diff
@@ -1,10 +1,18 @@
 name: Build
 
 on:
+  pull_request:
```
StyleShit, line 4: we're already building the code for other workflows
maybe we can somehow unify it and build only once so we don't spend time & resources?

matipojo: We need to do it, but it's for another PR.

StyleShit: Maybe open a task for that?

StyleShit (approved):

PR #24353, opened 2023-11-20
Internal: Added body background as a context for AI Containers [ED-12785]

matipojo:


modules/ai/assets/js/editor/utils/editor-integration.js
```diff
@@ -65,11 +65,19 @@ export const renderLayoutApp = ( options = {
...
+
 	ReactDOM.render(
 		<LayoutApp
 			mode={ options.mode }
 			isRTL={ isRTL }
 			colorScheme={ colorScheme }
+			currentContext={ {
+				body: {
+					backgroundColor: bodyStyle.backgroundColor,
+					backgroundImage: bodyStyle.backgroundImage,
```
StyleShit, line 78: what about a background image that doesn't have a public URL?

matipojo: It's not a problem, the context if for better color contrast, so, for a local image it will not run this improvement.

StyleShit (approved):

PR #24515, opened 2023-12-06
Fix: Multiple plugins and themes compatibility issue [ED-13020]

KingYes:


includes/base/controls-stack.php
```diff
@@ -388,19 +388,6 @@ public function get_controls_settings() {
 	 * @return bool True if control added, False otherwise.
 	 */
 	public function add_control( $id, array $args, $options = [] ) {
-		$should_optimize_controls = $this->should_optimize_controls();
-
-		$ui_controls = [
-			Controls_Manager::RAW_HTML,
-			Controls_Manager::DIVIDER,
-			Controls_Manager::HEADING,
```
louiswol94, line 396: @KingYes This might be the only line causing the issue:

`Controls_Manager::HEADING,`

Unless you want to play it safe and remove all this code :)

blackcolt (approved):

PR #24571, opened 2023-12-13
Fix: Features page - Typo in modal for activating dependency experiments [ED-12179]

elchugreeva:


louiswol94 (dismissed):

core/experiments/assets/js/admin/behaviors/experiments-messages.js
```diff
@@ -171,7 +171,9 @@ export default class ExperimentsMessages {
 	showDependenciesDialog( experimentId ) {
 		const experiment = this.getExperimentData( experimentId ),
 			experimentName = experiment.title,
-			dialogMessage = this.joinDependenciesNames( this.getExperimentDependencies( experimentId ).map( ( d ) => d.title ) );
+			dependenciesArray = this.getExperimentDependencies( experimentId ).map( ( d ) => d.title ),
+			glue = 1 < dependenciesArray.length ? ' ' : '',
+			dialogMessage = this.joinDependenciesNames( dependenciesArray, glue );
```
wittytzipporah, line 176: I think this is overcomplicated. From what I can see, `dependenciesArray.join(' ')` would return the desired result - a list of the dependencies separated by a space.

If the dependencies are an array of more than two, I think it would be kind of odd results, as it would print something like "one two threefour", since the joinDependenciesNames() function doesn't use the passed glue for the connection between the penultimate and ultimate items. 

Also, the comment on line 178 indicates an expected comma-separated list, and that would not be generated by this code, as the glue used is only a space.

wittytzipporah: I just found this functionality, which may give us the smoothest option for list formatting -https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat/format

louiswol94: `dialogMessage = this.joinDependenciesNames( this.getExperimentDependencies( experimentId ).map( ( d ) => d.title ), ' ');`
@Barcov Can you please check if this works?

Barcov: @louiswol94 Yes that works as expected. @elchugreeva this looks like the simplest solution for right now.

Screenshot 2023-12-14 at 13 26 17

louiswol94, line 176:
```suggestion
			dialogMessage = this.joinDependenciesNames( this.getExperimentDependencies( experimentId ).map( ( d ) => d.title ), ' ');
```

louiswol94 (dismissed):

core/experiments/assets/js/admin/behaviors/experiments-messages.js
wittytzipporah: OK, I missed that `joinDependenciesNames` is just calling `join`.  So why have this extra formatting function though? Why not just use `join`? I see that currently this is the only place this function is used, and we aren't sending it a `finalGlue` parameter, so the rest of the function is not in use at all.

My concerns above about the formatting of the list remain 
- the comment below indicates that a comma-separated list is expected, and this does not print one. 
- if this list contains multiple items, it should have "and" for readability

I looked more into that built-in JS list format function, and it seems it would take care all of this automatically. For example:
`new Intl.ListFormat('en').format(['one', 'two', 'three', 'four'])`
prints
`'one, two, three, and four'`

`new Intl.ListFormat('en').format(['one', 'two'])`
prints 
`'one and two'`

core/experiments/assets/js/admin/behaviors/experiments-messages.js
```diff
@@ -171,7 +171,7 @@ export default class ExperimentsMessages {
 	showDependenciesDialog( experimentId ) {
 		const experiment = this.getExperimentData( experimentId ),
 			experimentName = experiment.title,
-			dialogMessage = this.joinDependenciesNames( this.getExperimentDependencies( experimentId ).map( ( d ) => d.title ) );
+			dialogMessage = this.joinDependenciesNames( this.getExperimentDependencies( experimentId ).map( ( d ) => d.title ), ' ' );
```
louiswol94, line 174:
```suggestion
			dialogMessage = new Intl.ListFormat( 'en' ).format( this.getExperimentDependencies( experimentId ).map( ( d ) => d.title ) );
```

louiswol94 (approved):

PR #24590, opened 2023-12-17
Fix: Use Intl.ListFormat to format list of dependencies [ED-12179]

wittytzipporah:
testing


core/experiments/assets/js/admin/behaviors/experiments-messages.js
```diff
@@ -149,29 +149,10 @@ export default class ExperimentsMessages {
...
-			lastItem = clone.pop();
-
-		return clone.join( glue ) + finalGlue + lastItem;
-	}
-
 	showDependenciesDialog( experimentId ) {
 		const experiment = this.getExperimentData( experimentId ),
 			experimentName = experiment.title,
-			dialogMessage = this.joinDependenciesNames( this.getExperimentDependencies( experimentId ).map( ( d ) => d.title ), ', ' );
+			dialogMessage = new Intl.ListFormat( 'en' ).format( this.getExperimentDependencies( experimentId ).map( ( d ) => d.title ) );
```
louiswol94, line 155: Hey @wittytzipporah,
Very nice investigation regarding the failing tests!
My only concern is with hard coding the `en` parameter. The word `and` for example might never be translated.
Maybe we should use `get_user_locale()` and expose it to the JS side. (But then we'll have to make sure the Locales from WP and this method matches).
WDYT?

louiswol94 (changes requested):

core/experiments/assets/js/admin/behaviors/experiments-messages.js
wittytzipporah: After some research into get_user_locale and JS Intl object, I think it's likely to be fine in all or nearly all cases. They are apparently both based on the BCP 47 standard. There may be some variance between browser implementations, but that is not likely to be in any of the major languages that Elementor is translated to.

wittytzipporah: We also don't have to inject it ourselves, it's actually added in the <html> tag of the page by WP core - see wp-includes/template.php function _wp_admin_html_begin(). There don't appear to be any filters that we need to worry about removing it.

core/experiments/assets/js/admin/behaviors/experiments-messages.js
```diff
@@ -149,34 +149,26 @@ export default class ExperimentsMessages {
...
-
-		if ( 1 === array.length ) {
-			return array[ 0 ];
-		}
-
-		const clone = [ ...array ],
-			lastItem = clone.pop();
+	getSiteLanguageCode() {
+		const languageCode = document.querySelector( 'html' ).getAttribute( 'lang' );
+		return languageCode ?? 'en'; // Fallback to English if no language code found.
```
louiswol94, line 154:
```suggestion
		const languageCode = document.querySelector( 'html' ).getAttribute( 'lang' );

		return languageCode ?? 'en'; // Fallback to English if no language code found.
```

core/experiments/assets/js/admin/behaviors/experiments-messages.js
```diff
@@ -149,34 +149,27 @@ export default class ExperimentsMessages {
...
-		const clone = [ ...array ],
-			lastItem = clone.pop();
+		return languageCode ?? 'en'; // Fallback to English if no language code found.
+	}
 
-		return clone.join( glue ) + finalGlue + lastItem;
+	formatDependenciesList( dependencies ) {
+		const dependenciesTitles = dependencies.map( ( d ) => d.title );
+		const languageCode = this.getSiteLanguageCode();
+		return new Intl.ListFormat( languageCode ).format( dependenciesTitles );
```
louiswol94, line 161:
```suggestion
		const languageCode = this.getSiteLanguageCode();

		return new Intl.ListFormat( languageCode ).format( dependenciesTitles );
```

louiswol94 (approved):

PR #24614, opened 2023-12-20
Internal: Create test pages for Playwright from the WP RestApi [ED-13150]

hein-obox:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #


tests/playwright/pages/editor-page.ts
```diff
@@ -537,7 +537,7 @@ export default class EditorPage extends BasePage {
 	 */
 	async editCurrentPage() {
 		const postId = await this.getPageIdFromFrontEnd();
-		await expect( postId, 'No Post/Page ID returned when calling getPageIdFromFrontEnd().' ).toBeTruthy();
+		await expect.soft( postId, 'No Post/Page ID returned when calling getPageIdFromFrontEnd().' ).toBeTruthy();
```
hein-obox, line 540: I have replaced `expect()` with `expect.soft()` everywhere, so that it will give all errors in a test all at once. This is helpful e.g. when making structural updates to tests.

tests/playwright/pages/wp-admin-page.ts
```diff
@@ -28,41 +30,58 @@ export default class WpAdminPage extends BasePage {
...
-		const selector = 'editor_v2' === editorType
-			? 'button[aria-label="Post Settings"]'
-			: '#elementor-panel-footer-settings';
+	async createNewPostWithAPI() {
+		const request: APIRequestContext = this.page.context().request,
+			postDataInitial = {
+				title: 'Playwright Test Page - Uninitialized',
+				content: '',
+			},
+			postId = await create( request, 'posts', postDataInitial ),
```
hein-obox, line 53: I have created `posts` instead of `pages` because they caused fewer problems with the existing tests. I assume this has to do with the default templates that are used for pages and  posts. I am not sure.

Would you be happy with that? This is the biggest structural change.

tests/playwright/pages/wp-admin-page.ts
```diff
@@ -28,41 +30,58 @@ export default class WpAdminPage extends BasePage {
...
+			postDataUpdated = {
+				title: `Playwright Test Page #${ postId }`,
```
hein-obox, line 55: I have used this title so that on localhost in the Trash it is easier to identify which posts were generated as part of the playwright tests. This will help a lot when you want to clean up the existing posts or if you would like to check the results of a Playwright test on localhost.

tests/playwright/pages/wp-admin-page.ts
```diff
@@ -28,41 +30,58 @@ export default class WpAdminPage extends BasePage {
...
+			};
 
-		await this.page.locator( selector ).click();
+		await create( request, `posts/${ postId }`, postDataUpdated );
+		await this.page.goto( `/wp-admin/post.php?post=${ postId }&action=elementor` );
+	}
+
+	async createNewPostFromDashboard( setPageName : boolean ) {
```
hein-obox, line 62: For a few existing tests, it works better when we apply the old method.

tests/playwright/sanity/includes/elements/section/reverse-columns/reverse-columns.ts
```diff
@@ -41,7 +41,7 @@ export default class ReverseColumns {
 			container: false,
 		} );
 		await this.wpAdminPage.openNewPage();
-		await this.editor.getPreviewFrame().locator( '.elementor-add-section-inner' ).click( { button: 'right' } );
```
hein-obox, line 44: No longer necessary since the implementation of the API approach.

tests/playwright/sanity/editor-v2.test.ts
```diff
@@ -15,7 +15,7 @@ test.describe( 'Editor v2', () => {
 
 		await wpAdminPage.setExperiments( { editor_v2: true } );
 
-		editor = await wpAdminPage.openNewPage( 'editor_v2' );
```
hein-obox, line 18: No longer necessary with the API approach.

tests/playwright/pages/editor-page.ts
dmaziar: Please don't do it. It probably makes sense with screenshot, but not with e2e. Very often one step depends on another and if it was failing in step 1, playwright will fail fast and retry. But if you have soft assertions, it will go over broken test and only then will do a retry. It's a standard practise on most of the test automation frameworks.  Also, it is not a part of this PR. Such changes should come as separate

tests/playwright/pages/wp-admin-page.ts
dmaziar: This is a game changer! Would be nice to have pages as well, but for now it's ok

PR #24686, opened 2024-01-04
Internal: Add REST to support pages panel UI for permissions [ED-13216]

ranshoshani34:


modules/site-navigation/rest-fields/page-user-can.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+namespace Elementor\Modules\SiteNavigation\Rest_Fields;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+class Page_User_Can {
```
kobizz, line 9: why do you need a different endpoint for user can?

ranshoshani34: Its not an endpoint.
I'm adding a field to the 'pages' resource.
This field will be enriched when we request `/wp/v2/pages/?_fields=user_can`

kobizz (approved):

modules/site-navigation/data/endpoints/homepage.php
```diff
@@ -0,0 +1,35 @@
...
+
+	public function get_items( $request ) {
+		$homepage_id = get_option( 'page_on_front' );
+		$show_on_front = get_option( 'show_on_front' );
+
+		if ( 'page' !== $show_on_front ) {
+			return 0;
+		}
+
+		return intval( $homepage_id );
```
blackcolt, line 33: return 'page' !== $show_on_front ? 0 : intval( $homepage_id )

blackcolt (dismissed):

shai-shwartz (approved):

modules/site-navigation/data/endpoints/homepage.php
StyleShit: Be positive instead 😉

```PHP
return 'page' === $show_on_front ? intval( $homepage_id ) : 0;
```

blackcolt (approved):

PR #24725, opened 2024-01-07
Internal: AI - Shouldn't check status for not connected users [ED-13301]

matipojo:


modules/ai/module.php
```diff
@@ -143,15 +143,20 @@ private function enqueue_main_script() {
 			true
 		);
 
+		$config = [
+			'is_get_started' => User::get_introduction_meta( 'ai_get_started' ),
+			'connect_url' => $this->get_ai_connect_url(),
+		];
+
+		if ( $this->get_ai_app()->is_connected() ) {
+			// Use a cached version, don't call the API on every editor load.
```
DennisNerush, line 152: no need for the comment

DennisNerush (approved):

modules/ai/module.php
matipojo: It's important to understand why we need a cached response.

PR #24771, opened 2024-01-12
Fix: Disabling the Display Conditions experiment will show promotion to Advanced plan [ED-13355]

louiswol94:


includes/managers/controls.php
```diff
@@ -1216,6 +1216,10 @@ private function should_clean_stack( $stack ): bool {
 	}
 
 	public function add_display_conditions_controls( Controls_Stack $controls_stack ) {
+		if ( Utils::has_pro() ) {
```
louiswol94, line 1219: We should not add these controls if the user has Pro. If the user has Pro, then Pro will decide if the Feature or the Promotion will show. These controls here in `add_display_conditions_controls()` should only be added if the user only has the free plugin.

murzin-nikita (approved):

hein-obox (approved):

PR #24836, opened 2024-01-18
Tweak: Added new "Alert" control for Editor panels [ED-13432]

rami-elementor:


includes/controls/alert.php
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Elementor alert control.
+ *
+ * A base control for creating alerts in the Editor panels.
+ *
+ * @since 3.20.0
+ */
+class Control_Alert extends Base_UI_Control {
+
```
davseve, line 16:
```suggestion
const alert_type_info = 'info';

const alert_type_success = 'success';

const alert_type_warning = 'warning';

const alert_type_danger = 'danger';
```

includes/controls/alert.php
```diff
@@ -0,0 +1,73 @@
...
+	 *
+	 * @return array Control default settings.
+	 */
+	protected function get_default_settings() {
+		return [
+			'alert_type' => 'info', // info, success, warning, danger.
+			'heading' => '',
+			'content' => '',
+		];
+	}
```
davseve, line 72: I would suggest doing the following things:

set consts for `info`, `success`, `warning`, `danger`, see comment above. 

```suggestion
         * @param string $alert_type see alert_type_* constants
	 * @since 3.20.0
	 * @access protected
	 *
	 * @return array Control default settings.
	 */
	protected function get_default_settings( $alert_type ) {
		return [
			'alert_type' => $alert_type
			'heading' => '',
			'content' => '',
		];
	}
```

Then the function call will be:
`$this->get_default_settings( Control_Alert::alert_type_info )`.

bainternet (dismissed):

davseve (commented):
@rami-elementor Do we want tests here?

rami-elementor:
We don't have tests for controls...

davseve (dismissed):

bainternet (dismissed):

davseve (approved):

PR #24846, opened 2024-01-19
Fix: Button widget "Type" control is not working when global colors activated [ED-13437]

rami-elementor:


arielk:
Closed for now due to backwards compatibility concern.

PR #24849, opened 2024-01-19
Internal: Allow hosting to replace promotions [ED-13373]

Barcov:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #


core/role-manager/role-manager.php
```diff
@@ -155,14 +155,34 @@ private function display_role_controls( $role_slug, $role_data ) {
...
+
+		$promotion = [
+			'description' => __( 'Want to give access only to content?' ),
+			'upgrade_url' => $upgrade_url,
+			'upgrade_text' => __( 'Upgrade' ),
+		];
+
+		$promotion = apply_filters( 'elementor/role/restrictions/custom_promotion', $promotion );
+
+		if ( strpos( $promotion['upgrade_url'], 'elementor.com' ) === false ) {
```
TzviRabinovitch, line 179: can a user add elementor.com at the end of the URL ?
ex ...
https://www.google.com/?id=elementor.com

includes/base/widget-base.php
```diff
@@ -378,6 +378,10 @@ protected function get_initial_config() {
 			$config['tabs_controls'] = $this->get_tabs_controls();
 		}
 
+		if ( isset( $config['upsale_data']['upgrade_url'] ) && strpos( $config['upsale_data']['upgrade_url'], 'elementor.com' ) !== false ) {
```
TzviRabinovitch, line 381: maybe extract the logic of strpos ?
DRY

modules/promotions/admin-menu-items/custom-fonts-promotion-item.php
```diff
@@ -15,18 +15,42 @@ public function get_page_title() {
...
 	}
 
 	public function get_cta_url() {
-		return 'https://go.elementor.com/go-pro-custom-fonts/';
+		$upgrade_url = 'https://go.elementor.com/go-pro-custom-fonts/';
+		$promotion['upgrade_url'] = $upgrade_url;
+
+		$promotion = apply_filters( 'elementor/fonts/restrictions/custom_promotion', $promotion );
+
+		if ( strpos( $promotion['upgrade_url'] ?? '', 'elementor.com' ) === false ) {
```
TzviRabinovitch, line 51: same here

modules/promotions/admin-menu-items/go-pro-promotion-item.php
```diff
@@ -32,6 +32,17 @@ public function get_capability() {
 		return 'manage_options';
 	}
 
+	public function get_url() {
+		$promotion['upgrade_url'] = $this->url;
+		$filtered_url = apply_filters( 'elementor/adminmenuitems/restrictions/custom_promotion', $promotion )['upgrade_url'] ?? '';
+
+		if ( strpos( $filtered_url, 'elementor.com' ) !== false ) {
```
TzviRabinovitch, line 39: and here

modules/promotions/module.php
```diff
@@ -63,7 +63,7 @@ private function handle_external_redirects() {
 		}
 
 		if ( 'go_elementor_pro' === $_GET['page'] ) {
-			wp_redirect( Go_Pro_Promotion_Item::URL );
+			wp_redirect( ( new Go_Pro_Promotion_Item() )->get_url() );
```
TzviRabinovitch, line 66: looks like a hack 
we are creating an empty instance just to run the filter

includes/editor-templates/panel.php
```diff
@@ -189,12 +189,20 @@
...
+		<# } else { #>
 		<img class="elementor-nerd-box-icon" src="<?php Utils::print_unescaped_internal_string( ELEMENTOR_ASSETS_URL . 'images/go-pro.svg' ); ?>" loading="lazy" alt="<?php echo esc_attr__( 'Upgrade', 'elementor' ); ?>" />
+		<# } #>
 		<# if ( elementData.upsale_data.title ) { #>
 		<div class="elementor-nerd-box-title">{{ elementData.upsale_data.title }}</div>
 		<# } #>
 		<div class="elementor-nerd-box-message">{{{ elementData.upsale_data.description }}}</div>
-		<a class="elementor-button go-pro" href="{{ elementData.upsale_data.upgrade_url }}" target="_blank"><?php echo esc_html__( 'Upgrade Now', 'elementor' ); ?></a>
+		<# if ( elementData.upsale_data.upgrade_text ) { #>
+		<a class="elementor-button go-pro" href='{{ elementData.upsale_data.upgrade_url }}' target="_blank"><?php echo esc_html( '{{ elementData.upsale_data.upgrade_text }}', 'elementor' ); ?></a>
```
TzviRabinovitch, line 202: maybe move the if one level down ?
when you use the elementData.upsale_data.upgrade_text

TzviRabinovitch: also . maybe already have default values in the array before sending the filter ?

core/role-manager/role-manager.php
Barcov: Good point, Now only a elementor.com as the base domain works.

core/utils/promotions/validate-promotion.php
```diff
@@ -0,0 +1,26 @@
...
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+class Validate_Promotion {
+
+	public static function domain_is_on_elementor_dot_com( $url ):bool {
+		$domain = parse_url( $url, PHP_URL_HOST );
+		if ( isset( $domain ) ) {
```
TzviRabinovitch, line 13: maybe invert if ?
return early

core/utils/promotions/validate-promotion.php
```diff
@@ -0,0 +1,26 @@
...
+
+			$domain_segments = explode( '.', $domain );
+
+			if ( count( $domain_segments ) >= 2 ) {
```
TzviRabinovitch, line 17: same here

includes/base/widget-base.php
```diff
@@ -371,12 +372,20 @@ protected function get_initial_config() {
 			'upsale_data' => $this->get_upsale_data(),
 		];
 
+		$config['upsale_data'] = apply_filters( 'elementor/widgets/' . $this->get_name() . '/custom_promotion', $config['upsale_data'] ) ?? $this->get_upsale_data();
+		if ( isset( $config['upsale_data']['image'] ) ) {
```
TzviRabinovitch, line 376: maybe a space line before the if ?

includes/base/widget-base.php
```diff
@@ -371,12 +372,20 @@ protected function get_initial_config() {
...
+			sanitize_file_name( $config['upsale_data']['image'] );
+		}
+
 		$stack = Plugin::$instance->controls_manager->get_element_stack( $this );
 
 		if ( $stack ) {
 			$config['controls'] = $this->get_stack( false )['controls'];
 			$config['tabs_controls'] = $this->get_tabs_controls();
 		}
+		if ( isset( $config['upsale_data']['upgrade_url'] ) && false === Validate_Promotion::domain_is_on_elementor_dot_com( $config['upsale_data']['upgrade_url'] ) ) {
```
TzviRabinovitch, line 386: why false ?

modules/promotions/admin-menu-items/custom-fonts-promotion-item.php
```diff
@@ -15,18 +17,47 @@ public function get_page_title() {
 		return esc_html__( 'Custom Fonts', 'elementor' );
 	}
 
+	public function get_image_url() {
+			$default_image = 'images/go-pro-wp-dashboard.svg';
```
TzviRabinovitch, line 21: indentation?

modules/promotions/admin-menu-items/custom-fonts-promotion-item.php
```diff
@@ -15,18 +17,47 @@ public function get_page_title() {
 		return esc_html__( 'Custom Fonts', 'elementor' );
 	}
 
+	public function get_image_url() {
```
louiswol94, line 20: Hey @Barcov 

While there is nothing inherently wrong with using the same filter multiple times, I do believe we can refactor the `base-promotion-item.php` a bit and make this more efficient.

Can you try and change the `render()` method to something like this:

```
public function render() {
		$data = [
			'image_url' => $this->get_image_url(),
			'promotion_title' => $this->get_promotion_title(),
			'cta_url' => $this->get_cta_url(),
			'cta_text' => $this->get_cta_text(),
		];

		$data = apply_filters( 'elementor/promotion/' . $this->get_name() . '/custom_promotion', $data );
		?>
		<div class="wrap">
			<div class="elementor-blank_state">
				<img src="<?php echo esc_url( $data['image_url'] ); ?>" loading="lazy" />

				<h3><?php Utils::print_unescaped_internal_string( $data['promotion_title'] ); ?></h3>

				<p><?php $this->render_promotion_description(); ?></p>

				<a class="elementor-button go-pro" href="<?php echo esc_url( $data['cta_url'] ); ?>">
					<?php Utils::print_unescaped_internal_string( $data['cta_text']  ); ?>
				</a>
			</div>
		</div>
		<?php
}
```
I am basically just building the `$data` array with the methods from the child classes, and then running the filter once to override all the array. And then instead of calling those methods inside the html I just use `$data['whatever']`.

`$this->render_promotion_description()` will have to change/remove - I am not sure why it was done like this in the first place. (To just echo something).
`get_name()` has to be introduced in the base class and implemented in the child classes where it's needed.

WDYT?

includes/base/widget-base.php
Barcov: The filter runs and this condition checks if the url returned from the filter is invalid and replaces the url with the default for this promotion in that case.

includes/base/widget-base.php
```diff
@@ -3,6 +3,7 @@
 
 use Elementor\Core\Page_Assets\Data_Managers\Responsive_Widgets as Responsive_Widgets_Data_Manager;
 use Elementor\Core\Page_Assets\Data_Managers\Widgets_Css as Widgets_Css_Data_Manager;
+use Elementor\core\utils\promotions\Validate_Promotion;
```
louiswol94, line 6:
```suggestion
use Elementor\Core\Utils\promotions\Validate_Promotion;
```

assets/dev/scss/admin/_role-manager.scss
```diff
@@ -49,6 +49,7 @@
 			.elementor-role-go-pro {
 				display: flex;
 				align-items: center;
+				padding: 1em 0 0;
```
Barcov, line 52: New json upload notice in Role Manager was added yesterday without any padding between it and the promotion.

core/utils/promotions/validate-promotion.php
TzviRabinovitch: maybe return early ?

if ( ! isset( $domain ) ) {
return false ;
}

core/utils/promotions/validate-promotion.php
TzviRabinovitch: same here
if  (count( $domain_segments ) < 2)
return false ;

modules/promotions/admin-menu-items/custom-fonts-promotion-item.php
TzviRabinovitch: indent?

modules/promotions/admin-menu-items/custom-fonts-promotion-item.php
```diff
@@ -15,18 +21,23 @@ public function get_page_title() {
...
-		echo esc_html__(
-			'Custom Fonts allows you to add your self-hosted fonts and use them on your Elementor projects to create a unique brand language.',
-			'elementor'
-		);
+		return esc_html__( 'Custom Fonts allows you to add your self-hosted fonts and use them on your Elementor projects to create a unique brand language.', 'elementor' );
 	}
 
 	public function get_cta_url() {
-		return 'https://go.elementor.com/go-pro-custom-fonts/';
+		return esc_url( 'https://go.elementor.com/go-pro-custom-fonts/', 'elementor' );
```
TzviRabinovitch, line 41: why was this added?

modules/promotions/admin-menu-items/base-promotion-item.php
```diff
@@ -33,17 +33,31 @@ public function get_image_url() {
 	}
 
 	public function render() {
+		$config = [
+			'title' => $this->get_promotion_title(),
+			'description' => $this->get_promotion_description(),
```
Barcov, line 38: Signature of `render_promotion_description():void` changed to `get_promotion_description():string`. The base change affects several other classes that depend on this base and are now added to this PR.

louiswol94: @Barcov I don't think we can delete `render_promotion_description()` as it's a public function.

I suggest:

Adding a new one for:

```
public function get_promotion_description() {
		return esc_html__(
			'Custom Code is a tool gives you one place where you can insert scripts, rather than dealing with dozens of different plugins and deal with code.',
			'elementor' );
}
```

And modifying:

```
public function render_promotion_description() {
		echo $this->get_promotion_description()
}
```

modules/promotions/admin-menu-items/go-pro-promotion-item.php
```diff
@@ -3,14 +3,19 @@
...
 use Elementor\Core\Admin\Menu\Interfaces\Admin_Menu_Item_With_Page;
+use Elementor\Core\Utils\Promotions\Validate_Promotion;
 use Elementor\Settings;
 
 if ( ! defined( 'ABSPATH' ) ) {
 	exit; // Exit if accessed directly
 }
 
 class Go_Pro_Promotion_Item implements Admin_Menu_Item_With_Page {
-	const URL = 'https://go.elementor.com/pro-admin-menu/';
```
louiswol94, line 13: @Barcov I don't think we can totally remove a const also for BC/3rd party reasons. I suggest keeping the const and changing the logic of `get_url()` to something like this:

```
public static function get_url() {
        $url = self::URL;
        $filtered_url = apply_filters( 'elementor/admin_menu/promotion/upgrade_url', [ 'upgrade_url' => $url ] )['upgrade_url'] ?? '';

        if ( true === Validate_Promotion::domain_is_on_elementor_dot_com( $filtered_url ) ) {
            $url = $filtered_url;
        }

        return esc_url( $url );
    }
```

modules/promotions/admin-menu-items/go-pro-promotion-item.php
```diff
@@ -21,7 +26,8 @@ public function get_parent_slug() {
 	}
 
 	public function get_label() {
-		return esc_html__( 'Upgrade', 'elementor' );
+		$upgrade_text = __( 'Upgrade', 'elementor' );
+		return apply_filters( 'elementor/admin_menu/promotion/upgrade_text', ['upgrade_text' => $upgrade_text ] )['upgrade_text'] ?? $upgrade_text;
```
louiswol94, line 30:
```suggestion
		return apply_filters( 'elementor/admin_menu/promotion/upgrade_text', $upgrade_text ) ?? $upgrade_text;
```
I don't see why this has to be an array if you are filtering just one string here.

modules/promotions/admin-menu-items/go-pro-promotion-item.php
```diff
@@ -32,6 +38,16 @@ public function get_capability() {
 		return 'manage_options';
 	}
 
+	public static function get_url() {
+		$filtered_url = apply_filters( 'elementor/admin_menu/promotion/upgrade_url', ['upgrade_url' => self::$url ] )['upgrade_url'] ?? '';
```
louiswol94, line 42: Same here.

louiswol94 (changes requested):

modules/promotions/admin-menu-items/go-pro-promotion-item.php
Barcov: As mentioned before Uniformity of the Promotion filters abstracted. As things stand I'm not yet happy with the suggested splitting of the filter in this file. There's no real benefit and it makes usage of of promotion filters harder to understand. Creating unique arguments data types for each Promotion filter creates the same problem leading to the needs for individually documentation in what could be really cleaner.

Abstractions like these are ideally meant to hide the complexity and inconsistencies in the underlying implementation which in this case has naming inconsistency and data types.

modules/promotions/admin-menu-items/go-pro-promotion-item.php
Barcov: @louiswol94 Interesting suggestion.

modules/promotions/admin-menu-items/go-pro-promotion-item.php
```diff
@@ -21,7 +26,9 @@ public function get_parent_slug() {
 	}
 
 	public function get_label() {
-		return esc_html__( 'Upgrade', 'elementor' );
+		$upgrade_text = __( 'Upgrade', 'elementor' );
+		$promotion['upgrade_text'] = $upgrade_text;
+		return apply_filters( 'elementor/admin_menu/promotion/upgrade_text', $promotion )['upgrade_text'] ?? $upgrade_text;
```
Barcov: This requested change to have two filters here is not ideal from an abstraction point of view. The filter interface exposed to a dev mutating promotions should be as uniform as possible to avoid confusion. Promotions always have one or more of the same elements i.e. title, description, image, upgrade_text, upgrade_url.

modules/promotions/admin-menu-items/custom-code-promotion-item.php
```diff
@@ -19,6 +19,17 @@ public function get_promotion_title() {
...
+		);
+	}
+
+	/**
+	 * @deprecated use get_promotion_description instead
+	 * @return void
+	 */
 	public function render_promotion_description() {
 		echo esc_html__(
 			'Custom Code is a tool gives you one place where you can insert scripts, rather than dealing with dozens of different plugins and deal with code.',
```
louiswol94, line 35: @Barcov 

```
public function render_promotion_description() {
     echo $this->get_promotion_description();
}
```

Otherwise you are adding the text twice.

modules/promotions/admin-menu-items/go-pro-promotion-item.php
louiswol94: The Admin menu item is different than the other Promotions.

The other Promotions have one config array that you can filter all the items with one filter.

The Admin menu item has a separate Label and URL, with **two different filter names**, and is just a simple string that needs to be returned. You don't have to make this an array if we just need to filter a simple string and assign the string to some key.

modules/promotions/admin-menu-items/go-pro-promotion-item.php
```diff
@@ -3,14 +3,19 @@
...
+use Elementor\Core\Utils\Promotions\Validate_Promotion;
 use Elementor\Settings;
 
 if ( ! defined( 'ABSPATH' ) ) {
 	exit; // Exit if accessed directly
 }
 
 class Go_Pro_Promotion_Item implements Admin_Menu_Item_With_Page {
-	const URL = 'https://go.elementor.com/pro-admin-menu/';
+	CONST URL = 'https://go.elementor.com/pro-admin-menu/';
```
louiswol94, line 14:
```suggestion
	const URL = 'https://go.elementor.com/pro-admin-menu/';
```
There does not need to be a change here.

modules/promotions/admin-menu-items/base-promotion-item.php
```diff
@@ -33,17 +33,31 @@ public function get_image_url() {
...
+			'image' => $this->get_image_url(),
+			'upgrade_text' => $this->get_cta_text(),
+			'upgrade_url' => $this->get_cta_url(),
+		];
+
+		$config = apply_filters( 'elementor/' . $this->get_name() . '/restrictions/custom_promotion', $config );
```
louiswol94, line 44:
```suggestion
		$config = apply_filters( 'elementor/' . $this->get_name() . '/custom_promotion', $config );
```
I don't think `restrictions` make sense for all the promotions?

modules/promotions/admin-menu-items/go-pro-promotion-item.php
Barcov: I don't agree here. From the filter abstraction end this is one and the same thing. It's what abstractions are designed to do - hide underlying complexity. I've reduced this to a string in the function as you asked but the filter and the design of the filter abstraction is core to this PR and not something I'm willing to mutate into endless complexity for devs who implement this filter in the end.

modules/promotions/admin-menu-items/base-promotion-item.php
Barcov: > Do you need to create a get_name() in this base class?
Yes because you insisted on me creating `get_name()` @louiswol94. This PR started out perfectly functional and the refactor we made at your request in this file might be beneficial to future extension of related promotions. But it only affected one promotion in this PR. So it's inflated this PR quite a lot because this ended up dragging in unrelated files into the PR. But now we're just going in circles. 

> I don't think restrictions make sense for all the promotions?
Naming flows quite nicely at the moment except for the admin menu change requested. As you can see not all filters had 'restrction' only role manager to follow the naming of the action and fonts which is related. 
image

I'm shortening both font and role manager naming. 
image

modules/promotions/admin-menu-items/custom-code-promotion-item.php
Barcov: Honestly I think these changes should not have made it into this PR in the first place. This changes is exactly like you asked for earlier today. 
image

So we can go around again but it's wasting valuable company time.

louiswol94: @Barcov What do you mean? We could not just totally delete a public function 🤔

louiswol94 (approved):

includes/base/widget-base.php
```diff
@@ -371,12 +372,21 @@ protected function get_initial_config() {
 			'upsale_data' => $this->get_upsale_data(),
 		];
 
+		$config['upsale_data'] = apply_filters( 'elementor/widgets/' . $this->get_name() . '/custom_promotion', $config['upsale_data'] ) ?? $this->get_upsale_data();
```
KingYes, line 375: @TzviRabinovitch @Barcov 
Why do you have multiple calls to `$this->get_upsale_data()`?

TzviRabinovitch: Looks like a redundent call

louiswol94: What if someome unsets the variable using the filter? Then we fallback to get_upsale_data() because we never want the variable to be empty.

KingYes: So, store the original outside the filter. Why not call multiple times?

louiswol94: Understood.
@Barcov you can tweak this in ED-13558

PR #24962, opened 2024-01-29
Tweak: Make widget promotion area to be more flexible [ED-13533]

rami-elementor:


includes/base/widget-base.php
```diff
@@ -384,7 +394,8 @@ protected function get_initial_config() {
 			$config['controls'] = $this->get_stack( false )['controls'];
 			$config['tabs_controls'] = $this->get_tabs_controls();
 		}
-		if ( isset( $config['upsale_data']['upgrade_url'] ) && false === Validate_Promotion::domain_is_on_elementor_dot_com( $config['upsale_data']['upgrade_url'] ) ) {
```
Barcov, line 387: There was an acceptance criteria to only allow url's on elementor.com when a filter overrides the promotion. This conditional resets the filter return value with the default promotion url if it's invalid.

includes/editor-templates/panel.php
```diff
@@ -186,22 +186,22 @@
 		</div>
 	<# } #>
 
-	<?php if ( ! Utils::has_pro() ) : ?>
-	<# if ( elementData.upsale_data && elementData.upsale_data.description ) { #>
+	<# if ( elementData.upsale_data && elementData.upsale_data.condition ) { #>
 	<div class="elementor-nerd-box elementor-nerd-box--upsale">
 		<# if ( elementData.upsale_data.image ) { #>
-		<img class="elementor-nerd-box-icon" src="<?php Utils::print_unescaped_internal_string( '{{ elementData.upsale_data.image }}' ); ?>" loading="lazy" alt="<?php echo esc_attr__( 'Upgrade', 'elementor' ); ?>" />
```
Barcov, line 193: elementData.upsale_data.image is used by a filter in the hosting team to override images widget promotions for cloud hosting.

includes/widgets/button.php
```diff
@@ -77,6 +77,28 @@ public function get_categories() {
...
+	protected function get_upsale_data() {
+		return [
+			'condition' => ! Utils::has_pro(),
+			'image' => ELEMENTOR_ASSETS_URL . 'images/go-pro.svg',
+			'image_alt' => esc_attr__( 'Upgrade', 'elementor' ),
+			'title' => esc_html__( 'Convert visitors into customers', 'elementor' ),
+			'description' => esc_html__( 'Get the Call to Action widget and grow your toolbox with Elementor Pro.', 'elementor' ),
+			'upgrade_url' => esc_url( 'https://go.elementor.com/go-pro-button-widget/' ),
+			'upgrade_text' => esc_html__( 'Upgrade Now', 'elementor' ),
+		];
```
Barcov, line 99: @rami-elementor Love this

includes/editor-templates/panel.php
```diff
@@ -186,22 +186,22 @@
...
 		<# } #>
 		<# if ( elementData.upsale_data.title ) { #>
-		<div class="elementor-nerd-box-title">{{ elementData.upsale_data.title }}</div>
+		<div class="elementor-nerd-box-title">{{{ elementData.upsale_data.title }}}</div>
 		<# } #>
+		<# if ( elementData.upsale_data.description ) { #>
 		<div class="elementor-nerd-box-message">{{{ elementData.upsale_data.description }}}</div>
-		<a class="elementor-button go-pro" href="{{ elementData.upsale_data.upgrade_url }}" target="_blank"><?php echo esc_html( '{{ elementData.upsale_data.upgrade_text || `Upgrade Now` }}', 'elementor' ); ?></a>
+		<# } #>
+		<# if ( elementData.upsale_data.upgrade_url ) { #>
```
Barcov, line 200: This is good, maybe extend Validate_Promotion::domain_is_on_elementor_dot_com() to domain_is_on_elementor_dot_com_or_empty() to allow empty values for use in widget classes and filter overrides? Then this could allow removal of the button by filters or in select widgets.

includes/editor-templates/panel.php
```diff
@@ -186,22 +186,22 @@
 		</div>
 	<# } #>
 
-	<?php if ( ! Utils::has_pro() ) : ?>
```
Barcov, line 189: This line is important. These promotions only show if pro is not active. These promotions are hidden once Pro licensing is active. We have separate promotions in the pro repo for essential to advanced promotions.

rami-elementor: @Barcov 

This condition checks if Elementor Pro in not installed. However, external addon developers need to check whether their Premium plugins are not installer.

Therefor, I removed this from the editor-template, and now it's part of `get_upsale_data()` method. You have set `condition` arguments, with different activation condition, to control when to display promotions.

rami-elementor: Check the condition bellow, we now check for `elementData.upsale_data.condition`.

includes/base/widget-base.php
rami-elementor: What if the promotion is added by 3rd party addon? you can't reset their images.

includes/editor-templates/panel.php
rami-elementor: No, my intention was to allow 3rd party addon developers to create their own promotions.

includes/base/widget-base.php
Barcov: @rami-elementor Yes I see that's where you're headed. And I agree with the general direction to allow 3rd parties to use promotions on their own widgets.  Been advocating for keeping filters very standard to be kind to hosting and 3rd parties using this.

I'm just wondering how we keep 3rd parties from highjacking links on promotions for Elementor widgets.

includes/editor-templates/panel.php
Barcov: We're aligned on adding more flexibility.

includes/editor-templates/panel.php
Barcov: I'll consider a way to do this for Elementor widgets but not 3rd party widgets.

includes/editor-templates/panel.php
Barcov: Love that!

includes/base/widget-base.php
rami-elementor: Good point.

rami-elementor: Change reverted. For now.

includes/editor-templates/panel.php
rami-elementor: @Barcov Template updated.

KingYes (approved):

Issue #16473, opened 2021-10-04
invalid handling of content encoding


user982734:
The code attempting to use gzip content encoding in module.php should be removed.
It doesn't handle if the client doesn't accept gzip, nor if some other encoding is already enabled or requested by another wordpress plugin or the webserver itself.
It violates RFC.

https://github.com/elementor/elementor/blob/f9ad25409ac3704bdd8c2ea463ca10a8c3518a3b/core/common/modules/ajax/module.php#L264

phillip-elm:
Can confirm this. It is usually the web server / reverse proxy's job to handle response encoding. Modern web servers can also use brotli, which is preferable over gzip. 

At the very least this code should be checking to see if the client accepts gzip encoding first.

Dekadinious:
Can anyone please look at this? Getting tired of removing this in the plugin for each update...

Dekadinious:
We are approaching a year on this from my last comment. The only thing you need to do is remove an if-block and return the pure $json. The functionality of everything will be exactly the same as today, except that you will follow RFC and I will not have to manually edit the plugin code on each update.

The reverse proxy/server will still serve with the correct encoding.

Dekadinious:
Dear Elementor, here is your solution. Change this:

```
	/**
	 * Ajax success response.
	 *
	 * Send a JSON response data back to the ajax request, indicating success.
	 *
	 * @since 2.0.0
	 * @access protected
	 */
	private function send_success() {
		$response = [
			'success' => true,
			'data' => [
				'responses' => $this->response_data,
			],
		];

		$json = wp_json_encode( $response );

		while ( ob_get_status() ) {
			ob_end_clean();
		}

		echo $json; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped

		wp_die( '', '', [ 'response' => null ] );
	}
```

To this:

```
	/**
	 * Ajax success response.
	 *
	 * Send a JSON response data back to the ajax request, indicating success.
	 *
	 * @since 2.0.0
	 * @access protected
	 */
	private function send_success() {
                // Keeping this as I don't understand if you need it for some upstream stuff?
		while ( ob_get_status() ) {
			ob_end_clean();
		}

               wp_send_json_success( $this->response_data );
	}
```

Please take some time to see if this can be implemented.

Dekadinious:
Circling back after 5 more updates where I have to remove the code that is breaking the site. Will come back in 5 more updates.

Issue #17486, opened 2022-01-14
gzencode is eating too much memory


skito:
Hi There,

Line 265 in `core/common/modules/ajax/module.php` is eating too much memory. I noticed few cases on different websites now where it needs more than 500MB of memory limit to execute, which is totally unacceptable.

This is the snippet
```php
if ( function_exists( 'gzencode' ) ) {
	$response = gzencode( $json ); // Line 256. HERE IT BREAKS

	header( 'Content-Type: application/json; charset=utf-8' );
	header( 'Content-Encoding: gzip' );
	header( 'Content-Length: ' . strlen( $response ) );

	echo $response; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
} else {
	echo $json; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
}

```

In general having gzip like that is useless. Most hosting environments already support compression in one or another form. **Compressed response should be handled by the web server, not by the PHP app!**

I really hope you remove it in the next releases.

MichaelDarko:
+1

bart-prosite:
+1
I have been messing with this issue for a long time. This makes some websites unusable!
I couldn't find a solution. Maybe this should be given another title. This is the solution for the error 500 calling /wp-admin/ajax.php when opening a page with Elementor Editor and the editor not fully loading.

bart-prosite:
@skito : how would you rewrite the code? Then at least we can patch this and solve this for our own websites.

skito:
@bart-prosite Just comment the highlighted code above and leave only ``echo $json``

Issue #13170, opened 2020-11-26
Side Bar Loading issue - Ajax call failed


arasu1997:
Hi, In elementor edit pages, the sidebar is continuously loading.

While debugging it looks a ajax call with action 'get_widgets_config' is failing. The response coming for this ajax call is  {"message": "Internal server error"} with 502 error code.

In that ajax call, there is a logic to compress the response using "gzencode()" zlib's method. I've installed gd, zip, zlib extentions and also enabled gzip compression. But still this error occurs. 

If I directly return the response without compression, It is working fine. Kindly help us to fix this issue. 

FYI: Added screenshots of the issue.
https://imgur.com/a/4dhxAmo

System Info :					
== Server Environment ==
	Operating System: Linux
	Software: PHP 7.3.25 Development Server
	MySQL version: MySQL Community Server (GPL) v1.2.10
	PHP Version: 7.3.25
	PHP Max Input Vars: 1000
	PHP Max Post Size: 8M
	GD Installed: Yes
	ZIP Installed: Yes
	Write Permissions: There are some writing permissions issues with the following directories/files:
		 - WordPress root directory
	Elementor Library: Connected

== WordPress Environment ==
	Version: 5.5.1
	Site URL: https://drk3n9bxpb5o2.cloudfront.net
	Home URL: https://drk3n9bxpb5o2.cloudfront.net
	WP Multisite: No
	Max Upload Size: 2 MB
	Memory limit: 1024M
	Permalink Structure: Plain
	Language: en-US
	Timezone: 0
	Debug Mode: Active

== Theme ==
	Name: Twenty Twenty
	Version: 1.5
	Author: the WordPress team
	Child Theme: No

== User ==
	Role: administrator
	WP Profile lang: en_US
	User Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/87.0.4280.67 Safari/537.36

== Active Plugins ==
	Contact Form 7
		Version: 5.3
		Author: Takayuki Miyoshi

	Elementor
		Version: 3.0.12
		Author: Elementor.com

	S3 Uploads
		Version: 2.1.0-RC2
		Author: Human Made Limited

	WooCommerce
		Version: 4.6.0
		Author: Automattic

== Elements Usage ==
	
	wp-page : 1
		button : 1
		heading : 2

== Log ==

ptasz3k:
[...]

PR #25000, opened 2024-02-02
Fix: Remove gzip encoding in AJAX call [ED-20467]

Dekadinious:
## PR Checklist

## PR Type
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description
Fixes fatal error when trying to save Elementor pages on certain servers. It is the server's job to handle encoding. Encoding the response like this breaks Elementor's functionality completely on certain servers. We have to remove this after each update for Elementor to even work on our servers.

*

## Test instructions
Use a server that can't save pages because of the GZIP-encoding. Try to save. It will fail. Then test with the new code. You will be able to save without errors.

*

## Quality assurance

Fixes #
https://github.com/elementor/elementor/issues/16473

### ✨ PR Description
Purpose: Remove gzip compression from AJAX responses to prevent potential client-side compatibility issues.
Main changes:
- Removed conditional gzip encoding for AJAX response data
- Set consistent Content-Type header without gzip-specific headers


Dekadinious:
@louiswol94 Taking the liberty to tag you, because this breaks and breaks and breaks on every single update.

louiswol94:
Hi @Dekadinious 

Thanks for the tag.

Can you please explain the issue in further detail. What is the error you are getting?

As I see in the code it checks if `gzencode` is available on the server before it tries to call it?

Maybe @kobizz or @bainternet wants to chime in here? I see you worked on this part before.

Thanks!

Dekadinious:
@louiswol94 Yes, we get a 500-error: Backend responded with an invalid content-encoding gzip when saving pages when this if-block is in place. It's probably because the server decides to not encode in the way you are expecting, or maybe double encodes or something. It's not my area of expertise, and I know it does not happen for many hosts, but it happens on some. The host says that this if-block is not correct to have in the code. I will try to get more info from them.

Dekadinious:
@louiswol94 

So the answer is: "Checking if a php-function exists does not mean that a web client wants a certain response. This is also true for gzip-encoded responses. Therefore the code has [to be] been removed."

So it seems that you can't just send clients a certain encoding just because the function exists in PHP. The client in this case does not want that encoding, and so the whole thing breaks.

louiswol94:
> @louiswol94
> 
> So the answer is: "Checking if a php-function exists does not mean that a web client wants a certain response. This is also true for gzip-encoded responses. Therefore the code has [to be] been removed."
> 
> So it seems that you can't just send clients a certain encoding just because the function exists in PHP. The client in this case does not want that encoding, and so the whole thing breaks.

Can you share the debug log of the 500 error please.

Dekadinious:
It's not logging anything in the logs, but the response from the server is:

500-error: Backend responded with an invalid content-encoding gzip

This is plain text as a response to the AJAX call instead of what would be the actual response.

@louiswol94

louiswol94:
@Dekadinious thanks for your patience while I continue to investigate the issue.

I have a few more questions:

1. Please share your system information.
2. Is the issue on all browsers?

Dekadinious:
Got this from my host:

"
​
Du
​
Linux x64. WebCDN optimizer.

System information or browsers is not really relevant here.
The behaviour is out of spec with RFC7231 by not honoring the Accept-Encoding header which in this case clearly states that gzip encoding is not allowed.
Please see RFC7231 §5.3.4"

And as I see it the issue exists in all browsers tried.

Dekadinious:
@louiswol94 Any status on this? Getting tiresome to edit the plugin after each update :)

louiswol94:
Fixes https://github.com/elementor/elementor/issues/16473, Fixes https://github.com/elementor/elementor/issues/13170, Fixes https://github.com/elementor/elementor/issues/17486

louiswol94:
I heard from @kobizz from what he remembers this code fixed an encoding issue on servers that gzipping responses by default. We'll have to thoroughly test it on various server's config and see if there are any negative impact on Ajax requests.

Dekadinious:
It is good that this is fixed. I see the AI comes up with a possibly valid concern. I think this is easily fixed by using wp_send_json_success rather than just echoing json directly.

core/common/modules/ajax/module.php
```diff
@@ -266,17 +266,7 @@ private function send_success() {
...
-
-			header( 'Content-Type: application/json; charset=utf-8' );
-			header( 'Content-Encoding: gzip' );
-			header( 'Content-Length: ' . strlen( $response ) );
-
-			echo $response; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
-		} else {
-			echo $json; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
-		}
+		echo $json; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
louiswol94, line 270:
```suggestion
		header( 'Content-Type: application/json; charset=UTF-8' );
		echo $json; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
@Dekadinious Can you please apply the suggestion?

louiswol94:
@Dekadinious I am going to merge it into the `main` branch, but we need to keep a close eye on this as it is a sensitive area, and there are many different server environments and configurations. But seeing that this is a very old code block, I feel more confident that modern servers can handle the original issue better and hopefully we did not re-introduce an old bug by removing that code. But from my research I am of the opinion that we should let the server handle gzip internally and it might have been bad practice to skip Accept-Encoding negotiation.

louiswol94 (approved):

louiswol94:
> Fixes #16473, Fixes #13170, Fixes #17486

CC @nicholaszein

PR #25011, opened 2024-02-04
Internal: Allow hosting to replace more promotions [ED-13518]

TzviRabinovitch:


modules/promotions/admin-menu-items/form-submissions-promotion-item.php
```diff
@@ -21,15 +19,15 @@ public function get_page_title() {
 		return esc_html__( 'Submissions', 'elementor' );
 	}
 
-	public function get_promotion_title() {
+	protected function get_promotion_title(): string {
```
davseve, line 22: I see that the access modifiers were changed only in this file.
Maybe it is better to change all the access modifiers for the desired function in the rest of the desired classes?

modules/promotions/admin-menu-items/custom-icons-promotion-item.php
```diff
@@ -21,14 +20,14 @@ public function get_page_title() {
...
-		echo sprintf(
+		return sprintf(
 			/* translators: %s: br  */
 			esc_html( 'Enjoy creative freedom %s with Custom Icons', 'elementor' ),
 			'<br />'
 		);
 	}
 
-	public function set_list() {
+	public function get_content_lines(): array {
```
davseve, line 30: If we are already changing the API, why not to change the access modifiers as well?

modules/promotions/admin-menu-items/custom-icons-promotion-item.php
```diff
@@ -38,11 +37,11 @@ public function set_list() {
 		];
 	}
 
-	public function get_cta_url() {
+	public function get_cta_url(): string {
 		return 'https://go.elementor.com/go-pro-custom-icons/';
 	}
 
-	public function get_video_url() {
+	public function get_video_url(): string {
```
davseve, line 44: Please Change the access modifier

davseve (dismissed):

davseve (approved):

louiswol94:
@TzviRabinovitch @davseve @Barcov 
I don't know the full story, but so far I think it is unsafe that we changed these function names in Core while this base class is being used in Pro (core/pro dependency).

Barcov:
> @TzviRabinovitch @davseve @Barcov I don't know the full story, but so far I think it is unsafe that we changed these function names in Core while this base class is being used in Pro (core/pro dependency).

Yes agreed it's what I've been calling for prior to our decision to break the core/pro dependancy on Friday. But even then changing signatures is not ideal. David and I discussed keeping get_description() and injecting the <li> in. I implemented this on Friday's merge for Pro, and David would have implemented this in core, but the discussed changed never made it into core. Some of these PRO changes seems to have vanished by Monday morning ? So now we have a core dependancy again.

PR #25085, opened 2024-02-11
Tweak: Hide `size` control from Button Widget [ED-12944]

rami-elementor:


arielk (approved):

DataGhost:
Is there any rationale for this? It looks like there's a related ticket I can't see. Some people not tech savvy enough to find this PR just ran into an issue where they have an XL button on a page but can't add more of them without copy/pasting the existing one. So it's still working, just the functionality is hidden. Why?

arcaswebdesign:
I have to agree with @DataGhost, I'm not clear on the rationale here. Yes we can use the Style settings but this means having to set the style for each individual instance rather than simply being able to choose a button size then allow a stylesheet to set the styling for us, giving us more flexibility to adjust that styling later.

toineenzo:
> Is there any rationale for this? It looks like there's a related ticket on a separate tracker I can't see. Some people not tech savvy enough to find this PR just ran into an issue where they have an XL button on a page but can't add more of them without copy/pasting the existing one and they have no idea why. So it's still working, just the functionality is hidden and it seems to be causing needless frustration. Why?

I think this is done in preperation for the CSS Selectors functionality

rafadelnero:
What is the option now to do this configuration on Elementor? Do we have to do it manually now?

PR #25086, opened 2024-02-11
Internal: Allow hosting to replace the Elements Manager - Elementor Pro Elements [ED-13688]

davseve:


modules/element-manager/ajax.php
```diff
@@ -68,6 +79,50 @@ public function ajax_get_admin_page_data() {
...
+
+		return array_replace( $promotion_data, $filtered_data );
+	}
+
+	private function validate_promotion_data( $filtered_data, $promotion_data ) {
+		if ( Validate_Promotion::domain_is_on_elementor_dot_com( $filtered_data['url'] ) ) {
+			$filtered_data['url'] = esc_url( $filtered_data['url'] );
+		}
+
+		$filtered_data['text'] = $filtered_data['text'] && $promotion_data['text'];
```
louiswol94, line 121:
```suggestion
		$filtered_data['text'] = $filtered_data['text'] ?? $promotion_data['text'];
```
Maybe?

modules/element-manager/ajax.php
```diff
@@ -68,6 +79,50 @@ public function ajax_get_admin_page_data() {
...
+
+		$this->validate_promotion_data( $filtered_data, $promotion_data );
+
+		return array_replace( $promotion_data, $filtered_data );
+	}
+
+	private function validate_promotion_data( $filtered_data, $promotion_data ) {
+		if ( Validate_Promotion::domain_is_on_elementor_dot_com( $filtered_data['url'] ) ) {
+			$filtered_data['url'] = esc_url( $filtered_data['url'] );
+		}
```
louiswol94, line 119: @davseve Not sure if this part makes sense.

You are returning `$filtered_data`, so if `domain_is_on_elementor_dot_com()` dis not pass, it will just return the malicious URL.

I would expect something like:

```
if ( ! Validate_Promotion::domain_is_on_elementor_dot_com( $filtered_data['url'] ) ) {
		unset( $filtered_data['url'] );
}
```

This way it won't be replaced in `array_replace` and keep the original URL.

modules/element-manager/assets/js/app.js
```diff
@@ -590,7 +598,8 @@ export const App = () => {
 									</FlexItem>
 									<FlexItem>
 										<UpgradeButton
-											href="https://go.elementor.com/go-pro-element-manager/"
+											href={ elementManager.url }
+											text={ elementManager.text }
```
louiswol94, line 602: @davseve Should you not update the `UpgradeButton` component as well?

In `elementor/modules/element-manager/assets/js/upgrade-button.js` the text is still hard coded and not using the props.

davseve: The hard-coded text has been removed, by passing the `text` prop I update the text inside the button component which the `update_button` inherits from.

I've tested this manually and it works.

louiswol94: @davseve I did not see the changes in `upgrade-button.js`, maybe I missed it, sorry.

modules/element-manager/ajax.php
davseve: Good comment, I missed that.
Fixed.

louiswol94 (dismissed):

modules/element-manager/ajax.php
```diff
@@ -68,6 +79,48 @@ public function ajax_get_admin_page_data() {
...
+			'text' => esc_html__( 'Upgrade Now', 'elementor' ),
+			'url' => self::ELEMENT_MANAGER_PROMOTION_URL,
+		];
+
+		$filtered_data = apply_filters( 'elementor/element_manager/admin_app_data/promotion_data/element_manager', $promotion_data );
+
+		return $this->validate_promotion_data( $filtered_data, $promotion_data );
+	}
+
+	private function validate_promotion_data( $filtered_data, $promotion_data ) {
```
TzviRabinovitch, line 114: validate_promotion_data replaces the values?

murzin-nikita (dismissed):

modules/element-manager/ajax.php
TzviRabinovitch: maybe check that the returned array is not too big ?

louiswol94: > maybe check that the returned array is not too big ?

https://github.com/elementor/elementor/pull/25088/files#diff-f2b395a7fa05fe6f81da6c9c5b1616122fca1495d0a19bb07b11146f32dce463R239

davseve: @TzviRabinovitch @louiswol94 
I don't think checking the array length is a good solution.
What if the array length is the same but the keys are different?
I see two possible solutions:
1. Maybe we should obligate the end user to use all the keys in the base array.
2. We can map the base array and change the values only on the existing keys.

WDYT?

TzviRabinovitch: if the keys are different than we end up with x2 the size which is "ok" , the concern is that it might be x1000

modules/element-manager/ajax.php
```diff
@@ -68,6 +79,48 @@ public function ajax_get_admin_page_data() {
...
+		if ( ! Validate_Promotion::domain_is_on_elementor_dot_com( $filtered_data['url'] ) ) {
```
TzviRabinovitch, line 115: nice !!!
I will move this login to a common location in another PR

murzin-nikita (approved):

PR #25168, opened 2024-02-16
Internal: Add Upgrade button on each Pro section in the widget panel [ED-13722]

louiswol94:


murzin-nikita (dismissed):

core/base/document.php
```diff
@@ -192,6 +193,39 @@ public static function get_editor_panel_config() {
 		];
 	}
 
+	public static function get_filtered_editor_panel_categories(): array {
+		$categories = static::get_editor_panel_categories();
+
+		foreach ( $categories as $index => $cat ) {
```
hein-obox, line 199: Maybe?

```suggestion
		foreach ( $categories as $index => $category ) {
```

core/base/document.php
```diff
@@ -192,6 +193,39 @@ public static function get_editor_panel_config() {
...
+	 */
+	private static function get_panel_category_item( $promotion, $index, array $categories ): array {
+		if ( ! Utils::has_pro() ) {
+			$categories[ $index ]['promotion'] = Filtered_Promotions_Manager::get_filtered_promotion_data(
+				$promotion,
+				'elementor/panel/' . $index . '/custom_promotion',
+				'url'
+			);
+		} else {
+			unset( $categories[ $index ]['promotion'] );
```
hein-obox, line 223: Is this always set? Do we have to use `unset()`?

core/document-types/page-base.php
```diff
@@ -46,6 +46,9 @@ protected static function get_editor_panel_categories() {
 				'theme-elements-single' => [
 					'title' => esc_html__( 'Single', 'elementor' ),
 					'active' => false,
+					'promotion' => [
```
hein-obox, line 49: Or is this the answer to the previous question?

core/base/document.php
louiswol94: Done

core/base/document.php
louiswol94: @hein-obox Yes, it was easier to unset here in one place instead of building the categories array conditionally across various places.

core/document-types/page-base.php
louiswol94: @hein-obox answered above.

hein-obox (dismissed):

murzin-nikita (approved):

PR #25210, opened 2024-02-20
Internal: Improve Nested Elements repeaters performance - Accordion - Delete [ED-13762]

davseve:


assets/dev/js/editor/document/repeater/commands/remove.js
```diff
@@ -58,11 +58,16 @@ export class Remove extends $e.modules.editor.document.CommandHistoryBase {
 
 			// Remove from container and add to result.
 			result.push( repeaterContainer.children.splice( index, 1 ) );
-
 			collection.remove( model );
 
-			// Trigger render on widget but with the settings of the control.
-			repeaterContainer.render();
+			if ( 'Accordion' === container.label ) {
```
louiswol94, line 63: @davseve This base command should be generic and not aware of any specific feature.

Change this if statement to:

```
if ( isWidgetSupportNesting( widget_name ) && isWidgetSupportImprovedRepeaters( widget_name ) ) {
}
```

isWidgetSupportNesting() is already defined in `elementor/modules/nested-elements/assets/js/editor/utils.js`

You can create a function `isWidgetSupportImprovedRepeaters()` and then add some property on the widget config like `support_improved_repeaters`

Whatever else you need like `.e-n-accordion-item` can also be a part of the widget config. `protected function get_initial_config() {}`

louiswol94: @davseve This just manually removes the item title, so I just want to make sure the child Container is also removed?

louiswol94 (changes requested):

assets/dev/js/editor/document/repeater/commands/remove.js
davseve: Thanks for the detailed explanation, I'll do so.
And yes, the remove removes the entire element and its content.

louiswol94: For `support_improved_repeaters => true`, we can maybe instead use `repeater_action_type => 'single-node'`.
I will leave it up to you to put some thought into this.

louiswol94: @ArmanPurtoian Same for your task 🙏

louiswol94: @davseve Lastly, you need to think about how we can support the different nested structures. In your case we were lucky that both items (title & container) were wrapped in one element that can be removed. But in Mega Menu the Container is somewhere else (and will be removed by the above mentioned hook). I know the MM is a different widget, but I think we should support the different kind of structures now in this base already.

louiswol94: Note: this should be behind an experiment.

assets/dev/js/editor/document/repeater/commands/remove.js
```diff
@@ -61,8 +64,14 @@ export class Remove extends $e.modules.editor.document.CommandHistoryBase {
...
 			collection.remove( model );
 
-			// Trigger render on widget but with the settings of the control.
-			repeaterContainer.render();
+			if ( isWidgetSupportNesting( widgetType ) && isWidgetSupportImprovedRepeaters( widgetType ) ) {
+				const widgetContainer = container.document.$element[ 0 ];
+				const targetElement = widgetContainer.querySelectorAll( widgetNodes( widgetType ).targetContainer );
+				targetElement[ index ].remove();
+			} else {
+				// Trigger render on widget but with the settings of the control.
```
hein-obox, line 72: Maybe?
Extract this to a function so that we can avoid the comment?

modules/nested-elements/assets/js/editor/utils.js
```diff
@@ -24,3 +47,9 @@ export function findChildContainerOrFail( container, index ) {
 
 	return childView.getContainer();
 }
+
+export function shouldUseImprovedRepeaters( widgetType ) {
+	return elementorCommon.config.experimentalFeatures.e_nested_atomic_repeaters &&
+	isWidgetSupportNesting( widgetType ) &&
+	isWidgetSupportImprovedRepeaters( widgetType )
```
louiswol94, line 54:
```suggestion
	isWidgetSupportImprovedRepeaters( widgetType );
```

ArmanPurtoian (approved):

louiswol94 (approved):

PR #25216, opened 2024-02-20
New: Add Optimized Control Loading experiment [ED-13839]

KingYes:


core/frontend/performance.php
```diff
@@ -0,0 +1,40 @@
...
+
+	private static $use_style_controls = false;
+
+	private static $is_frontend = null;
+
+	private static $is_optimized_control_loading_feature_enabled = null;
+
+	public static function set_use_style_controls( $bool ): void {
+		static::$use_style_controls = (bool) $bool;
+	}
```
StyleShit, line 16:
```suggestion
	public static function set_use_style_controls( bool $bool ): void {
		static::$use_style_controls = $bool;
	}
```

core/frontend/performance.php
```diff
@@ -0,0 +1,40 @@
...
+
+		return static::$is_frontend;
+	}
+
+	public static function is_optimized_control_loading_feature_enabled() {
+		if ( null === static::$is_optimized_control_loading_feature_enabled ) {
+			static::$is_optimized_control_loading_feature_enabled = Plugin::$instance->experiments->is_feature_active( 'e_optimized_control_loading' );
+		}
+
+		return static::$is_optimized_control_loading_feature_enabled;
```
StyleShit, line 38: why do you need this caching mechanism?

includes/controls/groups/background.php
```diff
@@ -153,7 +153,7 @@ public function init_fields() {
 		$fields['background'] = [
 			'label' => esc_html__( 'Background Type', 'elementor' ),
 			'type' => Controls_Manager::CHOOSE,
-			'render_type' => 'ui',
+			//'render_type' => 'ui',
```
StyleShit, line 156: ?

includes/managers/controls.php
```diff
@@ -798,6 +801,15 @@ public function add_control_to_stack( Controls_Stack $element, $control_id, $con
 			'index' => null,
 		];
 
+		$control_type = 'controls';
+		if (
+			Performance::is_optimized_control_loading_feature_enabled()
+			&& Performance::should_optimize_controls()
```
StyleShit, line 807: why `should_optimize_controls` doesn't return whether the feature is enabled?

```suggestion
			Performance::should_optimize_controls()
```

includes/managers/controls.php
```diff
@@ -1337,4 +1349,26 @@ private function promotion_select_control( $title, $id ): string {
...
+			return true;
+		}
+
+		$render_type = $control_data['render_type'] ?? '';
+		if ( 'ui' === $render_type ) {
+			return true;
+		}
+
+		return false;
+	}
```
StyleShit, line 1373: isn't it enough to just check for `selectors`?

core/experiments/manager.php
```diff
@@ -455,6 +455,14 @@ private function add_default_features() {
...
 			],
 		] );
+
+		$this->add_feature( [
+			'name' => 'e_optimized_control_loading',
+			'title' => esc_html__( 'Optimized Control Loading', 'elementor' ),
+			'tag' => esc_html__( 'Performance', 'elementor' ),
+			'release_status' => self::RELEASE_STATUS_BETA,
+			'default' => self::STATE_INACTIVE,
+		] );
```
StyleShit, line 465: maybe it needs to be in the frontend performance module?
not sure about it though

core/files/css/base.php
```diff
@@ -674,7 +675,9 @@ protected function parse_content() {
 
 		Plugin::$instance->breakpoints->set_responsive_control_duplication_mode( $this->get_responsive_control_duplication_mode() );
 
+		Performance::set_use_style_controls( true );
 		$this->render_css();
+		Performance::set_use_style_controls( false );
```
StyleShit, line 680: maybe instead of changing to false, change to the previous value?

includes/managers/controls.php
```diff
@@ -798,6 +801,15 @@ public function add_control_to_stack( Controls_Stack $element, $control_id, $con
...
 		];
 
+		$control_type = 'controls';
+		if (
+			Performance::is_optimized_control_loading_feature_enabled()
+			&& Performance::should_optimize_controls()
+			&& $this->is_style_control( $control_data )
+		) {
+			$control_type = 'style_controls';
+		}
```
StyleShit, line 811: doesn't it break 3rd parties that expect the style controls to be under the `controls` key?

includes/managers/controls.php
StyleShit, line 1: how much performance do we gain from this change?

includes/managers/controls.php
KingYes: @StyleShit What did you do with controls with `selectors` and `prefix_classes` / `frontend_available`?

includes/managers/controls.php
KingYes: It's not part of the experiment.

includes/managers/controls.php
KingYes: 10%-20% depending on the page length.

blackcolt (dismissed):

arielk (approved):

PR #25248, opened 2024-02-22
Fix:  Tabs widget persists on both backend and frontend [ED-12791]

TzviRabinovitch:


includes/managers/widgets.php
```diff
@@ -3,10 +3,9 @@
 
 use Elementor\Core\Common\Modules\Ajax\Module as Ajax;
 use Elementor\Core\Utils\Collection;
-use Elementor\Core\Utils\Exceptions;
 use Elementor\Core\Utils\Force_Locale;
 use Elementor\Modules\NestedAccordion\Widgets\Nested_Accordion;
-use Elementor\Modules\NestedTabs\Widgets\NestedTabs;
+use Elementor\Modules\NestedTabs\Widgets\Nested_Tabs;
```
TzviRabinovitch, line 8: fixing the name of the class before the experiment goes into stable

includes/managers/widgets.php
```diff
@@ -45,7 +44,7 @@ class Widgets_Manager {
 	 * @var Widget_Base[]
 	 */
 	private $_promoted_widgets = [
-		'nested-elements' => NestedTabs::class,
+		'nested-tabs' => Nested_Tabs::class,
```
TzviRabinovitch, line 47: this is the actual fix

PR #25251, opened 2024-02-22
Fix: Tabs widget persists on both backend and frontend [ED-12791]

TzviRabinovitch:


modules/nested-tabs/widgets/nested-tabs.php
```diff
@@ -15,9 +15,6 @@
 use Elementor\Modules\NestedElements\Controls\Control_Nested_Repeater;
 use Elementor\Plugin;
 use Elementor\Repeater;
-use Elementor\Modules\DynamicTags\Module as TagsModule;
```
TzviRabinovitch, line 18: not used

modules/nested-accordion/module.php
```diff
@@ -19,12 +19,6 @@ public function get_name() {
 		return 'nested-accordion';
 	}
 
-	public function get_widgets() {
-		return [
```
TzviRabinovitch, line 23: not needed since we add it manually

includes/managers/widgets.php
```diff
@@ -97,6 +97,8 @@ private function init_widgets() {
 
 		$this->_widget_types = [];
 
+		$this->register_promoted_widgets();
```
TzviRabinovitch, line 100: fixed this one ( was broken )
https://elementor.atlassian.net/browse/ED-8909

includes/managers/widgets.php
```diff
@@ -667,4 +663,23 @@ public function register_ajax_actions( Ajax $ajax_manager ) {
...
+		}
+
+		if ( is_array( $classes ) ) {
+			foreach ( $classes as $class_name ) {
+				$this->register( new $class_name() );
+			}
+		} else {
+			$this->register( new $classes() );
+		}
+	}
```
davseve, line 684: Maybe, we can cast the `$classes` param to be an array then we can avoid the else condition at the bottom of this function
```suggestion
			public function register_promoted_active_widgets( string $experiment_name, array $classes ): void {
		if ( ! Plugin::$instance->experiments->is_feature_active( $experiment_name ) ) {
			return;
		}

		foreach ( $classes as $class_name ) {
			$this->register( new $class_name() );
		}
	}
```

includes/managers/widgets.php
```diff
@@ -43,16 +44,17 @@ class Widgets_Manager {
...
-		'nested-accordion' => Nested_Accordion::class,
+	private array $promoted_widgets = [
+		NestedElementsModule::EXPERIMENT_NAME => [ NestedTabs::class, Nested_Accordion::class ],
 	];
 
 	/**
 	 * Init widgets.
 	 *
-	 * Initialize Elementor widgets manager. Include all the the widgets files
+	 * Initialize Elementor widgets manager. Include all the widgets files
```
davseve, line 57: Not related fix

davseve (approved):

PR #25293, opened 2024-02-27
Tweak: Add units to Shape Divider [ED-13900]

rami-elementor:


includes/elements/section.php
```diff
@@ -985,6 +985,7 @@ protected function register_controls() {
 				[
 					'label' => esc_html__( 'Width', 'elementor' ) . ' (%)',
```
rodolphebertozzo, line 986:
```suggestion
					'label' => esc_html__( 'Width', 'elementor' ),
```

rodolphebertozzo (changes requested):

includes/elements/section.php
rami-elementor: Thank you @rodolphebertozzo

arielk (approved):

PR #25313, opened 2024-02-28
Tweak: Improve counter widget [ED-13915]

### LINKED ISSUE #9068: 💡  Add alignment controls for the Counter widget
## Prerequisites

- [x] I have searched for similar features requests in both open and closed tickets and cannot find a duplicate.
- [x] The feature is still missing in the latest stable version of Elementor ( Elementor Pro. )

**What problem is your feature request going to solve? Please describe.**
 Currently, the counter widget comes with centered content by default (number and title, however, many times it makes sense to have them left or right-aligned, depending the layout.

Screenshot 2019-09-13 at 12 35 35

**Describe the solution you'd like**
I think this alignment flexibility can be added painlessly with simply adding alignment controls in the Content tab.

In this quick mockup, you'll see the two alignment controls added.

Screenshot 2019-09-13 at 12 46 07

✌🏻 You could also fit them under the styles tab.

**Describe alternatives you've considered**
Can be forced with css, no problem, but a native alignment control would work better.

**Additional context**
Thanks for considering!

rami-elementor:


includes/widgets/counter.php
```diff
@@ -223,6 +244,36 @@ protected function register_controls() {
 			]
 		);
 
+		$this->add_responsive_control(
+			'counter-align',
```
rodolphebertozzo, line 248: A control name with a - (dash) character might cause unexpected behavior.

rodolphebertozzo (changes requested):

includes/widgets/counter.php
rami-elementor: Good catch!

rami-elementor: Fixed.

arielk (approved):

PR #25343, opened 2024-03-02
Fix: Icon Box widget renders redundant markup when it has no content [ED-13934]

rami-elementor:


includes/widgets/icon-box.php
```diff
@@ -693,6 +693,11 @@ protected function render() {
 		}
 
 		$has_icon = ! empty( $settings['icon'] );
+		$has_content = ! Utils::is_empty( $settings['title_text'] ) || ! Utils::is_empty( $settings['description_text'] );
```
blackcolt, line 696: can you extract to a function?

rami-elementor: See https://github.com/elementor/elementor/pull/25342

I prefer to use the same code convention in both widgets (image box and icon box).

blackcolt (approved):

PR #25360, opened 2024-03-04
Internal: Improve Nested Elements repeaters performance - Accordion - Sort [ED-13825]

davseve:


modules/nested-accordion/assets/js/frontend/handlers/nested-accordion.js
```diff
@@ -62,23 +62,41 @@ export default class NestedAccordion extends Base {
...
 			view = container.view.$el,
 			id = container.model.get( 'id' ),
 			currentId = this.$element.data( 'id' );
 
 		if ( id === currentId ) {
-			const containers = view.find( this.getSettings( 'selectors.accordionContentContainers' ) ),
-				accordionItems = view.find( this.getSettings( 'selectors.accordionItems' ) ),
-				lastContentContainer = containers[ containers.length - 1 ],
-				lastAccordionItem = accordionItems[ accordionItems.length - 1 ];
+			const [ accordionItem, contentContainer ] = 'undefined' === typeof targetIndex ? this.insert( view ) : this.move( view, targetIndex );
```
hein-obox, line 71: We probably will have to discuss how this works with our change for the duplication.

louiswol94 (approved):

modules/nested-accordion/assets/js/frontend/handlers/nested-accordion.js
```diff
@@ -67,26 +67,52 @@ export default class NestedAccordion extends Base {
...
-				targetIndex = isSpecificIndex ? index + 1 : accordionItems.length - 1,
-				targetContentContainer = targetContainer !== undefined ? targetContainer.view.$el[ 0 ] : containers[ containers.length - 1 ],
-				targetAccordionItem = accordionItems[ targetIndex ] || accordionItems[ accordionItems.length - 1 ];
-
-			targetAccordionItem.appendChild( targetContentContainer );
+			const accordionItems = view.find( this.getSettings( 'selectors.accordionItems' ) );
+			let accordionItem, contentContainer;
+
+			switch ( type ) {
+				case 'create':
```
hein-obox, line 80: Nice.

davseve: Thanks :)
We can make it even better,
I tried to add the strings as consts, but it caused lighthouse tests to fail so I had to give up on that.

hein-obox (approved):

PR #25368, opened 2024-03-05
Fix: RTL websites using a LTR UI are flipping between 'left' and 'right' (...) [ED-12220]

hein-obox:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #


tests/playwright/pages/wp-admin-page.ts
```diff
@@ -175,10 +175,42 @@ export default class WpAdminPage extends BasePage {
 		await this.page.click( '#submit' );
 	}
 
-	async setLanguage( language: string ) {
```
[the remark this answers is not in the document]
dmaziar: I have added api function
```
export async function setWPLanguage( request: APIRequestContext, language: string ) {
	await _post( request, 'settings', { language } );
}
```
But for some reason it works only locally. If it will work for you, then you can use it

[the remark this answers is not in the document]
hein-obox: @dmaziar I will stick with the old method for now.

tests/playwright/pages/editor-page.ts
```diff
@@ -509,13 +509,13 @@ export default class EditorPage extends BasePage {
...
 		await this.page.locator( 'button#elementor-panel-saver-button-publish' ).click();
 		await this.page.waitForLoadState();
-		await this.page.getByRole( 'button', { name: 'Update' } ).waitFor();
+		await this.page.locator( '#elementor-panel-saver-button-publish-label' ).waitFor();
 	}
 
 	async publishAndViewPage() {
 		await this.publishPage();
 		await this.page.locator( '#elementor-panel-header-menu-button i' ).click();
-		await this.page.getByRole( 'link', { name: 'View Page' } ).click();
```
hein-obox, line 518: What is the best way how to deal with text selector when they can be translated into another language (Hebrew in this case)?

hein-obox: @dmaziar For the moment I don't strictly need the translations. I have selected Hebrew as a language, so that I can test the RTL view, but the Hebrew translation isn't installed on our WordPress setup, so the text will still be displayed in English. In this way, we can test RTL, without having the language problem.

Once I download the translations from WordPress, then everything starts to break. I have removed that step for now.

I hope I am making sense.

But in general, I would like to know what would be the best approach here. It might be relevant in the future.

dmaziar: I think nothing we can do here, except using ids and classes. I'm pretty sure Playwright is not the right tool for localisation testing. We will have to think about the best solution.

tests/playwright/pages/wp-admin-page.ts
```diff
@@ -175,9 +175,28 @@ export default class WpAdminPage extends BasePage {
...
+		if ( ! await this.page.locator( 'html[lang=' + languageCheck + ']' ).isVisible() ) {
+			await this.page.selectOption( '#WPLANG', language );
+			await this.page.locator( '#submit' ).click();
+		}
+
+		// Set user profile language
+		const userProfileLanguage = null !== userLanguage ? userLanguage : language;
+
+		await this.page.goto( 'wp-admin/profile.php' );
+		await this.page.selectOption( '[name="locale"]', userProfileLanguage );
```
louiswol94, line 199: Maybe extract? `setUserProfileLanguage()`?

louiswol94 (dismissed):

rami-elementor (approved):

PR #25439, opened 2024-03-14
Internal: Un-attach Nested Accordion from Atomic Repeaters experiment [ED-14063]

davseve:


louiswol94:
@davseve I am thinking, do we really need to check the Experiment in the JS? If the JS config is not there from the PHP, it won't use it. And the JS config won't be added in the PHP if the Experiment is disabled.

louiswol94 (approved):

TzviRabinovitch (approved):

PR #25459, opened 2024-03-18
Internal: Nested infra issues when duplication and sorting items [ED-14084]

davseve:


modules/nested-accordion/assets/js/frontend/handlers/nested-accordion.js
```diff
@@ -120,10 +121,8 @@ export default class NestedAccordion extends Base {
 		} );
 	}
 
-	updateListeners( view ) {
-		this.elements.$accordionTitles = view.find( this.getSettings( 'selectors.accordionItemTitles' ) );
-		this.elements.$accordionItems = view.find( this.getSettings( 'selectors.accordionItems' ) );
-		this.elements.$accordionTitles.on( 'click', this.clickListener.bind( this ) );
+	updateListeners() {
+		elementorFrontend.elementsHandlxer.runReadyTrigger( this.$element[ 0 ] );
```
louiswol94, line 125: @ArmanPurtoian Had an issue with this, but if it works now it's okay. Just wanted you to be very sure :)

louiswol94 (approved):

modules/nested-accordion/assets/js/frontend/handlers/nested-accordion.js
davseve: Do you know which issues he had?
Actually, this solves some wired behaviors.

murzin-nikita (approved):

PR #25477, opened 2024-03-20
Internal: Nested Accordion event propagation when clicking a child nested element [ED-14111]

davseve:


louiswol94 (approved):

hein-obox (approved):

modules/nested-accordion/assets/js/frontend/handlers/nested-accordion.js
```diff
@@ -17,7 +17,8 @@ export default class NestedAccordion extends Base {
 				accordionItemTitles: '.e-n-accordion-item-title',
 				accordionItemTitlesText: '.e-n-accordion-item-title-text',
 				accordionContent: '.e-n-accordion-item > .e-con',
-				accordionWrapper: '.e-n-accordion-item',
+				directAccordionItems: '& > .e-n-accordion-item',
```
rloes, line 20: **This breaks all Accordion Widgets** please fix quickly.

I don't think `& >`  is valid css. `&` only works in sass. For some reason it works fine in most browsers. But it doesnt work in Safari on Mac and iOS. It throws the error `SynatxError: The string did not match the expected pattern`.

I'll try to open a ticket and issue in the next days. I just stumbled accross this and had to find the error. Hope this way you maybe react more quickly.  @davseve @hein-obox @nicholaszein

rloes: One idea would be to use :scope, instead of &

davseve: @rloes 
Thanks for your comment.
We will fix it ASAP.

PR #25489, opened 2024-03-21
Internal: Nested infra - provide repeater remove action the ability to update indexes [ED-14118]

davseve:


tests/playwright/sanity/modules/nested-accordion/helper.ts
```diff
@@ -124,7 +124,7 @@ export async function deleteItemFromRepeater( editor: EditorPage, accordionID: s
 		numberOfContents = await nestedAccordionItemContent.count();
 
 	// Act
-	await deleteItemButton.last().click();
+	await deleteItemButton.nth( 1 ).click();
```
davseve, line 127: I want tam deleting the 2nd item of three, and that I check ithe remaining items id's are updated.

murzin-nikita (approved):

louiswol94 (approved):

PR #25490, opened 2024-03-21
Internal: Home Screen - Create 'External Links' section [ED14010]

elchugreeva:


modules/home/assets/js/components/external-links-section.js
```diff
@@ -0,0 +1,35 @@
...
+
+const ExternalLinksSection = ( { ...props } ) => {
+	return (
+		<Paper elevation={ 0 } sx={ { px: 3 } }>
+			<List>
+				{
+					props.externalLinksData.map( ( item, index ) => {
+						return (
+							<>
+								<ListItemButton key={ index } href={ item.url } target="_blank" sx={ { '&:hover': { background: 'none' }, gap: 2, px: 0, py: 2 } }>
```
hein-obox, line 15:
```suggestion
								<ListItemButton key={ index } href={ item.url } target="_blank" sx={ { '&:hover': { backgroundColor: 'initial' }, gap: 2, px: 0, py: 2 } }>
```

hein-obox (dismissed):

modules/home/assets/js/components/external-links-section.js
```diff
@@ -0,0 +1,35 @@
...
+
+const ExternalLinksSection = ( { ...props } ) => {
+	return (
+		<Paper elevation={ 0 } sx={ { px: 3 } }>
+			<List>
+				{
+					props.externalLinksData.map( ( item, index ) => {
+						return (
+							<>
+								<ListItemButton key={ index } href={ item.url } target="_blank" sx={ { '&:hover': { backgroundColor: 'initial' }, gap: 2, px: 0, py: 2 } }>
```
ronenelementor, line 15: If I understand correctly, it's recommended to avoid using array index as key, but something constant, like item.label (assuming they're all unique).
You can read about this [here](https://stackoverflow.com/questions/59517962/react-using-index-as-key-for-items-in-the-list)

modules/home/assets/js/components/external-links-section.js
```diff
@@ -0,0 +1,35 @@
...
+									<Box component="img" src={ item.image } sx={ { width: '38px' } }></Box>
+									<ListItemText sx={ { color: 'text.secondary' } } primary={ item.label } />
+								</ListItemButton>
+								{
+									index < props.externalLinksData.length - 1 && <Divider />
```
ronenelementor, line 20:
```suggestion
									( index < props.externalLinksData.length - 1 ) && <Divider/>
```

modules/home/assets/js/components/external-links-section.js
elchugreeva: you're right, in this case it is not so critical, since we are not going to rerender this component. But fixed anyway.

modules/home/assets/js/components/external-links-section.js
elchugreeva: fixed

ronenboxer (dismissed):

hein-obox (approved):

PR #25509, opened 2024-03-24
Internal: Home Screen - Create 'Add-ons' section [ED-13986]

elchugreeva:


modules/home/assets/js/components/addons-section.js
```diff
@@ -0,0 +1,50 @@
...
+import Button from '@elementor/ui/Button';
+import Card from '@elementor/ui/Card';
+import CardActions from '@elementor/ui/CardActions';
+import CardContent from '@elementor/ui/CardContent';
+import CardMedia from '@elementor/ui/CardMedia';
+
+const Addons = ( { ...props } ) => {
+	const addonsArray = props.addonsData.repeater.filter( ( item ) => false === item.is_installed );
+	const cardsPerRow = 3 === addonsArray.length ? 3 : 2;
+	return (
```
hein-obox, line 13: Empty line.

modules/home/assets/js/components/addons-section.js
```diff
@@ -0,0 +1,51 @@
...
+import List from '@elementor/ui/List';
+import Link from '@elementor/ui/Link';
+import Button from '@elementor/ui/Button';
+import Card from '@elementor/ui/Card';
+import CardActions from '@elementor/ui/CardActions';
+import CardContent from '@elementor/ui/CardContent';
+import CardMedia from '@elementor/ui/CardMedia';
+
+const Addons = ( { ...props } ) => {
+	const addonsArray = props.addonsData.repeater.filter( ( item ) => false === item.is_installed );
```
hein-obox, line 11:
```suggestion
	const activeAddons = props.addonsData.repeater.filter( ( item ) => false === item.is_installed );
```

modules/home/assets/js/components/addons-section.js
```diff
@@ -0,0 +1,51 @@
...
+	const cardsPerRow = 3 === addonsArray.length ? 3 : 2;
```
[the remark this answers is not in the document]
elchugreeva: but that is the case only for 3, if it is more, and we start adding rows, we want 2 items in a row, if theres 5 add ons it is 2, 2 and 1

modules/home/assets/js/components/addons-section.js
```diff
@@ -0,0 +1,51 @@
...
+								<CardContent sx={ { display: 'flex', flexDirection: 'column', justifyContent: 'space-between', gap: 3 } }>
+									<Box>
+										<CardMedia image={ item.image } sx={ { height: '58px', width: '58px', mb: 3 } } />
+										<Box>
+											<Typography variant="subtitle2">{ item.title }</Typography>
+											<Typography variant="body2" color="text.secondary">{ item.description }</Typography>
+										</Box>
+									</Box>
+									<CardActions sx={ { p: 0 } }>
+										<Button variant="outlined" size="small" color="promotion" href={ item.url } target="_blank" sx={ { maxWidth: 'fit-content' } }>{ item.button_label }</Button>
```
hein-obox, line 34:
```suggestion
										<Button variant="outlined" size="small" color="promotion" href={ item.url } target="_blank">{ item.button_label }</Button>
```

modules/home/assets/js/components/addons-section.js
```diff
@@ -0,0 +1,51 @@
...
+									</CardActions>
+								</CardContent>
+							</Card>
+						);
+					} )
+				}
+			</List>
+			<Link variant="body2" color="info.main" underline="none">{ props.addonsData.footer.label }</Link>
```
hein-obox, line 42: I think the `a href` element is missing here.

modules/home/assets/js/components/addons-section.js
hein-obox, line 11: Not every item has the `is_installed` property.

```suggestion
	const addonsArray = props.addonsData.repeater.filter( ( item ) => true !== item?.is_installed );
```

modules/home/assets/js/components/home-screen.js
```diff
@@ -17,9 +18,10 @@ const HomeScreen = ( props ) => {
 					createNewPageUrl={ props.homeScreenData.create_new_page_url }
 				/>
 				<Box sx={ { display: 'flex', flexDirection: { xs: 'column', sm: 'row' }, justifyContent: 'space-between', gap: 3 } }>
-					<Box sx={ { flex: 1 } }>
+					<Container maxWidth="md" disableGutters={ true } sx={ { display: 'flex', flexDirection: 'column', gap: 3 } }>
```
hein-obox, line 21: Why do we have a `<Container>` here?
Can't we add the flex styling inside the `<Box>` component?
Or could we use the `<Stack>` instead?

I don't like seeing containers with a maxWidth if we don't really need that.

ronenboxer (approved):

hein-obox (approved):

PR #25534, opened 2024-03-26
Tweak: Merge "Improved Asset Loading" feature [ED-14124]

rami-elementor:


includes/base/widget-base.php
```diff
@@ -1228,9 +1228,8 @@ private function is_inline_css_mode() {
 		if ( null === $is_active ) {
 			$is_edit_mode = Plugin::$instance->editor->is_edit_mode();
 			$is_preview_mode = Plugin::$instance->preview->is_preview_mode();
-			$is_optimized_mode = Plugin::$instance->experiments->is_feature_active( 'e_optimized_css_loading' );
 
-			$is_active = ( Utils::is_script_debug() || $is_edit_mode || $is_preview_mode || ! $is_optimized_mode ) ? false : true;
+			$is_active = ( Utils::is_script_debug() || $is_edit_mode || $is_preview_mode ) ? false : true;
```
rotemee, line 1232: @rami-elementor - you have removed the "e_optimized_css_loading".

rotemee (changes requested):

includes/base/widget-base.php
rami-elementor: Reverted. Good catch!

rotemee:
@rami-elementor - What will happen if a user only upgrade the core but not the

rotemee (approved):

PR #25545, opened 2024-03-27
Internal: Reusable build workflow [ED-13639]

dmaziar:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #


.github/workflows/lighthouse.yml
```diff
@@ -11,52 +11,7 @@ concurrency:
...
-        run: npx grunt build
-      - name: Save build to cache
-        if: github.event.pull_request.title == null || steps.changelog_diff_files.outputs.diff
-        uses: actions/cache@v4
-        id: restore-build
-        with:
-          path: ./build/*
-          key: ${{ github.sha }}
-
+    uses: ./.github/workflows/build.yml
```
yotamselementor, line 14: Instead of this, can we cache/keep the results of the build job and reuse them?

yotamselementor (changes requested):

.github/workflows/lighthouse.yml
dmaziar: @yotamselementor do you have any suggestions? 
What I've investigated: 
1. We have an option to trigger `workflow_run` event after the build job is done, but this works only on default branch, so we will run playwright not on PR's branch, but on main. 
2. Trigger workflow run via API. This works fine, but then playwright workflow will be detached from a PR, so if it fails, user can still merge and we would need to add a trigger for any new workflow to build job.

yotamselementor (approved):

PR #25592, opened 2024-04-01
Internal: Publish elementor-editor-testing with elementor [ED-13026]

TzviRabinovitch:


louiswol94:
@TzviRabinovitch Do we need to run `composer update` so that the lock file also updates?

louiswol94 (approved):

PR #25730, opened 2024-04-14
New: Implement caching mechanism on render elements [ED-13883]

KingYes:


core/experiments/manager.php
```diff
@@ -479,6 +479,15 @@ private function add_default_features() {
 			'release_status' => self::RELEASE_STATUS_DEV,
 			'default' => self::STATE_INACTIVE,
 		] );
+
+		$this->add_feature( [
+			'name' => 'e_element_cache',
+			'title' => esc_html__( 'Element Cache', 'elementor' ),
+			'tag' => esc_html__( 'Performance', 'elementor' ),
+			'description' => esc_html__( 'Enable Element Cache to improve the performance of your site by caching the rendered content of Elementor widgets.', 'elementor' ),
```
rodolphebertozzo, line 487:
```suggestion
			'description' => esc_html__( 'Enable Element Cache to improve the performance of your site by caching the rendered content of Elementor elements.', 'elementor' ),
```

rodolphebertozzo (changes requested):

bkfino:
Works really nice, performance increase is hugh in my tests

assets/dev/js/editor/editor-base.js
```diff
@@ -285,6 +285,24 @@ export default class EditorBase extends Marionette.Application {
 			if ( ! this.widgetsCache[ widgetType ].commonMerged ) {
 				jQuery.extend( this.widgetsCache[ widgetType ].controls, this.widgetsCache.common.controls );
 
+				this.widgetsCache[ widgetType ].controls = elementor.hooks.applyFilters( 'elements/widget/controls/common', this.widgetsCache[ widgetType ].controls, widgetType, this.widgetsCache[ widgetType ] );
+
+				// TODO: Move this code to own file.
+				if ( this.widgetsCache[ widgetType ].controls?._element_cache ) {
+					let elementCacheDescription = __( 'The default behavior of this widget is:', 'elementor' ) + ' ';
+
+					elementCacheDescription += '<strong>';
```
rami-elementor, line 294:
```suggestion
					let elementCacheDescription = __( 'The default behavior of this widget is:', 'elementor' );

					elementCacheDescription += ' <strong>';
```

rami-elementor (dismissed):

assets/dev/js/editor/editor-base.js
```diff
@@ -286,6 +286,24 @@ export default class EditorBase extends Marionette.Application {
 			if ( ! this.widgetsCache[ widgetType ].commonMerged ) {
 				jQuery.extend( this.widgetsCache[ widgetType ].controls, this.widgetsCache.common.controls );
 
+				this.widgetsCache[ widgetType ].controls = elementor.hooks.applyFilters( 'elements/widget/controls/common', this.widgetsCache[ widgetType ].controls, widgetType, this.widgetsCache[ widgetType ] );
+
+				// TODO: Move this code to own file.
```
hein-obox, line 291: Could you implement this change now already?
We often forget to update such things at a later stage.

core/base/document.php
```diff
@@ -1801,6 +1859,53 @@ protected function print_elements( $elements_data ) {
...
+		$this->delete_meta( static::CACHE_META_KEY );
+	}
+
+	private function should_store_cache_elements() {
+		static $should_store_cache_elements = null;
+
+		if ( null === $should_store_cache_elements ) {
+			$should_store_cache_elements = (
+				! is_admin()
+				// && ! is_user_logged_in()
```
hein-obox, line 1901: Can we clean this up?

core/base/document.php
```diff
@@ -1801,6 +1859,53 @@ protected function print_elements( $elements_data ) {
 		}
 	}
 
+	public function set_document_cache( $value ) {
```
mykytamurzin, line 1862: This file is too big. Can we extract cache manipulation to a separate class or trait?

includes/base/element-base.php
```diff
@@ -523,6 +542,30 @@ public function print_element() {
...
+
+		if ( ! empty( $raw_data['settings']['_element_cache'] ) ) {
+			return 'yes' === $raw_data['settings']['_element_cache'];
+		}
+
+		if ( $this->is_dynamic_content() ) {
+			return true;
+		}
+
+		if ( empty( $raw_data['settings']['__dynamic__'] ) && empty( $raw_data['settings']['e_display_conditions'] ) ) {
```
mykytamurzin, line 562: Is it ok to have such display conditions reference in core?

core/base/document.php
mykytamurzin: maybe it should be within a cache module, I'm not sure

core/base/document.php
```diff
@@ -1783,6 +1788,59 @@ protected function save_settings( $settings ) {
...
+			global $wp_scripts, $wp_styles;
+			$scripts_ignored = $wp_scripts->queue;
+			$styles_ignored = $wp_styles->queue;
+
+			ob_start();
+
+			$this->do_print_elements( $elements_data );
+
+			$scripts_to_queue = array_values( array_diff( $wp_scripts->queue, $scripts_ignored ) );
+			$styles_to_queue = array_values( array_diff( $wp_styles->queue, $styles_ignored ) );
```
louiswol94, line 1811: @KingYes I might be reading this wrong, but it looks like these arrays will always be empty? 🤔

core/base/document.php
```diff
@@ -1783,6 +1788,59 @@ protected function save_settings( $settings ) {
...
 	protected function print_elements( $elements_data ) {
+		if ( ! Plugin::$instance->experiments->is_feature_active( 'e_element_cache' ) ) {
+			$this->do_print_elements( $elements_data );
+
+			return;
+		}
+
+		$cached_data = $this->get_document_cache();
+
+		if ( false === $cached_data ) {
```
louiswol94, line 1799: @KingYes We might be able to extract this a bit?

```
if ( false === $cached_data ) {
    $cached_data = $this->some_descriptive_method_name();
} else {
    $this->enqueue_styles_and_scripts();
}
```

core/base/document.php
```diff
@@ -1783,6 +1788,59 @@ protected function save_settings( $settings ) {
...
+
+			ob_start();
+
+			$this->do_print_elements( $elements_data );
+
+			$scripts_to_queue = array_values( array_diff( $wp_scripts->queue, $scripts_ignored ) );
+			$styles_to_queue = array_values( array_diff( $wp_styles->queue, $styles_ignored ) );
+
+			$cached_data = [
+				'content' => ob_get_clean(),
```
louiswol94, line 1814: Maybe `ob_get_clean()` right after `do_print_elements()`?

core/page-assets/loader.php
```diff
@@ -56,7 +56,11 @@ public function enable_assets( array $assets_data ) {
 
 		foreach ( $assets_data as $assets_type => $assets_list ) {
 			foreach ( $assets_list as $asset_name ) {
-				$this->assets[ $assets_type ][ $asset_name ]['enabled'] = true;
+				if ( 'scripts' === $assets_type ) {
+					wp_enqueue_script( $asset_name );
+				} else {
+					wp_enqueue_style( $asset_name );
+				}
```
louiswol94, line 63: @KingYes Should this be behind the experiment check or not?

includes/base/element-base.php
```diff
@@ -285,6 +285,20 @@ protected function should_print_empty() {
 		return true;
 	}
 
+	/**
+	 * Whether the element returns dynamic content.
+	 *
+	 * sed to determine whether to cache the element output of not.
```
louiswol94, line 291:
```suggestion
	 * set to determine whether to cache the element output or not.
```

includes/base/element-base.php
```diff
@@ -430,6 +444,11 @@ public function add_link_attributes( $element, array $url_control, $overwrite =
 	public function print_element() {
 		$element_type = $this->get_type();
 
+		if ( $this->should_render_shortcode() ) {
+			echo '[elementor-element data="' . esc_attr( base64_encode( wp_json_encode( $this->get_raw_data() ) ) ) . '"]';
```
louiswol94, line 448: Do we need `base64_encode()`?

louiswol94: @KingYes 
`echo do_shortcode('...');`?

includes/base/element-base.php
louiswol94: I agree. The Display conditions module should implement the `elementor/element/should_render_shortcode` filter.

louiswol94 (changes requested):
Amazing!

core/page-assets/loader.php
```diff
@@ -56,7 +56,11 @@ public function enable_assets( array $assets_data ) {
 
 		foreach ( $assets_data as $assets_type => $assets_list ) {
 			foreach ( $assets_list as $asset_name ) {
-				$this->assets[ $assets_type ][ $asset_name ]['enabled'] = true;
+				if ( 'scripts' === $assets_type ) {
+					wp_enqueue_script( $asset_name );
```
rotemee, line 60: @KingYes - Currently pro assets will not be loaded, see the usage of `assets_loader->add_assets` in the pro, it's adding the assets but not registering them:
Here in your implementation, you enqueue only by the handle, so all assets must be first registered, probably what's best is to change the `add_ssets` method to also register the assets that are being added.
+
Also, change the tests accordingly.

rotemee (changes requested):

core/base/document.php
KingYes: I thought about extracting it to the caching module. But we have a lots of codes connected to this class, so I just put it here for now.

core/base/document.php
KingYes: Yes!

core/page-assets/loader.php
KingYes: This is fixed about the old code.

includes/base/element-base.php
KingYes: Yes! If not, we can get a problem with the parsing args in the shortcode mechanism.

assets/dev/js/editor/editor-base.js
```diff
@@ -286,6 +286,24 @@ export default class EditorBase extends Marionette.Application {
...
+					elementCacheDescription += ' <strong>';
+					if ( this.widgetsCache[ widgetType ]?.is_dynamic_content ) {
+						elementCacheDescription += __( 'Dynamic', 'elementor' );
+					} else {
+						elementCacheDescription += __( 'Static', 'elementor' );
+					}
+					elementCacheDescription += '</strong>';
+
+					this.widgetsCache[ widgetType ].controls._element_cache.description = elementCacheDescription;
+				}
```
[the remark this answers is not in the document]
KingYes: 1. In the widgets we add this once in `common`, so in the description we want to show the widget type. In the editor, it's running once per type of widget.
2. If the experiment is off the control will not be there in the first place.

core/base/document.php
KingYes: After reading the code again, I will not change this. The `ob_get_clean()` runs exactly in the building of the array. I don't want to create another variable in the method scope.

arielk (dismissed):

core/page-assets/loader.php
```diff
@@ -75,6 +81,9 @@ public function add_assets( array $assets ) {
 		$this->assets = array_replace_recursive( $this->assets, $assets );
 	}
 
+	/**
+	 * @deprecated 3.22.0
+	 */
```
rotemee, line 86: @KingYes - Not sure I understand the latest changes related to the loader, I see that you've restored the call to enqueue_assets(), but why not go with the direction that you need?:
- enqueue_assets - will enqueue immediately (and not deprecated).
- enable_assets - will point to the enqueue_assets() and will be marked as deprecated.
- add_assets - will also register the assets.
In other words to get rid of the "pre-enabled" capability if it's not needed?

rotemee (dismissed):

includes/base/element-base.php
louiswol94, line 448: @KingYes Can you please check if you need to use `do_shortcode()` here? I see when the content is fetched via the REST API it renders `[elementor-element data="..."]`, instead of the actual HTML.

core/base/document.php
louiswol94: @KingYes As discussed on Slack, this seems to be causing the PlayWright failure:

This is causing the PlayWright failure:
```
Fatal error: Uncaught TypeError: array_diff(): Argument #1 ($array) must be of type array, null given in /var/www/html/wp-content/plugins/build/core/base/document.php:1812
230
Stack trace:
231
#0 /var/www/html/wp-content/plugins/build/core/base/document.php(1812): array_diff(NULL, NULL)
232
#1 /var/www/html/wp-content/plugins/build/core/base/document.php(1195): Elementor\Core\Base\Document->print_elements(Array)
233
#2 /var/www/html/wp-content/plugins/build/includes/frontend.php(1157): Elementor\Core\Base\Document->print_elements_with_wrapper(Array)
```

.wp-env.json
```diff
@@ -15,6 +15,8 @@
 	"config": {
 		"ELEMENTOR_SHOW_HIDDEN_EXPERIMENTS": true,
 		"SCRIPT_DEBUG": false,
-		"WP_DEBUG": false
+		"WP_DEBUG": true,
+		"WP_DEBUG_LOG": true,
+		"WP_DEBUG_DISPLAY": true
```
louiswol94, line 20: @KingYes You can revert this change if you want. It was only to debug the PlayWright issue.

.github/workflows/playwright.yml
```diff
@@ -50,7 +50,7 @@ jobs:
       matrix:
         testSuite: [
           'ai',
-          'onBoarding',
+          #'onBoarding',
```
louiswol94, line 53:
```suggestion
          'onBoarding',
```
Can we bring this back?

louiswol94 (approved):

arielk (approved):

PR #27326, opened 2024-05-14
Internal: Conversion Center - Handle empty pages and fix Library close

nicoladj77:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

* Make sure that if the document is empty the library opens. 

## Description

* Make sure that if the document is empty the library opens. 

## Test instructions

* Make sure that if the document is empty the library opens. 

## Quality assurance

Fixes #

### DIFF
--- a/assets/dev/scss/global/global.scss
+++ b/assets/dev/scss/global/global.scss
@@ -10,6 +10,10 @@
 	display: none;
 }
 
+.elementor-visibility-hidden {
+	visibility: hidden;
+}
+
 .elementor-screen-only,
 .screen-reader-text, // wp - common.css
 .screen-reader-text span, // wp - common.css
--- a/core/common/assets/js/views/modal/header.js
+++ b/core/common/assets/js/views/modal/header.js
@@ -35,5 +35,9 @@ export default class extends Marionette.LayoutView {
 
 	onCloseModalClick() {
 		this._parent._parent._parent.hideModal();
+
+		if ( elementor.config?.admin_conversion_center_url && 'links-page' === elementor.config.document?.type ) {
+			window.location.href = elementor.config.admin_conversion_center_url;
+		}
 	}
 }
--- a/includes/editor-templates/library-layout.php
+++ b/includes/editor-templates/library-layout.php
@@ -15,7 +15,7 @@ if ( ! defined( 'ABSPATH' ) ) {
 				<i class="eicon-close"
 				   aria-hidden="true"
 				   title="{{{ $e.components.get( 'document/elements' ).utils.getTitleForLibraryClose() }}}"></i>
-				<span class="elementor-screen-only"><?php echo esc_html__( 'Close', 'elementor' ); ?></span>
+				<span class="elementor-screen-only">{{{ $e.components.get( 'document/elements' ).utils.getTitleForLibraryClose() }}}</span>
 			</div>
 		<# } #>
 		<div id="elementor-template-library-header-tools"></div>
--- a/includes/editor-templates/library-layout.php
+++ b/includes/editor-templates/library-layout.php
@@ -8,7 +8,7 @@ if ( ! defined( 'ABSPATH' ) ) {
 	<div class="elementor-templates-modal__header__menu-area"></div>
 	<div class="elementor-templates-modal__header__items-area">
 		<# if ( closeType ) { #>
-			<div data-tooltip="Back to dashboard" class="elementor-templates-modal__header__close elementor-templates-modal__header__close--{{{ closeType }}} elementor-templates-modal__header__item">
+			<div class="elementor-templates-modal__header__close elementor-templates-modal__header__close--{{{ closeType }}} elementor-templates-modal__header__item">
 				<# if ( 'skip' === closeType ) { #>
 				<span><?php echo esc_html__( 'Skip', 'elementor' ); ?></span>
 				<# } #>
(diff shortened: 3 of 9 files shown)


core/common/assets/js/views/modal/header.js
```diff
@@ -35,5 +35,9 @@ export default class extends Marionette.LayoutView {
 
 	onCloseModalClick() {
 		this._parent._parent._parent.hideModal();
+
+		if ( elementor.config?.admin_conversion_center_url && 'links-page' === elementor.config.document?.type ) {
```
nicoladj77, line 39: @louiswol94 not the cleanest approach, let me what you think

nicoladj77: @louiswol94 no there is no command that I can hook into. How do I override a template?

louiswol94: @nicoladj77 In general, I see we can open the Library with a command `library/open`, so I am not sure why we don't close it with a command as well. Then we could have used a UI hook for this. But for now, maybe you can listen for this CustomEvent `elementor/routes/close` that is dispatched and perform the redirect?

louiswol94: @nicoladj77 So maybe something like this? (until we can create a proper close command).

Screenshot 2024-05-15 at 3 05 14 PM

includes/editor-templates/library-layout.php
```diff
@@ -12,7 +12,9 @@
 				<# if ( 'skip' === closeType ) { #>
 				<span><?php echo esc_html__( 'Skip', 'elementor' ); ?></span>
 				<# } #>
-				<i class="eicon-close" aria-hidden="true"></i>
+				<i class="eicon-close"
+				   aria-hidden="true"
+				   title="{{{ $e.components.get( 'document/elements' ).utils.getTitleForLibraryClose() }}}"></i>
 				<span class="elementor-screen-only"><?php echo esc_html__( 'Close', 'elementor' ); ?></span>
```
louiswol94, line 18: `<span class="elementor-screen-only"><?php echo esc_html__( 'Close', 'elementor' ); ?></span>`

Should this not also use the config?

includes/editor-templates/panel.php
```diff
@@ -70,13 +70,12 @@
 		<i class="elementor-icon eicon-menu-bar tooltip-target" aria-hidden="true" data-tooltip="<?php esc_attr_e( 'Menu', 'elementor' ); ?>"></i>
 		<span class="elementor-screen-only"><?php echo esc_html__( 'Menu', 'elementor' ); ?></span>
 	</button>
-	<# if ( $e.components.get( 'document/elements' ).utils.allowAddingWidgets() ) { #>
 	<h2 id="elementor-panel-header-title"></h2>
-	<button id="elementor-panel-header-add-button" class="elementor-header-button">
+	<# const extraClass = $e.components.get( 'document/elements' ).utils.allowAddingWidgets() ? '' : 'elementor-visibility-hidden'; #>
```
louiswol94, line 74: Why are you now just hiding it instead of not rendering it?

modules/conversion-center/module.php
```diff
@@ -96,6 +96,12 @@ public function __construct() {
 			return $this->admin_localize_settings( $settings );
 		} );
 
+		add_action( 'elementor/editor/localize_settings', function ( $data ) {
+			$menu_args = $this->get_menu_args();
+			$data['admin_conversion_center_url'] = admin_url( $menu_args['menu_slug'] );
+			return $data;
+		} );
```
louiswol94, line 103:
```suggestion
		add_action( 'elementor/editor/localize_settings', function ( $data ) {
			return return $this->editor_localize_settings( $data );
		} );
```
Extract

assets/dev/scss/global/global.scss
```diff
@@ -10,6 +10,10 @@
 	display: none;
 }
 
+.elementor-visibility-hidden {
+	visibility: hidden;
```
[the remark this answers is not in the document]
nicoladj77: @louiswol94 because visibility hidden keeps the layout, while display none breakes it. I also tweaked slightly the template because I need to display the name of the edited component.

includes/editor-templates/panel.php
nicoladj77: @louiswol94 as above, I was hiding one thing that I didn't want to hide, the `<h2>` . But to keep the h2 without breaking the layout I need to use the visibility thing, because a visibility:hidden element interacts with flex differently than display:none element

core/common/assets/js/views/modal/header.js
nicoladj77: @louiswol94 are you sure that this event is triggered?this hooks into the router but from what I see here we are just calling

	hideModal() {
		this.getModal().hide();
	}

which is in `elementor/core/common/assets/js/views/modal/layout.js`
but that to me looks like a jQuery method. 
I might be wrong but the above doesn't work and nothing is triggered from my tests.

louiswol94: Hi @nicoladj77 , it was working in my above screenshot with the Links Pages. I could not test with the Conversion Center, but I assumed it's the same.

nicoladj77: @louiswol94 so you are right, I don't know what I was doing wrong. The issue though is that the code you suggest is triggered every time the modal is closed, including when you insert something from the library. And given it's a custom event, I could not even try to use the target to "target" it somehow.

louiswol94 (dismissed):

modules/conversion-center/module.php
```diff
@@ -331,4 +335,11 @@ private function get_menu_args(): array {
 		];
 	}
 
+	public function editor_localize_settings( $data ) {
```
louiswol94, line 338:
```suggestion
	private function editor_localize_settings( $data ) {
```

louiswol94 (approved):

PR #27376, opened 2024-05-19
Internal: Undo action breaking Accordion Nested Element on the editor [ED-14309]

baghdasarovelementor:
## Description
Internal: Undo action breaking Accordion Nested Element on the editor 

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

*

## Quality assurance

### DIFF
--- a/assets/dev/js/editor/document/repeater/commands/insert.js
+++ b/assets/dev/js/editor/document/repeater/commands/insert.js
@@ -14,12 +14,14 @@ export class Insert extends $e.modules.editor.document.CommandHistoryBase {
 				model: data.model,
 				name: data.name,
 				options: { at: data.index },
+				isRestored: true,
 			} );
 		} else {
 			$e.run( 'document/repeater/remove', {
 				containers,
 				name: data.name,
 				index: data.index,
+				isRestored: true,
 			} );
 		}
 	}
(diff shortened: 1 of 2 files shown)


louiswol94 (approved):
So this means that on Undo, we should just re-render the whole widget again?

davseve (approved):

assets/dev/js/editor/document/repeater/commands/insert.js
```diff
@@ -14,12 +14,14 @@ export class Insert extends $e.modules.editor.document.CommandHistoryBase {
 				model: data.model,
 				name: data.name,
 				options: { at: data.index },
+				isRestored: true,
 			} );
 		} else {
 			$e.run( 'document/repeater/remove', {
```
davseve, line 20: This is the undo?

baghdasarovelementor: Yes

baghdasarovelementor:
@louiswol94 Yes

baghdasarovelementor:
> So this means that on Undo, we should just re-render the whole widget again?

To fix this issue without re-render of the whole widget we need much more time and resources, so I implemented it like this for now, we can make it work better later

PR #27377, opened 2024-05-19
Internal: Repeater - Implement min_items functionality

nicoladj77:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

* Add the ability to set a min number of items for a repeater

## Description

* Add the ability to set a min number of items for a repeater

## Test instructions

*

## Quality assurance

Fixes #


.gitignore
```diff
@@ -39,3 +39,6 @@ tests/playwright/blob-report
 **/allure-results
 playwright-test-results
 **/.env
+
+# phpunit
+.phpunit.result.cache
```
louiswol94, line 44: Please revert this change

assets/dev/js/editor/controls/repeater.js
```diff
@@ -113,11 +113,14 @@ ControlRepeaterItemView = ControlBaseDataView.extend( {
 	},
 
 	toggleMinRowsClass() {
-		if ( ! this.model.get( 'prevent_empty' ) ) {
-			return;
+		let minItems = 0;
+		if ( this.model.get( 'min_items' ) && Number.isInteger( this.model.get( 'min_items' ) ) ) {
+			minItems = this.model.get( 'min_items' );
+		} else if ( this.model.get( 'prevent_empty' ) ) {
+			minItems = 1;
```
louiswol94, line 120: Extract to `getMinItems()`

includes/controls/repeater.php
```diff
@@ -181,4 +182,20 @@ public function content_template() {
...
+				count( $control_data['default'] ) < $control_data['min_items']
+			) {
+				throw new \Exception(
+					__( 'In a Repeater control, if you specify a minimum number of items, you must also specify a default value that contains at least that number of items.', 'elementor' )
+				);
+			}
+		}
+
+		return true;
+	}
```
louiswol94, line 200: @nicoladj77 I am wondering if we should do this validation client side?

See: `elementor/assets/dev/js/editor/components/validator/number.js` as an example. This might not work specific for our scenario, but maybe just follow the same logic.

I am just wondering if it is dangerous to throw this kind of Exception that would totally break the Editor if a 3rd party does something wrong.

louiswol94 (changes requested):

includes/controls/repeater.php
nicoladj77: @louiswol94 this exception should only be thrown in the context of the Controls_Manager

		if ( $control_type_instance instanceof Has_Validation ) {
			try {
				$control_type_instance->validate( $control_data );
			} catch ( \Exception $e ) {
				_doing_it_wrong( sprintf( '%1$s::%2$s', __CLASS__, __FUNCTION__ ), esc_html( $e->getMessage() ), '3.23.0' );
				return false;
			}
		}

I totally understand your concern, I can of course to this differently, returning an array from validate() with two keys, valid and message. I wanted to run this by you because like this is cleaner from a coding perspective, but if you think it's not safe, let's do it the other way.

nicoladj77: @louiswol94 regarding FE vs BE, this to me looks similar to

		$control_type_instance = $this->get_control( $control_data['type'] );

		if ( ! $control_type_instance ) {
			_doing_it_wrong( sprintf( '%1$s::%2$s', __CLASS__, __FUNCTION__ ), sprintf( 'Control type "%s" not found.', esc_html( $control_data['type'] ) ), '1.0.0' );
			return false;
		}

which means that if a control is in a not valid state, it should not be added. But let me know if you think differently. if I had to do the check in the FE, I would add rows until they are reach the min_items value.

louiswol94 (approved):

PR #27525, opened 2024-05-28
Internal: Improve Nested Elements Dynamic Tags - Title [ED-13824]

davseve:


assets/dev/js/editor/elements/views/base.js
```diff
@@ -341,7 +341,23 @@ BaseElementView = BaseContainer.extend( {
 	},
 
 	attachElContent( html ) {
-		this.$el.empty().append( this.getHandlesOverlay(), html );
+		if ( elementor.previewView.isBuffering || ! elementorCommon.config.experimentalFeatures.e_nested_atomic_repeaters || 'nested-accordion' !== this?.model?.config?.name ) {
```
TzviRabinovitch, line 344: doing an action and returning is not the best practice 
maybe extract the actions to separate functions

 like :

if ( ShouldDoSomethig() ) {
DoSomething()
}else{
DoSomethingElse()
}

assets/dev/js/editor/elements/views/base.js
```diff
@@ -645,6 +652,13 @@ BaseElementView = BaseContainer.extend( {
 				}
 			} );
 
+
+			if ( isAtomicRepeater ) {
+				dynamicValue = this.getDynamicValue( settings );
+
+				this.renderDynamicValue( dynamicValue );
```
davseve, line 659: I am not sure that this is the best place for this function, 
Maybe it should be inside one of the render functions (renderHTML or render UI)

assets/dev/js/editor/elements/views/base.js
```diff
@@ -659,6 +673,35 @@ BaseElementView = BaseContainer.extend( {
 		this.renderHTML();
 	},
 
+	isAtomicDynamic( settingKey ) {
+		const widgetConfig = elementor.widgetsCache[ this.model.config.name ]
+		return settingKey.includes( '__dynamic__' ) &&
+			elementorCommon.config.experimentalFeatures.e_nested_atomic_repeaters &&
+			widgetConfig.supports_atomic_dynamic_titles; // May be redundant
```
davseve, line 680: This condition may be redundant.
We need to provide one solution for titles and links as well.
I have created a separate task for the MM dynamic links.

assets/dev/js/editor/elements/views/base.js
```diff
@@ -659,6 +673,35 @@ BaseElementView = BaseContainer.extend( {
...
+	},
+
+	getDynamicValue( settings ) {
+		const valueToParse = settings.attributes.__dynamic__.item_title,
+			dynamicSettings = { active: true };
```
davseve, line 685: I think that `dynamicSettings ` should be retrieved from somewhere. 
I couldn't find where it is coming from.

hein-obox (dismissed):

murzin-nikita (dismissed):

assets/dev/js/editor/elements/views/base.js
```diff
@@ -659,6 +659,31 @@ BaseElementView = BaseContainer.extend( {
...
 
+	isAtomicDynamic( dataBinding ) {
+		return !! ( dataBinding.el.hasAttribute( 'data-binding-dynamic' ) && elementorCommon.config.experimentalFeatures.e_nested_atomic_repeaters );
+	},
+
+	isTitle( bindingSetting ) {
+		const titleSettings = [ 'item_title', 'tab_title' ];
+
+		return titleSettings.includes( bindingSetting );
+	},
```
louiswol94, line 670: @davseve Why do we need this? It's not generic enough to use specific keys like `'item_title', 'tab_title'`, so maybe if I understand the reason we can think of something.

assets/dev/js/editor/elements/views/base.js
```diff
@@ -659,6 +659,31 @@ BaseElementView = BaseContainer.extend( {
...
+			try {
+				return elementor.dynamicTags.parseTagsText( valueToParse, dynamicSettings, elementor.dynamicTags.getTagDataContent );
+			} catch {
+				return false;
+			}
+		}
+
+		return settings.attributes[ bindingSetting ];
+	},
+
```
louiswol94, line 686: Can you please update the comment here:

https://github.com/elementor/elementor/blob/main/assets/dev/js/editor/elements/views/base.js#L681

```
	 * Current Limitation:
	 * Not working with dynamics, will required full re-render.
	 * UPDATE: Support for dynamics has experimentally been added in v3.23
```

louiswol94 (changes requested):

assets/dev/js/editor/elements/views/base.js
davseve: Good catch on this! I noticed a small mistake here:

Yes, we need this check. There are other controls that affect the item title. During manual testing, I found out that the mega menu dropdown is not working. The reason for that is our new code was treating it as dynamic.

When we toggle the mega menu dropdown, the settings.changed object contains a key named `item_dropdown_content`. We need to verify that the changed control is a title before treating it as dynamic.

in ` let change = settings.changed[bindingSetting];`
For example, while changing a title, `change` is undefined, so we treat it as dynamic. When we toggle the dropdown control, `change`  is also undefined and was being treated as dynamic, which was breaking the functionality. This is why I added the isTitle check.

assets/dev/js/editor/elements/views/base.js
davseve: Fixed.
Thanks :)

assets/dev/js/editor/elements/views/base.js
```diff
@@ -659,6 +659,31 @@ BaseElementView = BaseContainer.extend( {
 		this.renderHTML();
 	},
 
+	isAtomicDynamic( dataBinding ) {
+		return !! ( dataBinding.el.hasAttribute( 'data-binding-dynamic' ) && elementorCommon.config.experimentalFeatures.e_nested_atomic_repeaters );
+	},
+
+	isTitle( changedControl ) {
+		const titleSettings = [ 'item_title', 'tab_title' ];
```
hein-obox, line 667: A clarification question: are we bound to these titles for all nested widgets? 
Is there any way to make these parameters more dynamic?

davseve: We are not, Nested Tabs title is `tab_title`
And Nested Accordion and Mega Menu titles are `item_title`

We can pass them from the config, but it is more or less the same. 

WDYT?

hein-obox (dismissed):

assets/dev/js/editor/elements/views/base.js
davseve: Done

hein-obox (approved):

louiswol94 (approved):

PR #27542, opened 2024-05-29
Internal: Support PHP 7.4  in heading sanitization  [ED-14960] (#27537)

TzviRabinovitch:


murzin-nikita (approved):

mykytamurzin:
Hello @rloes , thanks for the tip. I think it's expected behavior. But we'll discuss it more next week.

rloes:
Hello @mykytamurzin,

Thanks for the reply. Please consider the following scenario in your discussions next week. 

I have designed a site for a client, where I used <span> elements inside the heading widget in order to highlight certain words in a different color. 

My client has an editor user account in order to only edit content and not destroy style and structure. I used this setup often without problem.

Now since this PR got merged. When he edits the site, the editor can still see the highlights and the classes inside the heading widget. However, on save they just disappear on the frontend without notice. Only after reloading the editor, one can see that something changed.

I took quite a time until I found the issue describing the source of the problem. I kept wondering why my classes keep disappearing.

I understand the necessity of content sanitation. However, this was a breaking change introduced quite silently and could break many sites. 

I hope you consider my thoughts.

I am looking forward to a good solution. 

Have a great week.

PR #27553, opened 2024-05-30
Fix: Links to Elementor settings are broken [ED-14577]

rami-elementor:
Elementor should not use hardcoded links to settings pages as the URL is based on the experiment.

Therefor, all the links should use generic function that returns the correct URL for the specific settings page tab.

### DIFF
--- a/includes/settings/settings-page.php
+++ b/includes/settings/settings-page.php
@@ -1,6 +1,8 @@
 <?php
 namespace Elementor;
 
+use Elementor\Plugin;
+
 if ( ! defined( 'ABSPATH' ) ) {
 	exit; // Exit if accessed directly.
 }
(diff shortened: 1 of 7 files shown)


includes/settings/settings-page.php
```diff
@@ -69,6 +71,27 @@ final public static function get_url() {
...
+	 * @return string Settings tab URL.
+	 */
+	final public static function get_settings_tab_url( $tab_id ): string {
+		$settings_page_id = Plugin::$instance->experiments->is_feature_active( 'home_screen' )
+			? 'elementor-settings'
+			: 'elementor';
+
+		return admin_url( "admin.php?page=$settings_page_id#tab-$tab_id" );
+	}
+
```
rami-elementor, line 94: The logic is in this new method.

arielk (approved):

PR #27637, opened 2024-06-08
Internal: Conversion Center - display conditions

nicoladj77:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

* Add support for display conditions.

## Description

*

## Test instructions

*

## Quality assurance

Fixes #


modules/conversion-center/module.php
```diff
@@ -45,6 +47,13 @@ public static function is_active(): bool {
 		return Plugin::$instance->experiments->is_feature_active( static::EXPERIMENT_NAME );
 	}
 
+	public static function is_pro_active(): bool {
```
nicoladj77, line 50: @louiswol94 I am sure there is a bebtter way but couldn't find it.

modules/conversion-center/assets/js/editor/hooks/ui/editor/document/select.js
```diff
@@ -0,0 +1,34 @@
...
+
+		return 'contact-buttons' === type;
+	}
+
+	apply( args ) {
+		const { container: { type = '' } } = args;
+		switch ( type ) {
+			case 'section':
+			case 'container':
+				$e.run( 'document/elements/deselect', args );
```
louiswol94, line 26: @nicoladj77 Can you rather just avoid the origin selection instead of selecting and de-selecting?

modules/conversion-center/module.php
```diff
@@ -8,6 +8,7 @@
 use Elementor\Core\Base\Module as BaseModule;
 use Elementor\Core\Documents_Manager;
 use Elementor\Core\Experiments\Manager;
+use Elementor\Core\Wp_Api;
```
louiswol94, line 11: Why do we need `Wp_Api`?

modules/conversion-center/module.php
louiswol94: You can use `Utils::has_pro()`

modules/conversion-center/module.php
```diff
@@ -96,6 +105,50 @@ public function __construct() {
 		add_action( 'wp_ajax_elementor_send_clicks', [ $this, 'handle_click_tracking' ] );
 		add_action( 'wp_ajax_nopriv_elementor_send_clicks', [ $this, 'handle_click_tracking' ] );
 
+		if ( ! static::is_pro_active() ) {
+			add_action( 'wp_footer', function () {
```
louiswol94, line 109: @nicoladj77

Can you please paste the output of the HTML here? I just want to make sure of something.

modules/conversion-center/module.php
```diff
@@ -96,6 +105,50 @@ public function __construct() {
...
+				$query = new \WP_Query( [
+					'post_type' => ConversionCenterModule::CPT_CONTACT_PAGES,
+					'posts_per_page' => -1,
```
louiswol94, line 112: Why not just `1` for `posts_per_page`?

modules/conversion-center/module.php
```diff
@@ -96,6 +105,50 @@ public function __construct() {
...
+			} );
+		}
+
+		add_filter( 'elementor-pro/theme-builder/conditions-manager/additional-cpts', function( array $cpts ) {
+			$cpts[] = self::CPT_CONTACT_PAGES;
+
+			return $cpts;
+		} );
+
+		add_filter( 'elementor-pro/theme-builder/allowed-documents', function( $allowed, $document ) {
```
louiswol94, line 144: I'm not sure about these filters yet. I don't really understand why you need them here in Core if you're just going to print the content in `wp_footer`.

core/common/assets/js/views/modal/header.js
```diff
@@ -38,7 +38,8 @@ export default class extends Marionette.LayoutView {
 
 		if (
 			window.elementor && elementor.config?.admin_conversion_center_contact_url &&
-			'contact-buttons' === elementor.config?.document?.type
+			'contact-buttons' === elementor.config?.document?.type &&
+			this.$el.find( '#elementor-template-library-header-menu' ).length
```
louiswol94, line 42: Can we remove this change?

louiswol94 (changes requested):

modules/conversion-center/module.php
nicoladj77: @louiswol94 this is to handle an edge case, which is that the user has pro installed, has set various display conditions, and then removes pro. in that case I want to only show the instance that was set to display and the whole site

modules/conversion-center/module.php
nicoladj77: @louiswol94 those filters are needed because I need to pass the contact buttons to enable display conditions in pro. Display conditions depend on extending a class which only exists in pro. I can't use the contact button class in pro.

modules/conversion-center/assets/js/editor/hooks/ui/editor/document/select.js
nicoladj77: @louiswol94 that looks a lot more complex because I think it's part of the import process. I also think it's cleaner because it actually prevents any possibility of selecting a section or a container.

modules/conversion-center/module.php
nicoladj77: @louiswol94 sorry I had imported it and forgot to delete it

core/common/assets/js/views/modal/header.js
nicoladj77: @louiswol94 this is needed as otherwise we close the display condition modal like we close the library modal, basically we return to the admin dashboard. This is not the cleanest approach, but closing the modal doesn't fire a command, which is why we have this code here. And the only difference between the two modals, without targetting a string which is prone to failure, is that the library modal had tabs, while the display conditions doesn't have them. If you have suggestions let me know.

modules/conversion-center/assets/js/editor/hooks/ui/editor/document/select.js
louiswol94: Okay I get you, if the import process selects this automatically, then it's ok to de-select it here with a Hook. nice.

modules/conversion-center/assets/js/editor/hooks/ui/editor/documents/open/add-contact-buttons-tab.js
```diff
@@ -17,7 +17,7 @@ export class ContactPageAddLibraryTab extends $e.modules.hookUI.After {
 			title: __( 'Floating Buttons', 'elementor' ),
 			filter: {
 				source: 'remote',
-				type: 'contact_button',
+				type: 'floating_button',
```
louiswol94, line 20: @nicoladj77 Maybe this should be in it's own PR. I think we're gonna have some back and forth on this PR.

core/common/assets/js/views/modal/header.js
louiswol94: @nicoladj77 I see. So closing the Display Conditions modal will also redirect to Admin?

Can you please extract these conditions to a method for readability? `isFloatingButtonLibraryClose`? Something like that, so we know what the situation is you are checking for.

modules/conversion-center/module.php
louiswol94: @nicoladj77 Just thinking here...

- Can we run the "register document" conditionally in Core? Only register the document in Core if Pro does not exist, otherwise register the Document in Pro?

nicoladj77:
Not Relevant anymore

PR #27860, opened 2024-06-27
Internal: Replace the image presented when 'apply new kit' flow is finished [ED-14216] 

louiswol94:


murzin-nikita (approved):

doronwolf (approved):
perfect! thank you

PR #27872, opened 2024-06-28
Fix: Tabs Widget - Tab stops being functional if we assign a CSS ID to a tab [ED-15160]

louiswol94:


modules/nested-tabs/assets/js/frontend/handlers/nested-tabs.js
```diff
@@ -52,7 +52,7 @@ export default class NestedTabs extends Base {
 		return {
 			selectors: {
 				widgetContainer: `[data-widget-number="${ widgetNumber }"]`,
-				tabTitle: `[id*="e-n-tab-title-${ widgetNumber }"]`,
+				tabTitle: `[aria-controls*="e-n-tab-content-${ widgetNumber }"]`,
```
louiswol94, line 55: We cannot use an `ID` selector here, because the user can change the ID from the controls.
I decided to use the `aria-controls` selector because it's the only one that worked well out of the box, without me having to add another attribute or class.

hein-obox (approved):

PR #27917, opened 2024-07-02
Tweak: Load nested elements styles only when they are in use [ED-15011]

rami-elementor:


KingYes (approved):

modules/nested-accordion/module.php
```diff
@@ -27,5 +26,21 @@ public function __construct() {
 				'nested-elements',
 			], ELEMENTOR_VERSION, true );
 		} );
+
+		add_action( 'elementor/frontend/after_register_styles', [ $this, 'register_style' ] );
```
rotemee, line 30: @rami-elementor - the before action is "/editor" while the after is "/frontend" is it intended?

modules/nested-accordion/module.php
```diff
@@ -27,5 +26,21 @@ public function __construct() {
...
+
+	/**
+	 * Register styles.
+	 *
+	 * @return void
+	 */
+	public function register_style() {
+		wp_register_style(
+			$this->get_name(),
+			$this->get_css_assets_url( 'frontend', 'assets/css/modules/nested-accordion/' ),
```
rotemee, line 41: @rami-elementor - is it tested in dev+prod+RTL?

modules/nested-accordion/widgets/nested-accordion.php
```diff
@@ -44,6 +44,10 @@ public function get_keywords() {
 		return [ 'nested', 'tabs', 'accordion', 'toggle' ];
 	}
 
+	public function get_style_depends(): array {
+		return [ 'nested-accordion' ];
```
rotemee, line 48: @rami-elementor - is it ok that it's going to be loaded in the footer? (layout shift etc..)

modules/nested-tabs/module.php
```diff
@@ -26,5 +26,21 @@ public function __construct() {
 				'nested-elements',
 			], ELEMENTOR_VERSION, true );
 		} );
+
+		add_action( 'elementor/frontend/after_register_styles', [ $this, 'register_style' ] );
```
rotemee, line 30: Same as above.

modules/nested-tabs/module.php
```diff
@@ -26,5 +26,21 @@ public function __construct() {
...
+	/**
+	 * Register styles.
+	 *
+	 * @return void
+	 */
+	public function register_style() {
+		wp_register_style(
+			$this->get_name(),
+			$this->get_css_assets_url( 'frontend', 'assets/css/modules/nested-tabs/' ),
+			[],
```
rotemee, line 42: @rami-elementor - Are you sure that no-dependency is needed? I see that there is a base-tabs.scss file for example.

rotemee (changes requested):

.grunt-config/sass.js
```diff
@@ -70,6 +70,20 @@ const sass = {
 				dest: 'assets/css/modules/apps',
 				ext: '.css',
 			},
+			{
```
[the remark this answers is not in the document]
louiswol94: Thanks @rotemee!

I actually prefer not to add a specific new entry for every module.

If we can create a rule like `modules/*/assets/sccs/*.sccs` - or something like this, and then just place the SCSS files in the correct location in each module, that would be great.

modules/nested-accordion/module.php
rami-elementor: The previous action is for JS, this one is for CSS.

When registering styles, there are specificity issues, therefor we need to position them after the all frontend styles are loaded.

Furthermore, when it was in the "_modules.scss" file, it was loaded in the end, after all fronend atyles are loaded.

modules/nested-accordion/widgets/nested-accordion.php
hein-obox:
```suggestion
		return [ $this->get_name() ];
```

modules/nested-accordion/module.php
hein-obox: @rami-elementor The use of this action hook is causing some specificity problems. That is why some of the nested tabs tests are failing.

I would prefer if we can enqueue the style in the `head `of the page, but after all other styles. I am not sure if we can do that?

hein-obox: @rami-elementor @rotemee @louiswol94 
I have worked on a possible solution to enqueue the widgets in this order:
- frontend.css
- conditionally loaded widgets
- inline widget styling

rami-elementor: @hein-obox how do you suggest to load the style in the head?

hein-obox: @rami-elementor Sorry, I think I had forgotten to share my test PR: 
https://github.com/elementor/elementor/pull/27944. 

My idea was to include the active widget CSS with this hook: `'elementor/frontend/enqueue_widgets'`

PR #27922, opened 2024-07-03
Internal: Add atomic widgets

Omerisra6:


includes/managers/widgets.php
```diff
@@ -100,18 +101,22 @@ private function init_widgets() {
 			'share-buttons',
 		];
 
+		$atomic_widgets_filenames = [
```
ronros-elementor, line 104: should we need it here? as it's also defined in `register_atomic_widgets`

includes/managers/widgets.php
```diff
@@ -145,6 +150,26 @@ private function init_widgets() {
...
+		$class_name = str_replace( '-', '_', $filename );
+
+		$class_name = __NAMESPACE__ . '\Widget_' . $class_name;
+
+		$this->register( new $class_name() );
+	}
+
+	private function register_atomic_widgets() {
+		$atomic_widgets_filenames = [
+			'atomic-heading',
```
ronros-elementor, line 163: shouldn't it be `'a-heading'` as it would have its way into the dom eventually? (or I might be confusing it with something else, not sure)

includes/base/atomic-widget-base.php
```diff
@@ -0,0 +1,47 @@
...
+		parent::__construct( $data, $args );
+
+		$this->version = $data['version'] ?? '0.0';
+	}
+
+	abstract public function get_atomic_controls(): array;
+
+	final public function get_controls( $aaa = null ) {
+		return [];
+	}
```
ronros-elementor, line 22: I believe we wanted it to return null if a specific `$control_id` is passed (to avoid faulty checks we saw throughout the code that might cause unexpected outcomes)
```suggestion
	final public function get_controls( $control_id = null ) {
		if(!empty($control_id)) {
			return null;
		}
		return [];
	}
```

ronros-elementor (changes requested):

includes/base/atomic-widget-base.php
Omerisra6: Done.

includes/managers/widgets.php
Omerisra6: Removed.

includes/managers/widgets.php
Omerisra6: We have decided to name all the 'atomic' related features as atomic-${name}, the only thing that should be shortened is the widget name. That array stores all the widget file names.

ronros-elementor (approved):

PR #27952, opened 2024-07-05
Tweak: Improve Menu widget performance when using Dynamic Tags in the URL field [ED-14967]

davseve:


louiswol94 (changes requested):
@davseve Can we avoid that Core knows about a specific feature like Mega Menu in Pro? You have very specific selectors and function names for the MM in this PR. Try and override these methods from this file `assets/dev/js/editor/elements/views/base.js` in `elementor-pro/modules/mega-menu/assets/js/editor/views/view.js`. I just added `renderOnChanges()` for example in this file, and it called the function in Pro.

davseve:
@louiswol94 I get what you say, If we'll add this ti the Pro repo we won't have a single source off truth to handle the dynamic item title tags. 
I am refactoring the code so it will be placed on the Pro mega menu view.js file as you suggested.

PR #27969, opened 2024-07-07
Tweak: Improve Nested Elements Dynamic Tags Title - Advanced [ED-15005]

ronenboxer:


assets/dev/js/editor/elements/views/base.js
```diff
@@ -1106,6 +1147,29 @@ BaseElementView = BaseContainer.extend( {
...
+	tryHandleDynamicTagsAdvancedTools( dataBinding, settings ) {
+		const attributeKeys = Object.keys( settings.attributes ),
+			shouldTryHandle = ! this.findUniqueKey( [ 'before', 'after', 'fallback' ], attributeKeys, true ).length,
+			changedControlKey = Object.keys( settings.changed )[ 0 ];
+
+		if ( ! shouldTryHandle || ! changedControlKey ) {
+			return false;
+		}
+
+		jQuery( dataBinding.el ).html( this.getAdvancedDynamicTitleChange(
```
ronenelementor, line 1164: Tried using `dataBinding.el.innerHTML = ...` but it didn't render changes until entire process was done, caused delays and failed tests

assets/dev/js/editor/elements/views/base.js
```diff
@@ -661,10 +661,10 @@ BaseElementView = BaseContainer.extend( {
 			dataBinding.el.getAttribute( 'data-binding-setting' ) === changedControl;
 	},
 
-	async getDynamicValue( settings, bindingSetting ) {
```
ronenelementor, line 664: I originally made changes in this function to address the task's AC in it, so in order for it to cover cases from previous logic and new logic, I had to make changes.
Since then, I decided to move my logic out because they don't rely on any dynamic content, so no need to wait for the async logic to happen to render them so these changes can be removed, however, I think we should consider leaving them if they make sense universally (regarding this function and the `renderDataBinding`, not to be confused with `renderDataBindings` )

murzin-nikita (approved):

louiswol94 (approved):

PR #28003, opened 2024-07-09
WIP

louiswol94:
WIP/POC - need to move the code to the appropriate asset loader classes and element iteration actions.


modules/nested-tabs/assets/scss/frontend.scss
```diff
@@ -1,4 +1,6 @@
 @import "../../../../assets/dev/scss/helpers/mixins";
+@import "../../../../assets/dev/scss/helpers/variables";
```
hein-obox, line 2: I am curious if this helps us to remove the need to import the helper files later on. I fear that this won't make a difference, but let's see.

PR #28076, opened 2024-07-15
Tweak: Merge "Lazy Load Background Images" feature and move to Settings [ED-15336]

rami-elementor:


modules/lazyload/module.php
```diff
@@ -100,4 +89,8 @@ public function init() {
 	private function should_lazyload() {
 		return ! is_admin() && ! Plugin::$instance->preview->is_preview_mode() && ! Plugin::$instance->editor->is_edit_mode();
 	}
+
+	private static function is_lazy_load_background_images_enabled() : bool {
```
KingYes, line 93:
```suggestion
	private static function is_lazy_load_background_images_enabled(): bool {
```

modules/lazyload/module.php
```diff
@@ -100,4 +89,8 @@ public function init() {
...
+		return (bool) get_option( 'elementor_lazy_load_background_images', '1' );
```
KingYes, line 94:
```suggestion
		return '1' === get_option( 'elementor_lazy_load_background_images', '1' );
```

KingYes (changes requested):

louiswol94 (approved):

KingYes (approved):

PR #28092, opened 2024-07-16
Tweak: Load `get_style_depends()` CSS files in the head to improve CLS [ED-15307]

KingYes:


rami-elementor (dismissed):

louiswol94:
@KingYes Should we remove if from here now? https://github.com/elementor/elementor/blob/main/includes/base/element-base.php#L516-L517

hein-obox (dismissed):

KingYes:
> @KingYes Should we remove if from here now? https://github.com/elementor/elementor/blob/main/includes/base/element-base.php#L516-L517

Yes, we should. But for some of BC, I prefer to keep it for a while.

louiswol94 (dismissed):

rami-elementor (approved):

PR #28111, opened 2024-07-17
Tweak: Improve Nested Elements Dynamic Tags - Title - Advanced [ED-15362]

ronenboxer:


assets/dev/js/editor/elements/views/base.js
```diff
@@ -834,6 +842,12 @@ BaseElementView = BaseContainer.extend( {
 			return;
 		}
 
+		if ( this.isRendering ) {
+			this.isRendering = false;
+
+			return;
+		}
```
ronenelementor, line 849: Changing the `before/after/fallback` inputs triggers `renderOnChange` twice. On the second trigger, the attributes are different and they lead to `renderDataBinding` which causes the entire widget to re render

davseve (approved):

baghdasarovelementor (approved):

PR #28121, opened 2024-07-18
Tweak: Load text-path styles only when it is in use [ED-15210]

elchugreeva:


modules/shapes/widgets/text-path.php
```diff
@@ -37,6 +37,10 @@ public function get_name() {
 		return 'text-path';
 	}
 
+	public function get_group_name() {
```
hein-obox, line 40: Nice catch

assets/dev/scss/frontend/_modules.scss
```diff
@@ -1,2 +1 @@
 // Shapes Module
-@import '../../../../modules/shapes/assets/scss/frontend';
```
hein-obox, line 2: I think now we can completely remove this file, and remove this reference from the frontend-lite.scss file as well.

hein-obox (approved):

louiswol94 (approved):

Batsirai-muchareva (approved):

PR #28126, opened 2024-07-18
Fix: Response code values are missing in Custom Code generated by AI [ED-15386]

DennisNerush:
image


modules/ai/assets/js/editor/pages/form-code/splitTextResult.js
```diff
@@ -0,0 +1,15 @@
+export function splitText( inputText ) {
+	if ( ! inputText ) {
+		return {};
+	}
+	const codeMatch = inputText.match( /```([\s\S]*?)```/ );
+	const code = codeMatch ? `\`\`\`${ codeMatch[ 1 ] }\`\`\``.trim() : '';
+
+	const detailsMatch = inputText.match( /```[\s\S]*?```([\s\S]*)/ );
+	const details = detailsMatch ? detailsMatch[ 1 ].trim() : '';
```
AdirD, line 9: instead of ternary you can do the nullish operator which looks simpler to understand (but you will end up with null if thats ok with the rest of the flow)

```suggestion
	const details = detailsMatch?.[ 1 ].trim();
```

AdirD (dismissed):

AdirD (approved):

PR #28128, opened 2024-07-18
Tweak: Conditionally load widget styles only if the widgets are used [ED-15384]

rami-elementor:


includes/widgets/accordion.php
```diff
@@ -78,6 +78,22 @@ protected function is_dynamic_content(): bool {
...
+	 * Get style dependencies.
+	 *
+	 * Retrieve the list of style dependencies the widget requires.
+	 *
+	 * @since 3.24.0
+	 * @access public
+	 *
+	 * @return array Widget style dependencies.
+	 */
+	public function get_style_depends(): array {
```
hein-obox, line 91: I would prefer if we could handle his logic in one of the widget base files.

Another question: could this lead to Core/Pro conflicts in any way?

assets/dev/scss/frontend/frontend.scss
```diff
@@ -1,5 +1,32 @@
-// Frontend Lite
-@import "frontend-lite";
+@import "../global/global";
+@import "breakpoints/proxy";
+@import "../helpers/mixins";
 
-// Widgets
-@import "widgets";
```
louiswol94, line 5: @rami-elementor What is the purpose of this PR? Are we now trying to optimise the default frontend CSS file, even without the "Improved CSS Loading"?

includes/frontend.php
```diff
@@ -238,7 +238,7 @@ public function init() {
 		$document = Plugin::$instance->documents->get( $this->post_id );
 
 		if ( is_singular() && $document && $document->is_built_with_elementor() ) {
-			add_action( 'wp_enqueue_scripts', [ $this, 'enqueue_styles' ] );
+			add_action( 'wp_enqueue_scripts', [ $this, 'enqueue_styles' ], 20 );
```
louiswol94, line 241: Can you add a const for this `20`? So that we don't have to guess what it's about :)
`AFTER_SOMETHING`

includes/frontend.php
```diff
@@ -535,9 +535,7 @@ public function register_styles() {
 
 		$direction_suffix = is_rtl() ? '-rtl' : '';
 
-		$frontend_base_file_name = $this->is_optimized_css_mode() ? 'frontend-lite' : 'frontend';
-
-		$frontend_file_name = $frontend_base_file_name . $direction_suffix . $min_suffix . '.css';
+		$frontend_file_name = "frontend{$direction_suffix}{$min_suffix}.css";
```
louiswol94, line 538: Ok, so are we removing `frontend-lite`? 🤔

includes/managers/widgets.php
```diff
@@ -552,6 +552,69 @@ public function get_widgets_frontend_settings_keys() {
...
+	 *
+	 * This method returns the list of all the widgets in the `/includes/`
+	 * folder that have styles.
+	 *
+	 * @since 3.24.0
+	 * @access public
+	 *
+	 * @return array The names of the widgets styles.
+	 */
+	public function widgets_styles_names(): array {
```
hein-obox, line 581: I am curious, how can this be extended by other plugins, including the Pro plugin?

includes/managers/widgets.php
```diff
@@ -552,6 +552,69 @@ public function get_widgets_frontend_settings_keys() {
...
+		$widgets_styles = [];
+
+		// Widgets inside the `/includes/widgets/` folder that have styles.
+		$widgets = [
```
louiswol94, line 585: @rami-elementor 

1. I don't know about this. It is really hard to maintain. I am thinking, the grunt process knows which widgets it created a CSS file for, maybe we should save it in a JSON - similar to `/assets/data/responsive-widgets.json`?
2. Do we need to add a hook here for 3rd parties? Hw can they leverage this loading method?

includes/preview.php
```diff
@@ -103,7 +103,7 @@ public function init() {
 		add_action( 'wp_enqueue_scripts', function() {
 			$this->enqueue_styles();
 			$this->enqueue_scripts();
-		} );
+		}, 20 );
```
louiswol94, line 106: Please add a const here.

includes/widgets/accordion.php
louiswol94: I agree, maybe in the base we can check if the CSS file exists, and if it does, load it in the `get_style_depends()`.

Maybe this way we also don't `need widgets_styles_names()`.

modules/floating-buttons/module.php
```diff
@@ -90,6 +90,8 @@ public function __construct() {
 		add_action( 'wp_ajax_elementor_send_clicks', [ $this, 'handle_click_tracking' ] );
 		add_action( 'wp_ajax_nopriv_elementor_send_clicks', [ $this, 'handle_click_tracking' ] );
 
+		add_action( 'wp_enqueue_scripts', [ $this, 'register_style' ] );
```
louiswol94, line 93: Just making sure this is definitely the correct hook to use? And there's not an Elementor hook we should rather use?

modules/floating-buttons/module.php
```diff
@@ -520,4 +522,20 @@ private function render_floating_buttons(): void {
...
+	 * @return void
+	 */
+	public function register_style() {
+		wp_register_style(
+			'widget-floating-buttons',
+			$this->get_css_assets_url( 'widget-floating-buttons', null, true, true ),
+			[],
+			ELEMENTOR_VERSION
+		);
+	}
```
louiswol94, line 540: This is not needed with the Improved CSS loading. So I guess this pertains to my first question, did we now decide to improve the general CSS loading as well?

modules/floating-buttons/widgets/contact-buttons.php
```diff
@@ -3,6 +3,7 @@
 namespace Elementor\Modules\FloatingButtons\Widgets;
 
 use Elementor\Modules\FloatingButtons\Base\Widget_Contact_Button_Base;
+use Elementor\Plugin;
```
louiswol94, line 6: Do we need this import?

modules/link-in-bio/module.php
```diff
@@ -41,4 +41,27 @@ public static function get_experimental_data(): array {
...
+
+	/**
+	 * Register styles.
+	 *
+	 * At build time, Elementor compiles `/modules/link-in-bio/assets/scss/frontend.scss`
+	 * to `/assets/css/widget-link-in-bio.min.css`.
+	 *
+	 * @return void
+	 */
+	public function register_style() {
```
louiswol94, line 59: Maybe we need a better solution for grouped widgets? 🤔

modules/shapes/module.php
```diff
@@ -8,6 +8,29 @@
...
+	 * @return void
+	 */
+	public function register_style() {
+		wp_register_style(
+			'widget-shapes',
+			$this->get_css_assets_url( 'widget-shapes', null, true, true ),
+			[],
+			ELEMENTOR_VERSION
+		);
+	}
```
louiswol94, line 32: Yeah, I think we need a similar method for these modules with widgets:

https://github.com/elementor/elementor/pull/28128/files#diff-25f13c7a23bc088a3c138855fa91482dd8a779a109f0adfea34208a36c81d956R563-R572

louiswol94 (changes requested):

includes/managers/widgets.php
rami-elementor: This is a unique solution for internal widgets in the `/includes/widgets/` folder. Widgets in the `/modules/` folder don't need this, so are third parties.

Third party developers can use the docs:
https://developers.elementor.com/docs/widgets/widget-dependencies/

assets/dev/scss/frontend/frontend.scss
rami-elementor: @louiswol94 You are correct.

The task is to eliminate `frontend-lite.min.css`. Making `frontend.min.css` as a lite file.

All the pages should load a light version of the frontend file, and widgets styles should always load conditionally, whether the experiment is active or not.

includes/frontend.php
rami-elementor: Requested by @KingYes, after a code review.

This way when you register styles in widgets, you can use the default 10 value.

To do that, when styles are enqueue, they should be above the default 10.

We want this value to be hardcoded.

includes/frontend.php
rami-elementor: YES!

includes/managers/widgets.php
rami-elementor: 1. We have two types of widgets, in the `/includes/widgets/` folder and in the `/modules/` folder. When using modules structure, you can register styles. Widgets in the "includes" folder can't. this is a workaround. The widget manager will register styles for them.

2. [Initially I added hooks](https://github.com/elementor/elementor/pull/28128/commits/bb628db7618100fe1a32a760bc566c0535ba6bf0) for third parties, but @KingYes prefers not to use hooks.

modules/floating-buttons/module.php
rami-elementor: Why does it matter? it's only for registration, not for enqueue.

modules/floating-buttons/module.php
rami-elementor: Yes, this PR improves the general CSS loading, both for `frontend.css` and for widgets styles.

modules/floating-buttons/widgets/contact-buttons.php
rami-elementor: Yes!

It fixed an error in the CI. See: https://github.com/elementor/elementor/pull/28128/commits/daa422531d436cf6761e2e13ecbd78ef11536ad5 commit.

rami-elementor: Don't remember which error, but it took me a while to find the issue.

includes/frontend.php
louiswol94: @rami-elementor Can you add a const or comment? Otherwise another developer would need to guess what `20` means.

includes/managers/widgets.php
louiswol94: @rami-elementor This method can create a JSON for the standalone widgets for example? Then we don't have to keep track of them here as well.

`https://github.com/elementor/elementor/blob/main/.grunt-config/widgets-css.js#L92-L109`

modules/floating-buttons/module.php
louiswol94: @rami-elementor What is the point then of the CSS Loading Experiment if Elementor will now be so optimised out of the box? It seems like it's now just a choice between if you want to load it in the <head> or inside the widget 🤔

louiswol94: I understand previously - having an experiment, to make a drastic improvement, but now it kind of just changes the location of the loading. So maybe the "loading location" can just be part of the widget config or something. I hope my thought pattern is making sense here :)

includes/managers/widgets.php
louiswol94: @rami-elementor This one is important ☝️ 
(Some of the other comments were just thoughts).

rami-elementor: Let me see if I understand correctly, this Grunt tasks, which runs on run on build time, generate the CSS. You want the task to generate addition JSON file with all the styles in the `/includes/widgets/` folder.

The `widgets_styles_names()` method is used for registering styles in PHP. You want register CSS files, from the JSON file.

The issue is that loading file from the disk is an expensive operation. It will increase server compute, negatively impacting on TTFB. We usually try to avoid these type of solutions.

louiswol94: @rami-elementor We do this for the Responsive CSS (Custom Breakpoints). I think it's here: `elementor/core/page-assets/data-managers/responsive-widgets.php`
My problem is just that when we add a new widget for example, we will never remember to come and add it here in `widgets_styles_names()`
So the Grunt task will just create this array for us (in JSON format), instead of us trying to maintain it manually.

modules/link-in-bio/module.php
rami-elementor: @louiswol94 What do you suggest?

louiswol94: @rami-elementor Let's leave it for now :)

includes/widgets/accordion.php
rami-elementor: @hein-obox  I'm trying to do the opposite. Not reduce the code, and hide everything in the widget base class.

I want the dependencies to be visible in the widget class, this way developers will see all the styles/scripts when looking at the widgets class. This would improve Developer Experience (DE), mainly for third-party developers.

As for Core/Pro conflicts, only if both widgets use the same name for style handles. But this is a general WordPress issue. The solutions is to use naming conventions.

includes/managers/widgets.php
rami-elementor: new widgets are added only using modules, not in the includes folder.

louiswol94 (dismissed):

includes/widgets/accordion.php
```diff
@@ -78,6 +78,22 @@ protected function is_dynamic_content(): bool {
...
+		return $this->load_widgets_styles_in_head()
+			? [ 'widget-accordion' ]
+			: [];
+	}
```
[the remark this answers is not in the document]
ronenelementor: This way we have one source to maintain and a unison convention for style-id. (We can also have the widget prefix set in a const)

includes/widgets/social-icons.php
```diff
@@ -78,6 +74,22 @@ protected function is_dynamic_content(): bool {
...
+	 * @since 3.24.0
+	 * @access public
+	 *
+	 * @return array Widget style dependencies.
+	 */
+	public function get_style_depends(): array {
+		return $this->load_widgets_styles_in_head()
+			? [ 'widget-social-icons', 'e-apple-webkit' ]
+			: [ 'e-apple-webkit' ];
+	}
```
ronenelementor, line 91: We can override my suggestion in edge cases

Gruntfile.js
```diff
@@ -105,10 +105,6 @@ module.exports = function( grunt ) {
 				src: [
 					'frontend.scss',
 					'frontend-rtl.scss',
-					'frontend-lite.scss',
-					'frontend-lite-rtl.scss',
-					'frontend-legacy.scss',
-					'frontend-legacy-rtl.scss',
```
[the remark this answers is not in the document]
rami-elementor: @ronenelementor They don't exist. Someone probably forgot to delete them.

includes/widgets/accordion.php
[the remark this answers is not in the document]
rami-elementor: This method will be very specific for our needs, but what about third-party plugins? what if they use a different naming convention? In any case, I prefer not to do that as we may change this naming convention in future releases. We plan to rewrite the grunt task responsible for compiling SCSS files to CSS.

ronenboxer (approved):

arielk (approved):

PR #28183, opened 2024-07-23
Tweak: Add heading and descriptions to admin settings pages [ED-15322]

rami-elementor:


KingYes (dismissed):

includes/settings/settings.php
```diff
@@ -252,6 +252,13 @@ protected function create_tabs() {
 				'label' => esc_html__( 'General', 'elementor' ),
 				'sections' => [
 					'general' => [
+						'label' => esc_html__( 'General', 'elementor' ),
+						'callback' => function() {
+							printf(
+								'<p>%s</p><br><hr><br>',
```
davseve, line 258:
```suggestion
								'<p>%s</p><br/><hr/><br/>',
```

includes/settings/settings.php
```diff
@@ -322,6 +329,13 @@ protected function create_tabs() {
 				'label' => esc_html__( 'Advanced', 'elementor' ),
 				'sections' => [
 					'advanced' => [
+						'label' => esc_html__( 'Advanced', 'elementor' ),
+						'callback' => function() {
+							printf(
+								'<p>%s</p><br><hr><br>',
```
davseve, line 335: Same as above

includes/settings/settings.php
rami-elementor: Not sure about this. This is an XHTML convention. HTML5 allows self-closing tags.

Read this reference: https://developer.mozilla.org/en-US/docs/Glossary/Void_element

> Self-closing tags (`<tag />`) do not exist in HTML.
>
> If a trailing `/` (slash) character is present in the start tag of an HTML element, HTML parsers ignore that slash character.

includes/settings/settings.php
rami-elementor: See my answer above.

arielk (approved):

PR #28237, opened 2024-07-29
Internal: Add new control type [ED-360] 

blackcolt:


modules/atomic-widgets/controls/types/input-control.php
```diff
@@ -0,0 +1,34 @@
...
+namespace Elementor\Modules\AtomicWidgets\Controls\Types;
+
+use Elementor\Modules\AtomicWidgets\Base\Atomic_Control_Base;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly
+}
+
+class Input_Control extends Atomic_Control_Base {
+	private string $placeholder;
```
Omerisra6, line 11: add default

Omerisra6 (changes requested):

Issue #28137, opened 2024-07-19
Icon List Widget Bug While Using it within Loop Item


ayoubkhan558:
### Prerequisites

- [X] I have searched for similar issues in open and closed tickets and cannot find a duplicate.
- [X] I have troubleshooted my issue, and it still exists against the latest stable version of Elementor.

### Description

When I Icon List Widget within Loop Item Template and Loop over posts, I select Text of Icon list item as Post Title using dynamic tags, But when it renders it renders only first post title in each loop item on front.

But when I use any other widget to render post title dynamically it renders it perfectly.

![image](https://github.com/user-attachments/assets/32b77a5c-0f96-4524-abe4-9d9b2739e3ea)

![image](https://github.com/user-attachments/assets/f93bb713-881f-469b-bd8c-cb650a694909)

### Steps to reproduce

1.  Create Loop Item template
2. Use Icon List widget within this loop
3. Select text of IconList Item as Post title using Dynamic Tags.
4. Now Loop Grid in any page and render post/custom post using this. 
5. Preview result, You will see, first post title in each IconList items text.

### Expected behavior

It should render any dynamic tag content within text or as icon. but It fails.

### Isolating the problem

- [X] This bug happens when only the Elementor (and Elementor Pro) plugins are active.
- [ ] This bug happens with the Hello Elementor theme active.
- [X] I can reproduce this bug consistently by following the steps I described above.

### Elementor System Info

```txt
== Server Environment ==
	Operating System: Linux
	Software: LiteSpeed
	MySQL version: MariaDB Server v10.11.8-MariaDB-cll-lve
	PHP Version: 8.1.27
	PHP Memory Limit: 512M
	PHP Max Input Vars: 5000
	PHP Max Post Size: 256M
	GD Installed: Yes
	ZIP Installed: Yes
	Write Permissions: All right
	Elementor Library: Connected

== WordPress Environment ==
	Version: 6.5.5
	Site URL: https://elitepropae.com/royalpin
	Home URL: https://elitepropae.com/royalpin
	WP Multisite: No
	Max Upload Size: 256 MB
	Memory limit: 40M
	Max Memory limit: 512M
	Permalink Structure: /%postname%/
	Language: en_US
	Timezone: 0
	Admin Email: muhammadinshadarshadbajwa@gmail.com
	Debug Mode: Inactive

== Theme ==
	Name: Hello Elementor
	Version: 3.1.0
	Author: Elementor Team
	Child Theme: No

== User ==
[...]

PR #28241, opened 2024-07-29
Fix: Element Caching prevents repeater control from loading Dynamic Tags (#28137) [ED-15421] (CP)

KingYes:


arielk (approved):

includes/base/element-base.php
```diff
@@ -561,7 +561,7 @@ protected function should_render_shortcode() {
 
 		$is_dynamic_content = apply_filters( 'elementor/element/is_dynamic_content', false, $raw_data, $this );
 
-		$has_dynamic_tag = ! empty( $raw_data['settings']['__dynamic__'] );
+		$has_dynamic_tag = $this->has_element_dynamic_tag( $raw_data['settings'] );
```
yotamselementor, line 564: Can we add a unit test for this?

PR #28274, opened 2024-08-01
Internal: Create The Marketing Checklist BE Infra [ED-15465]

ronenboxer:


modules/checklist/module.php
```diff
@@ -0,0 +1,178 @@
...
+use Elementor\Plugin;
+use Elementor\Utils;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+class Module extends BaseModule {
+	const EXPERIMENT_ID = 'launchpad-checklist';
+	const DB_OPTION_KEY = 'e-checklist';
```
mykytamurzin, line 17: should it be `elementor_checklist`? 
As I see other names are in format `elementor_*`

modules/checklist/module.php
```diff
@@ -0,0 +1,178 @@
...
+	 *
+	 * @param $step_id
+	 *
+	 * @return null|array {
+	 *      @type bool $is_marked_done
+	 *      @type bool $is_completed
+	 *  }
+	 */
+	public function get_step_progress( $step_id ) {
+		foreach ( $this->user_progress['steps'] as $id => $step ) {
```
mykytamurzin, line 90: can we just check for `$this->user_progress['steps'][$step_id]` ?

modules/checklist/steps-manager.php
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	/**
+	 * Marks a step as done, returns true if the step was found and marked or false otherwise
+	 *
+	 * @param string $step_id
+	 *
+	 * @return bool
+	 */
+	public function mark_step_as_done( $step_id ) : bool {
```
mykytamurzin, line 45: `string $step_id`

modules/checklist/steps-manager.php
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	/**
+	 * Unmarks a step as done, returns true if the step was found and unmarked or false otherwise
+	 *
+	 * @param string $step_id
+	 *
+	 * @return bool
+	 */
+	public function unmark_step_as_done( $step_id ) : bool {
```
mykytamurzin, line 64: `string $step_id`

modules/checklist/steps-manager.php
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	/**
+	 * Maybe marks a step as completed (depending on if source allows it), returns true if the step was found and marked or false otherwise
+	 *
+	 * @param $step_id
+	 *
+	 * @return bool
+	 */
+	public function maybe_mark_step_as_complete( $step_id ) : bool {
```
mykytamurzin, line 83: a bit confusing between `done` and `complete` namings

modules/checklist/module.php
```diff
@@ -0,0 +1,178 @@
...
+	private Steps_Manager $steps_manager;
+
+	private Wordpress_Adapter $wordpress_adapter;
+
+	/**
+	 * @param ?Wordpress_Adapter $wordpress_adapter
+	 *
+	 * @return void
+	 */
+	public function __construct( $wordpress_adapter = null ) {
```
mykytamurzin, line 30: `?Wordpress_Adapter_Interface $wordpress_adapter = null`

modules/checklist/steps/step-base.php
```diff
@@ -0,0 +1,150 @@
...
+		$this->wordpress_adapter = $wordpress_adapter ?? new Wordpress_Adapter();
+		$this->user_progress = $module->get_step_progress( $this->get_id() ) ?? $this->get_step_initial_progress();
+	}
+
+	/**
+	 * Marking a step as done based on user's desire
+	 *
+	 * @return void
+	 */
+	public function mark_as_done() : void {
```
mykytamurzin, line 66: should we return true/false when succeed or not?

modules/checklist/module.php
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	private function init_user_progress() {
+		$default_settings = [
+			'is_hidden' => false,
+			'last_opened_timestamp' => time(),
+			'steps' => [],
+		];
+
+		add_option( self::DB_OPTION_KEY, wp_json_encode( $default_settings ) );
```
mykytamurzin, line 172: looks like it will re-write options with default settings on each constructor call. Or maybe I wrong?

modules/checklist/module.php
ronenelementor: Sure!

modules/checklist/module.php
ronenelementor: Yes you're right

modules/checklist/steps-manager.php
ronenelementor: I see, I started a discussion on slack and we'll decide on better naming

modules/checklist/steps/step-base.php
ronenelementor: No real requirement for it, depends how polite we're feeling like
WDYT?

modules/checklist/module.php
ronenelementor: add_option doesn't update existing options, I can leave a comment with reference link if you think it's needed

modules/checklist/module.php
```diff
@@ -0,0 +1,172 @@
...
+		$this->wordpress_adapter = $wordpress_adapter ?? new Wordpress_Adapter();
+		parent::__construct();
+
+		$this->register_experiment();
+
+		if ( ! $this->is_experiment_active() ) {
+			return;
+		}
+
+		$this->setup();
```
TzviRabinovitch, line 41: why did you split it from the CTOR ?

modules/checklist/module.php
```diff
@@ -0,0 +1,172 @@
...
+	 *      @type array $steps {
+	 *          @type string $step_id => {
+	 *              @type bool $is_marked_completed
+	 *              @type bool $is_completed
+	 *          }
+	 *      }
+	 *  }
+	 */
+	public function get_user_progress_from_db() : array {
+		return json_decode( get_option( self::DB_OPTION_KEY ), true );
```
TzviRabinovitch, line 77: why not use WP adapter  ?

modules/checklist/module.php
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	private function init_user_progress() : void {
+		$default_settings = [
+			'is_hidden' => false,
+			'last_opened_timestamp' => time(),
+			'steps' => [],
+		];
+
+		add_option( self::DB_OPTION_KEY, wp_json_encode( $default_settings ) );
```
TzviRabinovitch, line 166: why not use WP adapter ?

modules/checklist/module.php
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	private function update_user_progress_in_db() : void {
+		update_option( self::DB_OPTION_KEY, wp_json_encode( $this->user_progress ) );
```
TzviRabinovitch, line 170: why not use WP adapter ?
update_option

modules/checklist/steps-manager.php
```diff
@@ -0,0 +1,140 @@
...
+	exit; // Exit if accessed directly.
+}
+
+class Steps_Manager {
+	/** @var Step_Base[] $step_instances */
+	private array $step_instances = [];
+
+	private Module $module;
+
+	public function __construct( Module $module ) {
```
TzviRabinovitch, line 18: step manager can be decoupled it if will get an interface of the module and not the actual module class in the CTOR

modules/checklist/steps-manager.php
```diff
@@ -0,0 +1,140 @@
...
+			}
+		}
+	}
+
+	public function get_step_by_id( string $step_id ) : ?Step_Base {
+		foreach ( $this->step_instances as $step ) {
+			if ( $step->get_id() === $step_id ) {
+				return $step;
+			}
+		}
```
louiswol94, line 94: Do you need a loop here? Just check if `$this->step_instances[ $step_id ]` is set, and then return it?

modules/checklist/steps-manager.php
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	/**
+	 * Getting the step instances array based on source's order
+	 *
+	 * @return void
+	 */
+	private function register_steps() : void {
+		foreach ( $this->get_step_ids() as $step_id ) {
+			$step_instance = $this->get_step_instance( $step_id );
```
louiswol94, line 106: Only register it if it has not been registered already.

modules/checklist/steps-manager.php
ronenelementor: The instances are not associative, I thought we should have them ordered but if you prefer we can make it associative

ronenelementor: I'll switch it to be associative array

modules/checklist/steps-manager.php
ronenelementor: Done

modules/checklist/steps-manager.php
ronenelementor: Done

modules/checklist/module.php
ronenelementor: United elements

modules/checklist/module.php

modules/checklist/module.php

modules/checklist/module.php

modules/checklist/steps-manager.php
ronenelementor: Done

TzviRabinovitch (approved):

PR #28285, opened 2024-08-04
Internal: Share-links - jQuery to vanilla JS [ED-9706]

nuritsha:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes

## Summary

* [tweak] share-link lib: replace the use of jQuery with vanilla JS

## Description

*

## Test instructions

*

## Quality assurance

Fixes #


assets/lib/share-link/share-link.js
```diff
@@ -58,32 +56,24 @@
...
 
 			settings.classPrefixLength = settings.classPrefix.length;
 		};
 
-		var initElements = function() {
-			$element = $( element );
-		};
-
-		var init = function() {
+		const init = function() {
```
hein-obox, line 67: Maybe convert to arrow function?

assets/lib/share-link/share-link.js
```diff
@@ -116,22 +106,20 @@
 		height: 480,
 	};
 
-	ShareLink.getNetworkLink = function( networkName, settings ) {
-		var link = ShareLink.networkTemplates[ networkName ].replace( /{([^}]+)}/g, function( fullMatch, pureMatch ) {
+	window.ShareLink.getNetworkLink = function( networkName, settings ) {
```
hein-obox, line 109: For clarification: why do we have to add the `window` object here?

Maybe arrow function?

assets/lib/share-link/share-link.js
```diff
@@ -144,9 +132,15 @@
...
+	};
+
+	NodeList.prototype.shareLink = function ( settings ) {
+		this.forEach( ( element ) => {
+			new ShareLink( element, settings );
 		} );
+		return this;
 	};
-} )( jQuery );
+} )();
```
hein-obox, line 146: Maybe?

```suggestion
} );
```

assets/lib/share-link/share-link.js
```diff
@@ -144,9 +132,15 @@
...
 
-	$.fn.shareLink = function( settings ) {
-		return this.each( function() {
-			$( this ).data( 'shareLink', new ShareLink( this, settings ) );
+	Element.prototype.shareLink = function ( settings ) {
+		new ShareLink( this, settings );
+		return this;
+	};
+
+	NodeList.prototype.shareLink = function ( settings ) {
```
hein-obox, line 140: Couldn't we do do this above? 
I am not sure.

assets/lib/share-link/share-link.js
```diff
@@ -1,25 +1,23 @@
 /**
  * By Elementor Team
  */
-( function( $ ) {
-	window.ShareLink = function( element, userSettings ) {
-		var $element,
-			settings = {};
+( function() {
```
hein-obox, line 4: Could we remove this function wrapper?
In that case should we create it as a class? 

Not sure. But this looks like a jQuery approach.

assets/lib/share-link/share-link.js
nuritsha: @hein-obox but don't we need to call the wrapping function? if we remove the `()` at the end, will the function get called ?

hein-obox: @nuritsha Could you check how we do it other places? Generally speaking we export the class or function, and we don't have to wrap it in inside the kind of jQuery function.

assets/lib/share-link/share-link.js
```diff
@@ -1,25 +1,23 @@
...
+	window.ShareLink = function( eElement, userSettings ) {
+		let element = {};
```
rami-elementor, line 6: Initially it was undefined, and now it's an object. Is that correct?

hein-obox: Why don't we use the eElement instead of element?
Can't we use the eElement object that is passed to this function?

Originally we needed to convert it from an `element` to `$element`, but now maybe we can use it as is?

assets/lib/share-link/share-link.js
rami-elementor: This approach called Immediately Invoked Function Expression (IIFE), it's an older convention. I agree with @hein-obox , we need to stop using IIFE.

Ref:
* https://developer.mozilla.org/en-US/docs/Glossary/IIFE
* https://en.wikipedia.org/wiki/Immediately_invoked_function_expression
* Also Google for "IIFE alternatives".

assets/lib/share-link/share-link.js
nuritsha: @hein-obox it can be an arrow function, but for the sake of style consistency throughout the ShareLink object, I kept it as a regular function

assets/lib/share-link/share-link.js
nuritsha: @hein-obox I'm not sure I understand what you mean 🤔
As far as I understand, in order to cover all the cases that are covered by the jQuery implementation, we need to add the method to both the prototypes, what am I missing?

PR #28287, opened 2024-08-05
Fix: Styles Command - history transaction [EDS-331]

ManorHazaz:


modules/atomic-widgets/assets/js/editor/commands/styles.js
```diff
@@ -91,38 +96,39 @@ export class Styles extends $e.modules.editor.document.CommandHistoryDebounceBas
...
 		}
 
 		if ( ! this.variantExists( style, meta ) ) {
 			$e.internal( 'document/atomic-widgets/create-variant', {
 				container,
-				styleDefId,
+				styleDefID,
 				meta,
 			} );
 		}
```
blackcolt, line 127: if ( ! styleDefID ) {
			style = $e.internal( 'document/atomic-widgets/create-style', {
				container,
				bind,
			} );

			styleDefID = style.id;
		} else if ( oldStyles[ styleDefID ] ) {
		    style = oldStyles[ styleDefID ];
		} else {
			throw new Error( 'Style Def not found' );
		}

blackcolt (approved):

PR #28311, opened 2024-08-07
Internal: POC / Load Swiper CSS through JS

hein-obox:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #

### DIFF
--- a/assets/dev/js/frontend/utils/assets-loader.js
+++ b/assets/dev/js/frontend/utils/assets-loader.js
@@ -42,10 +42,21 @@ export default class AssetsLoader {
 
 			element.onload = () => resolve( true );
 
-			const parent = 'head' === assetData.parent ? assetData.parent : 'body';
-
-			document[ parent ].appendChild( element );
+			this.appendAsset( assetData, element );
 		} );
+	}
+
+	appendAsset( assetData, element ) {
+		const beforeElement = document.querySelector( assetData.before );
+
+		if ( !! beforeElement ) {
+			beforeElement.insertAdjacentElement( 'beforebegin', element );
+			return;
+		}
+
+		const parent = 'head' === assetData.parent ? assetData.parent : 'body';
+
+		document[ parent ].appendChild( element );
 	}
 }
 
@@ -75,6 +86,7 @@ AssetsLoader.assets = {
 		swiper: {
 			src: swiperCssSource,
 			parent: 'head',
+			before: '#elementor-frontend-css',
 		},
 		'e-lightbox': {
 			src: `${ elementorFrontendConfig.urls.assets }css/conditionals/lightbox${ fileSuffix }.css?ver=${ elementorFrontendConfig.version }`,
--- a/assets/dev/js/frontend/utils/swiper.js
+++ b/assets/dev/js/frontend/utils/swiper.js
@@ -17,7 +17,8 @@ export default class Swiper {
 		container.closest( '.elementor-widget' )?.classList.add( 'e-widget-swiper' );
 
 		return new Promise( ( resolve ) => {
-			elementorFrontend.utils.assetsLoader.load( 'script', 'swiper' )
+			elementorFrontend.utils.assetsLoader.load( 'style', 'swiper' )
+				.then( () => elementorFrontend.utils.assetsLoader.load( 'script', 'swiper' ) )
 				.then( () => resolve( this.createSwiperInstance( container, this.config ) ) );
 		} );
 	}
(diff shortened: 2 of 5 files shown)


assets/dev/js/frontend/utils/assets-loader.js
```diff
@@ -42,11 +42,22 @@ export default class AssetsLoader {
...
-
-			document[ parent ].appendChild( element );
+			this.appendAsset( assetData, element );
 		} );
 	}
+
+	appendAsset( assetData, element ) {
+		const beforeElement = document.querySelector( assetData.before );
+
+		if ( !! beforeElement ) {
```
hein-obox, line 52: Load before the frontend styling, otherwise we get specificity issues.

assets/dev/js/frontend/utils/swiper.js
```diff
@@ -17,7 +17,8 @@ export default class Swiper {
 		container.closest( '.elementor-widget' )?.classList.add( 'e-widget-swiper' );
 
 		return new Promise( ( resolve ) => {
-			elementorFrontend.utils.assetsLoader.load( 'script', 'swiper' )
+			elementorFrontend.utils.assetsLoader.load( 'style', 'swiper' )
```
hein-obox, line 20: Load Swiper styling on the initiation of a new Swiper JS Class. This class is used in all Swiper widgets that are using the `elementorFrontend.utils.swiper` to instantiate a new Swiper.

assets/dev/scss/frontend/_swiper-preload.scss
```diff
@@ -0,0 +1,46 @@
+.swiper {
```
hein-obox, line 1: I have loaded this styling on every page, to avoid unnecessary loading glitches before the swiper.css has been loaded.

Is it acceptable to load this on every page?

includes/frontend.php
```diff
@@ -659,7 +659,7 @@ public function enqueue_styles() {
 
 			wp_enqueue_style( 'elementor-frontend' );
 
-			wp_enqueue_style( 'swiper' );
+//			wp_enqueue_style( 'swiper' );
```
hein-obox, line 662: I have tested this with the enqueue commented out, and the test results are very good.

assets/dev/scss/frontend/_swiper-preload.scss
louiswol94: @hein-obox IMO - if this helps with the layout shifts, and it's not a lot of code, I think it is worth the trade off.

rami-elementor: We try to avoid loading unnecessary styles on all pages. If it is required, I guess it will be acceptable. But if you have any other solution you should investigate a different approach.

rami-elementor: BTW, why not add the styles to `_swiper-customisations.scss` ?

hein-obox: @rami-elementor 

I suggested to Mugurel to load the _swiper-customisations.scss conditionally, and to remove from frontend.scss. What do you think?

rami-elementor: How will his PR affect your POC?

And if his PR loads `_swiper-customisations.scss` only when in use, why do we need this PR?

hein-obox: The main question: what will happen to existing websites that are using a custom swiper based on the elementorFrontend.utils.swiper class? They won't have the swiper.css file, once we remove it from the frontend.scss file.

This applies to 3rd party plugins but also to individual websites with custom solutions.

hein-obox: @doronwolf

doronwolf: @hein-obox @rami-elementor If I understand you correctly, it sounds like the risk is real and we don't have a solution for existing sites, I think we should try to find a way to active the new conditional style behavior only for new sites (and also inform developers and 3rd party plugins to give them time to prepare for the change), and as long as we don't have a solution for existing sites then we will need to put this task on hold

hein-obox: I will plan a meeting about this.

rami-elementor: @doronwolf you add additional and unnecessary complexity to the POC.

doronwolf: @rami-elementor There is no complexity - if we lack BC support then we need to find a solution before merging this PR to the version, let's discuss it in the meeting

PR #28350, opened 2024-08-11
Internal: Create Launchpad Checklist settings [ED-15438]

ronenboxer:


tests/playwright/sanity/modules/checklist/selectors.ts
elchugreeva, line 1: maybe also move to selectors folder?

ronenelementor: Discussed and decided to leave as is

modules/checklist/assets/js/editor.js
```diff
@@ -3,3 +3,15 @@ import Component from './component';
 
 $e.components.register( new Component() );
 editorV2();
+
+window.addEventListener( 'DOMContentLoaded', maybeHideToggleIcon );
+
+function maybeHideToggleIcon() {
```
louiswol94, line 9: @ronenelementor Maybe we should conditionally run `editorV2()` based on the setting value, instead of always registering it and maybe hiding it. WDYT?

modules/checklist/module.php
```diff
@@ -143,6 +146,14 @@ public function enqueue_editor_scripts() : void {
 		} );
 	}
 
+	public static function is_preference_switch_on() : bool {
+		$user_preferences = get_user_meta( get_current_user_id(), Preferences_Manager::META_KEY, true );
```
louiswol94, line 150: Try to use `$editor_preferences->get_model()->get_settings( self::VISIBILITY_SWITCH_ID )`

louiswol94 (changes requested):

modules/checklist/assets/js/editor.js
ronenelementor: I thought about it, but then if the user turns the settings switch on, we won't be able to show it as far as I understand, WDYT?

louiswol94: You are correct! A page refresh is probably not the best UX.

modules/checklist/module.php

louiswol94 (approved):

PR #28368, opened 2024-08-13
Tweak: Reduce the use of JS for smooth scroll with Modern CSS scroll [ED-14623]

Batsirai-muchareva:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #


core/experiments/manager.php
```diff
@@ -482,6 +482,17 @@ private function add_default_features() {
...
+
+		$this->add_feature( [
+			'name' => 'e_css_smooth_scroll',
+			'title' => esc_html__( 'CSS Smooth Scroll', 'elementor' ),
+			'tag' => esc_html__( 'Performance', 'elementor' ),
+			'description' => esc_html__( 'Use CSS Smooth Scroll to improve the user experience on your site. This experiment replaces the default JavaScript-based smooth scroll with a CSS-based solution.', 'elementor' ),
+			'release_status' => self::RELEASE_STATUS_DEV,
+			static::TYPE_HIDDEN => true,
+			'default' => self::STATE_INACTIVE,
+			'generator_tag' => true,
```
hein-obox, line 494: What do we use this tag for?

.grunt-config/sass.js
```diff
@@ -91,6 +91,13 @@ const sass = {
 				dest: 'assets/css',
 				ext: '.css',
 			},
+			{
+				expand: true,
```
hein-obox, line 95: Maybe you can move this scss file to `'assets/dev/scss/frontend/conditionals'`?

assets/dev/js/frontend/frontend.js
```diff
@@ -205,6 +205,14 @@ export default class Frontend extends elementorModules.ViewModule {
 			controls: new Controls(),
 		};
 
+		if ( ! this.config.experimentalFeatures.e_css_smooth_scroll ) {
```
hein-obox, line 208: Maybe use `if { } else { }` instead?

assets/dev/js/frontend/utils/anchor-scroll-margin.js
```diff
@@ -0,0 +1,127 @@
...
+
+	onInit() {
+		const waitForElements = new Promise( ( resolve ) => {
+			// Simulating an async operation
+			setTimeout( () => {
+				resolve();
+			}, 100 );
+		} );
+
+		waitForElements.then( () => {
```
hein-obox, line 19: Do we need this wait?
If so, then I would prefer to use a method that doesn't include a timeout.

assets/dev/js/frontend/utils/anchor-scroll-margin.js
```diff
@@ -0,0 +1,127 @@
...
+
+		this.organizeStickyAndAnchors( trackedElements );
+	},
+
+	organizeStickyAndAnchors( elements ) {
+		const stickyList = this.filterAndSortElementsByType( elements, 'sticky' );
+		const anchorList = this.filterAndSortElementsByType( elements, 'anchor' );
+
+		stickyList.forEach( ( sticky, index ) => {
+			// Define the range for the current sticky
```
hein-obox, line 60: Could you possibly extract this logic to a function, and give the function a self-explantory name, like `this.defineCurrentStickyRange()`

assets/dev/js/frontend/utils/anchor-scroll-margin.js
```diff
@@ -0,0 +1,127 @@
...
+			const nextStickyScrollPosition = ( index + 1 < stickyList.length )
+				? stickyList[ index + 1 ].scrollPosition
+				: Infinity;
+			sticky.anchor = anchorList.filter( ( anchor ) => {
+				const withinRange = anchor.scrollPosition > sticky.scrollPosition && anchor.scrollPosition < nextStickyScrollPosition;
+				if ( withinRange ) {
+					// Set scrollMarginTop for each anchor
```
hein-obox, line 67: Can we try to avoid comments?

.grunt-config/sass.js
```diff
@@ -88,6 +88,13 @@ const sass = {
 				expand: true,
 				cwd: 'assets/dev/scss/admin',
 				src: 'wc-product-editor.scss',
+				dest: 'assets/css/conditionals',
+				ext: '.css',
+			},
+			{
+				expand: true,
```
hein-obox, line 95: You probably can delete this.

All files in that folder are automatically added to the css folder.

Screenshot 2024-08-27 at 14 53 22

assets/dev/js/frontend/utils/anchor-scroll-margin.js
```diff
@@ -0,0 +1,153 @@
...
+				links: '.elementor-element a[href*="#"]',
+				stickyElements: '.elementor-element.elementor-sticky',
+			},
+		};
+	},
+
+	onInit() {
+		const that = this;
+
+		this.observeStickyElements( () => {
```
hein-obox, line 14: What happens when there are no sticky elements?

Do we only need to initialize this code when there are sticky elements?

assets/dev/js/frontend/utils/anchor-scroll-margin.js
```diff
@@ -0,0 +1,153 @@
...
+	},
+
+	defineCurrentStickyRange( sticky, index, stickyList, anchorList ) {
+		const nextStickyScrollPosition = ( index + 1 < stickyList.length )
+			? stickyList[ index + 1 ].scrollPosition
+			: Infinity;
+		sticky.anchor = anchorList.filter( ( anchor ) => {
+			const withinRange = anchor.scrollPosition > sticky.scrollPosition && anchor.scrollPosition < nextStickyScrollPosition;
+			if ( withinRange ) {
+				anchor.element.style.scrollMarginTop = `${ sticky.scrollMarginTop }px`;
```
hein-obox, line 95: Do we need to handle an offset value for the admin bar well? The admin bar adds something like 32px to the frontend, if it is visible. I assume that we will need to add that value conditionally.

includes/frontend.php
```diff
@@ -673,6 +673,15 @@ public function enqueue_styles() {
 				wp_enqueue_style( 'elementor-wp-admin-bar' );
 			}
 
+			if ( Plugin::$instance->experiments->is_feature_active( 'e_css_smooth_scroll' ) ) {
+				wp_enqueue_style(
```
louiswol94, line 677: Should we not only register the style here, and enqueue it only when needed?

hein-obox: @louiswol94 I am actually not sure what should be condition here. Only if an anchor is available? But we check that with JS. In that case we could load it with the `assets/dev/js/frontend/utils/assets-loader.js`.

Alternatively we could load it on every page load in one of the default frontend file (unconditionally). 

Wdyt?

hein-obox: @louiswol94 Maybe this is fine actually.

assets/dev/js/frontend/utils/anchor-scroll-margin.js
Batsirai-muchareva: @hein-obox do you have a reference for implementing this

assets/dev/js/frontend/utils/anchor-scroll-margin.js
Batsirai-muchareva: I have put that check in the functions below @hein-obox

assets/dev/js/frontend/utils/anchor-scroll-margin.js
hein-obox: @Batsirai-muchareva 
https://github.com/elementor/elementor/blob/9eeee9b0695700187cda24b97962a726c80a47cb/assets/dev/js/frontend/utils/anchors.js#L54

assets/dev/js/frontend/utils/anchor-scroll-margin.js
hein-obox: @Batsirai-muchareva 
Are those functions triggered if there are no sticky elements as well?

hein-obox (dismissed):

assets/dev/js/frontend/utils/anchor-scroll-margin.js
```diff
@@ -0,0 +1,160 @@
+module.exports = elementorModules.ViewModule.extend( {
```
louiswol94, line 1:
```suggestion
export default class extends elementorModules.ViewModule
```
Can you make this an ES6 module?

.grunt-config/sass.js
```diff
@@ -88,7 +88,7 @@ const sass = {
 				expand: true,
 				cwd: 'assets/dev/scss/admin',
 				src: 'wc-product-editor.scss',
-				dest: 'assets/css',
+				dest: 'assets/css/conditionals',
```
louiswol94, line 91: Why change the location of the WC Product Editor in this PR?

assets/dev/js/frontend/utils/anchor-scroll-margin.js
```diff
@@ -0,0 +1,160 @@
...
+	getDefaultSettings() {
```
louiswol94, line 2: Just a question, this is a lot of JS, how did we "reduce the use of JS"? :)

includes/frontend.php
louiswol94: Do you mean we can always load it?

louiswol94 (changes requested):

.grunt-config/sass.js
hein-obox: @Batsirai-muchareva I think this is good point.
It is not good practice to change unrelated aspects in a PR.

includes/frontend.php
hein-obox: @louiswol94 I think so.
It is just one line of css.

assets/dev/js/frontend/utils/anchor-scroll-margin.js
hein-obox: @louiswol94 If I understand it correctly, the code will only run once, when there are no sticky elements. Then it will only assign the scroll margin values on page load.

When there are sticky elements with a scroll margin top setting on the page, then this code will run more often. But only then. I think that is acceptable.

The scrolling itself is not handled by JS, which causes some of the Accessibility issues.

assets/dev/js/frontend/utils/anchor-scroll-margin.js
```diff
@@ -0,0 +1,166 @@
...
+		return {
+			selectors: {
+				links: '.elementor-element a[href*="#"]',
+				stickyElements: '.elementor-element.elementor-sticky',
+			},
+		};
+	}
+
+	onInit() {
+		const that = this;
```
louiswol94, line 12: @Batsirai-muchareva Can you avoid this because you are using an arrow function?

hein-obox (approved):

louiswol94 (approved):

afagard:
This causes an issue with people who want to use:

```
    elementorFrontend.on( 'components:init', function() {
        elementorFrontend.utils.anchors.setSettings( 'selectors.targets', '.dummy-selector' );
    } );
```

to disable this functionality and do their own.

PR #28383, opened 2024-08-14
Tweak: 'Edit with Elementor' button to Woo New Product Editor [ED-15359]

louiswol94:


modules/wc-product-editor/assets/js/e-wc-product-editor.js
```diff
@@ -0,0 +1,63 @@
...
+	const productId = useEntityId( 'postType', 'product' );
+	const { saveEntityRecord } = useDispatch( 'core' );
+
+	const postStatus = useSelect( ( select ) => {
+		return select( 'core' ).getEditedEntityRecord( 'postType', 'product', productId )?.status;
+	}, [ productId ] );
+
+	const handleClick = () => {
+		if ( 'auto-draft' === postStatus ) {
+			saveEntityRecord( 'postType', 'product', {
```
joshuatf, line 17: This should work fine, but one alternative to this it to use the [useProductManager](https://github.com/woocommerce/woocommerce/blob/8c523190d86aa9d628e146b6dca22f8fcaa4097a/packages/js/product-editor/src/hooks/use-product-manager/use-product-manager.ts#L47) hook exported from `@woocommerce/product-editor` as `__experimentalUseProductManager`.

This hook has a number of states and methods you can use:

```js
return {
	isValidating,
	isDirty,
	isSaving,
	isPublishing: isSaving,
	isTrashing,
	save,
	publish,
	trash,
	copyToDraft,
};
```

modules/wc-product-editor/assets/js/e-wc-product-editor.js
```diff
@@ -0,0 +1,63 @@
...
+				id: productId,
+				name: `Elementor #${ productId }`,
+				status: 'draft',
```
joshuatf, line 20: This probably works, but keep in mind that if the user has made any changes to the current product, those will be lost since this does not account for the edits in the entity record.

On the flip side, if you do include edits you'll also need to account for a potentially invalid state.

modules/wc-product-editor/assets/js/e-wc-product-editor.js
```diff
@@ -0,0 +1,63 @@
...
+				redirectToElementor( productId );
+			} ).catch( () => {} );
+		} else {
+			redirectToElementor( productId );
+		}
+	};
+
+	const redirectToElementor = ( id ) => {
+		const checkSaveStatus = () => {
+			if ( wp.data.select( 'core/editor' ).isSavingPost() ) {
```
joshuatf, line 31: Instead of creating a recursive timeout, you could grab this value outside of this block:

```js
const isSaving = wp.data.select( 'core/editor' ).isSavingPost();
```

Create a state for when you should be redirecting `isRedirecting` and set to `true` inside your `handleClick`.  Then add a `useEffect` to watch for those things:

```js
useEffect( () => {
    if ( isRedirecting && ! isSaving ) {
        // do redirect
    }
}, [ isRedirecting, isSaving ] );
```

This will be more performant and reduce any potential delays due to the timeout.

modules/wc-product-editor/assets/js/e-wc-product-editor.js
```diff
@@ -0,0 +1,63 @@
...
+		url.searchParams.set( 'action', 'elementor' );
+
+		return url.toString();
+	};
+
+	return (
+		<WooHeaderItem name="product">
+			<Button variant="primary" onClick={ handleClick } style={ { display: 'flex', alignItems: 'center' } }>
+				<i className="eicon-elementor-square" aria-hidden="true" style={ { paddingInlineEnd: '8px' } }></i>
+				Edit with Elementor
```
joshuatf, line 54: I'm not sure how this repo is set up, but you might consider translation if you have `@wordpress/i18n`.

```suggestion
				__( 'Edit with Elementor', 'text-domain' )
```

joshuatf (commented):
Have not tested this all the way through; just leaving some comments around the entity data store.  Hope that helps!

modules/wc-product-editor/assets/js/e-wc-product-editor.js
louiswol94: Of course! Updated.

modules/wc-product-editor/assets/js/e-wc-product-editor.js
louiswol94: @joshuatf I noticed this during manual testing, but then I looked at our current implementation with the classic product editor (not sure what you call it 😃), and it has the same problem. So I decided not to inflate the scope of this task. Thanks!

package.json
```diff
@@ -164,6 +168,10 @@
 		"@reach/router": {
 			"react": "$react",
 			"react-dom": "$react-dom"
+		},
+		"@woocommerce/admin-layout": {
+			"react": "^18.2.0",
+			"react-dom": "^18.2.0"
```
louiswol94, line 174: @joshuatf I had to add an `npm` override for the React version for `@woocommerce/admin-layout`. I didn't see any visible side effects 🤞 

```
npm install @woocommerce/admin-layout --save                   
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR! 
npm ERR! While resolving: elementor@3.25.0
npm ERR! Found: react@18.3.1
npm ERR! node_modules/react
npm ERR!   dev react@"^18.2.0" from the root project
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer react@"^17.0.2" from @woocommerce/admin-layout@1.1.0
npm ERR! node_modules/@woocommerce/admin-layout
npm ERR!   @woocommerce/admin-layout@"^1.1.0" from the root project
```

modules/wc-product-editor/assets/js/e-wc-product-editor.js
louiswol94: Hey @joshuatf, I think I prefer not to bring yet another package in just for the save, but thanks, it was helpful.

PS: Again got `ERESOLVE unable to resolve dependency tree` when trying to install `@woocommerce/product-editor`

Still keen to hear your feedback if you have further push back or thoughts :)

modules/wc-product-editor/assets/js/e-wc-product-editor.js
louiswol94: Missed that one. Thanks!

modules/wc-product-editor/assets/js/e-wc-product-editor.js
joshuatf: That should be fine!  It does the same thing for now.  The one benefit to `useProductManager` is that you'd get progressive updates to the hook functions as we go along, but it may not be necessary in your case.

Also, we have no plans to move away from the entity data store so this should continue to work without issue.

Batsirai-muchareva (approved):

mugurelLupu (approved):

PR #28421, opened 2024-08-18
Internal: Integrate with post css for rendering atomic style [EDS-350]

ManorHazaz:


blackcolt (dismissed):

modules/atomic-widgets/module.php
```diff
@@ -74,4 +79,26 @@ private function enqueue_scripts() {
 			true
 		);
 	}
+
+	/**
+	 * Add the CSS for the element.
+	 *
+	 * @param Post $post
+	 * @param Element_Base $element
+	 */
```
StyleShit, line 88: this PHPDoc doesn't help much (both in terms of descriptiveness and typings)
either explain properly or remove it completely

modules/atomic-widgets/module.php
```diff
@@ -74,4 +79,26 @@ private function enqueue_scripts() {
...
+	private function add_css( Post $post, Element_Base $element ) {
+		if ( ! ( $element instanceof Atomic_Widget_Base ) || Post::class !== get_class( $post ) ) {
+			return;
+		}
+
+		$styles = $element->get_raw_data()['styles'];
+
+		if ( empty( $styles ) ) {
+			return;
+		}
```
StyleShit, line 98: I'm not sure whether this piece of code should be a part of this PR or be in the next PR

StyleShit (commented):
Not sure what's the point of this PR TBH 🤷

modules/atomic-widgets/module.php
ManorHazaz: So the next PR will remove it.

PR #28457, opened 2024-08-20
Internal: Change atomic attachment control to atomic image control [EDS-370]

ManorHazaz:


modules/atomic-widgets/base/atomic-widget-base.php
```diff
@@ -182,17 +182,16 @@ private function transform_setting( $setting ) {
...
-				];
+				if ( isset( $setting['value']['attachmentId'] ) ) {
+					$attachment_url = wp_get_attachment_image_url( $setting['value']['attachmentId'] ) ?? null;
+				}
 
-			case 'image-url':
-				return [
-					'url' => $setting['value']['url'] ?? Utils::get_placeholder_image_src(),
-				];
+				return $attachment_url ?? Utils::get_placeholder_image_src();
```
StyleShit, line 194: shouldn't it be something like this?
(because you need to have either id or url)

```PHP
$url = '';

if ( isset( $value['id'] ) ) {
	$url = get_by_id( $value['id'] );
} else if ( isset( $value['url' ) ) {
	$url = $value['url'];
}

return $url ?? get_placeholder();
```

StyleShit (changes requested):

modules/atomic-widgets/base/atomic-widget-base.php
ManorHazaz: Done

modules/atomic-widgets/base/atomic-widget-base.php
```diff
@@ -183,15 +183,13 @@ private function transform_setting( $setting ) {
...
-
 				if ( isset( $setting['value']['attachmentId'] ) ) {
-					$attachment_url = wp_get_attachment_image_url( $setting['value']['attachmentId'] ) ?? null;
+					$url = wp_get_attachment_image_url( $setting['value']['attachmentId'] ) ?? null;
+				} elseif ( isset( $setting['value']['url'] ) ) {
+					$url = $setting['value']['url'];
 				}
 
-				return $attachment_url ?? Utils::get_placeholder_image_src();
+				return $url ?? Utils::get_placeholder_image_src();
```
StyleShit, line 192:
```suggestion
				return empty( $url ) ? Utils::get_placeholder_image_src() : $url;
```

to catch cases where the URL is falsy but not null

StyleShit (changes requested):

StyleShit (approved):

PR #28475, opened 2024-08-21
Fix: Build with AI fails with "The app is not available" error [ED-15586]

matipojo:


DennisNerush (approved):
Perhaps worth asking Tzvi/Yakir to review as well.

louiswol94:
@matipojo @DennisNerush I think this was a bug... needs to be:

```
opener.dispatchEvent( new CustomEvent( 'elementor/connect/success', {
       detail: <?php echo wp_json_encode( $data ); ?>
} ) );
```

`dispatchEvent()` does not have two arguments :)

matipojo:
> @matipojo @DennisNerush I think this was a bug... needs to be:
> 
> ```
> opener.dispatchEvent( new CustomEvent( 'elementor/connect/success', {
>        detail: <?php echo wp_json_encode( $data ); ?>
> } ) );
> ```
> 
> `dispatchEvent()` does not have two arguments :)

It looks like a bug...
But seems not to affect the app. Is something not working?

PR #28516, opened 2024-08-26
Internal: Validate atomic widgets props schema only in Editor [EDS-397]

StyleShit:


tests/phpunit/elementor/modules/atomic-widgets/test-atomic-widget-base.php
```diff
@@ -316,8 +226,23 @@ public function test_get_props_schema__is_serializable() {
...
+
+		$widget = $this->make_mock_widget( [ 'props_schema' => $schema ] );
+
+		// Expect.
+		$this->expectNotToPerformAssertions();
+
 		// Act.
+		$widget::get_props_schema();
+	}
+
```
StyleShit, line 243: the only new test, other has just been moved down and renamed

modules/atomic-widgets/base/atomic-widget-base.php
```diff
@@ -24,20 +24,23 @@ public function __construct( $data = [], $args = null ) {
...
+
+		// Validate the schema only in the Editor.
+		static::validate_schema( $schema );
 
-		return $this->get_valid_controls( $controls );
+		return $this->get_valid_controls( $schema, $controls );
 	}
 
-	private function get_valid_controls( array $controls ): array {
+	private function get_valid_controls( array $schema, array $controls ): array {
```
Omerisra6, line 35: Don't you think better approach will be to call the `get_props_schema` method?

tests/phpunit/elementor/modules/atomic-widgets/test-atomic-widget-base.php
```diff
@@ -316,8 +226,23 @@ public function test_get_props_schema__is_serializable() {
 		}', $serialized );
 	}
 
-	public function test_get_props_schema() {
+	public function test_get_props_schema__not_throwing_when_the_schema_is_invalid() {
```
Omerisra6, line 229: I don't think this test is necessary

Omerisra6 (changes requested):

modules/atomic-widgets/base/atomic-widget-base.php
StyleShit: It's in order to avoid calculating the schema many times per widget

tests/phpunit/elementor/modules/atomic-widgets/test-atomic-widget-base.php
StyleShit: it's to make sure that the `get_props_schema` doesn't validate the schema each time you call it, to ensure it's not running in the frontend

Omerisra6: There is no situation in which the method will throw an exception, so I think you should delete it.

Omerisra6 (approved):

PR #28542, opened 2024-08-28
Internal: Arrangement of style transformers [EDS-399]

ronros-elementor:
Adding Styles_Renderer to the Post_CSS rendering flow

+ Organized the first two style transformers, though expected to be changed later on once we have a solid implementation for Transformer_Base


modules/atomic-widgets/base/style-transformer-base.php
```diff
@@ -0,0 +1,24 @@
+<?php
+namespace Elementor\Modules\AtomicWidgets\Base;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+abstract class Style_Transformer_Base {
```
Nevoss, line 8: I think this will be handle by my side, but we can talk about it

modules/atomic-widgets/module.php
```diff
@@ -51,6 +68,12 @@ public function __construct() {
 			add_filter( 'elementor/widgets/register', fn( Widgets_Manager $widgets_manager ) => $this->register_widgets( $widgets_manager ) );
 			add_filter( 'elementor/editor/localize_settings', fn( array $settings ) => $this->add_prop_types_config( $settings ) );
 			add_action( 'elementor/editor/after_enqueue_scripts', fn() => $this->enqueue_scripts() );
+
+			add_filter(
+				'elementor/atomic-widgets/styles/transformers',
+				fn ( array $transformers) => $this->register_style_transformers( $transformers )
+			);
+			add_action( 'elementor/element/parse_css', fn( Post $post, Element_Base $element ) => $this->parse_atomic_widget_css( $post, $element ), 10, 2 );
```
Nevoss, line 76: Please move all the logic of the rendering to a different class, something like `Element_Parse_CSS`, the action could be here in the module but the logic can live outside

modules/atomic-widgets/module.php
```diff
@@ -110,4 +147,106 @@ private function enqueue_scripts() {
...
+
+	private function get_breakpoints(): array {
+		if ( count( $this->breakpoints ) === 0 ) {
+			$this->breakpoints = $this->build_breakpoints();
+		}
+
+		return $this->breakpoints;
+	}
+
+	private function build_breakpoints(): array {
```
Nevoss, line 179: Why do you need normalize the breakpoints?

Nevoss (dismissed):

modules/atomic-widgets/base/style-transformer-base.php
ronros-elementor: Totally agree, it's a placeholder for now but I based it on your work
One thing I need to align with you on is how can I invoke the value-transform logic from within a transformer (I passed a callback as its second argument)

ronros-elementor: Created [this task](https://elementor.atlassian.net/browse/EDS-424) to refactor once the main backend transformers task is done

modules/atomic-widgets/module.php
ronros-elementor: Done

modules/atomic-widgets/module.php
```diff
@@ -34,6 +34,7 @@ public function __construct() {
 
 		if ( Plugin::$instance->experiments->is_feature_active( self::EXPERIMENT_NAME ) ) {
 			( new Dynamic_Tags() )->register_hooks();
+			( new Widget_Styles() )->register_hooks();
```
Omerisra6, line 37: I think the atomic widgets module should register the hook for the widgets styles.

The `Widget_Styles` module should be a generic module that takes a style object and returns the css using the transformers.

modules/atomic-widgets/widget-styles.php
```diff
@@ -0,0 +1,101 @@
...
+use Elementor\Modules\AtomicWidgets\Styles\Styles_Renderer;
+use Elementor\Modules\AtomicWidgets\Styles\Transformers\Array_Transformer;
+use Elementor\Modules\AtomicWidgets\Styles\Transformers\Size_Transformer;
+use Elementor\Plugin;
+
+class Widget_Styles {
+	/**
+	 * @var array<string, Style_Transformer_Base> $transformers
+	 */
+	private array $style_transformers = [];
```
Omerisra6, line 18: I think it makes more sense that the transformers will be stored in the `Styles_Renderer`, I am not sure why we pass the renderer a config instead of storing it there...

Omerisra6 (changes requested):

modules/atomic-widgets/module.php
ronros-elementor: As discussed - since Styles_Renderer _is_ already that generic module which gets a styles array and returns a css string - adding the handler for `elementor/element/parse_css` within the general module might cause some unnecessary "noise" to that module.
But I did separate, within the `Atomic_Styles` module, the logic of the action handler, and the use of `Style_Renderer` into two different functions for better readability

Omerisra6 (changes requested):

modules/atomic-widgets/atomic-styles.php
```diff
@@ -0,0 +1,118 @@
...
+			return;
+		}
+
+		$styles = $element->get_raw_data()['styles'];
+
+		if ( empty( $styles ) ) {
+			return;
+		}
+
+		$post->get_stylesheet()->add_raw_css( $this->convert_styles_to_css( $styles ) );
```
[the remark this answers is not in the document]
ronros-elementor: Done

modules/atomic-widgets/styles/styles-renderer.php
```diff
@@ -132,15 +134,21 @@ private function wrap_with_media_query( string $breakpoint, string $css ): strin
 			return $css;
 		}
 
+		$breakpoint_config = $this->breakpoints[ $breakpoint ];
```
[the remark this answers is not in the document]
ronros-elementor: Good point, but lets settle on `breakpoint_id` instead of `breakpoint_name`, as this is its identifier 🙂 (and of course the `config` is dropped)

Omerisra6 (approved):

PR #28574, opened 2024-09-01
Internal: Remove `Atomic_Prop` class [EDS-408]

StyleShit:


modules/atomic-widgets/base/atomic-widget-base.php
```diff
@@ -136,33 +132,6 @@ public static function get_props_schema(): array {
...
-
-			if ( ! $prop_type ) {
-				Utils::safe_throw( "Prop type `{$prop->get_type()}` for prop `$key` does not exist in `{$widget_name}`." );
-			}
-
-			try {
-				$prop->validate( $prop->get_default() );
-			} catch ( \Exception $e ) {
-				Utils::safe_throw( "Default value for `$key` prop is invalid in `{$widget_name}` - {$e->getMessage()}" );
-			}
```
StyleShit, line 162: @Nevoss 
do you think we should keep this?

tests/phpunit/elementor/modules/atomic-widgets/test-atomic-widget-base.php
```diff
@@ -203,37 +195,39 @@ public function test_get_props_schema__is_serializable() {
...
+				]
 			},
 			"number_prop": {
 				"type": "number",
-				"constraints": [],
-				"default": 123
+				"default": 123,
+				"dynamic_categories": [
+					"number"
+				]
```
StyleShit, line 209: @Nevoss 

I don't like the fact that we check the "dynamic_categories" here
any idea how tp avoid that?

modules/atomic-widgets/prop-types/prop-type.php
```diff
@@ -20,9 +42,10 @@ public function get_dynamic_categories(): array {
...
 
 	public function jsonSerialize(): array {
-		return [
-			'type' => $this->get_type(),
+		return array_merge( [
+			'key' => $this->get_key(),
 			'dynamic_categories' => $this->get_dynamic_categories(),
-		];
+			'default' => $this->default,
+		], $this->settings );
```
StyleShit, line 49: the dynamic categories will now be passed to the editor on each prop in the schema, which means there will be duplicates. Do we care about it?

modules/atomic-widgets/base/atomic-widget-base.php
Nevoss: We should still validate the schema but without validating the default value, maybe?

modules/atomic-widgets/base/atomic-widget-base.php
```diff
@@ -26,9 +25,6 @@ public function get_atomic_controls() {
 		$controls = $this->define_atomic_controls();
 		$schema = static::get_props_schema();
 
-		// Validate the schema only in the Editor.
-		static::validate_schema( $schema );
```
Nevoss, line 30: I think we can keep it, but maybe make it simpler

modules/atomic-widgets/prop-types/prop-type.php
```diff
@@ -8,10 +8,32 @@
 
 abstract class Prop_Type implements \JsonSerializable {
 
-	abstract public function get_type(): string;
+	protected $default = null;
+
+	// TODO: Find a real name!
```
Nevoss, line 13: It is fine, remove 😄

modules/atomic-widgets/prop-types/prop-type.php
Nevoss: Can we do it differently? just pass some configuration on the "dynamic module" (we do not have a module for dynamic but we can create a class that is responsible for it) also the category should not be hardcoded in the prop type

modules/atomic-widgets/prop-types/string-prop-type.php
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	/**
+	 * @return array<string>
+	 */
+	public function get_dynamic_categories(): array {
+		return [
+			DynamicTags::TEXT_CATEGORY,
+		];
+	}
```
Nevoss, line 45: Remove

tests/phpunit/elementor/modules/atomic-widgets/test-atomic-widget-base.php
Nevoss: I feel it should live outside of the `Prop_Types` mechanism and be independent

Nevoss (changes requested):

modules/atomic-widgets/base/atomic-widget-base.php
StyleShit: I think validating the default is fine
added the validation back, maybe we'll remove it in the future

modules/atomic-widgets/prop-types/prop-type.php
StyleShit: Dםמק

modules/atomic-widgets/prop-types/prop-type.php
```diff
@@ -8,21 +8,35 @@
...
 
 	public function jsonSerialize(): array {
-		return [
-			'type' => $this->get_type(),
-			'dynamic_categories' => $this->get_dynamic_categories(),
-		];
+		return array_merge( [
+			'type' => static::get_key(),
+			'default' => $this->default,
+		], $this->settings );
```
StyleShit, line 40: put under settings key

modules/atomic-widgets/dynamic-tags.php
```diff
@@ -5,17 +5,22 @@
...
 	public function register_hooks() {
 		add_filter( 'elementor/editor/localize_settings', fn( $settings ) => $this->add_atomic_dynamic_tags_settings( $settings ) );
 	}
 
 	private function add_atomic_dynamic_tags_settings( $settings ) {
 		if ( isset( $settings['dynamicTags']['tags'] ) ) {
-			$settings['atomicDynamicTags'] = $this->convert_dynamic_tags_to_atomic( $settings['dynamicTags']['tags'] );
+			$settings['atomicDynamicTags'] = [
+				'tags' => $this->convert_dynamic_tags_to_atomic( $settings['dynamicTags']['tags'] ),
+				'prop_types_mapping' => $this->get_prop_types_mapping(),
```
StyleShit, line 22:
```suggestion
				'prop_types_to_dynamic' => $this->get_prop_types_mapping(),
```

Nevoss (approved):

Nevoss (approved):

PR #28606, opened 2024-09-04
Internal: Settings sanitization for atomic widgets [EDS-305]

ronros-elementor:


tests/phpunit/elementor/modules/atomic-widgets/test-atomic-widget-base.php
```diff
@@ -368,6 +368,65 @@ public function test_get_atomic_controls__schema_validation__throws_for_invalid_
...
+
+		// Assert.
+		$this->assertSame( [
+			'string_prop' => 'valid-string',
+			'number_prop' => 123,
+			'boolean_prop' => true,
+		], $sanitized_settings );
+	}
+
 	/**
```
Omerisra6, line 430: Maybe test the safe throw when an error is thrown from the sanitization?

Omerisra6 (changes requested):

Omerisra6 (approved):

PR #28685, opened 2024-09-12
Fix: Make sure the post-css are loaded after all core styles [ED-15769] (#28679)

KingYes:


core/files/css/post.php
```diff
@@ -37,6 +37,8 @@ class Post extends Base {
 	 */
 	private $post_id;
 
+	private $additional_style_dependencies = [];
```
louiswol94, line 40: Can you add a PHPDoc here?

louiswol94 (approved):

core/files/css/post.php
rami-elementor: no need, this is a cherry pick pr.

PR #28738, opened 2024-09-16
Internal: Add flag to re-generate CSS in WP CLI [ED-15763]

louiswol94:
`wp elementor flush-css --regenerate`

### DIFF
--- a/core/files/manager.php
+++ b/core/files/manager.php
@@ -206,4 +206,61 @@ class Manager {
 	private function reset_assets_data() {
 		delete_option( Page_Assets_Data_Manager::ASSETS_DATA_KEY );
 	}
+
+	/**
+	 * Generate CSS.
+	 *
+	 * Generates CSS for all posts build with Elementor.
+	 *
+	 * @since 3.25.0
+	 * @access public
+	 */
+	public function generate_css() {
+		$batch_size = apply_filters( 'elementor/core/files/generate_css/batch_size', 100 );
+		$processed_posts = 0;
+
+		while ( true ) {
+			$args = [
+				'post_type' => get_post_types(),
+				'posts_per_page' => $batch_size,
+				'meta_query' => [
+					[
+						'key' => Document_Base::BUILT_WITH_ELEMENTOR_META_KEY,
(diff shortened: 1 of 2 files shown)


core/files/manager.php
```diff
@@ -206,4 +206,60 @@ private function register_actions() {
...
+			$query = new \WP_Query( $args );
+
+			if ( empty( $query->posts ) ) {
+				break;
+			}
+
+			foreach ( $query->posts as $post_id ) {
+				$document = Plugin::$instance->documents->get_doc_for_frontend( $post_id );
+
+				if ( $document && $document->is_built_with_elementor() ) {
```
matipojo, line 245: It's unnecessary because you already do it in the `meta_query`
```suggestion
				if ( $document ) {
```

core/files/manager.php
```diff
@@ -206,4 +206,60 @@ private function register_actions() {
...
+					$css_file = Post_CSS::create( $post_id );
+					$css_file->update();
+				}
+			}
+
+			$processed_posts += $batch_size;
+
+			wp_reset_postdata();
```
matipojo, line 253: If you didn't call `the_post`, I think you don't need to reset postdata.

core/files/manager.php
```diff
@@ -206,4 +206,60 @@ private function register_actions() {
...
+		}
+
+		/**
+		 * Elementor Generate CSS files.
+		 *
+		 * Fires after Elementor generates new CSS files
+		 *
+		 * @since 3.250.0
```
matipojo, line 261:
```suggestion
		 * @since 3.25.0
```

core/files/manager.php
```diff
@@ -206,4 +206,60 @@ private function register_actions() {
...
+		 */
+		do_action( 'elementor/core/files/generate_css' );
```
matipojo, line 263: Maybe indicate that it's after the process is done.
```suggestion
		do_action( 'elementor/core/files/after_generate_css' );
```

matipojo (changes requested):

matipojo (approved):

PR #28802, opened 2024-09-23
Internal: Add linked dimensions transformer [EDS-427]

Omerisra6:


tests/phpunit/elementor/modules/atomic-widgets/style-transformers/test-linked-dimensions-transformer.php
```diff
@@ -0,0 +1,67 @@
...
+			'top' => '10px',
+			'right' => '20px',
+			'bottom' => '30px',
+			'left' => '40px',
+		], function( $value ) {
+			return $value;
+		} );
+
+		// Assert.
+		$this->assertSame( '10px 20px 30px 40px', $transformed_value );
```
ronros-elementor, line 31: If `isLinked === true` shouldn't all values be the same?
Though in theory I assume that if the "link" button is clicked - all values should be changes automatically from the UI, but maybe even if not but the value is true - enforce one of the values to be used for all edges?

Omerisra6: It is just a UI implementation, the linked property is meaningless to the transformer. The only reason it exists is that we need to preserve the state of the UI.

ronros-elementor (approved):

PR #28805, opened 2024-09-23
Fix: Elementor AI - "Reference a website" fails in Safari 17 [ED-15828]

matipojo:
Switched to work with a command from ai-integration.

Safari seems to block the direct `postMessage` from the top frame to the iframe.

The `ai-integration` command initiates communication from the iframe to the top frame, and it works in Safari as well.


DennisNerush (dismissed):
Can you explain more in detail in our slack channel about the solution and why it solves the issue?

DennisNerush (approved):

PR #28858, opened 2024-10-01
Fix: Background gradient color dynamic tag is not being rendered on the frontend [ED-15218]

louiswol94:


core/files/css/base.php
```diff
@@ -780,7 +787,9 @@ private function get_style_control_value( array $control, array $values ) {
 			return $this->get_selector_global_value( $control, $values['__globals__'][ $control['name'] ] );
 		}
 
-		$value = $values[ $control['name'] ];
+		$value = isset( $values[ $control['name'] ] )
+			? $values[ $control['name'] ]
+			: $this->preserved_dynamic_style_values[ $control['name'] ];
```
KingYes, line 792: @louiswol94 You should check isset on `$this->preserved_dynamic_style_values` too. No?

KingYes (changes requested):

core/files/css/base.php
KingYes: You're right, but still, how you can make sure the value is set there?

louiswol94: @KingYes Let me add null as a fallback? :)

KingYes: I think it's okay. Thanks.

louiswol94: @KingYes Updated.

KingYes (approved):

PR #28864, opened 2024-10-01
Internal: Fix - user with old pro gets "upgrade" promo [ED-15857]

mike-elementor:
Internal: Fix - user with an activated pro license still gets "upgrade to pro" banners [ED-15857]

 ## Description:

 Changed the logic so that:
 - user with an activated PRO 3.24 and prior will not see the step
 - user with no PRO (or deactivated pro) will see the promo
 - user with the latest PRO and CORE will see the Set up a header step (standard)

 ## Quality Assurance:

 - [ ] This code does not break any APIs
 - [ ] I have tested this code against all the relevant Core versions
 - [ ] I have tested this code against the relevant competitor plugins

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #

[ED-15857]: https://elementor.atlassian.net/browse/ED-15857?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/checklist/steps/setup-header.php
```diff
@@ -11,7 +11,7 @@
 }
 
 class Setup_Header extends Step_Base {
-	const STEP_ID = 'setup_header';
+	const STEP_ID = 'setup_header_promo';
```
mike-elementor, line 14: I do not like this, because we've implemented class-autoload as `new {namespace}\{step-id}` , but it's half of the story, we also know about this class in the PRO repository.
main idea behind the changes was to make them in such a way, that PRO repository doesn't have to be changed for all to work...

ronenboxer (dismissed):

mugurelLupu (dismissed):

modules/checklist/steps/setup-header.php
```diff
@@ -25,6 +26,14 @@ public function get_id() : string {
 		return self::STEP_ID;
 	}
 
+	public function is_visible() : bool {
```
louiswol94, line 29:
```suggestion
	public static function is_active() : bool {
```
Nitpick comment. But maybe consider calling it `is_active` to be consistent with other conditional loading throughout the code base.

Examples:
```
if ( $class_name::is_active() ) {
	$this->modules[ $module_name ] = $class_name::instance();
}
```

(Not a blocker).

elchugreeva (dismissed):

modules/checklist/steps/setup-header.php
mike-elementor: @louiswol94 the `static` for this method is more for a consistency sake, or do we have some other reasoning for this preference?

I don't mind changing the name...

though from what I see in the code-base, most of the time it's used in a context of a module, that can be "activated" / "deactivated", not sure our steps must follow in this case.

but again, when this goes public, it's public, so we need to consider wisely.. 

@TzviRabinovitch how do you feel about that? should we do it?

TzviRabinovitch: we do not need a static function here
I think is_visible is more relevant here 
a module may be active or not 
a step is hidden or not

louiswol94 (dismissed):

louiswol94 (approved):

ronenboxer (approved):

PR #28886, opened 2024-10-06
Internal: Update the Checklist experiment to be active only for new website [ED-15868]

elchugreeva:


tests/playwright/sanity/modules/checklist/checklist.test.ts
```diff
@@ -17,7 +17,6 @@ test.describe( 'Launchpad checklist tests', () => {
 
 		await wpAdmin.setExperiments( {
 			editor_v2: true,
-			'launchpad-checklist': true,
```
davseve, line 20: I think that `editor_v2` is active by default so maybe we can remove the `setExperiments` function.
As well we can maybe remove the `wpAdmin.resetExperiments();` function as well.
WDYT?

davseve (approved):

louiswol94 (approved):

PR #28893, opened 2024-10-07
Internal: Validate styles on save [EDS-523]

Omerisra6:


modules/atomic-widgets/styles/style-schema.php
```diff
@@ -0,0 +1,81 @@
...
+				100,
+				200,
+				300,
+				400,
+				500,
+				600,
+				700,
+				800,
+				900,
+			]),
```
Nevoss, line 48:
```suggestion
			'font-weight' => String::make()->enum([
				100,
				200,
				300,
				400,
				500,
				600,
				700,
				800,
				900,
			]),
```

Font weight can be also `normal`, `bold`, `lighther`. For now we are not supporting it, but, we should be prepared for it, so it be better to set it as string.

modules/atomic-widgets/prop-types/number-prop-type.php
```diff
@@ -16,5 +19,43 @@ public function validate( $value ): void {
...
+				->implode( ', ' );
+
+			throw new \Exception( "`$value` is not in the list of allowed values ($values)." );
+		}
+	}
+
+	private function all_are_integers( array $allowed_values ): bool {
+		return array_reduce(
+			$allowed_values,
+			fn ( $carry, $item ) => $carry && is_numeric( $item ),
```
Nevoss, line 57: Remove? lets talk about it if you wish

modules/atomic-widgets/prop-types/size-prop-type.php
```diff
@@ -0,0 +1,37 @@
...
+			throw new \Exception( 'Value must be an array, ' . gettype( $value ) . ' given.' );
+		}
+
+		if ( ! isset( $value['size'] ) || ! isset( $value['unit'] ) ) {
+			throw new \Exception( 'Value must include size and unit' );
+		}
+
+		if ( array_search( $value['unit'], $this->settings['units'], true ) === false ) {
+			throw new \Exception( 'unit is not valid' );
+		}
```
Nevoss, line 35:
```suggestion
if ( ! is_array( $value ) ) {
	throw new \Exception( 'Value must be an array, ' . gettype( $value ) . ' given.' );
}
if ( ! isset( $value['size'], $value['unit'] ) ) {
	throw new \Exception( 'Value must include size and unit' );
}

// One option:

if ( ! is_numeric( $value['size'] ) ) {
	throw new \Exception( 'size must be a number' );
}

if ( ! in_array( $value['unit'], $this->settings['units'], true ) ) {
	throw new \Exception( 'unit is not valid' );
}

// Another option:
$this->internl_types['size']->validate_with_additionals( $value['size'] );
$this->internl_types['unit']->validate_with_additionals( $value['unit'] );

```

modules/atomic-widgets/styles/style-schema.php
```diff
@@ -0,0 +1,81 @@
...
+			'font-size' => Size_Prop_Type::make(),
+			'color' => String_Prop_Type::make(),
```
Nevoss, line 50: Lets create `Color_Prop_Type`

modules/atomic-widgets/styles/style-schema.php
```diff
@@ -0,0 +1,81 @@
...
+				'right',
+				'justify',
+			]),
+			'font-style' => String_Prop_Type::make()->enum([
+				'normal',
+				'italic',
+				'oblique',
+			]),
+			// TODO: validate text-decoration in more specific way [EDS-524]
+			'text-decoration' => String_Prop_Type::make(),
```
Nevoss, line 65: For now it is just string right? `underline`, `wavy`? maybe for now it will be `enum`?

modules/atomic-widgets/validators/styles-validator.php
```diff
@@ -0,0 +1,91 @@
...
+	 */
+	public function validate( array $styles ): array {
+		$errors = [];
+
+		foreach ( $styles as $style_id => $style ) {
+			try {
+				$this->validate_id( $style['id'] );
+			} catch ( \Exception $e ) {
+				$errors[] = 'id';
+			}
```
Nevoss, line 40: can I even save style if the ID is not valid? maybe it just need to throws without catch?

modules/atomic-widgets/validators/styles-validator.php
```diff
@@ -0,0 +1,91 @@
...
+
+			foreach ( $style['variants'] as $variant_index => $variant ) {
+				try {
+					$this->validate_meta( $variant['meta'] );
+				} catch ( \Exception $e ) {
+					$errors[] = 'meta';
+				}
+
+				[,$validated_props, $variant_errors] = Props_Validator::make( $this->schema )->validate( $variant['props'] );
+				$styles[ $style_id ]['variants'][ $variant_index ]['props'] = $validated_props;
```
Nevoss, line 50: It will be much clearer if you will update new variable and not change the reference

modules/atomic-widgets/validators/styles-validator.php
```diff
@@ -0,0 +1,91 @@
...
+	 *     1: array<string, mixed>,
+	 *     2: array<string>
+	 * }
+	 */
+	public function validate( array $styles ): array {
+		$errors = [];
+
+		foreach ( $styles as $style_id => $style ) {
+			try {
+				$this->validate_id( $style['id'] );
```
Nevoss, line 37:
```suggestion
				$this->validate_id( $style['id'] );
```

isset? maybe just check that this is exists and it is a string

modules/atomic-widgets/validators/styles-validator.php
```diff
@@ -0,0 +1,91 @@
...
+		}
+	}
+
+	public function validate_meta( array $meta ) {
+		$valid_states = [
+			'hover',
+			'active',
+			'focus',
+			null,
+		];
```
Nevoss, line 77: Put it const up in the class

```
const VALID_STATES = [...]
```

modules/atomic-widgets/validators/styles-validator.php
```diff
@@ -0,0 +1,91 @@
...
+
+		if ( ! is_array( $meta ) ) {
+			throw new \Exception( 'Meta should be an array' );
+		}
+
+		if ( ( ! is_null( $meta['state'] ) && ! isset( $meta['state'] ) ) || ! in_array( $meta['state'], $valid_states, true ) ) {
```
Nevoss, line 83:
```suggestion
		if ( ! array_key_exists('state', $meta) || ! in_array( $meta['state'], $valid_states, true ) ) {
```

modules/atomic-widgets/validators/styles-validator.php
```diff
@@ -0,0 +1,91 @@
...
+			throw new \Exception( 'Meta state is required and should be one of: ' . implode( ', ', $valid_states ) );
+		}
+
+		if ( ! isset( $meta['breakpoint'] ) || ! is_string( $meta['breakpoint'] ) ) {
```
Nevoss, line 87: Can we validate the breakpoints?

Nevoss (changes requested):

modules/atomic-widgets/styles/style-schema.php
Omerisra6: It's saved as a string and not as an array, for example: 'underline overline'

modules/atomic-widgets/validators/styles-validator.php
Omerisra6: It won't save, the errors are shown when trying to save

modules/atomic-widgets/styles/style-schema.php
Omerisra6: color could be any string (named colors), I don't think we should validate it...

modules/atomic-widgets/validators/styles-validator.php
Omerisra6: I will open a task for it

modules/atomic-widgets/prop-types/size-prop-type.php
```diff
@@ -0,0 +1,35 @@
...
+		$this->internal_types['size'] = Number_Prop_Type::make();
+		$this->internal_types['unit'] = String_Prop_Type::make()->enum( $this->default_units );
+	}
+
+	public static function get_key(): string {
+		return 'size';
+	}
+
+	public function units( array $units ): self {
+		$this->internal_types['unit'] = $this->internal_types['unit']->enum( $units );
```
Nevoss, line 22:
```suggestion
		$this->internal_types['unit']->enum( $units );
```

it should work, it is a change by reference

modules/atomic-widgets/styles/transformers/color-transformer.php
```diff
@@ -0,0 +1,25 @@
...
+
+namespace Elementor\Modules\AtomicWidgets\Styles\Transformers;
+
+use Elementor\Modules\AtomicWidgets\Base\Style_Transformer_Base;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+class Color_Transformer extends Style_Transformer_Base {
```
Nevoss, line 11: You dont need this transformer, you can bind the type to `Primitive_Transformer`

modules/atomic-widgets/validators/styles-validator.php
```diff
@@ -0,0 +1,85 @@
...
+	 *     1: array<string, mixed>,
+	 *     2: array<string>
+	 * }
+	 */
+	public function validate( array $styles ): array {
+		foreach ( $styles as $style_id => $style ) {
+			if ( ! isset( $style['id'] ) || ! is_string( $style['id'] ) ) {
+				$this->errors_bag[] = 'id';
+				continue;
+			}
```
Nevoss, line 45: If it failed but I still have access to the styles in the return value, why did we validated from the first place? lets talk about it

Nevoss (changes requested):

modules/atomic-widgets/styles/transformers/color-transformer.php
Omerisra6: how? Don't forget we are still using the old style transformers system.

Nevoss: Right sorry

Nevoss (approved):

PR #28898, opened 2024-10-07
Tweak: Script Module loading for Nested Tabs / POC [ED-15863]

hein-obox:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #


assets/dev/js/frontend/handlers/accessibility/nested-title-keyboard-handler.js
```diff
@@ -1,6 +1,4 @@
-import Base from '../base';
```
hein-obox, line 1: The module approach, requires us to load all the imports as script modules as well.

assets/dev/js/frontend/handlers/accessibility/nested-title-keyboard-handler.js
```diff
@@ -214,3 +212,5 @@ export default class NestedTitleKeyboardHandler extends Base {
 
 	handleContentElementTabEvents() {}
 }
+
+window.NestedTitleKeyboardHandler = NestedTitleKeyboardHandler;
```
hein-obox, line 216: This is a recommended approach to load script modules on the frontend. Probably we should introduce some kind of unique name for these handlers. 

```suggestion
window.elementorFrontend.NestedTitleKeyboardHandler = NestedTitleKeyboardHandler;
```

includes/wordpress/class-wp-script-modules.php
```diff
@@ -0,0 +1,359 @@
+<?php
```
hein-obox, line 1: This class was only introduced in WordPress 6.5. I have loaded it conditionally for users with older versions.

modules/nested-tabs/assets/js/frontend/handlers/nested-tabs.js
```diff
@@ -464,3 +452,5 @@ export default class NestedTabs extends Base {
 		};
 	}
 }
+
+window.elementorModules.frontend.handlers.NestedTabs = NestedTabs;
```
hein-obox, line 456: For now this will cause problems for the Mega Menu.

We will need to create an experiment for the Mega Menu to conditionally load the old or new NestedTabs module for BC reasons (we can remove this experiment after 2 versions).

modules/nested-tabs/module.php
```diff
@@ -49,4 +56,32 @@ public function register_styles() {
...
 		);
 	}
+
+	/**
+	 * Register script modules.
+	 *
+	 * @return void
+	 */
+	public function register_script_modules() {
+		$this->wordpress_adapter->wp_register_script_module(
```
hein-obox, line 66: The `wp_register_script_module()` method requires us to register and enqueue the module files directly.

core/base/document.php
```diff
@@ -1809,9 +1817,12 @@ protected function print_elements( $elements_data ) {
 			$scripts_to_queue = [];
 			$styles_to_queue = [];
 
-			global $wp_scripts, $wp_styles;
+			global $wp_scripts, $wp_script_modules, $wp_styles;
+
+			$printed_script_modules = $this->cache_script_module( $wp_script_modules );
```
hein-obox, line 1822: I need to update the class, to support module caching.

We have a challenge here, the new script module method doesn't work 100% in the same way as the script and style enqueue methods.

For the moment I am caching the printed version of the script module tag.

includes/wordpress/class-wp-script-modules.php
hein-obox: An alternative would be to enqueue the JS files as normal scripts, and to use a filter to add the type="module" reference. Possibly this implementation is a bit more straight forward.

See https://stackoverflow.com/questions/73967285/how-do-i-make-my-script-type-module-with-wordpress

Wdyt?

assets/dev/js/frontend/elements-handlers-manager.js
```diff
@@ -26,8 +26,8 @@ module.exports = function( $ ) {
 		'wp-widget-media_audio.default': () => import( /* webpackChunkName: 'wp-audio' */ './handlers/wp-audio' ),
 	};
 
-	if ( elementorFrontendConfig.experimentalFeatures[ 'nested-elements' ] ) {
-		this.elementsHandlers[ 'nested-tabs.default' ] = () => import( /* webpackChunkName: 'nested-tabs' */ 'elementor/modules/nested-tabs/assets/js/frontend/handlers/nested-tabs' );
+	if ( elementorModules?.frontend?.handlers?.NestedTabs ) {
```
KingYes, line 29: If you move this to the own module, you should move the register too.

PR #28909, opened 2024-10-08
Internal: Hide checklist if kit is installed [ED-15451]

ronenboxer:


modules/checklist/module.php
```diff
@@ -24,23 +24,30 @@ class Module extends BaseModule implements Checklist_Module_Interface {
...
 	 */
-	public function __construct( ?Wordpress_Adapter_Interface $wordpress_adapter = null, ?Kit_Adapter_Interface $kit_adapter = null ) {
+	public function __construct(
+		?Wordpress_Adapter_Interface $wordpress_adapter = null,
+		?Kit_Adapter_Interface $kit_adapter = null,
+		?Elementor_Counter_Adapter_Interface $counter_adapter = null
+	) {
 		$this->wordpress_adapter = $wordpress_adapter ?? new Wordpress_Adapter();
 		$this->kit_adapter = $kit_adapter ?? new Kit_Adapter();
+		$this->counter_adapter = $counter_adapter ?? Elementor_Counter::instance();
```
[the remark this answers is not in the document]

modules/elementor-counter/module.php
```diff
@@ -0,0 +1,71 @@
...
+
+	private ?Wordpress_Adapter_Interface $wordpress_adapter = null;
+
+	public function get_name() {
+		return 'elementor-counter';
+	}
+
+	public function __construct( ?Wordpress_Adapter_Interface $wordpress_adapter = null ) {
+		parent::__construct();
+		$this->wordpress_adapter = $wordpress_adapter ?? new Wordpress_Adapter();
```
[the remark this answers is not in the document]
ronenelementor: Done

modules/checklist/assets/js/commands/toggle-popup.js
```diff
@@ -15,9 +15,9 @@ export class TogglePopup extends $e.modules.CommandBase {
 			this.mount();
 		} else {
 			this.unmount();
-			updateUserProgress( { [ USER_PROGRESS.EDITOR_VISIT_COUNT ]: -1 } );
 		}
 
+		updateUserProgress( { [ USER_PROGRESS.LAST_OPENED_TIMESTAMP ]: TogglePopup.isOpen ? Date.now() : null } );
```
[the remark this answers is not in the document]
ronenelementor: Done

davseve (dismissed):

modules/elementor-counter/module.php
```diff
@@ -0,0 +1,76 @@
...
+use Elementor\Core\Isolation\Elementor_Counter_Adapter_Interface;
+use Elementor\Core\Isolation\Wordpress_Adapter;
+use Elementor\Core\Isolation\Wordpress_Adapter_Interface;
+use Elementor\Core\Base\Module as BaseModule;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+class Module extends BaseModule implements Elementor_Counter_Adapter_Interface {
```
louiswol94, line 14: I am wondering if it makes sense to have this as a 'module', as it's not a feature. It could be split into a 'component' of the checklist?

core/isolation/kit-adapter.php
```diff
@@ -12,4 +13,14 @@ public function get_kit_settings() {
...
 	}
+
+	public function is_active_kit_default() : bool {
+		$kit_id = Plugin::$instance->kits_manager->get_active_id();
+
+		if ( false === $kit_id || null === $kit_id ) {
+			return false;
+		}
+
+		return esc_html__( 'Default Kit', 'elementor' ) === get_post( $kit_id )->post_title;
```
louiswol94, line 24: Should we make sure `post_title` exists?

louiswol94: Are you not going to add active_kit === previous_kit?

core/settings/editor-preferences/model.php
```diff
@@ -185,7 +185,7 @@ protected function register_controls() {
 					'type' => Controls_Manager::SWITCHER,
 					'label_on' => esc_html__( 'Show', 'elementor' ),
 					'label_off' => esc_html__( 'Hide', 'elementor' ),
-					'default' => ChecklistModule::is_preference_switch_on() ? 'yes' : '',
+					'default' => Plugin::$instance->modules_manager->get_modules( 'checklist' )->is_preference_switch_on() ? 'yes' : '',
```
louiswol94, line 188: Why did we need this change? I actually think the previous version is less risky haha

core/utils/constants.php
```diff
@@ -1,16 +0,0 @@
-<?php
```
louiswol94, line 1: I wonder why we added this and never used it? 🤔

modules/elementor-counter/module.php
```diff
@@ -0,0 +1,76 @@
...
+		return 'elementor-counter';
+	}
+
+	public function __construct( ?Wordpress_Adapter_Interface $wordpress_adapter = null ) {
+		parent::__construct();
+		$this->wordpress_adapter = $wordpress_adapter ?? new Wordpress_Adapter();
+
+		if ( self::$should_count_editor ) {
+			add_action( 'elementor/editor/init', function () {
+				$this->increment( self::EDITOR_COUNTER_KEY );
```
louiswol94, line 30: Does your other calls to `elementor/editor/init` have an execution order dependancy? Like, does set_user_preferences() have to run before increment()? If so, maybe add priority?

tests/phpunit/elementor/modules/checklist/classes/test-kit-installation.php
```diff
@@ -0,0 +1,56 @@
...
+			'get_user_preferences' => [ $this, 'get_user_preferences' ],
+		] );
+
+		$this->mock_editor_visit();
+
+		$this->assertTrue( $this->checklist_module->is_preference_switch_on() );
+	}
+
+	public function test__custom_kit_installed() {
+		parent::setup();
```
louiswol94, line 28: Why do you run `parent::setup();` in every method? Can you not have a setup method in this class?

modules/elementor-counter/module.php
ronenelementor: Tzvi said that we should change it to be more global than the checklist because by definition the editor counter is something that we should count regardless of it. We might want to later expand that counter to other areas

core/isolation/wordpress-adapter.php
```diff
@@ -51,6 +55,22 @@ public function add_option( $option_key, $option_value ) : void {
 		add_option( $option_key, $option_value );
 	}
 
+	public function get_user_preferences( $preference_key ) {
```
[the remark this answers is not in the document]
ronenelementor: We can change the kit adapter to be elementor adapter and put it there, WDYT?

core/isolation/kit-adapter.php
ronenelementor: We checked this and it's is redundant. Could we have missed something?

core/utils/constants.php
ronenelementor: Changes were made and this wasn't removed...

modules/elementor-counter/module.php
ronenelementor: Good!!!

core/settings/editor-preferences/model.php
ronenelementor: We needed the wordpress adapters inside the `is_preference_switch_on` function and this needed the instance
`public function is_preference_switch_on() : bool {
		if ( $this->should_switch_preferences_off() ) {
			return false;
		}

		$user_preferences = $this->wordpress_adapter->get_user_preferences( self::VISIBILITY_SWITCH_ID );

		return 'yes' === $user_preferences || $this->wordpress_adapter->is_new_installation();
	}`

tests/phpunit/elementor/modules/checklist/classes/test-kit-installation.php
ronenelementor: I want that executed before each test
Does the `setup` method run before each test?

davseve (approved):

PR #28913, opened 2024-10-08
Internal: Implement DI container in the checklist [ED-15890]

davseve:


core/utils/isolation-manager.php
```diff
@@ -0,0 +1,19 @@
...
+	exit; // Exit if accessed directly.
+}
+
+class Isolation_Manager {
+	/**
+	 * @param string $adapter_name
+	 */
+	public static function get_adapter( string $adapter_name ) {
+		$container = PLugin::$instance->elementor_container();
+		return $container->get( $adapter_name );
```
ronenelementor, line 17: I would wrap it with `try/catch` to avoid errors, and return null as a fallback. WDYT?

ronenelementor: This can also be an abstract class, could't it?

core/utils/isolation-manager.php
```diff
@@ -0,0 +1,19 @@
...
+	}
```
ronenelementor, line 18: Since we're using the WP adapter and the Kit adapter a lot, I would create a getter for them. Something like this perhaps? And then you will be able to use that anywhere without additional imports and long lines
```suggestion
		public static function get_wordpress_adatper() : Wordpress_Adapter_Interface {
			return self::get_adapter( Wordpress_Adapter::class );
		}
	}
```

Batsirai-muchareva (approved):

ronenboxer (approved):

PR #28956, opened 2024-10-13
Internal: Fix Dependency Injection template file [ED-15947]

davseve:


php-scoper/dependency-injection.inc.php
```diff
@@ -14,4 +14,20 @@
 	'exclude-files' => [
```
Batsirai-muchareva, line 14: Can you remove this

Batsirai-muchareva (dismissed):

Batsirai-muchareva (approved):

PR #28962, opened 2024-10-14
Internal: Add multi prop support [EDS-477]

Omerisra6:


modules/atomic-widgets/props-resolver/transformer-base.php
```diff
@@ -6,5 +6,13 @@
 }
 
 abstract class Transformer_Base {
-	abstract public function transform( $value );
+	abstract public function transform( $value, $key );
+
+	protected function multi( $value ) {
```
Nevoss, line 11:
```suggestion
	protected function multi_props( $value ) {
```

modules/atomic-widgets/props-resolver/transformer-base.php
```diff
@@ -6,5 +6,13 @@
...
+		return [
+			'$$multi' => true,
```
Nevoss, line 13:
```suggestion
			'$$multi_props' => true,
```

modules/atomic-widgets/props-resolver/transformer-base.php
Nevoss, line 11: Remove from the base transformer and create it inside of some kind of util `Multi_Props` then it can have:
`generate`: create the asoc array
`is`: check if the value is multi

modules/atomic-widgets/props-resolver/props-resolver.php
```diff
@@ -127,6 +144,14 @@ private function is_transformable( $value ): bool {
 		);
 	}
 
+	private function is_multi( $value ): bool {
+		return (
+			! empty( $value['$$multi'] ) &&
+			true === $value['$$multi'] &&
+			array_key_exists( 'value', $value )
+		);
+	}
```
Nevoss, line 153: Remove and move to some kind of class util

modules/atomic-widgets/props-resolver/props-resolver.php
```diff
@@ -78,13 +78,25 @@ public function resolve( array $schema, array $props ): array {
...
+			$transformed = $this->transform( $value, $key );
+
+			if ( $this->is_multi( $transformed ) ) {
+				unset( $transformed_props[ $key ] );
+
+				$transformed_props = array_merge( $transformed_props, $transformed['value'] );
+				continue;
+			}
+
+			$transformed_props[ $key ] = $transformed;
```
Nevoss, line 93: You can do something like that:
```php
	private function assign_values( $values ) {
		$resolved = [];

		foreach ( $values as $key => $value ) {
			if ( Multi_Props::is( $value ) ) {
				foreach ( Multi_Props::extract( $value ) as $k => $v ) {
					$resolved[ $k ] = $v;
				}

				continue;
			}

			$resolved[ $key ] = $value;
		}

		return $resolved;
	}
```

then also use it inside the `transform` function

Nevoss (changes requested):

modules/atomic-widgets/props-resolver/props-resolver.php
Omerisra6: I think it makes things complicated.

Nevoss (approved):

PR #28999, opened 2024-10-16
Tweak: Optimized Dom [ED-15929]

louiswol94:


core/experiments/manager.php
```diff
@@ -441,7 +441,7 @@ private function add_default_features() {
 			'tag' => esc_html__( 'Performance', 'elementor' ),
 			'description' => esc_html__( 'Reduce the DOM size by eliminating HTML tags in various elements and widgets. This experiment includes markup changes so it might require updating custom CSS/JS code and cause compatibility issues with third party plugins.', 'elementor' ),
 			'release_status' => self::RELEASE_STATUS_ALPHA,
-			'default' => self::STATE_INACTIVE,
+			'default' => self::STATE_ACTIVE,
```
rami-elementor, line 444:
```suggestion
			'default' => self::STATE_INACTIVE,
			'new_site' => [
				'default_active' => true,
				'minimum_installation_version' => '3.26.0',
			],
```

louiswol94: @rami-elementor This is not real, I just did this to see what breaks :) But I will update it if this is the requirement, thanks!

core/kits/documents/tabs/settings-layout.php
```diff
@@ -100,6 +100,16 @@ protected function register_tab_controls() {
...
 
+		$optimized_markup = Plugin::instance()->experiments->is_feature_active( 'e_optimized_markup' );
+
+		$widget_spacing = $optimized_markup
+			? 'content: ""; display:block; height: {{ROW}}{{UNIT}}'
+			: 'margin-block-end: {{ROW}}{{UNIT}}';
+
+		$widget_spacing_selector = $optimized_markup
+			? '.elementor-widget:not(:last-child)::after'
+			: '.elementor-widget:not(:last-child)';
```
louiswol94, line 111: @hein-obox @rami-elementor I found an issue with `Sections`. The global widget spacing was using `margin` on the main wrapper, and the margin from the Advanced tab on the inner div. But now that the Advanced tab will apply it on the main wrapper as well, the two margins are conflicting. I saw the layout shifts in the plugins-tester. I think I fixed it with an `::after` pseudo element and giving it a `height`. WDYT? If we can't use this, please help with other suggestions, or the Optimized DOM would need the Container as a dependency. Everything works fine in the Container, because it uses the `gap` property and not margin.

hein-obox: @louiswol94 
I understand the complexity. 

I am thinking could we leave the sections as is? In that way users could be motivated more to move from Sections to Containers if they value performance. That could be better approach, then to implement potentially risky styling to the Sections.

Wdyt?

hein-obox (approved):

louiswol94:
@rami-elementor I merged this one so long, which is basically just the name change. We'll talk about the other issues offline.

PR #29090, opened 2024-10-28
Internal: Check that the post type is correct when tracking clicks on floating elements [ED-15702]

nicoladj77:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

* When updating meta for click tracking, check that the post we are updating is a floating element

## Description

*

## Test instructions

*

## Quality assurance

Fixes #


modules/floating-buttons/module.php
```diff
@@ -287,6 +287,10 @@ public function handle_click_tracking() {
 		}
 
 		foreach ( $posts_to_update as $post_id => $clicks ) {
+			if ( self::CPT_FLOATING_BUTTONS !== get_post_type( $post_id ) ) {
```
louiswol94, line 290: @nicoladj77 Can you please explain this? It's been a while since we talked about it. AFAIU we said we only allow processing a post with post status set to `publish`?

nicoladj77: @louiswol94 so the discussion was all about restricting the ability to set meta values on wrong posts. The gold standard would be to actually refactor click tracking so that each button has a nonce created with it's own post id and that nonce is validated when adding clicks, in this way it would not be possible to do anything different from clicking on the button, that is more complex though, so I at least restricted the ability to set arbitrary meta data only on floating elements. Given that the only thing that can be done is set a number in the meta for a post, something which is not exploitable, I think this is a good trade off, but let me know what you think.

louiswol94: @nicoladj77 Can you send a build to patchstack and ask for confirmation if the issue is patched?

nicoladj77: @louiswol94 ok sending it, I also added a check that the post is published.

louiswol94 (dismissed):

louiswol94 (approved):

PR #29101, opened 2024-10-29
Internal: Add setup command for mac [EDS-558]

blackcolt:


yotamselementor (approved):

package.json
```diff
@@ -9,6 +9,7 @@
 	"author": "Elementor Team",
 	"homepage": "https://elementor.com/",
 	"scripts": {
+		"test:setup:playwright:mac": "npm run build && npm run start-local-server && npm run test:setup:playwright && npx playwright install",
```
yotamselementor, line 12: `playwright install` -> `playwright install chromium`

```suggestion
		"test:setup:playwright:mac": "npm run build && npm run start-local-server && npm run test:setup:playwright && npx playwright install chromium",
```

yotamselementor (changes requested):

package.json
blackcolt: Do we work only with chrome?

yotamselementor (approved):

PR #29182, opened 2024-11-04
Internal: Add global classes experiment [EDS-571]

Omerisra6:


modules/editor-global-classes/module.php
```diff
@@ -0,0 +1,51 @@
...
+			add_filter( 'elementor/editor/v2/packages', fn( $packages ) => $this->add_packages( $packages ) );
+		}
+	}
+
+	private function register_experiment() {
+		Plugin::$instance->experiments->add_feature( [
+			'name' => self::EXPERIMENT_NAME,
+			'title' => esc_html__( 'Global Classes', 'elementor' ),
+			'description' => esc_html__( 'Enable global CSS classes.', 'elementor' ),
+			'hidden' => true,
```
Nevoss, line 42: Add TODO that when this feature will not be hidden anymore, the code should be changed and we should be depended on `atomic widgets`

core/modules-manager.php
```diff
@@ -123,6 +123,7 @@ public function get_modules_names() {
 			'content-sanitizer',
 			'editor-events',
 			'atomic-widgets',
+			'editor-global-classes',
```
Nevoss, line 126:
```suggestion
			'global-classes',
```
 
why did we add the `editor` prefix?

Nevoss (changes requested):

core/modules-manager.php
Omerisra6: Do you mean here or also in the class name?

Nevoss (approved):

PR #29202, opened 2024-11-06
Internal: Allow Filtering Common Controls [ED-16100]

nuritsha:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #


ramiy (commented):
not sure, but if you can change the value, shouldn't we use 'let'.

nuritsha:
> not sure, but if you can change the value, shouldn't we use 'let'.

@ramiy `const` should work here, the value can be changed by the filter, but the `commonControls` variable is initialized only after the filter returns. I don't think we want to allow modifying this value after the filter.

assets/dev/js/editor/editor-base.js
```diff
@@ -287,7 +287,8 @@ export default class EditorBase extends Marionette.Application {
 			}
 
 			if ( ! this.widgetsCache[ widgetType ].commonMerged && ! this.widgetsCache[ widgetType ].atomic_controls ) {
-				jQuery.extend( this.widgetsCache[ widgetType ].controls, this.widgetsCache.common.controls );
+				const commonControls = elementor.hooks.applyFilters( 'elements/widget/controls/common/default', this.widgetsCache.common.controls, widgetType );
```
rami-elementor, line 290:
```suggestion
				let commonControls = this.widgetsCache.common.controls;

				/**
				 * Filter widgets common controls.
				 *
				 * @param array  commonControls - An array of the default common controls.
				 * @param string widgetType     - The widget type.
				 */
				commonControls = elementor.hooks.applyFilters( 'elements/widget/controls/common/default', commonControls, widgetType );
```

louiswol94, line 290: @nuritsha , I would be interested to see the use case for this :) But anyway, I think this filter might be too early. In the "Optimized Markup" experiment, we introduced a new Advanced tab `common-optimized`, so you might need to add the filter after that. Maybe just double check?

nuritsha: @louiswol94 I don't see any reference to that experiment in the js code, can you please direct me to the relevant place?

louiswol94: @nuritsha The experiment name is `e_optimized_markup`. You will see on [line 302](https://github.com/elementor/elementor/blob/e321132bc41340a2a4af4012aabaa8ec85e6afdc/assets/dev/js/editor/editor-base.js#L302) we merge those controls in as well. So I am thinking the filter should only be added after that?

nuritsha: yep, u r right

louiswol94 (approved):

PR #29204, opened 2024-11-06
Internal: Create repository for `global-classes` [EDS-572]

Omerisra6:


modules/global-classes/repository.php
```diff
@@ -0,0 +1,79 @@
...
+			throw new \Exception( "Global class with id ${id} not found" );
+		}
+
+		$this->kit->update_json_meta( self::META_KEY, [
+			'data' => $all->get_data()->except( [ $id ] )->all(),
+			'order' => $all->get_order()->filter( fn( $item ) => $item !== $id )->all(),
+		] );
+	}
+
+	public function patch( string $id, array $updated ) {
```
Nevoss, line 42: What do we do, if the `$updated` is not valid data?

modules/global-classes/repository.php
```diff
@@ -0,0 +1,79 @@
...
+
+		$this->kit->update_json_meta( self::META_KEY, [
+			'data' => $all->get_data()->merge( [ $id => $updated ] )->all(),
+			'order' => $all->get_order()->all(),
+		] );
+
+		return true;
+	}
+
+	public function create( array $class ) {
```
Nevoss, line 57: same here

Nevoss (changes requested):

modules/global-classes/repository.php
Omerisra6: Shouldn't it be handled at the API part?

modules/atomic-widgets/styles/utils.php
```diff
@@ -0,0 +1,17 @@
...
+
+class Utils {
+	public static function generate_id( string $prefix = '', $existing_ids = [] ): string {
+		$existing_ids = array_flip( $existing_ids );
+
+		$id = $prefix . wp_rand( 1000000, 9999999 );
+
+		while ( isset( $existing_ids[ $id ] ) ) {
+			$id = $prefix . wp_rand( 1000000, 9999999 );
+		}
```
Nevoss, line 13: use do while

```suggestion
		$id;

		do () {
			$id = $prefix . wp_rand( 1000000, 9999999 );
		} while ( in_array( $id, $existing_ids ) )
```

modules/global-classes/repository.php
```diff
@@ -0,0 +1,105 @@
...
+		}
+
+		return $this->all()->get_order()->all();
+	}
+
+	private function generate_global_class_id() {
+		$existing_ids = array_keys( $this->all()->get_items()->all() );
+		$kit_id = $this->kit->get_id();
+
+		return Atomic_Styles_Utils::generate_id( 'g-' . $kit_id . '-', $existing_ids );
```
Nevoss, line 103:
```suggestion
		return Atomic_Styles_Utils::generate_id( 'g-' . $kit_id . '-', array_keys( $existing_ids ) );
```

Nevoss (changes requested):

modules/global-classes/repository.php
Omerisra6: why?

modules/atomic-widgets/styles/utils.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace Elementor\Modules\AtomicWidgets\Styles;
+
+class Utils {
+	public static function generate_id( string $prefix = '', $existing_ids = [] ): string {
+		$existing_ids = array_flip( $existing_ids );
```
Nevoss, line 7: Why there is a need for flip? do you pass an array or ids? `['id-1', 'id-2', 'id-3']` if yes (it seems like you pass it the right way) there is no need to flip

Nevoss (changes requested):

Nevoss (approved):

PR #29254, opened 2024-11-11
Fix: Global widgets fatal error after permanent deletion [ED-16130]

ronenboxer:


davseve (dismissed):

louiswol94 (dismissed):

core/isolation/elementor-adapter.php
```diff
@@ -42,6 +42,10 @@ public function is_key_allowed( $key ): bool {
 	}
 
 	public function get_template_type( $template_id ): string {
-		return Plugin::$instance->documents->get( $template_id )->get_template_type();
+		$document = Plugin::$instance->documents->get( $template_id );
```
louiswol94, line 45:
```suggestion
		return Source_local::get_template_type( $template_id );
```
Maybe? You can decide.

louiswol94 (approved):

PR #29260, opened 2024-11-12
Internal: Create REST API for `global-classes` [EDS-573]

Omerisra6:


modules/atomic-widgets/validators/styles-validator.php
```diff
@@ -38,49 +36,81 @@ public static function make( array $schema ): self {
 	 * }
 	 */
 	public function validate( array $styles ): array {
+		$errors_bag = [];
 		foreach ( $styles as $style_id => $style ) {
-			if ( ! isset( $style['id'] ) || ! is_string( $style['id'] ) ) {
-				$this->errors_bag[] = 'id';
+			if ( $style_id !== $style['id'] ) {
```
Nevoss, line 41: What if there is not `id` in `$style` var, and what if this is not a `string` key and it is a `number`?

modules/atomic-widgets/validators/styles-validator.php
```diff
@@ -38,49 +36,81 @@ public static function make( array $schema ): self {
 	 * }
 	 */
 	public function validate( array $styles ): array {
+		$errors_bag = [];
```
Nevoss, line 39: It still can be a property member in the class, but you have to reset it after done with the validation.

modules/atomic-widgets/validators/styles-validator.php
```diff
@@ -38,49 +36,81 @@ public static function make( array $schema ): self {
...
 
-			foreach ( $style['variants'] as $variant_index => $variant ) {
-				if ( ! isset( $variant['meta'] ) ) {
-					$this->errors_bag[] = 'meta';
-					continue;
-				}
+			[, $style, $errors_bag] = $this->validate_style( $style, $errors_bag );
+
+			$styles[ $style_id ] = $style;
+			$errors_bag = array_merge( $errors_bag );
```
Nevoss, line 50: It feels like you should keep the property `error_bag` in the class level

modules/global-classes/api.php
```diff
@@ -0,0 +1,204 @@
+<?php
+
+namespace Elementor\Modules\GlobalClasses;
+
+use Elementor\Core\Kits\Documents\Kit;
+use Elementor\Modules\AtomicWidgets\Styles\Style_Schema;
+use Elementor\Modules\AtomicWidgets\Validators\Styles_Validator;
+
```
Nevoss, line 8:
```suggestion
if ( ! defined( 'ABSPATH' ) ) {
	exit; // Exit if accessed directly.
}
```

modules/global-classes/api.php
```diff
@@ -0,0 +1,204 @@
...
+			$updated = $this->repository->arrange( $order );
+		} catch ( \Exception $e ) {
+			return new \WP_Error( 'unexpected_error', 'Arranging global classes failed unexpectedly', [ 'status' => 500 ] );
+		}
+
+		return new \WP_REST_Response( $updated, 200 );
+	}
+
+	// TODO: Add sanitization when implemented on prop types [EDS-574]
+	private function register_routes() {
```
Nevoss, line 130: Move up

modules/global-classes/api.php
```diff
@@ -0,0 +1,204 @@
...
+
+	public function register_hooks() {
+		add_action( 'rest_api_init', fn() => $this->register_routes() );
+	}
+
+	private function all() {
+		try {
+			$all = $this->repository->all();
+		} catch ( \Exception $e ) {
+			return new \WP_Error( 'unexpected_error', 'Getting global classes failed unexpectedly', [ 'status' => 500 ] );
```
Nevoss, line 24:
```suggestion
			return new \WP_Error( 'unexpected_error', __( 'Getting global classes failed unexpectedly', 'elementor' ), [ 'status' => 500 ] );
```

also fo others.

maybe it will be easier to create a const or custom `WP_Error` which represent `Global_Classes_Error`

modules/global-classes/api.php
```diff
@@ -0,0 +1,204 @@
...
+		$this->repository = new Repository( $kit );
+	}
+
+	public function register_hooks() {
+		add_action( 'rest_api_init', fn() => $this->register_routes() );
+	}
+
+	private function all() {
+		try {
+			$all = $this->repository->all();
```
Nevoss, line 22:
```suggestion
			$classes = $this->repository->all();
```

modules/global-classes/api.php
```diff
@@ -0,0 +1,204 @@
...
+		}
+
+		return $all->get();
+	}
+
+	private function get( $request ) {
+		$id = $request->get_param( 'id' );
+
+		try {
+			$one = $this->repository->get( $id );
```
Nevoss, line 34:
```suggestion
			$class = $this->repository->get( $id );
```

modules/global-classes/api.php
```diff
@@ -0,0 +1,204 @@
+<?php
+
+namespace Elementor\Modules\GlobalClasses;
+
+use Elementor\Core\Kits\Documents\Kit;
+use Elementor\Modules\AtomicWidgets\Styles\Style_Schema;
+use Elementor\Modules\AtomicWidgets\Validators\Styles_Validator;
+
+class API {
```
Nevoss, line 9:
```suggestion
class Global_Classes_REST_API {
```

modules/global-classes/api.php
```diff
@@ -0,0 +1,204 @@
...
+	private Repository $repository;
```
Nevoss, line 10:
```suggestion
	private Global_Classes_Repository $repository;
```

modules/global-classes/api.php
```diff
@@ -0,0 +1,204 @@
...
+
+		if ( null === $one ) {
+			return new \WP_Error( 'entity_not_found', 'Global class not found', [ 'status' => 404 ] );
+		}
+
+		try {
+			$this->repository->delete( $id );
+		} catch ( \Exception $e ) {
+			return new \WP_Error( 'unexpected_error', 'Deleting global class failed unexpectedly', [ 'status' => 500 ] );
+		}
```
Nevoss, line 63: Lets talk about middleware for all the routes to catch errors.

modules/global-classes/api.php
```diff
@@ -0,0 +1,204 @@
...
+
+		return new \WP_REST_Response( null, 204 );
+	}
+
+	private function patch( \WP_REST_Request $request ) {
+		$id = $request->get_param( 'id' );
+		$updated = $request->get_params();
```
Nevoss, line 70:
```suggestion
		$values = $request->get_params();
```

modules/global-classes/api.php
```diff
@@ -0,0 +1,204 @@
...
+		] );
+
+		register_rest_route( Module::API_NAMESPACE, '/' . Module::API_BASE . '/(?P<id>[\w-]+)', [
+			[
+				'methods' => \WP_REST_Server::EDITABLE,
+				'callback' => fn( $request ) => $this->patch( $request ),
+				'validate_callback' => function( \WP_REST_Request $request ) {
+					[, , $errors ] = Styles_Validator::make( Style_Schema::get() )->validate_style( $request->get_body_params() );
+
+					return empty( $errors ) || [ 'id' ] === $errors;
```
Nevoss, line 174: Something feels wrong here, lets talk about it

modules/global-classes/api.php
```diff
@@ -0,0 +1,204 @@
...
+				'permission_callback' => fn() => current_user_can( 'manage_options' ),
+			],
+		] );
+
+		register_rest_route( Module::API_NAMESPACE, '/' . Module::API_BASE . '/(?P<id>[\w-]+)', [
+			[
+				'methods' => \WP_REST_Server::EDITABLE,
+				'callback' => fn( $request ) => $this->patch( $request ),
+				'validate_callback' => function( \WP_REST_Request $request ) {
+					[, , $errors ] = Styles_Validator::make( Style_Schema::get() )->validate_style( $request->get_body_params() );
```
Nevoss, line 172: Why not use `is_valid`?

modules/global-classes/api.php
Nevoss: Maybe you should use `get_body_params` and maybe you should remove the id in `sanitize_callback`

modules/global-classes/api.php
```diff
@@ -0,0 +1,204 @@
...
+			$all = $this->repository->all();
+		} catch ( \Exception $e ) {
+			return new \WP_Error( 'unexpected_error', 'Arranging global classes failed unexpectedly', [ 'status' => 500 ] );
+		}
+
+		$differences = array_merge( $all->get_order()->diff( $order ), array_diff( $order, $all->get_order()->all() ) );
+
+		if ( ! empty( $differences ) ) {
+			return new \WP_Error( 'invalid_content', 'Global classes order invalid', [ 'status' => 422 ] );
+		}
```
Nevoss, line 118: Maybe it should be in `validate_callback`?

Nevoss: Why not using collection all the way? and seem the error message is not correct.
also I cant understand from the code what does it checks.

modules/global-classes/api.php
```diff
@@ -0,0 +1,204 @@
...
+						'required' => true,
+					],
+				],
+				'permission_callback' => fn() => current_user_can( 'manage_options' ),
+			],
+		] );
+
+		register_rest_route( Module::API_NAMESPACE, '/' . Module::API_BASE . '/(?P<id>[\w-]+)', [
+			[
+				'methods' => \WP_REST_Server::EDITABLE,
```
Nevoss, line 169: I think it will be better to just write `PUT` see:
https://github.com/WordPress/wordpress-develop/blob/7b294142e2b0db42a0bd37b81886ce7dd7965c74/src/wp-includes/rest-api/class-wp-rest-server.php#L40

modules/global-classes/module.php
```diff
@@ -12,7 +12,11 @@
 }
 
 class Module extends BaseModule {
-	const EXPERIMENT_NAME = 'global_classes';
+	const NAME = 'global_classes';
+	const API_NAMESPACE = 'elementor/v1';
+	const API_BASE = 'global-classes';
```
Nevoss, line 17: Why not putting the consts in the API class?

modules/global-classes/module.php
```diff
@@ -26,18 +30,28 @@ public function __construct() {
...
 
 		// TODO: When the `Atomic_Widgets` feature is not hidden, add it as a dependency
 		if ( $is_feature_active && $is_atomic_widgets_active ) {
 			add_filter( 'elementor/editor/v2/packages', fn( $packages ) => $this->add_packages( $packages ) );
+			$kit = Plugin::$instance->kits_manager->get_active_kit();
+
+			if ( $kit->get_id() === 0 ) {
+				Plugin::$instance->kits_manager->create_default_kit();
+				$kit = Plugin::$instance->kits_manager->get_active_kit();
+			}
```
Nevoss, line 44: Lets talk about it.

Nevoss (changes requested):

modules/atomic-widgets/validators/styles-validator.php
Omerisra6: then we add it to error bag,
WDYM?

modules/atomic-widgets/validators/styles-validator.php
Omerisra6: I feel like it is easier to track changes that way

modules/global-classes/api.php
Omerisra6: I am not sure we want to force users to remove the id

modules/global-classes/api.php
Omerisra6: There is a difference here, in my opinion, when the data is not in the format, we expect it's not valid, but when the the content is in the right format but wrong, it's invalid content. For example, if the ID the user sent us (when getting an ID) is in the right format, but we can't find the class we don't send validation error.

Omerisra6: I am checking if there are differences between the two arrays, I think it is understandable (I am using a variable called `differences`)

modules/global-classes/api.php
Omerisra6: Shouldn't we use wordpress format?

modules/global-classes/module.php
```diff
@@ -12,7 +12,9 @@
 }
 
 class Module extends BaseModule {
-	const EXPERIMENT_NAME = 'global_classes';
+	const NAME = 'global_classes';
+
+	private Global_Classes_REST_API $api;
```
Nevoss, line 17: Why do we put it on a property?

modules/global-classes/global-classes-rest-api.php
```diff
@@ -0,0 +1,189 @@
...
+		$this->get_repository()->delete( $id );
+
+		return new \WP_REST_Response( null, 204 );
+	}
+
+	private function patch( \WP_REST_Request $request ) {
+		$id = $request->get_param( 'id' );
+		$values = $request->get_params();
+
+		unset( $values['id'] );
```
Nevoss, line 153: Add commend to explain

Nevoss (changes requested):

modules/atomic-widgets/base/atomic-widget-base.php
```diff
@@ -148,13 +149,25 @@ private static function validate_schema( array $schema ) {
...
 
 	private function sanitize_atomic_styles( array $styles ) {
-		[ , $validated, $errors ] = Styles_Validator::make( Style_Schema::get() )->validate( $styles );
+		$errors_bag = [];
+		foreach ( $styles as $style_id => $style ) {
+			if ( ! isset( $style['id'] ) || ! is_string( $style['id'] ) ) {
+				$errors_bag[] = 'id';
+				$styles[ $style_id ] = [];
+				continue;
+			}
```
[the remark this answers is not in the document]
Omerisra6: Mistake, but should we check if the ids match?

modules/atomic-widgets/base/atomic-widget-base.php
```diff
@@ -148,13 +149,25 @@ private static function validate_schema( array $schema ) {
...
 
-		if ( ! empty( $errors ) ) {
-			throw new \Exception( 'Styles validation failed. Invalid keys: ' . join( ', ', $errors ) );
+			[, $sanitized_style, $style_errors_bag] = Style_Validator::make( Style_Schema::get() )->validate( $style );
+
+			$styles[ $style_id ] = $sanitized_style;
+			$errors_bag = array_merge( $style_errors_bag );
```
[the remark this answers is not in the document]
Omerisra6: We do have a test but I don't think it introduces a bug, but it is a mistake

modules/global-classes/module.php
Omerisra6: We thought that the name might change, but I will delete it for now

modules/atomic-widgets/validators/style-validator.php
```diff
@@ -26,45 +26,56 @@ public static function make( array $schema ): self {
...
-				[,$validated_props, $variant_errors] = Props_Validator::make( $this->schema )->validate( $variant['props'] );
-				$styles[ $style_id ]['variants'][ $variant_index ]['props'] = $validated_props;
-
-				$this->errors_bag = array_merge( $this->errors_bag, $variant_errors );
-			}
+			$this->errors_bag = array_merge( $this->errors_bag, $variant_errors );
 		}
 
-		$is_valid = empty( $this->errors_bag );
+		$is_valid = empty( $errors_bag );
```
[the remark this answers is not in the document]
Omerisra6: I updated the tests

modules/global-classes/global-classes-rest-api.php
```diff
@@ -109,7 +109,9 @@ private function register_routes() {
 					}
 
 					$classes = $this->get_repository()->all();
-					return Collection::make( $order )->every( fn( $id ) => $classes->get_order()->contains( $id ) );
+
+					return Collection::make( $order )
+							->every( fn( $id ) => $classes->get_order()->contains( $id ) );
```
Nevoss, line 114: `$classes->get_data()->contains()`

Nevoss (changes requested):

Nevoss (approved):

PR #29295, opened 2024-11-14
Tweak: Added Gutenberg blocks support in frontend [ED-16078]

KingYes:


louiswol94 (approved):
Very nice!

louiswol94:
Can we test what happens if plugins used `do_blocks()` in the widget `render()` method?

https://gist.github.com/ManiruzzamanAkash/b8cafe6815dfd15a146847d354bef75b#file-elementor-nested-surecart-product-widget-php-L322

KingYes:
> Can we test what happens if plugins used `do_blocks()` in the widget `render()` method?
> 
> https://gist.github.com/ManiruzzamanAkash/b8cafe6815dfd15a146847d354bef75b#file-elementor-nested-surecart-product-widget-php-L322

After my PR it's unnecessary. The other widgets should return the plain Gutenberg template without any `do_blocks()`. When you return the native plain Gutenberg template, it will support out-of-box our Caching mechanism.

louiswol94:
> > Can we test what happens if plugins used `do_blocks()` in the widget `render()` method?
> > https://gist.github.com/ManiruzzamanAkash/b8cafe6815dfd15a146847d354bef75b#file-elementor-nested-surecart-product-widget-php-L322
> 
> After my PR it's unnecessary. The other widgets should return the plain Gutenberg template without any `do_blocks()`. When you return the native plain Gutenberg template, it will support out-of-box our Caching mechanism.

@KingYes I understand that. I would just like to see what happens in this case, because they are already using it. If it causes any kind of breakage, I assume we need to give them some time to update their plugins and only release this in 3.26. But if everything still looks ok, then I think we can release it as part of a 3.25 patch like Ariel asked.

louiswol94:
I see it does not cause any breakages. I assume because when they call it, it already brings back the HTML so your `do_blocks` won't get executed. All good.

PR #29304, opened 2024-11-17
Internal: Enqueue font family in the frontend [EDS-509]

ManorHazaz:


modules/atomic-widgets/styles/atomic-styles.php
```diff
@@ -58,9 +58,31 @@ private function parse_element_style( Post $post, Element_Base $element ) {
...
+	 *         props: array<string, mixed>,
+	 *         meta: array<string, mixed>
+	 *     }>
+	 * }> $styles
+	 */
+	private function styles_enqueue_fonts( array $styles ): void {
+		foreach ( $styles as $style ) {
+			foreach ( $style['variants'] as $variant ) {
+				if ( isset( $variant['props']['font-family'] ) ) {
+					Plugin::$instance->frontend->enqueue_font( $variant['props']['font-family'] );
```
ronros-elementor, line 80: Is it possible to mock `Plugin::$instance->frontend->enqueue_font` in order to properly test it? We have a particular [test suite for atomic styles](https://github.com/elementor/elementor/blob/cf9dee8010da804f2d5fd00051379b3d28a74ca9/tests/phpunit/elementor/modules/atomic-widgets/styles/test-atomic-styles.php) where this could be used at

Alternatively we could use
```
ob_start();
// ...
$output = ob_get_clean();
```
and make sure the output contains a responsive font import statement, similar to how it was tested [here](https://github.com/elementor/elementor-pro/blob/89d924a903c3947c5241ea50f5fde8d3a8fae780/tests/phpunit/elementor-pro/modules/loop-builder/files/css/test-loop-css-trait.php#L51-L65)

ronros-elementor (changes requested):

ronros-elementor (approved):

PR #29409, opened 2024-11-25
Internal: Add sanitization to atomic widget prop types [EDS-574]

ronros-elementor:


modules/atomic-widgets/prop-types/base/array-prop-type.php
```diff
@@ -75,6 +75,24 @@ protected function validate_value( $value ): bool {
 		return true;
 	}
 
+	public function sanitize( $value ) {
+		if ( ! $this->validate( $value ) ) {
+			return $this->get_default();
+		}
```
Nevoss, line 81: Why do we need to validate when we sanitize?

Nevoss (changes requested):
This is temp review, I will continue later.

Nice work!

modules/atomic-widgets/prop-types/base/plain-prop-type.php
```diff
@@ -37,6 +37,20 @@ public function validate( $value ): bool {
 		);
 	}
 
+	public function sanitize( $value ) {
+		if ( ! $this->validate( $value ) ) {
+			return $this->get_default();
+		}
```
Nevoss, line 43: Do we need to validate before sanitize?

modules/atomic-widgets/prop-types/base/plain-prop-type.php
```diff
@@ -37,6 +37,20 @@ public function validate( $value ): bool {
...
+
+		if ( $this->is_transformable( $value ) ) {
+			$value['value'] = $this->sanitize_value( $value['value'] );
+		} else {
+			$value = $this->sanitize_value( $value );
+		}
```
Nevoss, line 49: Maybe this if else should live inside the `Supports_Shorthanded_Value` trait?

modules/atomic-widgets/prop-types/image-attachment-id-prop-type.php
```diff
@@ -17,4 +17,8 @@ public static function get_key(): string {
 	protected function validate_value( $value ): bool {
 		return is_numeric( $value ) && Plugin::$instance->wp->wp_attachment_is_image( $value );
 	}
+
+	protected function sanitize_value( $value ) {
+		return $value;
```
Nevoss, line 22: We need to `wp_kses_post` every value

modules/atomic-widgets/prop-types/primitives/boolean-prop-type.php
```diff
@@ -18,4 +18,8 @@ public static function get_key(): string {
 	protected function validate_value( $value ): bool {
 		return is_bool( $value );
 	}
+
+	protected function sanitize_value( $value ) {
+		return $this->validate_value( $value ) ? (bool) $value : null;
```
Nevoss, line 23: We should assume we are using `sanitize` after we already validated

modules/atomic-widgets/prop-types/union-prop-type.php
```diff
@@ -78,6 +78,24 @@ protected function validate_prop_types( $value ): bool {
...
+	}
+
+	protected function sanitize_prop_types( $value ) {
+		foreach ( $this->get_prop_types() as $prop_type ) {
+			if ( $prop_type->validate( $value ) ) {
+				return $prop_type->sanitize( $value );
+			}
+		}
+
+		return $this->get_default();
```
Nevoss, line 96: Why do we need `get_default` here? is it related to `sanitize`?

modules/atomic-widgets/prop-types/union-prop-type.php
```diff
@@ -78,6 +78,24 @@ protected function validate_prop_types( $value ): bool {
...
+		}
+
+		return $this->sanitize_prop_types( $value );
+	}
+
+	protected function sanitize_prop_types( $value ) {
+		foreach ( $this->get_prop_types() as $prop_type ) {
+			if ( $prop_type->validate( $value ) ) {
+				return $prop_type->sanitize( $value );
+			}
```
Nevoss, line 93: Maybe we can understand what the `Prop_Type` is based on the `$$type`?

modules/atomic-widgets/prop-types/url-prop-type.php
```diff
@@ -16,4 +16,8 @@ public static function get_key(): string {
 	protected function validate_value( $value ): bool {
 		return ! ! wp_http_validate_url( $value );
 	}
+
+	protected function sanitize_value( $value ) {
+		return $value;
```
Nevoss, line 21: We should sanitize URL with `sanitize_url `

modules/global-classes/global-classes-rest-api.php
```diff
@@ -72,9 +72,9 @@ private function register_routes() {
 				'methods' => 'PUT',
 				'callback' => fn( $request ) => $this->route_wrapper( fn() => $this->put( $request ) ),
 				'validate_callback' => function( \WP_REST_Request $request ) {
-					[ $is_valid ] = Style_Validator::make( Style_Schema::get() )
+					[ $is_valid ] = Style_Parser::make( Style_Schema::get() )
 						->without_id()
-						->validate( $request->get_body_params() );
+						->parse( $request->get_body_params() );
```
Nevoss, line 77: Feels like we should have `validate` and `sanitize` methods on the `Style_Parser` and also have `parse`.

+ we should add `sanitize_callback` to the routes params.

modules/global-classes/global-classes-rest-api.php
```diff
@@ -87,9 +87,9 @@ private function register_routes() {
 				'methods' => 'POST',
 				'callback' => fn( $request ) => $this->route_wrapper( fn() =>  $this->create( $request ) ),
 				'validate_callback' => function( \WP_REST_Request $request ) {
-					[ $is_valid ] = Style_Validator::make( Style_Schema::get() )
+					[ $is_valid ] = Style_Parser::make( Style_Schema::get() )
 						->without_id()
-						->validate( $request->get_body_params() );
+						->parse( $request->get_body_params() );
```
Nevoss, line 92: Same as above

Nevoss (changes requested):

modules/atomic-widgets/parsers/props-parser.php
```diff
@@ -46,12 +48,15 @@ public function validate( array $props ): array {
 
 			if ( ! $is_valid ) {
 				$this->errors_bag[] = $key;
+				$validated[ $key ] = false;
```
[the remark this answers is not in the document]
ronros-elementor: The `validated` is with only the flags of the validity of each field, while `sanitized` holds the actual values that should be used/stored, so they serve different purposes
I could maybe unify them into one object of `value`, `sanitized_value`, `is_valid`, did you mean something like that? It could even have error messages, instead of the `error_bag` we use

modules/atomic-widgets/parsers/props-parser.php
```diff
@@ -50,7 +50,7 @@ public function validate( array $props ): array {
 				continue;
 			}
 
-			if ( $value ) {
+			if ( ! is_null( $value ) ) {
```
Nevoss, line 53: Cool catch! 🚀

modules/atomic-widgets/prop-types/base/object-prop-type.php
```diff
@@ -85,6 +85,30 @@ protected function validate_value( $value ): bool {
 		return true;
 	}
 
+	public function sanitize( $value ) {
+		if ( ! $this->is_transformable( $value ) ) {
+			return null;
+		}
```
Nevoss, line 91: Is it still required?

modules/atomic-widgets/prop-types/base/plain-prop-type.php
```diff
@@ -37,6 +37,16 @@ public function validate( $value ): bool {
 		);
 	}
 
+	public function sanitize( $value ) {
+		if ( $this->is_transformable( $value ) ) {
+			$value['value'] = $this->sanitize_value( $value['value'] );
+		} else {
+			$value = $this->sanitize_value( $value );
+		}
```
Nevoss, line 45: We removed the shorthand right?

modules/atomic-widgets/prop-types/concerns/has-atomic-base.php
```diff
@@ -68,30 +68,33 @@ private static function validate_schema( array $schema ) {
 		}
 	}
 
-	private function sanitize_atomic_styles( array $styles ) {
-		foreach ( $styles as $style ) {
-			[$is_valid, $sanitized, $errors_bag] = Style_Validator::make( Style_Schema::get() )->validate( $style );
+	private function sanitize_atomic_styles( array $styles ): array {
```
Nevoss, line 71: Maybe we should change the name for `parse_atomic_styles?`

Nevoss: Can we use the `parse` method?

modules/atomic-widgets/prop-types/concerns/has-atomic-base.php
```diff
@@ -68,30 +68,33 @@ private static function validate_schema( array $schema ) {
...
-			$styles[ $sanitized['id'] ] = $sanitized;
+			$styles[ $style_id ] = $style_parser->sanitize( $validated );
 		}
 
 		return $styles;
 	}
 
 	private function sanitize_atomic_settings( array $settings ): array {
 		$schema = static::get_props_schema();
+		$props_parser = Props_Parser::make( $schema );
```
Nevoss, line 89: Same as above

modules/atomic-widgets/prop-types/image-attachment-id-prop-type.php
```diff
@@ -17,4 +17,8 @@ public static function get_key(): string {
 	protected function validate_value( $value ): bool {
 		return is_numeric( $value ) && Plugin::$instance->wp->wp_attachment_is_image( $value );
 	}
+
+	protected function sanitize_value( $value ): string {
+		return wp_kses_post( $value );
```
Nevoss, line 22: maybe just cast it to `int`?

modules/atomic-widgets/prop-types/primitives/string-prop-type.php
```diff
@@ -65,4 +65,8 @@ private function validate_enum( $value ): bool {
 	private function validate_regex( $value ): bool {
 		return preg_match( $this->settings['regex'], $value );
 	}
+
+	protected function sanitize_value( $value ) {
+		return ! empty( $value ) ? sanitize_text_field( $value ) : null;
```
Nevoss, line 70: maybe we should keep the `wp_kses_post`? not sure whats better

modules/atomic-widgets/prop-types/primitives/supports-shorthanded-value.php
```diff
@@ -17,14 +17,21 @@ public function validate( $value ): bool {
 			return ! $this->is_required();
 		}
 
-		if ( $this->validate_value( $value ) ) {
-			return true;
+		if ( ! $this->is_transformable( $value ) ) {
+			return $this->validate_value( $value );
```
Nevoss, line 21: This class should be remove

modules/atomic-widgets/prop-types/union-prop-type.php
```diff
@@ -45,6 +45,26 @@ public function get_prop_type( $type ): ?Transformable_Prop_Type {
...
+			return $this->get_prop_type( 'number' );
+		}
+
+		if ( is_bool( $value ) ) {
+			return $this->get_prop_type( 'boolean' );
+		}
+
+		if ( is_string( $value ) ) {
+			return $this->get_prop_type( 'string' );
+		}
```
Nevoss, line 63: Should we remove it?

modules/global-classes/global-classes-rest-api.php
```diff
@@ -87,7 +92,7 @@ private function register_routes() {
 				'methods' => 'POST',
 				'callback' => fn( $request ) => $this->route_wrapper( fn() =>  $this->create( $request ) ),
 				'validate_callback' => function( \WP_REST_Request $request ) {
-					[ $is_valid ] = Style_Validator::make( Style_Schema::get() )
+					[ $is_valid ] = Style_Parser::make( Style_Schema::get() )
```
Nevoss, line 95: We should also sanitize in the `create` flow

Nevoss, line 95: We should also sanitize in the `create` flow

Nevoss, line 95: We should also sanitize in the `create` flow

Nevoss (changes requested):

modules/atomic-widgets/prop-types/base/plain-prop-type.php
ronros-elementor: Done

modules/atomic-widgets/prop-types/base/object-prop-type.php
ronros-elementor: No - removed

modules/atomic-widgets/prop-types/concerns/has-atomic-base.php
ronros-elementor: Following our discussion - changed to parse, moved all errors to be inside the props_parser->parse method

(in other words - done 😝)

modules/atomic-widgets/prop-types/concerns/has-atomic-base.php
ronros-elementor: Done

modules/atomic-widgets/prop-types/primitives/string-prop-type.php
ronros-elementor: The `sanitize_text_field` method removes *any* HTML element, this is why I'm currently testing the use of 
```
<b>...</b>
```
`wp_kses_post` on the other hand would remove only the dissallowed tags, such as a 
```
<script />
```

modules/atomic-widgets/prop-types/union-prop-type.php
ronros-elementor: Done + fixed a test which still had a shorthand value

modules/global-classes/global-classes-rest-api.php
ronros-elementor: Done

ronros-elementor: Done (including tests)

ronros-elementor: Done

modules/atomic-widgets/prop-types/primitives/supports-shorthanded-value.php
ronros-elementor: We'll take it in a designated PR as it requires some changes to our test-atomic-widget-base tests, and there's some consultation to have with the team regarding how to target use-cases like Size_Prop_Type (where both `unit` and `size` keys should also be changed to the non-shorthanded string/number values - if we choose to go straightforward with this approach)

modules/atomic-widgets/prop-types/base/plain-prop-type.php
```diff
@@ -50,4 +56,8 @@ public function jsonSerialize(): array {
 	abstract public static function get_key(): string;
 
 	abstract protected function validate_value( $value ): bool;
+
+	protected function sanitize_value( $value ) {
+		return ! is_null( $value ) ? wp_kses_post( $value ) : null;
```
Nevoss, line 61: how can it be null?

modules/atomic-widgets/prop-types/base/object-prop-type.php
```diff
@@ -85,6 +85,30 @@ protected function validate_value( $value ): bool {
...
+		return $value;
+	}
+
+	public function sanitize_value( $value ) {
+		foreach ( $this->get_shape() as $key => $prop_type ) {
+			if ( ! isset( $value[ $key ] ) ) {
+				continue;
+			}
+
+			$sanitized_value = $prop_type->sanitize( $value[ $key ] );
```
Nevoss, line 100: Why `sanitize` return null?

modules/global-classes/global-classes-repository.php
```diff
@@ -45,15 +47,18 @@ public function put( string $id, array $value ) {
...
-			throw new \Exception( "Global class with id ${id} not found" );
+			throw new \Exception( "Global class with id {$id} not found" );
 		}
 
 		if ( $value === $all->get_items()[ $id ] ) {
 			return $value;
 		}
 
+		[, $parsed] = Style_Parser::make( Style_Schema::get() )
+			->parse( $value );
```
Nevoss, line 58: it validate the values twice, once in the `validate_callback` and once here, maybe we should use `sanitize_callback` instead of this code?

modules/atomic-widgets/prop-types/primitives/string-prop-type.php
```diff
@@ -65,4 +65,8 @@ private function validate_enum( $value ): bool {
 	private function validate_regex( $value ): bool {
 		return preg_match( $this->settings['regex'], $value );
 	}
+
+	protected function sanitize_value( $value ) {
+		return ! empty( $value ) ? wp_kses_post( $value ) : null;
```
Nevoss, line 70:
```suggestion
		return wp_kses_post( $value );
```

modules/atomic-widgets/parsers/style-parser.php
```diff
@@ -0,0 +1,153 @@
...
+	 * the style object to parse
+	 *
+	 * @return array{
+	 *      0: bool,
+	 *      1: array<string, mixed>,
+	 *      2: array<string>
+	 *  }
+	 */
+	public function parse( array $style ): array {
+		$props_parser = Props_Parser::make( $this->schema );
```
Nevoss, line 139: Why create 2 parser at the same process?

modules/atomic-widgets/prop-types/primitives/number-prop-type.php
```diff
@@ -18,4 +18,8 @@ public static function get_key(): string {
 	protected function validate_value( $value ): bool {
 		return is_numeric( $value );
 	}
+
+	protected function sanitize_value( $value ) {
+		return ! is_null( $value ) ? (int) $value : null;
```
Nevoss, line 23: Why is `is_null`?

modules/atomic-widgets/prop-types/primitives/string-prop-type.php
Nevoss: OK, so I guess `sanitize_text_field` is better

Nevoss (changes requested):

Nevoss (approved):

PR #29419, opened 2024-11-26
Tweak: Remove Core/Pro dependencies in version 3.27 [ED-15717]

hein-obox:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #


core/experiments/manager.php
```diff
@@ -439,19 +439,6 @@ private function add_default_features() {
 			'release_status' => self::RELEASE_STATUS_ALPHA,
 			'default' => self::STATE_ACTIVE,
 		] );
-
-		// TODO: Possibly remove experiment in v3.27.0 [ED-15717].
```
hein-obox, line 443: @rami-elementor I believe it is safe to remove this experiment, as it was hidden and set to immutable for 2 versions already.

rami-elementor: I agree

rami-elementor (dismissed):

Batsirai-muchareva (dismissed):

louiswol94:
Regarding deleting the `global-css.php` file, I know we should remove it at some point and we only support two versions, but maybe we should push it out even longer... The issue is that it triggers a fatal error on version upgrade if the file is not found. WDYT?

hein-obox:
> Regarding deleting the `global-css.php` file, I know we should remove it at some point and we only support two versions, but maybe we should push it out even longer... The issue is that it triggers a fatal error on version upgrade if the file is not found. WDYT?

@louiswol94 Can you please guide us? This time, the Playwright tests seem to be passing. I don't really understand why we got the errors in the past.
What is your proposal?
Should we create a separate task for this?

louiswol94:
> > Regarding deleting the `global-css.php` file, I know we should remove it at some point and we only support two versions, but maybe we should push it out even longer... The issue is that it triggers a fatal error on version upgrade if the file is not found. WDYT?
> 
> @louiswol94 Can you please guide us? This time, the Playwright tests seem to be passing. I don't really understand why we got the errors in the past. What is your proposal? Should we create a separate task for this?

@hein-obox I just saw that with older versions it throws an error. And because it's a fatal error, I was wondering how many users are on older (than 2) versions. And maybe we should keep it a bit longer until most of our users are on 3.24/3.25 - can't remember which version removed the references to this file.

hein-obox:
> > > Regarding deleting the `global-css.php` file, I know we should remove it at some point and we only support two versions, but maybe we should push it out even longer... The issue is that it triggers a fatal error on version upgrade if the file is not found. WDYT?
> > 
> > 
> > @louiswol94 Can you please guide us? This time, the Playwright tests seem to be passing. I don't really understand why we got the errors in the past. What is your proposal? Should we create a separate task for this?
> 
> @hein-obox I just saw that with older versions it throws an error. And because it's a fatal error, I was wondering how many users are on older (than 2) versions. And maybe we should keep it a bit longer until most of our users are on 3.24/3.25 - can't remember which version removed the references to this file.

@louiswol94 Shall we for safety, postpone this removal to v3.28?

louiswol94 (approved):

PR #29444, opened 2024-11-28
Internal: Load http package and its env variables [EDS-603]

Omerisra6:


core/editor/loader/v2/editor-v2-loader.php
```diff
@@ -57,6 +58,17 @@ public function init() {
...
+		add_filter( 'elementor/editor/v2/scripts/env', function( $env ) {
+			$env['@elementor/http'] = [
+				'base_url' => rest_url( 'elementor/v1' ),
+				'headers' => [
+					'X-WP-Nonce' => wp_create_nonce( 'wp_rest' ),
+				],
+			];
+
+			return $env;
+		} );
```
StyleShit, line 70: should we add a `register_hooks` method and call it in the editor itself?

Omerisra6: Lets talk about it

StyleShit (approved):

PR #29466, opened 2024-12-02
Internal: V2 background repeaters overlay color [ED-16118]

elchugreeva:


hein-obox:
@elchugreeva Can we close this PR?

PR #29507, opened 2024-12-05
Internal: [V2] Display control > Add flex gap control [ED-16187]

davseve:


ronenboxer (dismissed):

IshayMaya (dismissed):

IshayMaya:
@davseve we need to add a transformer here as well
see the `Linked_Dimensions_Transformer` as example

ronenboxer (approved):

IshayMaya (approved):

PR #29559, opened 2024-12-11
Internal: [V2] Div Block move CSS file [ED-16369]

baghdasarovelementor:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #


modules/atomic-widgets/module.php
```diff
@@ -156,4 +157,17 @@ private function enqueue_scripts() {
...
+
+	private function enqueue_styles() {
+		wp_register_style(
+			'div-block',
+			$this->get_css_assets_url( 'div-block', 'assets/css/' ),
+			[],
+			ELEMENTOR_VERSION
+		);
+
+		wp_enqueue_style(
```
louiswol94, line 169: @baghdasarovelementor Can you try to only "register" the style here, and add `get_style_depends()` in the element? This way it will only be enqueued if there is actually a div-block on the page.

baghdasarovelementor: @louiswol94 I tried it. Not working properly on Editor.

baghdasarovelementor: in `modules/atomic-widgets/elements/div-block.php`
Added:

`public function get_style_depends() {
		return [ 'div-block' ];
	}`
	
and remove 
`wp_enqueue_style(
			'div-block'
		);`
		
Part from `modules/atomic-widgets/module.php`

louiswol94: Can you debug it a little? It should work.

modules/atomic-widgets/module.php
```diff
@@ -155,4 +156,17 @@ private function enqueue_scripts() {
 			true
 		);
 	}
+
+	private function enqueue_styles() {
+		wp_register_style(
+			'div-block',
+			$this->get_css_assets_url( 'div-block', 'assets/css/' ),
```
louiswol94, line 163:
```suggestion
			$this->get_css_assets_url( 'div-block', 'assets/css/', null, true, true ),
```
I see this in other widgets. Can you check if it's needed here?

modules/atomic-widgets/module.php
```diff
@@ -155,4 +156,17 @@ private function enqueue_scripts() {
...
+			[],
```
louiswol94, line 164:
```suggestion
			[ 'elementor-frontend' ],
```
Maybe?

louiswol94 (changes requested):
check comments

modules/atomic-widgets/module.php
```diff
@@ -155,4 +156,17 @@ private function enqueue_scripts() {
 			true
 		);
 	}
+
+	private function enqueue_styles() {
```
rami-elementor, line 160:
```suggestion
	public function register_styles() {
```

The module should only register styles, and the element should declare a style dependency.

modules/atomic-widgets/module.php
baghdasarovelementor: Not needed for this case.

rami-elementor (approved):

louiswol94 (approved):

PR #29565, opened 2024-12-11
Internal: Add link control to atomic heading, create some infra [ED-16302]

murzin-nikita:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #


modules/atomic-widgets/widgets/atomic-heading.php
```diff
@@ -44,6 +54,27 @@ protected function render() {
...
+			'href' => esc_url( $settings['link']['href'] ) ?? '',
+			'target' => $settings['link']['targetBlank'] ? '_blank' : '',
+		];
+
+		return sprintf(
+			'<%1$s %2$s><a %4$s>%3$s</a></%1$s>',
+			Utils::validate_html_tag( $tag ),
+			Utils::render_html_attributes( $attrs ),
+			esc_html( $title ),
+			Utils::render_html_attributes( array_filter( $link_attrs ) )
```
ronenelementor, line 74: Can we maybe switch it to be 3 before 4?

ronenboxer (dismissed):

modules/atomic-widgets/prop-types/link-prop-type.php
```diff
@@ -0,0 +1,26 @@
...
+class Link_Prop_Type extends Object_Prop_Type {
+	public static function get_key(): string {
+		return 'link';
+	}
+
+	protected function define_shape(): array {
+		return [
+			'enabled' => Boolean_Prop_Type::make(),
+			'href' => Url_Prop_Type::make(),
+			'targetBlank' => Boolean_Prop_Type::make(),
```
baghdasarovelementor, line 23: maybe `isTargetBlank ` ?

baghdasarovelementor (dismissed):

modules/atomic-widgets/widgets/atomic-heading.php
```diff
@@ -44,6 +54,27 @@ protected function render() {
 		);
 	}
 
+	public function get_link_template( array $settings ): string {
```
louiswol94, line 57: Seems like we are duplicating a lot of the logic here. We could just conditionally render an `<a>` tag inside the original render method? 🤔

modules/atomic-widgets/widgets/atomic-heading.php
```diff
@@ -28,6 +32,12 @@ public function get_icon() {
 
 	protected function render() {
 		$settings = $this->get_atomic_settings();
+		$is_link_enabled = ! empty( $settings['link']['enabled'] ) && ! empty( $settings['link']['href'] );
```
ronros-elementor, line 35: Following my comment on the right prop resolver to use, instead of checking the `enabled` flag, we would probably check that `href` is set (as there shouldn't really be an `enabled` attribute there)

modules/atomic-widgets/widgets/atomic-heading.php
```diff
@@ -44,6 +54,27 @@ protected function render() {
...
 
+	public function get_link_template( array $settings ): string {
+		$tag = $settings['tag'];
+		$title = $settings['title'];
+		$attrs = array_filter( [
+			'class' => $settings['classes'] ?? '',
+		] );
+
+		$link_attrs = [
+			'href' => esc_url( $settings['link']['href'] ) ?? '',
```
ronros-elementor, line 65: Since `href` is a Url_Prop_Type - it would be sanitized as part of my [PR](https://github.com/elementor/elementor/pull/29409) so it would probably be an overkill to esc

ronros-elementor (changes requested):

modules/atomic-widgets/widgets/atomic-heading.php
louiswol94: I would prefer if we escape it during render as well. There could be other processes that interfere with the data after it was saved to the DB.

modules/atomic-widgets/widgets/atomic-heading.php
```diff
@@ -29,19 +32,40 @@ public function get_icon() {
 	protected function render() {
 		$settings = $this->get_atomic_settings();
 
+		$format = $this->get_heading_template( ! empty( $settings['link']['href'] ) );
+		$args = $this->get_template_args( $settings );
+
+		echo sprintf( $format, ...$args );
+	}
+
+	private function get_heading_template(bool $is_link_enabled): string {
```
louiswol94, line 41:
```suggestion
	private function get_heading_template( bool $is_link_enabled ): string {
```

mykytamurzin: @louiswol94 done

modules/atomic-widgets/props-resolver/transformers/settings/link-transformer.php
```diff
@@ -0,0 +1,28 @@
...
+		if ( empty( $value['enabled'] ) ) {
+			return [];
+		}
+
+		if ( empty( $value['href'] ) ) {
+			throw new \Exception( 'Url is not provided.' );
+		}
+
+		$link_attrs = [
+			'href' => esc_url( $value['href'] ) ?? '',
```
louiswol94, line 22:
```suggestion
			'href' => esc_url( $value['href'] ),
```
`esc_url` will return an empty string anyway.

louiswol94 (dismissed):

ronros-elementor:
Sorry totally missed the fact there are no tests

Please add tests [here](https://github.com/elementor/elementor/blob/6aee927fb091f685d0df02b065808f9bc70b4fa0/tests/phpunit/elementor/modules/atomic-widgets/prop-types) to the new link prop type, and [here](https://github.com/elementor/elementor/blob/58c9b7ff8e2b181c63253a9bb1db5f55661aa695/tests/phpunit/elementor/modules/atomic-widgets/test-atomic-widget-base.php) - I suggest adding a valid link value into `get_atomic_settings_data_provider` -> `basic`, to make sure the transformer behaves as expected (the first test focuses on the validation of this prop type's values

ronros-elementor (approved):

PR #29636, opened 2024-12-19
Internal: Remove leftovers from the swiper upgrade [ED-16656]

rami-elementor:


m1ga:
please also consider putting one swiper.min.js back in the lib folder (https://github.com/orgs/elementor/discussions/29612)

KingYes (approved):

PR #29676, opened 2024-12-23
Internal: Div block aria-hidden error [ED-16392]

mugurelLupu:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #


baghdasarovelementor (commented):
I can't reproduce it. Maybe we can ask STR.

modules/atomic-widgets/elements/div-block.php
```diff
@@ -105,6 +105,10 @@ protected function add_render_attributes() {
 				$settings['classes'] ?? '',
 			],
 		] );
+
+		if ( isset( $settings['aria_hidden'] ) && $settings['aria_hidden'] ) {
+			$this->add_render_attribute( '_wrapper', 'aria-hidden', 'true' );
+		}
```
davseve, line 111: Are there cases in which the div block won't have an `aria-hiddent` attr?
Where do we set the `aria-hidden`?

ronenelementor: Do we know what purpose `aria-hidden` has in this context? Is it needed, is it an SEO/accessibility issue?

davseve: @ronenelementor As far as I understand,  the purpose of aria hidden is to prevent from using an element that is not visible.
In our case there is a focused element that is inside an element that contains an `aria-hidden` attr set to true.
Which makes no sense.

PR #29719, opened 2024-12-30
Internal: V2 textarea should not be null error [ED-16718]

mugurelLupu:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #

### DIFF
--- a/modules/atomic-widgets/controls/types/textarea-control.php
+++ b/modules/atomic-widgets/controls/types/textarea-control.php
@@ -10,6 +10,8 @@ if ( ! defined( 'ABSPATH' ) ) {
 class Textarea_Control extends Atomic_Control_Base {
 	private $placeholder = null;
 
+	private $value = '';
+
 	public function get_type(): string {
 		return 'textarea';
 	}
@@ -20,9 +22,16 @@ class Textarea_Control extends Atomic_Control_Base {
 		return $this;
 	}
 
+	public function set_value( string $value ): self {
+		$this->value = $value;
+
+		return $this;
+	}
+
 	public function get_props(): array {
 		return [
 			'placeholder' => $this->placeholder,
+			'value' => $this->value,
 		];
 	}
 }
--- a/modules/atomic-widgets/controls/types/textarea-control.php
+++ b/modules/atomic-widgets/controls/types/textarea-control.php
@@ -22,8 +22,8 @@ class Textarea_Control extends Atomic_Control_Base {
 		return $this;
 	}
 
-	public function set_value( string $value ): self {
-		$this->value = $value;
+	public function set_default_value( string $value = null ): self {
+		$this->value = $value ?? '';
 
 		return $this;
 	}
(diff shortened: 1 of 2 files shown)


asafdl (dismissed):

modules/atomic-widgets/controls/types/textarea-control.php
```diff
@@ -20,9 +22,16 @@ public function set_placeholder( string $placeholder ): self {
 		return $this;
 	}
 
+	public function set_value( string $value ): self {
```
davseve, line 25: maybe?
```suggestion
	public function set_value( string $value = '' ): self {
```

modules/atomic-widgets/widgets/atomic-heading.php
```diff
@@ -104,8 +104,8 @@ protected function define_atomic_controls(): array {
 
 					Textarea_Control::bind_to( 'title' )
 						->set_label( __( 'Title', 'elementor' ) )
-						->set_placeholder( __( 'Type your title here', 'elementor' ) ),
-
+						->set_placeholder( __( 'Type your title here', 'elementor' ) )
+						->set_value( '' ),
```
davseve, line 108: If we add a fallback in the function itself than it can be removed from here 
```suggestion
						->set_value(),
```

modules/atomic-widgets/controls/types/textarea-control.php
```diff
@@ -20,9 +22,16 @@ public function set_placeholder( string $placeholder ): self {
...
+		$this->value = $value;
+
+		return $this;
+	}
```
davseve, line 29: maybe adding a default value?
WDYT?
```suggestion
	public function set_value( string $value = null ): self {
		$this->value = $value ?? '';

		return $this;
	}
```

modules/atomic-widgets/controls/types/textarea-control.php
Omerisra6, line 1: I think we should handle this error in the `elementor-packages` repo and just set the default state to empty string

Omerisra6 (changes requested):

modules/atomic-widgets/controls/types/textarea-control.php
davseve: It seems that this issue was already solved
Thanks :)

modules/atomic-widgets/controls/types/textarea-control.php
```diff
@@ -20,9 +22,16 @@ public function set_placeholder( string $placeholder ): self {
 		return $this;
 	}
 
+	public function set_default_value( string $value = null ): self {
```
eyalmrejen-elementor, line 25: Why not set the `$value` to an empty string in the function arguments and lose that complicated condition on line 26? This is what this capability is for

modules/atomic-widgets/widgets/atomic-heading.php
```diff
@@ -104,8 +104,8 @@ protected function define_atomic_controls(): array {
 
 					Textarea_Control::bind_to( 'title' )
 						->set_label( __( 'Title', 'elementor' ) )
-						->set_placeholder( __( 'Type your title here', 'elementor' ) ),
-
+						->set_placeholder( __( 'Type your title here', 'elementor' ) )
+						->set_default_value(),
```
eyalmrejen-elementor, line 108: I know it's a small bug, but I would add a small test for this. Once a test is there, this bug will NEVER come back

PR #29749, opened 2025-01-02
Internal: Fixed lint comments [EDS-1186]

blackcolt:


TzviRabinovitch (approved):

modules/element-manager/ajax.php
```diff
@@ -122,7 +122,7 @@ private function force_enabled_all_elements() {
 	}
 
 	private function get_plugin_name_from_widget_instance( $widget ) {
-		if ( in_array( 'wordpress', $widget->get_categories() ) ) {
+		if ( in_array( 'WordPress', $widget->get_categories(), true ) ) {
```
TzviRabinovitch, line 125: @blackcolt 
Im not sure if this is the same as the previous check

PR #29762, opened 2025-01-04
Internal: Remove "Upgrade Swiper Library" experiment [ED-15983]

rami-elementor:


hein-obox (dismissed):

hein-obox (dismissed):

core/experiments/manager.php
```diff
@@ -638,7 +633,7 @@ private function render_feature_dependency( $feature ) {
 		<?php
 	}
 
-	private function has_non_existing_dependency( $feature ) {
+	private function has_non_existing_dependency( $feature ): bool {
```
blackcolt, line 636: if its an exsisting function that add a return type for lint (if its a new function its ok)

blackcolt (approved):

PR #29852, opened 2025-01-09
Internal: Fix PHP lint issues [ED-17737]

blackcolt:


core/admin/canary-deployment.php
```diff
@@ -127,7 +127,7 @@ private function check_group( $group ) {
 			// Reset results for each condition.
 			$result = false;
 			switch ( $condition['type'] ) {
-				case 'wordpress': // phpcs:ignore WordPress.WP.CapitalPDangit.Misspelled
+				case 'WordPress': // phpcs:ignore WordPress.WP.CapitalPDangit.Misspelled
```
TzviRabinovitch, line 130: the value might ber case sensitive

blackcolt: thanks.

modules/element-manager/ajax.php
```diff
@@ -122,7 +122,7 @@ private function force_enabled_all_elements() {
 	}
 
 	private function get_plugin_name_from_widget_instance( $widget ) {
-		if ( in_array( 'wordpress', $widget->get_categories() ) ) {
+		if ( in_array( 'WordPress', $widget->get_categories() ) ) {
```
rami-elementor, line 125: @blackcolt are we sure it's won't break anything?

blackcolt: its without the strict, so its not suppose to, by I will revert it.

PR #29853, opened 2025-01-09
Tweak: Load Text Editor widget styles on control level rather than widget level [ED-17735]

rami-elementor:


includes/widgets/text-editor.php
```diff
@@ -95,13 +95,17 @@ public function get_keywords() {
...
+	 * The 'widget-text-editor' style is required only when the drop cap is used.
+	 * Therefor, style should not be loaded on the widget level, rather only on
+	 * control level when the drop cap is active.
+	 *
 	 * @since 3.24.0
 	 * @access public
 	 *
 	 * @return array Widget style dependencies.
 	 */
 	public function get_style_depends(): array {
```
hein-obox, line 107: Maybe remove this function completely?

rami-elementor: Yes, initially I did, but then I thought that we need to keep track on the changes so that in the future developers won't remove this optimization.

hein-obox: You can decide about it.

hein-obox (approved):

includes/widgets/text-editor.php
rami-elementor: Let's keep it.

PR #29864, opened 2025-01-10
Tweak: Create infrastructure for conditional loading of frontend JS handlers [ED-16264]

hein-obox:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #


elementor.php
```diff
@@ -105,3 +105,33 @@ function elementor_fail_wp_version() {
...
+		wp_register_script( $bc_handle, $src, $dependencies, $version, true );
+
+		add_filter( 'script_loader_tag', function ( $tag, $script_handle ) use ( $bc_handle ) {
+			if ( $script_handle === $bc_handle ) {
+				$tag = str_replace( '<script ', '<script type="module" ', $tag );
+			}
+			return $tag;
+		}, 10, 2 );
+	}
+}
```
rami-elementor, line 137: Copied from core?

includes/frontend.php
```diff
@@ -344,6 +345,46 @@ public function remove_content_filter() {
...
+		 * Fires after Elementor frontend script modules are registered.
+		 *
+		 * @since 3.28.0
+		 */
+		do_action( 'elementor/frontend/after_register_script_modules' );
+
+		$this->enqueue_script_modules();
+	}
+
+
```
rami-elementor, line 387:
```suggestion
```

core/experiments/manager.php
```diff
@@ -393,6 +393,14 @@ private function add_default_features() {
...
 			'default' => self::STATE_INACTIVE,
 		] );
+
+		// TODO: Remove this experiment in v3.32 [ED-17712].
+		$this->add_feature( [
+			'name' => 'e_load_js_modules_conditionally',
+			'title' => esc_html__( 'Load JS Module Script files conditionally', 'elementor' ),
+			'default' => self::STATE_INACTIVE,
+			static::TYPE_HIDDEN => true,
+		] );
```
rami-elementor, line 403: @hein-obox  This is for Core/Pro compatibility?

elementor.php
hein-obox: It is not copied. WordPress has created an additional class for this. This is a simplified version to support module script enqueues.

rami-elementor (changes requested):
@hein-obox 

Any change we can move the logic from `register_script_modules()` method to the end of `register_scripts()` method?

Same for `enqueue_script_modules()`, moving it to `enqueue_scripts()`?

I prefer not to further inflate the registration/enqueue methods.

includes/frontend.php
```diff
@@ -344,6 +345,45 @@ public function remove_content_filter() {
...
+	 * Registers script modules.
+	 *
+	 * Registers all the frontend script modules.
+	 *
+	 * Fired by `wp_enqueue_scripts` action.
+	 *
+	 * @since 3.28.0
+	 * @access public
+	 */
+	public function register_script_modules(): void {
```
hein-obox, line 358: @rami-elementor Yes, I believe it is possible to merge this logic to the register_scripts() method.

I will leave it for me. I would like to make changes only, once we have Yakir's feedback as well.

assets/dev/js/frontend/utils/swiper.js
```diff
@@ -89,3 +89,7 @@ export default class SwiperHandler {
 		return config;
 	}
 }
+
+window.elementorFrontend = window.elementorFrontend || {};
+window.elementorFrontend.utils = window.elementorFrontend.utils || {};
+window.elementorFrontend.utils.swiper = SwiperHandler;
```
hein-obox, line 95: I am still studying this approach. I am not sure yet if we should attach the modules to the `window` object.

I would love to hear your thoughts about this.

ChatGpt says: 

> Caveats of Using window
> Global Namespace Pollution: Attaching too many properties to window can clutter the global namespace and potentially cause conflicts with other scripts.
> Modularity: By avoiding imports, you lose some benefits of modularity, such as tree-shaking (removing unused code) and better dependency management.
> If you're building a modern application, prefer using ES modules with imports unless you specifically need global access for compatibility. For legacy WordPress environments or simpler use cases, attaching to window is a practical approach.

hein-obox: I have discussed this aspect with @Batsirai-muchareva and @mike-elementor, and they indicated that we should be safe here. The `window` object is also used heavily in e.g. `elementor.widgetsCache` inside v4.

PR #29872, opened 2025-01-12
Fix: Background slideshow option doesn't work for columns after Elementor 3.26 update [ED-16726]

nami-p:


ronenboxer (approved):

tests/playwright/sanity/modules/column/column.test.ts
```diff
@@ -0,0 +1,53 @@
...
+import WpAdminPage from '../../../pages/wp-admin-page';
+
+test.describe( 'Column tests @column', () => {
+	test.beforeAll( async ( { browser, apiRequests }, testInfo ) => {
+		const context = await browser.newContext();
+		const page = await context.newPage();
+		const wpAdmin = new WpAdminPage( page, testInfo, apiRequests );
+		await wpAdmin.setExperiments( {
+			container: false,
+			'nested-elements': true,
```
rami-elementor, line 12: @nami-p nested is dependent on containers, if you deactivate containers then the nested will be deactivated too. and when you activate nested, then containers are activated.

PR #29873, opened 2025-01-12
Internal: Fix PHP lint issues [ED-17752]

blackcolt:


modules/compatibility-tag/compatibility-tag.php
```diff
@@ -30,12 +30,12 @@ class Compatibility_Tag extends Base_Object {
 	public function __construct( $header ) {
 		$this->header = $header;
 	}
-
+
```
KingYes, line 33: Remove it

KingYes (changes requested):

modules/compatibility-tag/compatibility-tag.php
blackcolt: Done

KingYes (dismissed):

ManorHazaz (approved):

PR #29881, opened 2025-01-13
Internal: Fix PHP lint issues [ED-17760]

blackcolt:


modules/history/revisions-manager.php
```diff
@@ -6,6 +6,7 @@
 use Elementor\Core\Files\CSS\Post as Post_CSS;
 use Elementor\Plugin;
 use Elementor\Utils;
+use mysql_xdevapi\Exception;
```
KingYes, line 9: @blackcolt What is it??

KingYes (changes requested):

modules/history/revisions-manager.php
blackcolt: good catch, thanks!

KingYes (approved):

PR #29893, opened 2025-01-13
Internal: Fixed lint error [ED-17694]

KingYes:


app/modules/import-export/runners/import/elementor-content.php
```diff
@@ -95,7 +94,7 @@ private function import_post( array $post_settings, array $post_data, $post_type
 		);
 
 		if ( is_wp_error( $new_document ) ) {
-			throw new \Exception( $new_document->get_error_message() );
+			throw new \Exception( esc_html( $new_document->get_error_message() ) );
```
blackcolt, line 97: not suppose to be esc_html__ for translation?

KingYes: Did you see any translate here?

blackcolt (approved):

PR #29912, opened 2025-01-14
Internal: Regenerate style IDs on duplicated widgets [ED-17002]

ronros-elementor:
## Problem

When duplicating (copy-paste/duplicate/import) an atomic widget that already has a local style applied to - we would end up with a "shared" local style for both widgets, meaning any modification to any of the widgets' local style would affect both widgets.

## Implementation Details

The fix should not "pollute" v1, hence it should be as part of the atomic-widgets component (`modules/atomic-widgets/assets/js/editor/component.js`)

This PR introduces new data hooks, as part of the atomic-widgets module, bound to each of the problematic commands - 
- `"document/elements/duplicate"`
- `"document/elements/paste"`


modules/atomic-widgets/assets/js/editor/hooks/data/duplicated-styles/base-duplicated-styles-detection.js
```diff
@@ -0,0 +1,98 @@
...
+		// Update local styles
+		container.model.set( 'styles', newStyles );
+	}
+
+	/**
+	 * @param {StyledElementData[]} styledElements
+	 * @param {Container}           model
+	 * @return {StyledElementData[]}
+	 */
+	getDuplicatedStyledElements(
```
Nevoss, line 58: Check if third argument `result` can help

Nevoss (changes requested):

modules/atomic-widgets/assets/js/editor/utils/get-random-style-id.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * @typedef {import('elementor/assets/dev/js/editor/container/container')} Container
+ */
+
+/**
+ * @param {Container} container
+ * @return {string}
+ */
+export function getRandomStyleId( container ) {
+	return `s-${ container.id }-${ elementorCommon.helpers.getUniqueId() }`;
```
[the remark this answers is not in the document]
ronros-elementor: I actually think that atomic styles are prefixed with `'s-`, based on [this](https://github.com/elementor/elementor/blob/e41ac24835bc9fc4b04ec271cd489209a37eb9c4/modules/atomic-widgets/assets/js/editor/commands-internal/create-style.js#L24)

modules/atomic-widgets/assets/js/editor/commands/styles-update.js
```diff
@@ -0,0 +1,13 @@
+export class StylesUpdate extends $e.modules.CommandBase {
```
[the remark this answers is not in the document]
ronros-elementor: Removed

modules/atomic-widgets/assets/js/editor/hooks/data/duplicated-styles/base-duplicated-styles-detection.js
ronros-elementor: It _did_, thanks! 😊

modules/atomic-widgets/assets/js/editor/hooks/data/duplicated-styles/base-duplicated-styles-detection.js
```diff
@@ -0,0 +1,98 @@
+import { classesPropTypeUtil } from '@elementor/editor-props';
```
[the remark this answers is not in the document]
ronros-elementor: Removed

modules/atomic-widgets/assets/js/editor/commands-internal/create-style.js
```diff
@@ -20,16 +21,19 @@ export class CreateStyle extends $e.modules.editor.CommandContainerInternalBase
 		}
 	}
 
-	randomId( containerId ) {
-		return `s-${ containerId }-${ elementorCommon.helpers.getUniqueId() }`;
+	randomId( container ) {
```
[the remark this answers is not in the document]
ronros-elementor: Removed this changed

modules/atomic-widgets/assets/js/editor/utils/get-random-style-id.js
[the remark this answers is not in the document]
Nevoss: We just changed it, `e` for `element` styles and `g` for `global` styles.
@StyleShit Has PR that will remove the commands from the plugin and move all the logic to the packages, in this PR he changed it. (in the tests, it was mistake)

modules/atomic-widgets/assets/js/editor/utils/handle-duplicated-styles.js
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * @typedef {import('assets/dev/js/editor/container/container')} Container
+ */
+
+/**
+ * Get a container - iterate over its children, find all styled atomic widgets and update their style ids
+ * @param {Container} duplicatedElement
+ */
+export function handleDuplicatedStyles( duplicatedElement ) {
+	const allElements = getElementChildren( duplicatedElement );
```
Nevoss, line 13: Do you nee to split all those functions into small files or is it OK to put it on the same file?

tests/jest/unit/modules/atomic-widgets/create-container.js
```diff
@@ -11,6 +11,9 @@ export default function createContainer( {
 	const container = {
 		id,
 		label,
+		elType,
+		widgetType,
+		type: elType,
```
Nevoss, line 16: Why do we need `type` and `elType`?

tests/jest/unit/modules/atomic-widgets/hooks/data/duplicated-styles/detect-duplicated-styles.test.js
```diff
@@ -0,0 +1,238 @@
+import { createContainer, addChildToContainer } from '../../../create-container';
+import { setGlobalContainers } from './utils/set-global-containers';
+
+describe( 'Detect duplicated styles', () => {
```
Nevoss, line 4: Is the name of the test and file are correct?

tests/jest/unit/modules/atomic-widgets/hooks/data/duplicated-styles/utils/set-global-containers.js
```diff
@@ -0,0 +1,15 @@
+export function setGlobalContainers( containers = [] ) {
```
Nevoss, line 1: Do we need another file?

Nevoss: If it is a general util, maybe it should be outside of duplicated-styles dir and please re-use it when ever need

modules/atomic-widgets/assets/js/editor/utils/handle-duplicated-styles.js
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * @typedef {import('assets/dev/js/editor/container/container')} Container
+ */
+
+/**
+ * Get a container - iterate over its children, find all styled atomic widgets and update their style ids
+ * @param {Container} duplicatedElement
+ */
+export function handleDuplicatedStyles( duplicatedElement ) {
```
Nevoss, line 12: Maybe the name is not correct? because when `import` there are no duplicate styles. 
Maybe `syncStyleIdsWithElementIds` or something like that?

modules/atomic-widgets/assets/js/editor/utils/update-style-id.js
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Update the style id of the container.
+ * @param {Container} container
+ */
+export function updateStyleId( container ) {
+	const widgetsCache = window.elementor?.widgetsCache ?? {};
+
+	const elType = container.model.get( 'elType' );
+	const widgetType = elType !== 'widget' ? elType : container.model.get( 'widgetType' );
```
Nevoss, line 15: I am not sure it is gonna work, did you check if it works with `div-block`? another idea is not the check against the prop schema at all, just run over the settings and check if there is any transformable object with `classes` on its type.
what do you think?

modules/atomic-widgets/assets/js/editor/utils/update-style-id.js
```diff
@@ -0,0 +1,55 @@
...
+	const originalStyles = container.model.get( 'styles' );
+	const settings = container.settings?.toJSON() ?? {};
+
+	const elementType = widgetType ? widgetsCache?.[ widgetType ] : null;
+
+	const [ classesPropKey ] = Object.entries( elementType?.atomic_props_schema ?? {} ).find(
+		( [ , propType ] ) => 'plain' === propType.kind && 'classes' === propType.key,
+	) ?? [];
```
Nevoss, line 23: What if I have multiple `classes` props?

modules/atomic-widgets/assets/js/editor/utils/update-style-id.js
```diff
@@ -0,0 +1,55 @@
...
+		const newStyleId = getRandomStyleId( container );
+
+		newStyles[ newStyleId ] = structuredClone( { ...style, id: newStyleId } );
+
+		changedIds[ originalStyleId ] = newStyleId;
+	} );
+
+	const newClasses = {
+		...settings[ classesPropKey ],
+		value: originalClasses.map( ( className ) => changedIds[ className ] ?? className ),
```
Nevoss, line 42: Cool 🚀

Nevoss (changes requested):

modules/atomic-widgets/assets/js/editor/utils/get-random-style-id.js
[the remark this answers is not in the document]
ronros-elementor: Ohh got it, will change 👍

modules/atomic-widgets/assets/js/editor/utils/handle-duplicated-styles.js
ronros-elementor: Good point, also for the purpose of the util function - I don't mind what the reason is to create new style ids, I just iterate over any styled element and regenerate an id for it.

As `sync` may mistakenly hint on some execution constraint (as in sync/async), or a pub/sub-like design pattern, and the element id association to the style id is an implementation detail for the `getRandomStyleId`, I though of going with `regenerateLocalStyleIds` - as we also only care about the local styles (as I only look for styles that are set in the `styles` attribute of the widget)

wdyt?

modules/atomic-widgets/assets/js/editor/utils/update-style-id.js
ronros-elementor: It _does_ work with div-blocks of course, including any nested styled atomic widget, and the duplicated styled div-blocks themselves.

But regardless - your suggestion makes sense and will probably be simpler, will take it 😃

tests/jest/unit/modules/atomic-widgets/create-container.js
ronros-elementor: You're right, I went so deep with this task, in so many different directions, I ended up accidentally polluting this mock function with stuff I no longer need apparently 🫣 removed all of those added attributes

modules/atomic-widgets/assets/js/editor/utils/update-style-id.js
ronros-elementor: Going forward with your suggestion - I'd also address this topic

modules/atomic-widgets/assets/js/editor/utils/regenerate-local-style-ids.js
```diff
@@ -0,0 +1,70 @@
+import { getElementChildren } from './get-element-children';
+import { getRandomStyleId } from './get-random-style-id';
+
+/**
+ * @typedef {import('assets/dev/js/editor/container/container')} Container
+ */
+
+function isClassesProp( prop ) {
+	return prop.$$type && Array.isArray( prop.value ) && prop.value.length > 0;
```
Nevoss, line 9:
```suggestion
	return prop.$$type && prop.$$type === 'classes' && Array.isArray( prop.value ) && prop.value.length > 0;
```

modules/atomic-widgets/assets/js/editor/hooks/data/regenerate-local-style-ids/duplicate-element.js
```diff
@@ -0,0 +1,18 @@
+import { regenerateLocalStyleIds } from '../../../utils/regenerate-local-style-ids';
+
+export class DuplicateElementHook extends $e.modules.hookData.After {
```
Nevoss, line 3: Seems like we can use `create` command and not hook into all those command

Nevoss (changes requested):

modules/atomic-widgets/assets/js/editor/utils/regenerate-local-style-ids.js
```diff
@@ -0,0 +1,70 @@
...
+}
+
+/**
+ * Get a container - iterate over its children, find all styled atomic widgets and update their style ids
+ * @param {Container} container
+ */
+export function regenerateLocalStyleIds( container ) {
+	const allElements = getElementChildren( container );
+
+	const styledElements = allElements.filter( ( element ) => Object.keys( element.model.get( 'styles' ) ).length > 0 );
```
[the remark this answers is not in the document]
ronros-elementor: Done

modules/atomic-widgets/assets/js/editor/hooks/data/regenerate-local-style-ids/duplicate-element.js
ronros-elementor: Yep sounds like it indeed, it stayed quite the same as the other 3 hooks, but removed those and also updated the test to only check the create hook instead

modules/atomic-widgets/assets/js/editor/hooks/data/regenerate-local-style-ids/create-element.js
```diff
@@ -0,0 +1,20 @@
...
+	getId() {
+		return 'regenerate-local-style-ids--document/elements/duplicate';
+	}
+
+	apply( args, result ) {
+		const containers = Array.isArray( result ) ? result : [ result ];
+
+		containers.forEach( regenerateLocalStyleIds );
+
+		console.log( 'END HOOK' );
```
Nevoss, line 17:
```suggestion
```

Nevoss (dismissed):

Nevoss (approved):

PR #29913, opened 2025-01-14
Internal: Added deactivation action [ED-16178]

KingYes:


includes/api.php
```diff
@@ -53,21 +53,7 @@ class Api {
...
-	 * @since 2.0.0
-	 * @access private
-	 * @static
-	 *
-	 * @param bool $force_update Optional. Whether to force the data retrieval or
-	 *                                     not. Default is false.
-	 *
-	 * @return array|false Info data, or false.
-	 */
-	private static function get_info_data( $force_update = false ) {
```
rami-elementor, line 70: Why remove the phpDocs?

nuritsha (approved):

includes/api.php
KingYes: It's so useless, we should remove all other comments from our code :D

PR #29974, opened 2025-01-21
Internal: Run long tests after merge 

TzviRabinovitch:


.github/workflows/playwright.yml
```diff
@@ -156,20 +156,20 @@ jobs:
         if: ${{ needs.Playwright.result == 'failure' }}
         uses: actions/checkout@v4
       - name: Send slack message
-        if: ${{ needs.Playwright.result == 'failure' && github.event_name == 'schedule' }}
+        if: ${{ needs.Playwright.result == 'failure'}}
```
Svitlana-Dykun, line 159: I think we shouldn't send notifications for every failed run, maybe only for 'schedule' and 'push' triggers?

blackcolt (dismissed):

Svitlana-Dykun (dismissed):

blackcolt (dismissed):

.github/workflows/playwright.yml
```diff
@@ -1,7 +1,7 @@
 name: Playwright
 
 on:
-  pull_request:
+  push:
```
Svitlana-Dykun, line 4: I noticed that workflow is still triggered and cancelled on every push to a branch
We should specify branches : 
on:
  push:
    branches:
      - 'main'
      - '3.*'

Svitlana-Dykun (changes requested):

.github/workflows/playwright.yml
Svitlana-Dykun, line 1: We also need to change trigger for build-plugin job in this file
Something like this should work
build-plugin:
  if: ${{ github.event_name == 'push' && ( github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/3') ) || github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}

.github/workflows/playwright.yml
Svitlana-Dykun: Or as an alternative,   
pull_request:
    types:
      - closed

Svitlana-Dykun (approved):

PR #30027, opened 2025-01-26
Tweak: Add filter hook for Home JSON [ED-17800]

elchugreeva:


TzviRabinovitch (dismissed):

modules/home/api.php
```diff
@@ -3,6 +3,7 @@
 
 use Elementor\Includes\EditorAssetsAPI;
 use Elementor\Modules\Home\Classes\Transformations_Manager;
+use Elementor\Core\Utils\Promotions\Filtered_Promotions_Manager;
```
eyalmrejen-elementor, line 6: Is this needed? Maybe a clean up

modules/home/api.php
```diff
@@ -14,6 +15,8 @@ public function __construct( EditorAssetsAPI $editor_assets_api ) {
 	public function get_home_screen_items( $force_request = false ): array {
 		$assets_data = $this->editor_assets_api->get_assets_data( $force_request );
 
+		$assets_data = apply_filters( 'elementor/core/admin/homescreen', $assets_data );
```
eyalmrejen-elementor, line 18: Is there a way to add a test to ensure this filter was applied? So that if it's removed by mistake, this test will save you

elchugreeva: @TzviRabinovitch should it be hosting team worry?

TzviRabinovitch: @eyalmrejen-elementor 
this will be testing wordpress capabilities 
the only test that I can think of is to validate that the apply filer was called once with this filter name

eyalmrejen-elementor: I don't mean to test WP - but rather that this filter was set - this specific one - not to test the `add_filter` method

modules/home/api.php
eyalmrejen-elementor: do we have a php linter to test unused imports?

TzviRabinovitch (approved):

PR #30139, opened 2025-02-04
Internal: [V4] Background Repeaters - Overlay - Image - Size customs [ED-16334]

nami-p:


tests/phpunit/elementor/modules/atomic-widgets/test-atomic-widget-base.php
```diff
@@ -1263,8 +1272,20 @@ public function test_get_data_for_save__throws_on_styles_background_color_overla
 																'url' => null
 															],
 														],
-														'size' => 'cover',
-                                                        'position' => 'center left',
+														'size' => [
```
elchugreeva, line 1275: we can remove additional check for Background_Image_Overlay_Size_Scale_Prop_Type validation here, the current function fails on wrong color input, while scale prop type allows input for only 1 dimention, and is a combination of size inputs, that where checked separately.

elchugreeva (changes requested):

elchugreeva (dismissed):

tests/phpunit/elementor/modules/atomic-widgets/test-atomic-widget-base.php
nami-p: This is necessary because the transformer receives a custom value as a string and then converts it to css with 'custom' value

mike-elementor (dismissed):

elchugreeva (approved):

modules/atomic-widgets/module.php
```diff
@@ -171,6 +173,7 @@ private function register_styles_transformers( Transformers_Registry $transforme
 		$transformers->register( Image_Attachment_Id_Prop_Type::get_key(), new Primitive_Transformer() );
 		$transformers->register( Url_Prop_Type::get_key(), new Primitive_Transformer() );
 		$transformers->register( Background_Image_Overlay_Prop_Type::get_key(), new Background_Image_Overlay_Transformer() );
+		$transformers->register( Background_Image_Overlay_Size_Scale_Prop_Type::get_key(), new Background_Image_Overlay_Size_Scale_Transformer() );
```
IshayMaya, line 176: are you adding it to the canvas render as well? didn't see it there

nami-p: yes , 
here :https://github.com/elementor/elementor-packages/pull/558/files#diff-33d5a31b7ff224d2b5fb0df2439a78fd0881666f32ef4dc40398cf205208bb55

IshayMaya (approved):

PR #30165, opened 2025-02-06
Internal: Implement search input in the cloud library source [ED-18007]

louiswol94:


assets/dev/js/editor/components/template-library/views/parts/templates.js
```diff
@@ -26,6 +27,7 @@ TemplateLibraryCollectionView = Marionette.CompositeView.extend( {
 		myFavoritesFilter: '#elementor-template-library-filter-my-favorites',
 		orderInputs: '.elementor-template-library-order-input',
 		orderLabels: 'label.elementor-template-library-order-label',
+		icon: '#elementor-template-library-filter-text-wrapper i',
```
max-zu, line 30: let's maybe rename `icon` to something more meaningful and descriptive ( `searchInputIcon` or something like that)

modules/cloud-library/connect/cloud-library.php
```diff
@@ -26,6 +26,10 @@ public function get_resources( $args = [] ): array {
 			$endpoint .= '?parentId=' . $args['template_id'];
 		}
 
+		if ( ! empty( $args['search'] ) ) {
+			$endpoint .= '?search=' . $args['search'];
+		}
+
```
max-zu, line 32: let's maybe use `http_build_query ` to build the query_string ?

assets/dev/js/editor/components/template-library/views/parts/templates.js
louiswol94: I thought about that, but it is a search icon and then a spinner icon, so thats why I just said `icon`. But I can name it `searchInputIcon` since that's the initial state and it's easy to understand 👍

modules/cloud-library/connect/cloud-library.php
louiswol94: I also thought about that hehe, but, we don't want more than one param in the query string in this case. We don't want a `parentId` and a `search` term at the same time. Wdyt?

max-zu (dismissed):

ManorHazaz (dismissed):
Super!

max-zu (approved):

PR #30168, opened 2025-02-06
Internal: [V4] SVG Widget > Upload SVG [ED-16702]

mugurelLupu:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #


ronenboxer (dismissed):

modules/atomic-widgets/controls/types/svg-control.php
```diff
@@ -3,19 +3,31 @@
...
+			'src' => [
+				'id' => [
+					'$$type' => 'image-attachment-id',
+					'value' => null,
+				],
+				'url' => [
+					'$$type' => 'image-url',
+					'value' => null,
+				],
+			],
```
Nevoss, line 30: Why do you need those props? does it come from default values?

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
```diff
@@ -31,82 +30,73 @@ public function get_icon() {
...
+		$svg_content = ( strpos( $svg_url, self::SVG_PREFIX ) === 0 )
+			? urldecode( substr( $svg_url, strlen( self::SVG_PREFIX ) ) )
+			: wp_remote_retrieve_body( wp_safe_remote_get( $svg_url ) );
 
 		$svg = new \WP_HTML_Tag_Processor( $svg_content );
 
-		if ( ! $svg->next_tag() ) {
-			return;
+		if ( $svg->next_tag( 'svg' ) ) {
+			$this->set_svg_attributes( $svg, $settings );
```
Nevoss, line 48: I think that all the logic of "transforming" a prop into something that can be rendered should live inside a transformer, maybe consider creating a specific transformer for svg

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
```diff
@@ -31,82 +30,73 @@ public function get_icon() {
...
 
-		$svg_data = $settings['svg'];
-
-		$svg_content = isset( $svg_data['url'] ) ? $svg_data['url'] : '';
-
-		if ( strpos( $svg_content, self::SVG_ENCODING ) === 0 ) {
-			$svg_content = urldecode( substr( $svg_content, self::SVG_START_INDEX ) );
-		}
+		$svg_content = ( strpos( $svg_url, self::SVG_PREFIX ) === 0 )
+			? urldecode( substr( $svg_url, strlen( self::SVG_PREFIX ) ) )
```
Nevoss, line 42: If you really want, you can create a union which sometimes be a URL and sometimes be a a real svg, then you can have 2 transformers for each prop type of the union

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
```diff
@@ -31,82 +30,73 @@ public function get_icon() {
...
+			: wp_remote_retrieve_body( wp_safe_remote_get( $svg_url ) );
```
Nevoss, line 43: What about performance here? if I have a lot of SVG widgets is it gonna be slow?

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
```diff
@@ -31,82 +30,73 @@ public function get_icon() {
...
+		printf( '%s', $valid_svg ? $valid_svg : self::get_default_svg() ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
 	}
 
-	private function set_svg_attributes( $svg, $settings ) {
+	private function set_svg_attributes( \WP_HTML_Tag_Processor $svg, $settings ) {
 		$svg->set_attribute( 'fill', $settings['color'] ?? 'currentColor' );
-		$svg->set_attribute( 'width', $settings['width'] ?? '100%' );
-		$svg->set_attribute( 'height', $settings['height'] ?? '100%' );
+		$svg->set_attribute( 'width', $settings['width'] ?? '100px' );
+		$svg->set_attribute( 'height', $settings['height'] ?? '100px' );
```
Nevoss, line 64: First, seems like you put default values, so you do not need a fallback here. 
Second, why not just using regular CSS? why there is need for real attribute on the element?

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
```diff
@@ -31,82 +30,73 @@ public function get_icon() {
...
 		return [
-			$content_section,
+			Section::make()
+				->set_label( esc_html__( 'Content', 'elementor' ) )
+				->set_items( [ Svg_Control::bind_to( 'svg' ) ] ),
 		];
 	}
 
+	protected static function get_default_svg(): string {
+		return '<div style="margin-inline: auto"><svg width="100" height="100" xmlns="http://www.w3.org/2000/svg">
```
Nevoss, line 77: Do you need this div? or the `height` and `width`? also there is a fill on the `g` tag

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
```diff
@@ -31,82 +30,73 @@ public function get_icon() {
...
-                    <rect width="6" height="6" fill="#505050"/>
-                    <rect width="3" height="3" fill="#999999"/>
-                    <rect x="3" y="3" width="3" height="3" fill="#999999"/>
-                </pattern>
-            </defs>
-
-            <rect width="100%" height="100%" fill="url(#checkered)"/>
-
-            <g transform="translate(100, 100) scale(0.8)" fill="#fff">
+		$default_placeholder_svg = '<svg width="200" height="200" xmlns="http://www.w3.org/2000/svg">
```
Nevoss, line 85: Why there is a need for placeholder and default?

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
```diff
@@ -31,82 +30,73 @@ public function get_icon() {
...
-				->default( [] ),
-			'width' => String_Prop_Type::make()
-				->default( '100%' ),
-			'height' => String_Prop_Type::make()
-				->default( '100%' ),
-			'svg' => Image_Prop_Type::make()
-				->default_url( self::get_placeholder() ),
+			'classes' => Classes_Prop_Type::make()->default( [] ),
+			'width' => String_Prop_Type::make()->default( '100px' ),
+			'height' => String_Prop_Type::make()->default( '100px' ),
```
Nevoss, line 97: Should width and height be a string?

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
```diff
@@ -31,82 +30,73 @@ public function get_icon() {
...
+			'margin-inline' => String_Prop_Type::make()->default( 'auto' ),
```
Nevoss, line 98: Also a CSS property which should not be part of the settings

modules/atomic-widgets/prop-types/svg-prop-type.php
```diff
@@ -0,0 +1,47 @@
...
+use Elementor\Modules\AtomicWidgets\PropTypes\Base\Object_Prop_Type;
+use Elementor\Modules\AtomicWidgets\PropTypes\Primitives\String_Prop_Type;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+class Svg_Prop_Type extends Object_Prop_Type {
+	public static function get_key(): string {
+		return 'image/svg+xml';
```
Nevoss, line 15: This is the `key` for the prop type and not a mime type, maybe `svg` will fit better here.

modules/atomic-widgets/prop-types/svg-prop-type.php
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	protected function define_shape(): array {
+		return [
+			'src' => Image_Src_Prop_Type::make()->required(),
```
Nevoss, line 20: Is it an image? maybe it should be SVG_Src_Prop_Type?

modules/atomic-widgets/prop-types/svg-prop-type.php
```diff
@@ -0,0 +1,47 @@
...
+			'size' => String_Prop_Type::make()->enum( Image_Sizes::get_keys() )->required(),
```
Nevoss, line 21: Is size relevant here?

modules/atomic-widgets/prop-types/svg-prop-type.php
```diff
@@ -0,0 +1,47 @@
...
+		] );
+
+		return $this;
+	}
+
+	public function default_size( string $size ): self {
+		$this->get_shape_field( 'size' )->default( $size );
+
+		return $this;
+	}
```
Nevoss, line 46: is it in use?

modules/atomic-widgets/prop-types/svg-prop-type.php
Nevoss: And again maybe you can use a union with `SVG_Src_Prop_Type`, `SVG_Attachment_Id_Prop_Type`

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
```diff
@@ -31,82 +30,73 @@ public function get_icon() {
...
-			->set_label( esc_html__( 'Content', 'elementor' ) )
-			->set_items( [
-				Svg_Control::bind_to( 'svg' ),
-			] );
-
 		return [
-			$content_section,
+			Section::make()
+				->set_label( esc_html__( 'Content', 'elementor' ) )
+				->set_items( [ Svg_Control::bind_to( 'svg' ) ] ),
```
Nevoss, line 72: Where is the check if the user actually can upload an SVG? Is it in the frontend

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
```diff
@@ -31,82 +30,73 @@ public function get_icon() {
...
+		if ( $svg->next_tag( 'svg' ) ) {
+			$this->set_svg_attributes( $svg, $settings );
 		}
 
-		$this->set_svg_attributes( $svg, $settings );
-
-		$svg->add_class( $settings['classes'] ?? '' );
+		while ( $svg->next_tag( [ 'path', 'rect', 'circle', 'g' ] ) ) {
+			$svg->remove_attribute( 'fill' );
+		}
```
Nevoss, line 53: Feels like all those staff can be done in the upload process and save the SVG in the DB and just fetch it right here already ready for render

Nevoss (dismissed):

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
mugurelLupu: I need this part in order to reset all the other fill attributes, so that the user can change the svg fill color, this is done via svg attributes and not only css.

modules/atomic-widgets/prop-types/svg-prop-type.php
mugurelLupu: I used this to prevent some code duplication, because it uses the same mechanism.

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
mugurelLupu: I dont see any reason why not.

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
mugurelLupu: The placeholder is the default for the control image, and the default is the default image we render in template.

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
mugurelLupu: Yes, without the sizing the svg is not defined. Also the fill is the default color of the svg that was requested by the product team.

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
mugurelLupu: We need some default values, otherwise the image will be 100%.

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
mugurelLupu: I tried creating a transformer and using it, but then i lose the ability to control the svg, for some reason i lose the ability to control fill color, and also the height.

Nevoss: Maybe those attribute should come from the style tab and not from the settings tab?

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
Nevoss: but why with attributes? you can maybe use the style tab

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
Nevoss: you can use a new feature we just added, called base style. please talk with @Omerisra6 about it

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
Nevoss: both of them are the same SVG? maybe the default is the placeholder?

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
Nevoss: why not usings `Size_Prop_Type`? is there any control for this Prop Type?

modules/atomic-widgets/prop-types/svg-prop-type.php
Nevoss: It does something else, when loading an SVG you actually read the file and print the content on the page, that is why it should be a Prop Type with its own transformer.

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
Nevoss: Yeah, but what about the performance here? are we going to read from a file each user request and manipulate the dom element each time? maybe you can do it once?

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
```diff
@@ -31,82 +34,80 @@ public function get_icon() {
...
+				->add_variant(
+					Style_Variant::make()
+						->add_prop( 'width', $width )
+						->add_prop( 'height', $height )
+				),
+		];
+	}
 
-            <g transform="translate(100, 100) scale(0.8)" fill="#fff">
+	protected static function get_default_svg(): string {
```
Omerisra6, line 99: Please remove this part; the default should come from the prop type default URL

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
```diff
@@ -31,82 +34,80 @@ public function get_icon() {
...
-			'classes' => Classes_Prop_Type::make()
-				->default( [] ),
-			'width' => String_Prop_Type::make()
-				->default( '100%' ),
-			'height' => String_Prop_Type::make()
-				->default( '100%' ),
-			'svg' => Image_Prop_Type::make()
-				->default_url( self::get_placeholder() ),
+			'classes' => Classes_Prop_Type::make()->default( [] ),
+			'svg' => Svg_Prop_Type::make()->default_url( ELEMENTOR_ASSETS_URL . 'images/defaultsvg.svg' ),
```
Omerisra6, line 110: You didn't move the default to the file

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
```diff
@@ -31,82 +34,80 @@ public function get_icon() {
...
 		}
 
-		$svg_data = $settings['svg'];
-
-		$svg_content = isset( $svg_data['url'] ) ? $svg_data['url'] : '';
-
-		if ( strpos( $svg_content, self::SVG_ENCODING ) === 0 ) {
-			$svg_content = urldecode( substr( $svg_content, self::SVG_START_INDEX ) );
-		}
+		$svg_content = ( strpos( $svg_url, self::SVG_PREFIX ) === 0 )
```
Omerisra6, line 45: You have this logic inside the transformer it should be removed from here

Omerisra6 (changes requested):

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
```diff
@@ -3,19 +3,22 @@
...
+use Elementor\Modules\AtomicWidgets\Styles\Style_Variant;
 
 if ( ! defined( 'ABSPATH' ) ) {
 	exit; // Exit if accessed directly.
 }
 
 class Atomic_Svg extends Atomic_Widget_Base {
-	const SVG_ENCODING = 'data:image/svg+xml;charset=utf-8,';
-	const SVG_START_INDEX = 26;
+	const SVG_PREFIX = 'data:image/svg+xml;charset=utf-8,';
```
Omerisra6, line 20: Is it used anywhere?

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
```diff
@@ -29,84 +32,80 @@ public function get_icon() {
...
-		}
-
-		$svg_data = $settings['svg'];
-
-		$svg_content = isset( $svg_data['url'] ) ? $svg_data['url'] : '';
-
-		if ( strpos( $svg_content, self::SVG_ENCODING ) === 0 ) {
-			$svg_content = urldecode( substr( $svg_content, self::SVG_START_INDEX ) );
+		if ( ! $svg_url && isset( $settings['svg']['id'] ) ) {
+			$attachment = wp_get_attachment_image_src( $settings['svg']['id'], 'full' );
```
Omerisra6, line 44: extract the "full" string to const (DEFAULT_SIZE)

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
```diff
@@ -29,84 +32,80 @@ public function get_icon() {
...
+			$svg_url = isset($attachment[0]) ? $attachment[0] : null;
 		}
 
-		$svg = new \WP_HTML_Tag_Processor( $svg_content );
-
-		if ( ! $svg->next_tag() ) {
+		if ( ! $svg_url ) {
+			// Render default SVG if no URL is provided
+			printf( '%s', wp_safe_remote_get( $this->get_default_svg_path() ) ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
Omerisra6, line 50: You can remove this part as it is handled by the prop type default

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
```diff
@@ -29,84 +32,80 @@ public function get_icon() {
...
-			'classes' => Classes_Prop_Type::make()
-				->default( [] ),
-			'width' => String_Prop_Type::make()
-				->default( '100%' ),
-			'height' => String_Prop_Type::make()
-				->default( '100%' ),
-			'svg' => Image_Prop_Type::make()
-				->default_url( self::get_placeholder() ),
+			'classes' => Classes_Prop_Type::make()->default( [] ),
+			'svg' => Image_Src_Prop_Type::make()->default_url(  ELEMENTOR_ASSETS_URL . 'images/defaultsvg.svg' ),
```
Omerisra6, line 108: I still can't see the new file. When you add it, please rename it to: `a-default-svg.svg`

modules/atomic-widgets/prop-types/svg-prop-type.php
```diff
@@ -0,0 +1,32 @@
+<?php
```
Omerisra6, line 1: you can remove this file

Omerisra6 (changes requested):

modules/atomic-widgets/controls/types/svg-control.php
```diff
@@ -3,14 +3,15 @@
...
+use Elementor\Modules\AtomicWidgets\Image_Sizes;
 
 if ( ! defined( 'ABSPATH' ) ) {
 	exit; // Exit if accessed directly.
 }
 
 class Svg_Control extends Atomic_Control_Base {
 	public function get_type(): string {
-		return 'image';
+		return 'svg';
```
Omerisra6, line 14: call it `svg-media` instead

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
```diff
@@ -29,84 +32,74 @@ public function get_icon() {
...
-		$this->set_svg_attributes( $svg, $settings );
-
-		$svg->add_class( implode( ' ', $settings['classes'] ?? [] ) );
-
 		$valid_svg = ( new Svg_Sanitizer() )->sanitize( $svg->get_updated_html() );
 
-		// we need this line in order to render the svg, otherwise it will be rendered as a string
-		echo ( false === $valid_svg ) ? '' : $valid_svg; //phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+		// Render the SVG content
+		printf( '%s', $valid_svg ?? wp_safe_remote_get( $this->get_default_svg_path() ) ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
Omerisra6, line 60: Is it possible for the SVG to be invalid?

Omerisra6 (changes requested):

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
mugurelLupu: Yes, when there is no attachment for that url

tests/phpunit/elementor/modules/atomic-widgets/test-atomic-svg.php
```diff
@@ -5,29 +5,48 @@
...
+				'id' => 123,
+				'url' => ELEMENTOR_ASSETS_URL . 'images/a-default-svg.svg',
+			],
+		],
+		'widgetType' => 'a-svg',
+	];
+
+	protected $instance;
+
+	public function setUp(): void {
```
Omerisra6, line 24: Create a `test__render_svg_from_url` and `test__render_svg_from_id` to test both cases.

Use the same logic we used before this change.

Omerisra6 (changes requested):

Omerisra6 (dismissed):

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
```diff
@@ -29,84 +32,74 @@ public function get_icon() {
...
 
-		$svg_data = $settings['svg'];
-
-		$svg_content = isset( $svg_data['url'] ) ? $svg_data['url'] : '';
-
-		if ( strpos( $svg_content, self::SVG_ENCODING ) === 0 ) {
-			$svg_content = urldecode( substr( $svg_content, self::SVG_START_INDEX ) );
+		$svg = new \WP_HTML_Tag_Processor( file_get_contents( $svg_url ) );
+		if ( $svg->next_tag( 'svg' ) ) {
+			$this->set_svg_attributes( $svg, $settings );
```
ronenelementor, line 50: Please add an empty line before the `if`, shouldn't we check that the svg is_url set in the `if` above actually has an svg before going to `next_tag`?

tests/phpunit/elementor/modules/atomic-widgets/test-atomic-svg.php
```diff
@@ -5,29 +5,84 @@
...
+		parent::setUp();
+
+		add_filter('wp_get_attachment_image_src', function($image, $attachment_id, $size) {
```
ronenelementor, line 37: Please verify linting in this file (mainly spaces in the parenthesis, cmd+option+L should do it if you imported the PHP schema in editor code style)

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
mugurelLupu: https://github.com/elementor/elementor/pull/30168#discussion_r1957260101

ronenelementor: The url could be set, but it might be outdated, faulty or corrupt, you should check that the $svg variable was successfully instanciated, I added another comment

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
```diff
@@ -29,84 +32,76 @@ public function get_icon() {
...
+		if ( $svg && $svg->next_tag( 'svg' ) ) {
+			$this->set_svg_attributes( $svg, $settings );
 		}
 
-		$svg = new \WP_HTML_Tag_Processor( $svg_content );
-
-		if ( ! $svg->next_tag() ) {
-			return;
+		while ( $svg->next_tag( [ 'path', 'rect', 'circle', 'g' ] ) ) {
+			$svg->remove_attribute( 'fill' );
```
ronenelementor, line 56: Same check here

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
```diff
@@ -29,84 +32,76 @@ public function get_icon() {
...
 		}
 
-		$this->set_svg_attributes( $svg, $settings );
-
-		$svg->add_class( implode( ' ', $settings['classes'] ?? [] ) );
-
 		$valid_svg = ( new Svg_Sanitizer() )->sanitize( $svg->get_updated_html() );
```
ronenelementor, line 59: And same check here, $svg could be null

ronenboxer (dismissed):

ronenboxer (approved):

Omerisra6 (approved):

PR #30189, opened 2025-02-09
Internal: Add custom escaper for template renderer [ED-18015]

Nevoss:


modules/atomic-widgets/elements/atomic-heading/atomic-heading.html.twig
```diff
@@ -0,0 +1,12 @@
+{# TODO: Need to add base classes #}
+<{{ settings.tag | e('html_tag') }}
```
StyleShit, line 2: shouldn't it come sanitized from the props resolver?

StyleShit (changes requested):

modules/atomic-widgets/elements/atomic-heading/atomic-heading.html.twig
Nevoss: We have some validations, but as for today we validate it and also when printing sanitize it, just like `full_url`, this is the same. not sure about it, what do you think? I think if we are going to remove it, we also need to remove `full_url`

StyleShit (approved):

PR #30204, opened 2025-02-10
Internal: Add support for base styles in twig templates [ED-18016]

Nevoss:


modules/atomic-widgets/elements/has-base-styles.php
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	public function get_base_styles_dictionary() {
+		$result = [];
+
+		foreach ( $this->get_base_styles() as $key => $style ) {
+			$result[ $key ] = $style['id'];
+		}
+
+		return $result;
```
StyleShit, line 36: map doesn't work here?

StyleShit (approved):

PR #30206, opened 2025-02-10
Tweak: Change placeholder image for V4 [ED-17912]

MiryamOren:
Atomic image


m1ga:
size can be reduced by ~50% using tinypng.com

blackcolt (approved):

modules/atomic-widgets/elements/atomic-image/atomic-image.php
```diff
@@ -52,7 +52,7 @@ protected static function define_props_schema(): array {
 				->default( [] ),
 
 			'image' => Image_Prop_Type::make()
-				->default_url( Utils::get_placeholder_image_src() )
+				->default_url( Utils::get_placeholder_image_src( 'v4' ) )
```
StyleShit, line 55: why do we need to use the same function instead of inlining the URL?
kinda weird to add a param that lets this "core" function know about features ("v4")

StyleShit (changes requested):

includes/utils.php
```diff
@@ -337,10 +337,17 @@ public static function is_post_type_support( $post_type ) {
...
 	 */
-	public static function get_placeholder_image_src() {
-		$placeholder_image = ELEMENTOR_ASSETS_URL . 'images/placeholder.png';
+	public static function get_placeholder_image_src( $version = 'v3' ) {
+		$placeholder_image_by_version = [
+			'v3' => 'placeholder.png',
+			'v4' => 'placeholder-v4.png',
+		];
+
+		$placeholder_image = ELEMENTOR_ASSETS_URL . "images/" . ( $placeholder_image_by_version[ $version ] );
```
StyleShit, line 350: For future implementations: you're not checking that the key actually exists, which means it might break when the array changes (which will, in turn, break the website)

includes/utils.php
```diff
@@ -337,10 +337,17 @@ public static function is_post_type_support( $post_type ) {
...
+	 * @param string $version The placeholder image version. Defaults to 'v3'. Options: 'v3', 'v4'.                     Options: 'v3', 'v4'.
+	 *
 	 * @return string The source of the default placeholder image used by Elementor.
 	 */
-	public static function get_placeholder_image_src() {
-		$placeholder_image = ELEMENTOR_ASSETS_URL . 'images/placeholder.png';
+	public static function get_placeholder_image_src( $version = 'v3' ) {
+		$placeholder_image_by_version = [
+			'v3' => 'placeholder.png',
+			'v4' => 'placeholder-v4.png',
```
StyleShit, line 347: IMO, the v4 placeholder should live inside the atomic module, since it's part of the feature and not a system-wide image

StyleShit (changes requested):

PR #30228, opened 2025-02-12
Internal: Refactor global classes REST to support syncing [ED-18162]

StyleShit:


modules/global-classes/global-classes-repository.php
Omerisra6, line 1: Sad Cry GIF by SpongeBob SquarePants

modules/global-classes/global-classes-rest-api.php
```diff
@@ -42,69 +42,51 @@ private function register_routes() {
...
+				'permission_callback' => fn() => current_user_can( 'manage_options' ),
 				'args' => [
-					'id' => [
-						'type' => 'string',
+					'items' => [
 						'required' => true,
+						'type' => 'object',
+						'additionalProperties' => false,
+						'patternProperties' => [
+							'^g-[a-z0-9]+$' => [
```
Omerisra6, line 56: Do we want to fail if a user passes am ID that starts without the letter 'g'?

modules/global-classes/global-classes-rest-api.php
```diff
@@ -42,69 +42,51 @@ private function register_routes() {
...
+										'required' => true,
+									],
+									'variants' => [
+										'type' => 'array',
+										'required' => true,
+									],
+									'type' => [
+										'type' => 'string',
+										'enum' => [ 'class' ],
+										'required' => true,
```
Omerisra6, line 71: Maybe we need to set the type to class instead of failing on nonclass types (and change this to not required)?

modules/global-classes/global-classes-rest-api.php
```diff
@@ -117,109 +99,77 @@ private function all() {
...
+	private function sanitize_items( array $items ) {
+		$sanitized_items = [];
+
+		foreach ( $items as $item_id => $item ) {
+			[$is_valid, $sanitized, $errors] = Style_Parser::make( Style_Schema::get() )->parse( $item );
+
+			if ( ! $is_valid ) {
+				return [
+					null,
+					Error_Builder::make( 'invalid_item' )
```
Omerisra6, line 137: Wouldn't it be better to throw an error and, in the put method, return the response?

Omerisra6 (changes requested):

modules/global-classes/global-classes-rest-api.php
StyleShit: Done

modules/global-classes/global-classes-rest-api.php
StyleShit: I'm strict

modules/global-classes/global-classes-rest-api.php
StyleShit: I'm strict

Omerisra6 (approved):

PR #30248, opened 2025-02-13
Fix: Reset style did not properly reset global settings [ED-18235]

KingYes:


assets/dev/js/editor/document/elements/commands/reset-settings.js
```diff
@@ -30,6 +30,8 @@ export class ResetSettings extends $e.modules.editor.document.CommandHistoryBase
 				defaultValues[ controlName ] = control.default;
 			} );
 
+			defaultValues.__globals__ = {};
```
StyleShit, line 33: unrelated to this PR, but does setting the default value to the control mean we're saving it to the database rather than just resetting it?
shouldn't it be the same solution as what you've done with the globals? (basically removing the data and letting the system set the defaults)

StyleShit, line 33: I saw these things:

https://github.com/elementor/elementor/blob/9eeee9b0695700187cda24b97962a726c80a47cb/assets/dev/js/editor/document/elements/commands/paste-style.js#L119-L123

https://github.com/elementor/elementor/blob/281850f765a66f31d86acc6fd6f599f60776fec2/assets/dev/js/editor/document/globals/commands/settings.js#L59-L63

https://github.com/elementor/elementor/blob/39ffa61c3da41fc4ae87e97ba0a33ee3d6f51b8d/assets/dev/js/editor/document/globals/commands/disable.js#L32-L35

And at this point, I have no idea which is the "right" way...

StyleShit (dismissed):

nuritsha (approved):

PR #30289, opened 2025-02-18
Tweak: Use CSS logical properties in safe mode toast [ED-18268]

rami-elementor:
Use **CSS logical properties** (inset) instead on **CSS physicals properties** (top/bottom/left/right) in Elementor Editor safe mode toast.

### DIFF
--- a/modules/safe-mode/module.php
+++ b/modules/safe-mode/module.php
@@ -155,23 +155,19 @@ class Module extends \Elementor\Core\Base\Module {
 			.elementor-safe-mode-toast {
 				position: absolute;
 				z-index: 10000; /* Over the loading layer */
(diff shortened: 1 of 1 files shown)


modules/safe-mode/module.php
```diff
@@ -441,7 +425,7 @@ public function print_try_safe_mode() {
 					}
 
 					if ( ! $notice.data( 'visible' ) ) {
-						$notice.show().data( 'visible', true );
+						$notice.attr( 'style', 'display: flex;' );
```
rami-elementor, line 428: Previously, it added `display:block`, now it's `display:flex`.

KingYes (approved):

PR #30301, opened 2025-02-18
Tweak: Set placeholder image for V4 [ED-17912] 

MiryamOren:
image


composer.json
```diff
@@ -28,6 +28,7 @@
 	},
 	"scripts": {
 		"lint": "vendor/bin/phpcs --extensions=php --standard=./ruleset.xml",
+		"format": "vendor/bin/phpcbf --extensions=php --standard=./ruleset.xml",
```
hein-obox, line 31: Do we need to add this in this PR, or could we handle this in a separate task?

modules/atomic-widgets/elements/atomic-image/atomic-image.php
ronros-elementor, line 1: I think you can remove `use Elementor\Utils;` as we're not using it anymore

composer.json
MiryamOren: Deleted

modules/atomic-widgets/elements/atomic-image/atomic-image.php
MiryamOren: Done. Thanks for the catch

ronros-elementor (approved):

PR #30309, opened 2025-02-19
Internal: Fix renderOnChange for div-block [ED-18288] 

ronros-elementor:
### This PR fixes 2 things -
1. The main issue - when duplicating a `div-block` widget, the duplicated widget style doesn't seem to affect the editor preview, though all settings are configured correctly. Upon a refresh - the issue is resolved.
2. An acciental catch - when changing the style of a `div-block` - it re-renders the whole widget even though it is absolutely redundant, as the CSS is rendered elsewhere.

### Solution
Trigger the base view's renderOnChange only if there are actual changes, and replace the whole `'class'` attribute of the DOM element with an updated class list, to remove the original widget's local style class (i.e. the widget we duplicated)

#### Before:
https://github.com/user-attachments/assets/0075cd37-c507-41fa-86d0-f17d57d980d5

#### After:
https://github.com/user-attachments/assets/88cf63a8-35c2-49dc-acd7-25d1a2920cf6


modules/atomic-widgets/assets/js/editor/div-block-view.js
```diff
@@ -57,8 +57,23 @@ const DivBlockView = BaseElementView.extend( {
 		return width.toFixed( 1 ) + '%';
 	},
 
-	renderOnChange() {
-		BaseElementView.prototype.renderOnChange.apply( this, arguments );
+	renderOnChange( settings ) {
+		const changed = settings.changedAttributes();
+
+		if ( ! changed ) {
+			// If nothing change - avoid the redundant re-rendering
```
ronenelementor, line 64: I think this makes a lot of sense even without the comment, feels unneeded

ronenelementor (dismissed):

Batsirai-muchareva (dismissed):

ronenelementor (approved):

PR #30326, opened 2025-02-20
Tweak: Add a button “create a new folder” in the cloud library source [ED-18210]

ManorHazaz:


includes/editor-templates/templates.php
```diff
@@ -116,10 +116,25 @@
 			<?php endif; ?>
 			</div>
 		<# } #>
-		<div id="elementor-template-library-filter-text-wrapper">
-			<label for="elementor-template-library-filter-text" class="elementor-screen-only"><?php echo esc_html__( 'Search Templates:', 'elementor' ); ?></label>
-			<input id="elementor-template-library-filter-text" placeholder="<?php echo esc_attr__( 'Search', 'elementor' ); ?>">
-			<i class="eicon-search"></i>
+
+		<div id="elementor-template-library-filter-toolbar-side-actions">
```
louiswol94, line 120: Is this correct? According to the design the icon should be next to the dropdown? (Not next to the search)

includes/editor-templates/templates.php
```diff
@@ -116,10 +116,25 @@
...
+			<?php if ( Plugin::$instance->experiments->is_feature_active( 'cloud-library' ) ) : ?>
+				<# if ( 'cloud' === activeSource ) {
+					var isInRootFolder = elementor.templates.getFilter('parent') === null;
```
louiswol94, line 123:
```suggestion
					const isInRootFolder = elementor.templates.getFilter('parent') === null;
```

includes/editor-templates/templates.php
```diff
@@ -116,10 +116,25 @@
...
+
+					if ( isInRootFolder ) { #>
```
louiswol94, line 125: Do you still need this check?

includes/editor-templates/templates.php
```diff
@@ -116,10 +116,25 @@
...
+						<div id="elementor-template-library-add-new-folder" class="">
+							<i class="eicon-folder" aria-hidden="true" title="Create a New Folder"></i>
+							<span class="elementor-screen-only">Create a New Folder</span>
```
louiswol94, line 128: Translations?

includes/template-library/manager.php
```diff
@@ -639,6 +639,22 @@ public function load_more_templates( array $args ) {
...
+		$validate_args = $this->ensure_args( [ 'source', 'title' ], $args );
+
+		if ( is_wp_error( $validate_args ) ) {
+			return $validate_args;
+		}
+
+		$source = $this->get_source( $args['source'] );
+
+		if ( ! $source ) {
+			return new \WP_Error( 'template_error', 'Template source not found.' );
```
louiswol94, line 652:
```suggestion
			return new \WP_Error( 'folder_error', 'Template source not found.' );
```
Maybe? You decide.

includes/editor-templates/templates.php
ManorHazaz: Yes.
Take a look on the final design :)

includes/editor-templates/templates.php
ManorHazaz: Correct

assets/dev/js/editor/components/template-library/manager.js
```diff
@@ -251,6 +253,77 @@ const TemplateLibraryManager = function() {
 		} );
 	};
 
+	this.createFolder = function( folderData, options ) {
```
louiswol94, line 256: It would probably be a separate task, but I wonder if we can improve the UX when the confirm button is clicked. It feels like the screen just freezes for a few seconds and then reloads. Maybe the add new folder icon could be replaced with a spinner while the request is happening (similar to the delete/rename actions).

assets/dev/js/editor/components/template-library/views/parts/templates.js
```diff
@@ -339,6 +341,22 @@ const TemplateLibraryCollectionView = Marionette.CompositeView.extend( {
...
+		if ( 'cloud' !== activeSource ) {
+			return;
+		}
+
+		elementor.templates.createFolder( {
+			source: activeSource,
+		},
+		{
+			onSuccess: () => {
+				elementor.templates.showTemplates();
```
louiswol94, line 356: I think we have a problem with `elementor.templates.showTemplates();`, the Tabs become unresponsive after it.s been called. It looks like this is the issue: https://github.com/elementor/elementor/blob/main/assets/dev/js/editor/components/template-library/manager.js#L598 - everything still works fine if I delete this line `self.layout.setHeaderDefaultParts()`;, but once again it's a BC concern that we might not know if we remove it what it will break.

louiswol94 (approved):

assets/dev/js/editor/components/template-library/views/parts/templates.js
louiswol94, line 356:
```suggestion
				$e.routes.refreshContainer( 'library' );
```
Following on my [previous comment](https://github.com/elementor/elementor/pull/30326/files#r1970791907), it seems like this actually works better.

louiswol94: We'll need to apply this fix on the delete action as well in Local and Cloud.

assets/dev/scss/editor/_templates-modal.scss
```diff
@@ -275,6 +275,30 @@ label.elementor-template-library-order-label {
...
+	border-radius: 2px;
+	color: var(--e-a-color-txt);
+
+	&:hover {
+		background-color: var(--e-a-bg-hover);
+	}
+}
+
+#elementor-template-library-create-new-folder-dialog {
+	.elementor-create-folder-template-dialog__p {
```
louiswol94, line 297:
```suggestion
#elementor-template-library-create-new-folder-dialog {

	.elementor-create-folder-template-dialog__p {
```

includes/template-library/sources/cloud.php
```diff
@@ -101,6 +101,20 @@ public function save_item( $template_data ): int {
 		return (int) $response['id'];
 	}
 
+	public function save_folder( array $folder_data = [] ) {
+		$app = $this->get_app();
+
+		$resource_data = [
+			'title' => $folder_data['title'] ?? esc_html__( 'New Folder', 'elementor' ),
+			'type' => self::FOLDER_RESOURCE_TYPE,
```
louiswol94, line 109: `templateType` is missing

Screenshot 2025-02-26 at 4 51 32 AM

louiswol94 (changes requested):

includes/template-library/sources/cloud.php
ManorHazaz: Good one

assets/dev/js/editor/components/template-library/views/parts/templates.js
ManorHazaz: Done but it feels like a workaround and not like solving the bug.

louiswol94 (approved):

PR #30345, opened 2025-02-24
Internal: Create REST API for user introductions [ED-16881]

Omerisra6:


modules/editor-user/module.php
```diff
@@ -0,0 +1,40 @@
...
+		parent::__construct();
+
+		$is_atomic_widgets_active = Plugin::$instance->experiments->is_feature_active( Atomic_Widgets_Module::EXPERIMENT_NAME );
+
+		if ( $is_atomic_widgets_active ) {
+			$this->register_hooks();
+		}
+	}
+
+	public function register_hooks() {
```
TzviRabinovitch, line 32: does register_hooks have to be public?

PR #30401, opened 2025-03-02
Internal: Unselect removed widgets [ED-17913]

ronros-elementor:
Small fix to deselect widgets upon deletion, if they are in the `elementor.selection` items

Though the issue affects visually V4 elements (as the editing panel is not closed upon deletion) this fix is also relevant to V3, as the actual selection list is not emptied of deleted elements, both V3 and V4

### DIFF
--- a/assets/dev/js/editor/document/elements/commands/delete.js
+++ b/assets/dev/js/editor/document/elements/commands/delete.js
@@ -50,6 +50,10 @@ export class Delete extends $e.modules.editor.document.CommandHistoryBase {
 
 			container.model.destroy();
 			container.panel.refresh();
+
+			if ( elementor.selection.has( container ) ) {
+				$e.run( 'document/elements/deselect', { container } );
+			}
 		} );
 
 		if ( 1 === containers.length ) {


assets/dev/js/editor/document/elements/commands/delete.js
```diff
@@ -50,6 +50,10 @@ export class Delete extends $e.modules.editor.document.CommandHistoryBase {
 
 			container.model.destroy();
 			container.panel.refresh();
+
+			if ( elementor.selection.has( container ) ) {
```
Omerisra6, line 54: can we somehow test it?

Omerisra6 (changes requested):

assets/dev/js/editor/document/elements/commands/delete.js
ronros-elementor: It would be way too complicated based on what I saw in the current playwrite tests, as this area would need to be fully mocked, both the new editing panel and the `selection` module, relevant commands and all - too pricey for this bug fix in my opinion

Omerisra6 (approved):

PR #30410, opened 2025-03-03
Internal: Link in Link restriction > Determine whether an Element/Widget is Linkable [ED-18346]

baghdasarovelementor:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #


assets/dev/js/editor/utils/helpers.js
```diff
@@ -731,4 +731,20 @@ module.exports = {
...
+	 * @param {HTMLElement} element - The parent element to search within.
+	 * @return {HTMLAnchorElement | null} An anchor element with an 'href' attribute, or null if none are found.
+	 */
+	getAnchorWithHrefInElement( element ) {
+		if ( ! element ) {
+			return null;
+		}
+
+		return element.querySelector( 'a[href]' );
+	},
```
davseve, line 748: In the A.C of the task it is written that this task should identify the also the element parent link

assets/dev/js/editor/utils/helpers.js
```diff
@@ -731,4 +731,20 @@ module.exports = {
...
+	 *
+	 * @param {HTMLElement} element - The parent element to search within.
+	 * @return {HTMLAnchorElement | null} An anchor element with an 'href' attribute, or null if none are found.
+	 */
+	getAnchorWithHrefInElement( element ) {
+		if ( ! element ) {
+			return null;
+		}
+
+		return element.querySelector( 'a[href]' );
```
ronenelementor, line 747:
```suggestion
		return element?.querySelector( 'a[href]' ) || null;
```

asafdl: The [Document](https://developer.mozilla.org/en-US/docs/Web/API/Document) method querySelector() returns the first [Element](https://developer.mozilla.org/en-US/docs/Web/API/Element) within the document that matches the specified [CSS selector](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_selectors), or group of CSS selectors. **If no matches are found, null is returned**.

@ronenelementor @baghdasarovelementor

asafdl: this can return a group of elements aswell, should handle this case, and also test for it

asafdl: and the || null is redundant, if not found, will return null

asafdl: ah sorry misread, won't return a group of elements

baghdasarovelementor: @ronenelementor There's no need to write ``document.querySelector() || null`` because ``document.querySelector()`` already returns null if no matching element is found.

@asafdl 
``document.querySelector()`` returns only the first matching element from the DOM. It does not return multiple elements.
If multiple elements match the selector, only the first one appears in the result.
To get a group of elements, ``document.querySelectorAll() `` should be used instead, which returns a NodeList containing all matching elements.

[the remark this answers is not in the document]
baghdasarovelementor: We can call ``elementor.helpers.getAnchorWithHrefInElement(element)``
We can have another file for this but I don't know we need this in this point or not.

assets/dev/js/editor/utils/helpers.js
baghdasarovelementor: @davseve I changed the code.

assets/dev/js/editor/utils/helpers.js
ronenelementor: Just to avoid undefined or never if the element is nullish, but it's no big deal to me

assets/dev/js/editor/utils/helpers.js
```diff
@@ -731,4 +731,33 @@ module.exports = {
...
+	 *
+	 * @param {HTMLElement} element - The element to check.
+	 * @return {HTMLAnchorElement | null} The closest anchor element with an 'href' attribute, or null if none is found.
+	 */
+	getClosestAnchorWithHref( element ) {
+		if ( ! element ) {
+			return null;
+		}
+
+		return element.closest( 'a[href]' );
```
ronenelementor, line 761: I guess we can use the same shorthand here

ronenboxer (approved):

davseve (approved):

PR #30442, opened 2025-03-05
Internal: Post REST implementation fixes and updates

ronenboxer:


modules/wp-rest/classes/post-query.php
```diff
@@ -181,45 +167,39 @@ private function get_posts( \WP_REST_Request $request ) {
 	}
 
 	/**
-	 * Hooks into the flow of wordpress's get_post querying.
-	 *
 	 * @return void
 	 */
 	private function add_filter_to_customize_query() {
-		add_filter( 'posts_search', fn ( $search_term, $wp_query ) => $this->customize_search( $search_term, $wp_query ), 10, 2 );
+		add_filter( 'posts_search', [ $this, 'customize_search' ], 10, 2 );
```
eyalmrejen-elementor, line 173: What is 10, and what is 2?
I think it would be more readable if we have a meaningful constant here - now I need to go into the function to understand what the values mean.
So I googled it, priority,y, and accepted args. Had I had these named constants, it would have saved me the search, and children in Africa would have more internet.

modules/wp-rest/classes/post-query.php
```diff
@@ -181,45 +167,39 @@ private function get_posts( \WP_REST_Request $request ) {
...
 	}
 
 	/**
-	 * Hooks out of the flow of wordpress's get_post querying.
-	 *
 	 * @return void
 	 */
 	private function remove_filter_to_customize_query() {
-		remove_filter( 'posts_search', fn ( $search_term, $wp_query ) => $this->customize_search( $search_term, $wp_query ), 10, 2 );
+		remove_filter( 'posts_search', [ $this, 'customize_search' ], 10, 2 );
```
eyalmrejen-elementor, line 180: same as before

modules/wp-rest/classes/post-query.php
```diff
@@ -181,45 +167,39 @@ private function get_posts( \WP_REST_Request $request ) {
...
 	}
 
 	/**
-	 * Arguments for registering an endpoint.
-	 *
 	 * @return array
 	 */
 	private function get_args() {
```
eyalmrejen-elementor, line 186: Actually, `get_arges` is not as informative as the comment - I would think about renaming it to `get_endpoint_registration_args`()

modules/wp-rest/classes/post-query.php
```diff
@@ -181,45 +167,39 @@ private function get_posts( \WP_REST_Request $request ) {
...
 		return [
-			self::EXCLUDED_POST_TYPES_KEY => [
+			self::EXCLUDED_POST_TYPE_KEYS => [
```
eyalmrejen-elementor, line 188: Regarding this method ( I am not sure, so I am asking since you are better than me at PHP)
This method always returns the same structure - right? No variables. So what about not having it as private property and returning it and not having it computed each time this method is called? It's basically static data, no?
It's a small thing - no real need to change it - it was just a thought.

tests/phpunit/elementor/modules/wp-rest/base/post-query-base.php
```diff
@@ -0,0 +1,199 @@
...
+namespace Elementor\Tests\Phpunit\Elementor\Modules\WpRest\Base;
+
+use Elementor\Modules\WpRest\Classes\Post_Query;
+use ElementorEditorTesting\Elementor_Test_Base;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+class Post_Query_Base extends Elementor_Test_Base {
```
eyalmrejen-elementor, line 12: Since this is a testing class (and I see that it is already in a testing folder with the name `Base` in it - I see two issues
1. The folder name is `Base`, so why have it again in the class name
2. The name actually seems like a real production class - and when I `shift shift` in phpStorm from the name of it, it would not be immediately clear that it's a testing class and not one to use on production code
I would remove the name, or change it to something like `Post_Query_Test_Suite`? Now, it's very obvious what this class is, and it won't be selected by mistake

tests/phpunit/elementor/modules/wp-rest/base/post-query-base.php
```diff
@@ -0,0 +1,199 @@
...
+	protected array $post_types = [];
+	protected array $posts = [];
+
+	protected function clean() {
+		foreach ( $this->posts as $post ) {
+			wp_delete_post( $post->ID, true );
```
eyalmrejen-elementor, line 18: I hate boolean arguments in a function - since it basically does two things - but it's not our API, so we can't do anything, right? But I can suggest two things for readability
I don't know what the `true` value means - I needed to google it, so
1. use `FORCE_DELETE` constant with the value of `true`
2. Or Create and use a wrapper function to it - hiding the need for the boolean argument with `wp_force_delete_post()` - now the use of it is even simpler since the next developer to use it, understand from the name what it means, and has less arguments to use when calling it - less is better

eyalmrejen-elementor (dismissed):
Very nice and clear PR - nicly done
Just some small suggestions - fix and push

modules/wp-rest/classes/post-query.php
ronenelementor: Done

modules/wp-rest/classes/post-query.php
ronenelementor: Done

modules/wp-rest/classes/post-query.php
ronenelementor: Done

tests/phpunit/elementor/modules/wp-rest/base/post-query-base.php
ronenelementor: Done / This file was replaced and I forgot to remove it

tests/phpunit/elementor/modules/wp-rest/base/post-query-base.php
ronenelementor: Used FORCE const and ALL_POSTS const for -1

eyalmrejen-elementor (approved):

eyalmrejen-elementor (approved):

PR #30470, opened 2025-03-09
Tweak: Add method to enqueue global scripts [ED-18391]

KingYes:


includes/base/element-base.php
```diff
@@ -120,6 +120,12 @@ public function get_script_depends() {
 		return $this->depended_scripts;
 	}
 
+	public function get_global_scripts() {
+		return [
+			'elementor-frontend',
+		];
```
rami-elementor, line 126:
```suggestion
		return [ 'elementor-frontend' ];
```

StyleShit (dismissed):

modules/atomic-widgets/elements/atomic-widget-base.php
```diff
@@ -23,6 +23,10 @@ public function __construct( $data = [], $args = null ) {
 
 	abstract protected function define_atomic_controls(): array;
 
+	public function get_global_scripts() {
+		return [];
+	}
+
```
Nevoss, line 29: Move to `Has_Atomic` trait

tests/phpunit/elementor/includes/test-frontend.php
```diff
@@ -118,4 +137,35 @@ public function test_get_early_access_google_font_urls() {
 		// Assert
 		$this->assertContains( 'https://fonts.googleapis.com/earlyaccess/opensanshebrew.css', $font_urls );
 	}
+
+	public function test_widget_with_global_scripts() {
+		// Arrange
+		$frontend = new Frontend();
+		$frontend->register_scripts();
+
+		$widget_with_global_scripts = $this->elementor()->widgets_manager->get_widget_types( 'text-editor' );
```
Nevoss, line 146: Maybe mention with comment or with the variable name, that this widget should be V3 widget and it does not matter which widget it should be. also for the V4 widget

tests/phpunit/elementor/includes/test-frontend.php
```diff
@@ -118,4 +137,35 @@ public function test_get_early_access_google_font_urls() {
...
+		$this->assertTrue( wp_script_is( 'jquery' ) );
+	}
+
+	public function test_widget_without_global_scripts() {
+		// Arrange
+		$frontend = new Frontend();
+		$frontend->register_scripts();
+
+		// Act
+		$widget_without_global_scripts = $this->elementor()->widgets_manager->get_widget_types( 'a-heading' );
```
Nevoss, line 163: Do we have some test for the config that should be printed with the `frontend` script?

Nevoss, line 163: Do we have some test for the config that should be printed with the `frontend` script?

Nevoss (changes requested):

Nevoss (approved):

PR #30499, opened 2025-03-12
POC: Atomic Import Export

Nevoss:


includes/template-library/sources/local.php
```diff
@@ -534,8 +534,13 @@ public function save_item( $template_data ) {
 			$template_data['content'] = $this->replace_elements_ids( $template_data['content'] );
 		}
 
+		$elements = apply_filters(
```
louiswol94, line 537: Do we need to check the cloud source as well?

PR #30522, opened 2025-03-13
Internal: Create fallback transformer [ED-18502]

Nevoss:


modules/atomic-widgets/props-resolver/transformers-registry.php
```diff
@@ -21,4 +23,14 @@ public function register( string $key, Transformer_Base $transformer ): self {
...
+
+	public function register_fallback( Transformer_Base $transformer ): self {
+		$this->fallback = $transformer;
+
+		return $this;
+	}
+
+	public function get( $key, $fallback = null ) {
+		return parent::get( $key, $fallback ) ?? $this->fallback;
+	}
```
StyleShit, line 35:
```suggestion
	public function get( $key ) {
		return parent::get( $key, $this->fallback );
	}
```

StyleShit (changes requested):

StyleShit (approved):

PR #30552, opened 2025-03-16
Internal: [V4] Div block height blocks user's min height [ED-18499]

ronenboxer:


Nevoss (approved):

modules/atomic-widgets/elements/div-block/div-block.php
```diff
@@ -176,7 +176,7 @@ protected function define_base_styles(): array {
 					Style_Variant::make()
 						->add_prop( 'display', $display_value )
 						->add_prop( 'padding', $padding_value )
-						->add_prop( 'height', $height_value )
+						->add_prop( 'min-height', $min_height_value )
```
eyalmrejen-elementor, line 179: I see that this is the convention in the previous lines, but the suffix of value feels redundant
You don't write
```
const User = {
   nameValue: "Ronen"
}
```

ronenelementor: I have to agree

PR #30559, opened 2025-03-17
Internal: V4 Remove extra margin in the atomic Heading [ED-18422]

TzviRabinovitch:


modules/atomic-widgets/elements/atomic-heading/atomic-heading.php
```diff
@@ -115,6 +115,7 @@ protected function define_base_styles(): array {
 						->add_prop( 'font-size', $font_size_value )
 						->add_prop( 'line-height', $line_height_value )
 						->add_prop( 'font-weight', $font_weight_value )
+						->add_prop( 'margin', 0 )
```
hein-obox, line 118: I would only overwrite the default margin block styles, which are for the top and bottom:

```suggestion
						->add_prop( 'margin-block', 0 )
```

hein-obox (dismissed):

hein-obox (dismissed):

hein-obox (approved):

PR #30568, opened 2025-03-17
Internal: Fix div block empty view display [ED-18499]

ronenboxer:


murzin-nikita (dismissed):

assets/dev/scss/editor/preview/_container.scss
StyleShit, line 1: why all the code is inside the container css file?

ronenelementor: This is where it was located
Do you want me to create a file for the div-block?

assets/dev/scss/editor/preview/_container.scss
```diff
@@ -87,8 +87,13 @@
 		}
 	}
 
-	&.e-div-block-base {
+	&.div-block-base {
```
mykytamurzin, line 90: should use `e-div-block-base` here

StyleShit (approved):

PR #30578, opened 2025-03-18
Fix: Ensure the page scrolls to the added widget after clicking to add a widget [ED-18260]

MiryamOren:
https://github.com/user-attachments/assets/735b2317-bbe8-4c10-b1d2-46cfb7e7640d


assets/dev/js/editor/elements/views/base.js
```diff
@@ -1085,7 +1085,7 @@ BaseElementView = BaseContainer.extend( {
 		return helper;
 	},
 
-	getDraggableElement() {
+	getElement() {
```
Nevoss, line 1088: Maybe `getDOMElement` because `Element` is also Elementor element and we have function that called `getElementType`

Nevoss (changes requested):

Nevoss (approved):

PR #30580, opened 2025-03-18
Fix: SVG default not working [ED-18570]

Nevoss:


modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
```diff
@@ -109,11 +113,52 @@ protected function render() {
...
+	private static function get_default_svg( string $context ) {
+		if ( 'path' === $context ) {
+			return ELEMENTOR_ASSETS_PATH . self::DEFAULT_SVG;
+		}
+
+		if ( 'url' === $context ) {
+			return ELEMENTOR_ASSETS_URL . self::DEFAULT_SVG;
+		}
+
+		return null;
```
StyleShit, line 162: why not just making these 2 consts at the top of the class?

```PHP
const DEFAULT_SVG_PATH = ELEMENTOR_ASSETS_PATH . self::DEFAULT_SVG;
const DEFAULT_SVG_URL = ELEMENTOR_ASSETS_URL . self::DEFAULT_SVG;
```

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
```diff
@@ -109,11 +113,52 @@ protected function render() {
...
 
-	private function set_svg_attributes( \WP_HTML_Tag_Processor $svg, $settings ) {
-		$svg->set_attribute( 'fill', 'currentColor' );
-		$string_classes = implode( ' ', $settings['classes'] );
-		$svg->add_class( $string_classes );
-		$base_styles = $this->get_base_styles_dictionary()[ self::BASE_STYLE_KEY ];
-		$svg->add_class( $base_styles );
+	private function get_svg_content( $settings ) {
+
+		if ( isset( $settings['svg']['id'] ) ) {
```
StyleShit, line 118:
```suggestion
		if ( isset( $settings['svg']['id'] ) ) {
```

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
```diff
@@ -109,11 +113,52 @@ protected function render() {
...
+			);
+
+			if ( ! is_wp_error( $content ) ) {
+				return $content['body'];
+			}
+		}
+
+		$content = Utils::file_get_contents(
+			static::get_default_svg( 'path' )
+		);
```
StyleShit, line 143: should we fallback to the default if the user has set an invalid value?
I mean... it's like setting a default value when there is 404 from an image URL

I think the default should be set only when it came as a default from the schema

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
```diff
@@ -81,25 +81,29 @@ protected function define_base_styles(): array {
...
+
+		$classes = array_filter( array_merge(
+			[ $this->get_base_styles_dictionary()[ self::BASE_STYLE_KEY ] ],
+			$settings['classes']
+		) );
 
-		$svg_html = $svg_html ?? file_get_contents( self::DEFAULT_SVG_PATH );
+		$svg->add_class( implode( ' ', $classes ) );
+
+		$svg_html = ( new Svg_Sanitizer() )->sanitize( $svg->get_updated_html() );
```
StyleShit, line 106: can we move it to a transformer already?

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
```diff
@@ -109,11 +113,52 @@ protected function render() {
...
+
+		return $content ? $content : null;
+	}
+
+	/**
+	 * @param string $context 'url' | 'path'
+	 *
+	 * @return string
+	 */
+	private static function get_default_svg( string $context ) {
```
StyleShit, line 153: Well, technically:

```suggestion
	 * @return string | null
	 */
	private static function get_default_svg( string $context ) {
```

and you can also:

```suggestion
	private static function get_default_svg( string $context ): ?string {
```

But I this whole function might be deleted based on one of my previous comments, so... 🤷

tests/phpunit/elementor/modules/atomic-widgets/test-atomic-svg.php
```diff
@@ -15,16 +15,16 @@ class Test_Atomic_Svg extends Elementor_Test_Base {
...
+		add_filter( 'get_attached_file', function( $path, $attachment_id ) {
 			if ( $attachment_id === 123 ) {
-				return [ self::TEST_RESOURCES_DIR . 'test.svg' ];
+				return self::TEST_RESOURCES_DIR . 'test.svg';
 			}
 
-			return $image;
-		}, 10, 3 );
+			return $path;
+		}, 10, 2 );
```
StyleShit, line 24:
```suggestion
		} );
```

StyleShit (changes requested):

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
Nevoss: Done

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
Nevoss: 😮 Done

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
Nevoss: Done

modules/atomic-widgets/elements/atomic-svg/atomic-svg.php
Nevoss: Deleted

tests/phpunit/elementor/modules/atomic-widgets/test-atomic-svg.php
Nevoss: I need the second arg, so must set the number of args

StyleShit: For some reason I thought 10,2 is the default (which doesn't make sense actually...)

never mind

StyleShit (approved):

PR #30632, opened 2025-03-24
Tweak: Add "Copy to" functionality on a template [ED-18340]

louiswol94:


assets/dev/js/editor/components/template-library/views/parts/save-template.js
```diff
@@ -72,10 +76,21 @@ const TemplateLibrarySaveTemplateView = Marionette.ItemView.extend( {
...
+	handleCopyContextUiState() {
+		this.ui.templateNameInput.val( this.model.get( 'title' ) );
+
+		const fromSource = this.model.get( 'source' );
+
+		if ( 'local' === fromSource ) {
+			this.$( '.source-selections-input #cloud' ).prop( 'checked', true );
+			this.ui.localInput.addClass( 'disabled' );
+		}
+	},
```
ManorHazaz, line 88: Can it be the same as `handleContextUiStateChecboxes`?

includes/template-library/manager.php
```diff
@@ -362,27 +365,51 @@ private function move_template_item( array $args ) {
...
+
+		return $source->get_item( $template_id );
 	}
 
 	private function is_action_to_same_source( $args ) {
 		return $args['source'] === $args['from_source'];
 	}
 
-	private function format_args_for_move_template_from_local_to_cloud( $args ) {
+	private function format_args_for_single_action_from_local_to_cloud( $args ) {
```
ManorHazaz, line 402: I'm wondering if we can move it to the source...
Feels weird when the manager knows the sources type.

includes/template-library/manager.php
```diff
@@ -397,7 +424,7 @@ private function format_args_for_move_template_from_local_to_cloud( $args ) {
 		return $args;
 	}
 
-	private function format_args_for_move_template_from_cloud_to_local( $args ) {
+	private function format_args_for_single_action_from_cloud_to_local( $args ) {
```
ManorHazaz, line 427: Same here

includes/template-library/manager.php
```diff
@@ -1108,6 +1136,13 @@ private function format_args_for_bulk_action_from_local( $args ) {
 		$bulk_args = [];
 
 		foreach ( $args['from_template_id'] as $from_template_id ) {
+			if ( ! $this->is_allowed_to_read_template( [
```
ManorHazaz, line 1139: Nice catch

ManorHazaz (changes requested):

includes/template-library/manager.php
louiswol94: @ManorHazaz I agree! But this adds complications to the permissions checks which I think is more important at this stage. I opened ED-18647 as a tech debt for this.

ManorHazaz (approved):

PR #30634, opened 2025-03-24
Tweak: Improve social-icons widget accessibility with `role` attributes [ED-18648]

rami-elementor:
Before:

image

After:

image

### DIFF
--- a/includes/widgets/social-icons.php
+++ b/includes/widgets/social-icons.php
@@ -607,7 +607,7 @@ class Widget_Social_Icons extends Widget_Base {
 		$migration_allowed = Icons_Manager::is_migration_allowed();
 
 		?>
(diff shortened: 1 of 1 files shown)


includes/widgets/social-icons.php
```diff
@@ -607,7 +607,7 @@ protected function render() {
 		$migration_allowed = Icons_Manager::is_migration_allowed();
 
 		?>
-		<div class="elementor-social-icons-wrapper elementor-grid">
+		<div class="elementor-social-icons-wrapper elementor-grid" role="list">
```
arielk, line 610: @rami-elementor same question like in the share-buttons, what if the user add only one button

rami-elementor: @arielk Fixed.

arielk (approved):

PR #30643, opened 2025-03-25
Internal: [V4] Remove URL structure validation (without hurting security measures) [ED-18513]

baghdasarovelementor:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #


murzin-nikita (dismissed):

tests/phpunit/elementor/modules/atomic-widgets/prop-types/test-url-prop-type.php
```diff
@@ -0,0 +1,58 @@
...
+	public function valid_urls_provider()
+	{
+		return [
+			[ "url" => "google.com" ],
+			[ "google" ],
+			[ "google.com/about-us/#section1" ],
+			[ "http://example.com" ],
+			[ "https://example.com" ],
+			[ "192.168.1.1" ],
+			[ "2001:db8::ff00:42:8329" ],
```
davseve, line 43: Can we add a `#`?

baghdasarovelementor: No only # we can't

davseve: I think it is one of the task requirements
@doronwolf wdyt?

davseve: @baghdasarovelementor 
We must support the `#` as a link

doronwolf: Hi @baghdasarovelementor - just as @davseve wrote - we must support `#` as well (it might be used as an anchor like `#anchor-name` and sometimes just `#` on it's on (which will cause to browser to scroll to the top of the page)

tests/phpunit/elementor/modules/atomic-widgets/prop-types/test-url-prop-type.php
```diff
@@ -0,0 +1,58 @@
...
+	public function test_validate__fail_when_value_is_not_a_allowed_url( $url ) {
+		// Act.
+		$result = Url_Prop_Type::validate_url( $url );
+
+		// Assert.
+		$this->assertFalse( $result );
+	}
+
+	public function valid_urls_provider()
+	{
```
davseve, line 35:
```suggestion
	public function valid_urls_provider() {
```

tests/phpunit/elementor/modules/atomic-widgets/prop-types/test-url-prop-type.php
davseve: Can we add the `#` scenario to the test?

davseve:
https://elementor.slack.com/files/U06Q4JCV2P7/F08KAFT5NDB/screen_recording_2025-03-26_at_11.59.12.mov

baghdasarovelementor:
Screen record how it works.
https://drive.google.com/file/d/1wVsSOBC2sBw3rTbt1mrjHzSYNlFA2hnT/view?usp=sharing

modules/atomic-widgets/prop-types/link-control-url-prop-type.php
```diff
@@ -0,0 +1,34 @@
...
+class Link_Control_Url_Prop_Type extends Plain_Prop_Type {
+	public static function get_key(): string {
+		return 'url';
+	}
+
+	public static function validate_url( $value ): bool {
+		$url = trim( $value );
+
+		// Single regex to handle full URLs, domains, IPs, paths, and fragments
+		if ( preg_match( '/^(https?:\/\/[a-zA-Z0-9\-\.]+(\.[a-zA-Z]{2,})?|#([\w\-]*)?|([a-zA-Z0-9-]+(\.[a-zA-Z]{2,})?|(\d{1,3}\.){3}\d{1,3}|[a-fA-F0-9:]+)?(\/[\w\-\/]*)?(#[\w\-]*)?)$/', $url ) ) {
```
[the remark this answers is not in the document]
mykytamurzin: I think it will be easier to maintain if we will separate this checks case by case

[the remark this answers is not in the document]
baghdasarovelementor: @mykytamurzin updated. Can you check new version is more understandable or not?

murzin-nikita (dismissed):

modules/atomic-widgets/prop-types/link-control-url-prop-type.php
```diff
@@ -0,0 +1,49 @@
...
+
+		// Allow full URLs with scheme (http, https)
+		if ( filter_var( $url, FILTER_VALIDATE_URL ) !== false ) {
+			return $url;
+		}
+
+		// Allow fragments (e.g., "#", "#my-div")
+		if ( preg_match( '/^#([\w\-]*)?$/', $url ) ) {
+			return $url;
+		}
```
ronenelementor, line 27: I would put the expression in a `$fragment_hash_only_pattern = `'/^#([\w\-]*)?$/'`

tests/phpunit/elementor/modules/atomic-widgets/prop-types/test-link-control-url-prop-type.php
```diff
@@ -0,0 +1,57 @@
...
+	 */
+	public function test_validate( $url ) {
+		// Act.
+		$result = Link_Control_Url_Prop_Type::validate_url( $url );
+
+		// Assert.
+		$this->assertTrue( $result );
+	}
+
+	/** @dataProvider invalid_urls_provider */
```
ronenelementor, line 25:
```suggestion
	/**
	 *	@dataProvider invalid_urls_provider
	 */
```

tests/phpunit/elementor/modules/atomic-widgets/prop-types/test-link-control-url-prop-type.php
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	public function invalid_urls_provider() {
+		return [
+			[ "example.com/?id=1<script>alert(1)</script>" ], // XSS injection
+			[ "' OR 1=1 --" ], // SQL injection
+			[ "<script>alert('hacked')</script>" ], // JavaScript injection
+			[ "http://" ],
+			[ "http:///example.com" ],
+		];
```
ronenelementor, line 55:
```suggestion
		$xss_injected_url = 'example.com/?id=1<script>alert(1)</script>';
		$sql_injected_url = "' OR 1=1 --";
		$js_injected_url = "<script>alert('hacked')</script>";

		return [
			[ xss_injected_url ],
			[ sql_injected_url ],
			[ js_injected_url ],
			[ 'http://' ],
			[ 'http:///example.com' ],
		];
```

ronenboxer (changes requested):

ronenboxer (dismissed):

modules/atomic-widgets/prop-types/link-control-url-prop-type.php
```diff
@@ -0,0 +1,65 @@
+<?php
+
```
asafdl, line 2: I have an issue with trying to solve it like this, I think a better approach is just to take a regex https://mathiasbynens.be/demo/url-regex
and test it.
however, url validation is a very tricky business, it seems that almost anything you do will violate some browser or some convention and will give false positive or false negatives.
after speaking with @doronwolf I think that the major issue that this task has, is no false positives, and no malicious code.
if the user has inserted a bad url, we shouldn't stop him.

so that being said i think you should `trim` and `filter_var($url, FILTER_SANITIZE_URL);` and move this param forward, no need to validate for correct URL

doronwolf: Hi @baghdasarovelementor - I want to reinforce the words of @asafdl 
If the user has inserted a bad URL, we shouldn't stop him, we should keep sanitizing the input and make sure the user can't insert malicious code, but we don't mind if he enters a URL with a bad structure

modules/atomic-widgets/prop-types/link-control-url-prop-type.php
```diff
@@ -0,0 +1,29 @@
...
+class Link_Control_Url_Prop_Type extends Plain_Prop_Type {
+	public static function get_key(): string {
+		return 'url';
+	}
+
+	public static function validate_url( $value ): bool {
+		return true;
+	}
+
+	protected function validate_value( $value ): bool {
```
asafdl, line 20: do we have to implement an empty validate function?

tests/phpunit/elementor/modules/atomic-widgets/prop-types/test-link-control-url-prop-type.php
```diff
@@ -0,0 +1,48 @@
...
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+class Test_Link_Control_Url_Prop_Type extends Elementor_Test_Base {
+
+	/**
+	 *	@dataProvider urls_provider
+	 */
+	public function test_validate( $url ) {
```
asafdl, line 17: tests should check for sanitize not validate since its an empty function

modules/atomic-widgets/prop-types/link-control-url-prop-type.php
baghdasarovelementor: I want to keep this and the test for the future.

davseve: We don't need to validate the value.
Only to sanitize the data that is added to the input field.
The initial definition of this task was misleading.

If we will need this functionality in the future, we will ad it. (YAGNI)

asafdl (dismissed):

modules/atomic-widgets/prop-types/link-control-url-prop-type.php
```diff
@@ -0,0 +1,29 @@
...
+	exit; // Exit if accessed directly.
+}
+
+class Link_Control_Url_Prop_Type extends Plain_Prop_Type {
+	public static function get_key(): string {
+		return 'url';
+	}
+
+	public static function validate_url( $value ): bool {
+		return true;
```
ronenelementor, line 17: Doesn't this need to br reverted back to the original wp url check?

baghdasarovelementor: No the thing is allow every type of URL

ronenboxer (dismissed):

modules/atomic-widgets/prop-types/link-control-url-prop-type.php
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Elementor\Modules\AtomicWidgets\PropTypes;
+
+use Elementor\Modules\AtomicWidgets\PropTypes\Base\Plain_Prop_Type;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+class Link_Control_Url_Prop_Type extends Plain_Prop_Type {
```
eyalmrejen-elementor, line 11: Why do u need Prop_Type in the class name when it's in the namespace and inheritance chain as well?

modules/atomic-widgets/prop-types/link-control-url-prop-type.php
```diff
@@ -0,0 +1,29 @@
...
+	public static function get_key(): string {
+		return 'url';
+	}
+
+	public static function validate_url( $value ): bool {
```
eyalmrejen-elementor, line 16: Why public static? This opens the class to the outside - do you really need it?
The rule of thumb is always to start with private and open only as needed - otherwise, what does OPP mean?

Nevoss (dismissed):

modules/atomic-widgets/prop-types/link-control-url-prop-type.php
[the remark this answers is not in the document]
baghdasarovelementor: `URL_Prop_Type ` Used on different validations for example image url validation.

baghdasarovelementor: I keep the same structure we had before.

modules/atomic-widgets/prop-types/link-control-url-prop-type.php
baghdasarovelementor: @eyalmrejen-elementor Fixed

modules/atomic-widgets/prop-types/link-control-url-prop-type.php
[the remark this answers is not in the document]
baghdasarovelementor: But is it not a issue allow "#" as a image URL?

modules/atomic-widgets/prop-types/link-control-url-prop-type.php
```diff
@@ -0,0 +1,25 @@
...
+namespace Elementor\Modules\AtomicWidgets\PropTypes;
+
+use Elementor\Modules\AtomicWidgets\PropTypes\Base\Plain_Prop_Type;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+class Link_Control_Url_Prop_Type extends Plain_Prop_Type {
+	public static function get_key(): string {
```
eyalmrejen-elementor, line 12: Same as the `validate_value`? Do we need it `public static`?

baghdasarovelementor: Yes we need them both. They are abstract methods.

modules/atomic-widgets/prop-types/link-control-url-prop-type.php
eyalmrejen-elementor: Why should we perpetuate this pattern if it's false - the next dev will copy it again, and again,  and our code base will never get better

modules/atomic-widgets/prop-types/link-control-url-prop-type.php
eyalmrejen-elementor: I mean, why have them `public static` - it's the same fix you did for line 16 where you agreed it was better - is this not the same case?

baghdasarovelementor: @eyalmrejen-elementor We can create Tweak task for this. But on `Plain_Prop_Type` class we have abstract method `abstract public static function get_key(): string;`. I don't know why we did it. @Nevoss Can answer better. If I not mistaken he is created the structure.

modules/atomic-widgets/prop-types/link-control-url-prop-type.php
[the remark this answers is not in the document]
ronros-elementor: The idea is that we could add some url-specific configuration, similar to how we've extended the string prop type, for example -
```
String_Prop_Type::make()->enum([
	...
])
```

The string prop type class has a unique enum method to add constraints to the value

WDYT about such an approach? To enrich the options for the url prop type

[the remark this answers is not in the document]
baghdasarovelementor: You mean modify sanitize method rule with unique method? @ronros-elementor

modules/atomic-widgets/prop-types/link-control-url-prop-type.php
Nevoss: We want to get the key of the `Prop_Type` without the need to create a new instance. the `key` is really `static` it should not be changed between instances This is really an "identifier" for `Prop_Type`.

modules/atomic-widgets/prop-types/link-control-url-prop-type.php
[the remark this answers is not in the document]
ronros-elementor: What that unique method would do is, for example, to set a private member of `Url_Prop_Type`, so it would look something like -
```
class Url_Prop_Type extends Plain_Prop_Type {
	public static function get_key(): string {
		return 'url';
	}

	public static function loose_validation(): self {
		$this->settings['is_loose'] = true;

	public static function validate_url( $value ): bool {
		if( $this->settings['is_loose'] ) {
			return true;
		}

		return (bool) wp_http_validate_url( $value );
	}

	...
}
```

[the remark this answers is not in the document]
baghdasarovelementor: I think it is creating unnecessary complexity.

ronros-elementor (approved):

ronenboxer (approved):

PR #30687, opened 2025-03-27
Fix: ED-18634 Remove "Save as default" from context-menu for V4 elements

YaelAvitan:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #


modules/kit-elements-defaults/assets/js/editor/module.js
```diff
@@ -30,6 +30,11 @@ export default class Module extends elementorModules.editor.utils.Module {
 						return group;
 					}
 
+					const widgetType = view.options?.model.get( 'widgetType' );
```
hein-obox, line 33:
```suggestion
					const isAtomicWidget = view.options?.model.get( 'widgetType' )?.startsWith( 'e-' );
```

modules/kit-elements-defaults/assets/js/editor/module.js
```diff
@@ -30,6 +30,11 @@ export default class Module extends elementorModules.editor.utils.Module {
...
+					if ( widgetType && widgetType.startsWith( 'e-' ) ) {
```
hein-obox, line 34:
```suggestion
					if ( isAtomicWidget ) {
```

hein-obox (approved):

modules/kit-elements-defaults/assets/js/editor/module.js
```diff
@@ -30,6 +30,11 @@ export default class Module extends elementorModules.editor.utils.Module {
 						return group;
 					}
 
+					const isAtomicWidget = view?.options?.model?.get( 'widgetType' )?.startsWith( 'e-' );
```
KingYes, line 33: @YaelAvitan @hein-obox I'm not sure this is the correct way to check if it's an Atomic element. The `e-` prefix doesn't mean anything!

KingYes (changes requested):

modules/kit-elements-defaults/assets/js/editor/module.js
hein-obox: @KingYes That's a good point. I will think about of an alternative approach as well.

YaelAvitan: I talked to someone else on the team who knows this area of ​​code better
and he recommended this way
It actually works
Do you want to check it out yourself?

hein-obox: @YaelAvitan There are potential issues with the implementation.

E.g.
A 3rd party plugin could be using the 'e-' for their own widgetTypes. That could cause problems.
In future it might be possible that our own Atoms will start with another prefix, e.g. 'en-' etc.
We need to keep in mind that our solution need to so robust that it still can be used in 10 years from now.

I will discuss possible other alternatives with Batsirai tomorrow.

hein-obox: @YaelAvitan 

If my PR will approved, you should be able to check if a widget is 'atomic' by using this code, but I am not sure yet if it will be approved by the core team.

`view.options?.model?.get('atomic')`

KingYes: It looks better, but keep in mind that it’s not part of atomic-element-base either

StyleShit: This module shouldn't know about atomic anyway
The action removal should happen from outside

hein-obox: @StyleShit This PR is no longer relevant, because we have handled this inside packages already. I have closed this PR. 

Apologies for the confusion.

PR #30759, opened 2025-04-01
Internal: Remove unnecessary sanitization comment

Omerisra6:


modules/global-classes/global-classes-rest-api.php
```diff
@@ -27,9 +27,6 @@ private function get_repository() {
 		return $this->repository;
 	}
 
-	/**
-	 * TODO: Add sanitization when implemented on prop types [EDS-574]
-	 */
```
IshayMaya, line 32: why?

IshayMaya (approved):

modules/global-classes/global-classes-rest-api.php
Omerisra6: Angry Season 4 GIF by The Office

PR #30785, opened 2025-04-02
Internal: V4 Variables: prop-type - transformer [ED-18669]

mike-elementor:
* allow to register the prop-type - transformer registration when the experiment is active (enabled)

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #


modules/variables/prop-types/color-variable-prop-type.php
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Elementor\Modules\Variables\PropTypes;
+
+use Elementor\Modules\AtomicWidgets\PropTypes\Base\Plain_Prop_Type;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+class Color_Variable_Prop_Type extends Plain_Prop_Type {
```
eyalmrejen-elementor, line 11: maybe not use `Prop_Type` again in the class name? We have it in the import line and the `extends` one

modules/variables/prop-types/color-variable-prop-type.php
```diff
@@ -0,0 +1,23 @@
...
+	public static function get_key(): string {
+		return 'global-color-variable';
+	}
+
+	protected function validate_value( $value ): bool {
+		return ! empty( $value );
+	}
+
+	protected function sanitize_value( $value ) {
+		return $value;
```
eyalmrejen-elementor, line 21: what is this?

eyalmrejen-elementor (dismissed):

modules/variables/hooks.php
```diff
@@ -0,0 +1,23 @@
...
+		add_action( 'elementor/atomic-widgets/styles/transformers/register', function ( $transformers ) {
+			$this->register_style_transformers( $transformers );
+		} );
+
+		return $this;
+	}
+
+	private function register_style_transformers( $transformers ) {
+		( new Style_Schema() )->append_to( $transformers );
+	}
```
Omerisra6, line 22: I would delete that and the style schema file and put that inline

modules/variables/module.php
```diff
@@ -36,6 +36,8 @@ public function __construct() {
 		if ( ! $this->is_experiment_active() ) {
 			return;
 		}
+
+		( new Hooks() )->register();
```
Omerisra6, line 40: I would define a `register_hooks` method at the module file (that's how we usually do that)

modules/variables/prop-types/color-variable.php
```diff
@@ -0,0 +1,23 @@
...
+		return 'global-color-variable';
+	}
+
+	protected function validate_value( $value ): bool {
+		return ! empty( $value );
+	}
+
+	protected function sanitize_value( $value ) {
+		return $value;
+	}
```
Omerisra6, line 22: I think we should define the structure here and do a real sanitization.

It's not clear from the prop type what we expect the value to be.

Omerisra6 (changes requested):

modules/variables/hooks.php
Omerisra6: Still, I think these are very general components and should be visible in the module.

The Transformers have no connection to the style schema, it's a different component in our system.

modules/variables/module.php
Omerisra6: I know it's more of a styling thing, but, in my opinion, we should keep the same styling standard so that things look familiar to other developers.

But, again it's more of a styling comment, so if you still want to go with your approach I will resolve that.

modules/variables/prop-types/color-variable.php
Omerisra6: These values should still be sanitized, a user can send malicious values when the experiment is on, and we need to prevent them from being saved.

Omerisra6 (changes requested):

Omerisra6 (dismissed):

Omerisra6 (approved):

PR #30790, opened 2025-04-03
Internal: Set maximum amount of global classes creation [ED-18642]

IshayMaya:


modules/global-classes/global-classes-repository.php
```diff
@@ -54,6 +56,12 @@ public function put( array $items, array $order ) {
 			return;
 		}
 
+		$items_count = count( $updated_value['items'] );
+
+		if ( $items_count > static::MAX_ITEMS ) {
```
Omerisra6, line 61: add tests

Omerisra6 (changes requested):

modules/global-classes/global-classes-rest-api.php
```diff
@@ -119,6 +121,18 @@ private function put( \WP_REST_Request $request ) {
 		$items_result = $parser->parse_items(
 			$request->get_param( 'items' )
 		);
+		
+		$items_count = count( $items_result->unwrap() );
+
+		if ( $items_count > static::MAX_ITEMS ) {
```
Omerisra6, line 127:
```suggestion
		if ( $items_count >= static::MAX_ITEMS ) {
```

Omerisra6 (changes requested):

Omerisra6 (approved):

PR #30850, opened 2025-04-08
Internal: Update global classes usage [ED-18798]

Omerisra6:


modules/global-classes/usage/applied-global-classes-usage.php
```diff
@@ -18,34 +18,38 @@ class Applied_Global_Classes_Usage {
...
 
-		foreach ( $elementor_posts as $post ) {
-			$document = Plugin::$instance->documents->get( $post->ID );
-			$elements_data = $document->get_json_meta( Document::ELEMENTOR_DATA_META_KEY );
+		if ( empty( $global_class_ids ) ) {
+			return [];
+		}
 
-			$count_per_type = $this->get_classes_count_per_element_type( $elements_data, $global_class_ids );
+		Plugin::$instance->db->iterate_elementor_documents( function( $document, $elements_data ) use ( &$total_count_per_class_id, $global_class_ids ) {
```
IshayMaya, line 31: maybe extract it to a function?

modules/global-classes/usage/applied-global-classes-usage.php
```diff
@@ -54,44 +58,35 @@ private function get_classes_count_per_element_type( $elements_data, $global_cla
...
 			/** @var Atomic_Element_Base | Atomic_Widget_Base $element_instance */
-			$classes_count = $this->get_classes_count_for_element( $element_instance->get_props_schema(), $element_data, $global_class_ids );
+			$applied_classes_per_element = $this->get_applied_global_classes_per_element( $element_instance->get_props_schema(), $element_data, $global_class_ids );
 
-			if ( 0 !== $classes_count ) {
-				$count_per_type[ $element_type ] ??= 0;
-				$count_per_type[ $element_type ] += $classes_count;
+			foreach ( $applied_classes_per_element as $global_class_id => $count ) {
+				$count_per_class[ $global_class_id ] ??= 0;
+				$count_per_class[ $global_class_id ] += $count;
```
IshayMaya, line 65:
```suggestion
				$count_per_class[ $global_class_id ] ??= $count;
```
can't we just do this?

modules/global-classes/usage/applied-global-classes-usage.php
Omerisra6: I only use that once

modules/global-classes/usage/applied-global-classes-usage.php
Omerisra6: No, because we are adding it and not setting it

IshayMaya (dismissed):

IshayMaya (approved):

PR #30881, opened 2025-04-14
Tweak: List and Grid sorters [ED-18486]

louiswol94:


assets/dev/js/editor/components/template-library/views/parts/templates.js
```diff
@@ -316,6 +322,32 @@ const TemplateLibraryCollectionView = Marionette.CompositeView.extend( {
...
+
+		elementor.templates.loadMore( {
+			onUpdate: () => {
+				elementor.templates.layout.hideLoadingView();
+			},
+			search: this.ui.textFilter.val(),
+			orderby: by,
+			order,
+			offset: 0,
+			limit: totalCount,
```
max-zu, line 347: why do we need to pass all these parameters if they are available from the closure of the loadMore method (in manager.js) ? 🤔

assets/dev/js/editor/components/template-library/manager.js
```diff
@@ -846,16 +854,20 @@ const TemplateLibraryManager = function() {
...
-				templatesCollection.add( collection.models, { merge: true } );
-
-				self.layout.addTemplates( collection.models );
+				if ( 0 === offset ) {
+					templatesCollection.reset( collection.models );
+					self.layout.updateViewCollection( templatesCollection.models );
+				} else {
+					templatesCollection.add( collection.models, { merge: true } );
+					self.layout.addTemplates( collection.models );
+				}
```
max-zu, line 870: The loadMore function should only load more items and add them to the collection. I think this logic should be moved to the`loadTemplates` and `searchTemplates` functions.
Also, orderby, order, offset, and limit should be taken from the closure, not passed from outside.

louiswol94: Discussed that we need to use `loadMore` because for BC we cannot change the signature of the PHP method that `loadTemplates` goes to.

max-zu (approved):

PR #30891, opened 2025-04-15
Tweak: Added accessibility statement notice [ED-19046] [APP-1302]

bainternet:


core/admin/admin-notices.php
```diff
@@ -654,12 +703,66 @@ public function admin_notices() {
...
+		$campaign_data = [
+			'source' => sanitize_key( $_GET['plg_source'] ),
+			'campaign' => sanitize_key( $_GET['plg_campaign'] ),
+			'medium' => sanitize_key( $_GET['plg_medium'] ),
+		];
+
+		set_transient( sanitize_key( $_GET['plg_campaign_name'] ), $campaign_data, 30 * DAY_IN_SECONDS );
+	}
+
+	public static function add_plg_campaign_data( $url, $campaign_data ) {
```
KingYes, line 744: Why is it public?

core/admin/admin-notices.php
```diff
@@ -654,12 +703,66 @@ public function admin_notices() {
...
+		return add_query_arg( [
+			'plg_campaign_name' => $campaign_data['name'],
+			'plg_campaign' => $campaign_data['campaign'],
+			'plg_source' => $campaign_data['source'],
+			'plg_medium' => $campaign_data['medium'],
+			'plg_campaign_nonce' => wp_create_nonce( $campaign_data['name'] ),
```
KingYes, line 750: I don't see any of the checks variable before adding to the array

core/admin/admin-notices.php
```diff
@@ -654,12 +703,66 @@ public function admin_notices() {
 		}
 	}
 
+	/**
+	 * Used to log campaigns for plugins
+	 */
```
KingYes, line 708: Remove ALL your comments

core/admin/admin-notices.php
```diff
@@ -654,12 +703,61 @@ public function admin_notices() {
...
+	public function maybe_log_campaign() {
+		if ( empty( $_GET['plg_campaign'] ) || empty( $_GET['plg_campaign_name'] ) ) {
+			return;
+		}
+		$allowed_plgs = [
```
KingYes, line 713: new line below

KingYes (changes requested):

core/admin/admin-notices.php
bainternet: so we can use it from external classes?

KingYes: When you need to use it external just change it, not now..

KingYes (changes requested):

KingYes (approved):

PR #30933, opened 2025-04-20
Internal: V4 New Endpoints [ED-18755]

elchugreeva:


modules/atomic-widgets/parsers/style-parser.php
```diff
@@ -124,6 +128,16 @@ private function sanitize( array $style ): Parse_Result {
...
 	}
 
+	private function sanitize_label( $label ): string {
+		if ( ! is_string( $label ) ) {
+			return '';
+		}
+
+		$sanitized_label = sanitize_text_field( $label );
+
+		return esc_html( $sanitized_label );
```
ronros-elementor, line 138: Looking at the string-prop-type as a reference - sanitization there includes only the `sanitize_text_field` so were wondering - do we really need this esc_html call in addition?

ronros-elementor (approved):

mike-elementor (approved):

PR #30935, opened 2025-04-20
Fix: [V4] The 'empty view' height conflicts parent height [ED-18956]

ronenboxer:


asafdl (dismissed):

modules/atomic-widgets/elements/div-block/div-block.php
```diff
@@ -168,7 +168,6 @@ protected function define_base_styles(): array {
 					Style_Variant::make()
 						->add_prop( 'display', $display )
 						->add_prop( 'padding', $this->get_base_padding() )
-						->add_prop( 'min-height', $this->get_base_height() )
```
davseve, line 171: If we remove this we should also remove the function itself

assets/dev/scss/editor/preview/_container.scss
```diff
@@ -290,9 +278,15 @@
 	}
 }
 
-.e-div-block-base:has( .elementor-empty-view ),
-.e-flexbox-base:has( .elementor-empty-view ) {
-	height: 100px;
+.e-div-block-base:has( > .elementor-empty-view ),
+.e-flexbox-base:has( > .elementor-empty-view ) {
+	display: grid !important;
```
davseve, line 283: Is there a better way than using `important`?

modules/atomic-widgets/elements/div-block/div-block.php
ronenelementor: Probably, nice catch

assets/dev/scss/editor/preview/_container.scss
ronenelementor: It gets overridden by a lot of things, so Evyatar® suggested it and he said there shouldn't be a reason for it to conflict with others

davseve (approved):

baghdasarovelementor (approved):

PR #30943, opened 2025-04-21
Fix: Atomic dynamic tag reload page is not working [ED-18846]

Omerisra6:


modules/atomic-widgets/dynamic-tags/dynamic-prop-types-mapping.php
```diff
@@ -108,7 +109,7 @@ private function get_related_categories( Transformable_Prop_Type $prop_type ): a
 			return [ V1_Dynamic_Tags_Module::TEXT_CATEGORY ];
 		}
 
-		if ( $prop_type instanceof Url_Prop_Type ) {
+		if ( $prop_type instanceof Url_Prop_Type || $prop_type instanceof Link_Control_Url_Prop_Type ) {
```
IshayMaya, line 112: can we test it?

IshayMaya (approved):

PR #30964, opened 2025-04-23
Fix: Widgets are not properly sorted upon drag&drop in the canvas [ED-19003] 

ronros-elementor:
## Before:
https://github.com/user-attachments/assets/ea433dc1-2134-46b6-8fb3-4ced88709511

## After:
https://github.com/user-attachments/assets/a435d578-c23e-4b6e-b58c-08f2ea3ad46c


modules/atomic-widgets/assets/js/editor/div-block-view.js
```diff
@@ -210,60 +210,55 @@ const DivBlockView = BaseElementView.extend( {
...
+				// Triggering the drag end manually, since it won't fire above the iframe
 				elementor.getPreviewView().onPanelElementDragEnd();
 
 				const draggedView = elementor.channels.editor.request( 'element:dragged' ),
-					draggingInSameParent = ( draggedView?.parent === this ),
+					draggedElement = draggedView?.getContainer().view.el,
 					containerSelector = event.currentTarget.parentElement;
 
-				let $elements = jQuery( containerSelector ).find( '> .elementor-element' );
+				const $elements = jQuery( containerSelector ).find( '> .elementor-element' );
```
hein-obox, line 220: Maybe overkill to define `$elements` for this?

modules/atomic-widgets/assets/js/editor/div-block-view.js
```diff
@@ -210,60 +210,55 @@ const DivBlockView = BaseElementView.extend( {
...
+				const elements = $elements.toArray();
```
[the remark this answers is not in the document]
hein-obox: Maybe group it together with the other `const` definitions?

modules/atomic-widgets/assets/js/editor/div-block-view.js
```diff
@@ -305,22 +300,38 @@ const DivBlockView = BaseElementView.extend( {
 		return editTools;
 	},
 
-	shouldIncrementIndex( side ) {
-		if ( ! this.draggingOnBottomOrRightSide( side ) ) {
-			return false;
+	getDropIndex( container, side, index, selfIndex ) {
+		const styles = window.getComputedStyle( container );
```
hein-obox, line 304: Do we need to add the `window` object?
Just asking :).

modules/atomic-widgets/assets/js/editor/div-block-view.js
```diff
@@ -305,22 +300,38 @@ const DivBlockView = BaseElementView.extend( {
...
+
+		const isFlex = 'flex' === styles.display;
+		const isFlexReverse = isFlex &&
+			[ 'column-reverse', 'row-reverse' ].includes( styles.flexDirection );
+
+		const isRtl = elementorCommon.config.isRTL;
+
+		const isReverse = isFlexReverse !== isRtl; 	// If not flex reversed + RTL, or if it is flex reversed + LTR
```
hein-obox, line 312: We always try to avoid using comments.
But in this case I wouldn't know how to do that.

We would use a variable name like `isVisualReverse`, but that is still very cryptic.

modules/atomic-widgets/assets/js/editor/div-block-view.js
```diff
@@ -305,22 +300,38 @@ const DivBlockView = BaseElementView.extend( {
...
-		if ( ! this.draggingOnBottomOrRightSide( side ) ) {
-			return false;
+	getDropIndex( container, side, index, selfIndex ) {
+		const styles = window.getComputedStyle( container );
+
+		const isFlex = 'flex' === styles.display;
+		const isFlexReverse = isFlex &&
+			[ 'column-reverse', 'row-reverse' ].includes( styles.flexDirection );
+
+		const isRtl = elementorCommon.config.isRTL;
```
hein-obox, line 310: We sometimes have challenges with the `isRTL` check in the editor.
Can you check if this works correct if the website language is English and the page content in Hebrew, or the other way around? We have had issues in the past where the languages are mixed inside the editor.
I can't remember how we can reproduce this... But I know that it is possible the mix the languages inside the editor.

modules/atomic-widgets/assets/js/editor/div-block-view.js
```diff
@@ -305,22 +300,38 @@ const DivBlockView = BaseElementView.extend( {
...
+
+		const isReverse = isFlexReverse !== isRtl; 	// If not flex reversed + RTL, or if it is flex reversed + LTR
+
+		if ( ( isReverse === this.draggingOnBottomOrRightSide( side ) ) ) {
+			// The element should be placed BEFORE the current target
+			// if is reversed + side is bottom/right OR not is reversed + side is top/left
```
hein-obox, line 316: I would assume that we only have to swop `left` and `right`, and leave `top` and `bottom` as is. Wdyt?

modules/atomic-widgets/assets/js/editor/div-block-view.js
```diff
@@ -210,60 +210,55 @@ const DivBlockView = BaseElementView.extend( {
...
 
-				// Exclude the dragged element from the indexing calculations.
-				if ( draggingInSameParent ) {
-					$elements = $elements.not( draggedView.$el );
-				}
-
-				const widgetsArray = Object.values( $elements );
-
-				let newIndex = widgetsArray.indexOf( event.currentTarget );
+				// User is dragging an element from the panel.
```
hein-obox, line 224: Maybe we can refactor it more (if you are not scared, because of the lack of tests...).
Suggestion by ChatGpt:

`onDropping: ( side, event ) => {
				event.stopPropagation();

				// Triggering the drag end manually, since it won't fire above the iframe
				elementor.getPreviewView().onPanelElementDragEnd();

				const draggedView = elementor.channels.editor.request( 'element:dragged' ),
					draggedElement = draggedView?.getContainer().view.el,
					containerSelector = event.currentTarget.parentElement;

				const $elements = jQuery( containerSelector ).find( '> .elementor-element' );
				const elements = $elements.toArray();
				const targetIndex = elements.indexOf( event.currentTarget );

				// User is dragging an element from the panel.
				if ( ! draggedView || ! draggedElement ) {
					this.onDrop( event, { at: targetIndex } );

					return;
				}

				// Prevent the user from dragging a parent container into its own child container
				const draggedId = draggedView.getContainer().id;

				let currentTargetParentContainer = this.container;

				while ( currentTargetParentContainer ) {
					if ( currentTargetParentContainer.id === draggedId ) {
						return;
					}

					currentTargetParentContainer = currentTargetParentContainer.parent;
				}

				const selfIndex = elements.indexOf( draggedElement );

				if ( targetIndex === selfIndex ) {
					return;
				}

				const dropIndex = this.getDropIndex( containerSelector, side, targetIndex, selfIndex );

				// Reset the dragged element cache.
				elementor.channels.editor.reply( 'element:dragged', null );

				$e.run( 'document/elements/move', {
					container: draggedView.getContainer(),
					target: this.getContainer(),
					options: {
						at: dropIndex,
					},
				} );
			},`
			
			
`
isPanelElement( draggedView, draggedElement ) {
	return !draggedView || !draggedElement;
},

isParentElement( draggedId ) {
	let current = this.container;

	while ( current ) {
		if ( current.id === draggedId ) {
			return true;
		}
		current = current.parent;
	}

	return false;
},

moveDroppedItem( draggedView, dropIndex ) {
	elementor.channels.editor.reply( 'element:dragged', null );

	$e.run( 'document/elements/move', {
		container: draggedView.getContainer(),
		target: this.getContainer(),
		options: {
			at: dropIndex,
		},
	} );
},
`

modules/atomic-widgets/assets/js/editor/div-block-view.js
ronros-elementor: I'd personally prefer that as it give away the source of this method, which could be helpful for anyone who may not be familiar with it (otherwise they'd  go on looking for some globally defined method or something)

so just as a way to give more semantics, no impact on performance what so ever
+ it could potentially be overridden in case you'd define your own method with the same name (if possible

modules/atomic-widgets/assets/js/editor/div-block-view.js
ronros-elementor: It's still an important data point if the display is of a flex column/reverse-column

modules/atomic-widgets/assets/js/editor/div-block-view.js
ronros-elementor: Haha I _am_ afraid, but will go for it 😄

modules/atomic-widgets/assets/js/editor/div-block-view.js
ronros-elementor: Agree, went with the vanilla approach

modules/atomic-widgets/assets/js/editor/div-block-view.js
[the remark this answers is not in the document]
ronros-elementor: Not a big fan of those consts grouping to begin with, but I do see it's quite common in the core repo, so caved in 😄

modules/atomic-widgets/assets/js/editor/div-block-view.js
```diff
@@ -305,22 +300,38 @@ const DivBlockView = BaseElementView.extend( {
 		return editTools;
 	},
 
-	shouldIncrementIndex( side ) {
-		if ( ! this.draggingOnBottomOrRightSide( side ) ) {
-			return false;
+	getDropIndex( container, side, index, selfIndex ) {
+		const styles = window.getComputedStyle( container );
+
+		const isFlex = 'flex' === styles.display;
```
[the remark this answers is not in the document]
ronros-elementor: Great input, added

modules/atomic-widgets/assets/js/editor/div-block-view.js
ronros-elementor: This code got quite changed, but anyway - comments are off :)

modules/atomic-widgets/assets/js/editor/div-block-view.js
ronros-elementor: I've tested it, found an issue that is of a bit of another aspect of the dnd, opened a [ticket](https://elementor.atlassian.net/browse/ED-19086) for it)

modules/atomic-widgets/assets/js/editor/div-block-view.js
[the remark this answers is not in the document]
hein-obox: I understand. We used to this all the time.
How would you prefer to do it? I am thinking we could have created an object with all the properties?

hein-obox (approved):

PR #30996, opened 2025-04-24
Internal: Only close latest dialog on Esc key [ED-18844]

louiswol94:


louiswol94:
https://github.com/elementor/elementor/actions/runs/14652505974/

max-zu (approved):

PR #31038, opened 2025-04-29
Internal: Add cssid control to atomic widget

asafdl:


murzin-nikita (dismissed):

baghdasarovelementor (dismissed):

baghdasarovelementor (dismissed):

modules/atomic-widgets/elements/atomic-heading/atomic-heading.html.twig
```diff
@@ -1,5 +1,6 @@
 {% if settings.title is not empty %}
-	<{{ settings.tag | e('html_tag') }} class="{{ settings.classes | merge( [ base_styles.base ] ) | join(' ') }}">
+	{% set id_attribute = settings.cssid is not empty ? 'id="' ~ settings.cssid | e('html_attr') ~ '"' : '' %}
```
Omerisra6, line 2: Shouldn't you add it to all elements?

modules/atomic-widgets/elements/div-block/div-block.php
```diff
@@ -90,8 +90,9 @@ protected function define_atomic_controls(): array {
 							],
 						]),
 
-					Link_Control::bind_to( 'link' ),
-
+					Link_Control::bind_to( 'link' )->set_meta( [
+						'topDivider' => true,
+					] ),
```
Omerisra6, line 95: can we set this as the default for link control?

modules/atomic-widgets/module.php
```diff
@@ -123,13 +124,22 @@ private function register_feature() {
...
-		Plugin::$instance->experiments->add_feature([
+		Plugin::$instance->experiments->add_feature( [
+			'name' => self::CSSID_EXPERIMENT_NAME,
+			'title' => esc_html__( 'V4 - CSS ID', 'elementor' ),
+			'description' => esc_html__( 'Enable CSS ID control for V4.', 'elementor' ),
+			'hidden' => true,
+			'default' => Experiments_Manager::STATE_INACTIVE,
+			'release_status' => Experiments_Manager::RELEASE_STATUS_ALPHA,
+		] );
+
```
Omerisra6, line 135: Please open a task for deleting the feature flag

Omerisra6 (changes requested):

modules/atomic-widgets/elements/atomic-heading/atomic-heading.html.twig
asafdl: different ticket

modules/atomic-widgets/elements/div-block/div-block.php
asafdl: we can, im not sure that we want to though

Omerisra6 (approved):

PR #31096, opened 2025-05-07
Internal: [V4] Display control > Display none [ED-18639]

ronenboxer:


modules/atomic-widgets/module.php
```diff
@@ -118,15 +118,36 @@ public function __construct() {
...
 			'description' => esc_html__( 'Enable atomic widgets.', 'elementor' ),
 			'hidden' => true,
 			'default' => Experiments_Manager::STATE_INACTIVE,
 			'release_status' => Experiments_Manager::RELEASE_STATUS_ALPHA,
-		]);
+		];
+	}
+
+	private function register_experimental_features() {
+		if ( self::is_active() ) {
```
Omerisra6, line 133: I would move to the constructor and include all experiments

Omerisra6 (changes requested):

Omerisra6 (approved):

PR #31128, opened 2025-05-08
Internal: Add cssID control to all widgets

asafdl:


modules/atomic-widgets/assets/js/editor/div-block-view.js
```diff
@@ -100,13 +110,22 @@ const DivBlockView = BaseElementView.extend( {
...
 		this.handleLink();
+		this.handleCssID();
 
 		// Defer to wait for everything to render.
 		setTimeout( () => {
 			this.droppableInitialize();
 		} );
 	},
 
+	handleCssID() {
```
louiswol94, line 121:
```
attributes() {
	return {
		...
	};
},
```
Maybe?

https://github.com/elementor/elementor/blob/main/assets/dev/js/editor/elements/views/base.js#L42-L48

davseve (approved):

PR #31170, opened 2025-05-13
Tweak: Disable using Elements caching shortcode by the interface [ED-18986]

KingYes:


modules/element-cache/module.php
```diff
@@ -52,12 +56,27 @@ public static function get_experimental_data(): array {
...
+		return $unique_id;
+	}
+
 	private function register_shortcode() {
 		add_shortcode( 'elementor-element', function ( $atts ) {
 			if ( empty( $atts['data'] ) ) {
 				return '';
 			}
 
+			if ( empty( $atts['k'] ) || $atts['k'] !== $this->get_unique_id() ) {
```
hein-obox, line 76: Maybe:
```suggestion
			if ( empty( $atts['k'] ) || ! hash_equals( $this->get_unique_id(), $atts['k'] ) ) {
```

hein-obox:
Yakir is still working on the test(s).

TzviRabinovitch (dismissed):

hein-obox (approved):

modules/element-cache/module.php
KingYes: Why?

hein-obox: I think we can leave this here. I think this more relevant in the context of password encryption.

> Checks whether two strings are equal without leaking information about the contents of known_string via the execution time.
> 
> This function can be used to mitigate timing attacks. Performing a regular comparison with === will take more or less time to execute depending on whether the two values are different or not and at which position the first difference can be found, thus leaking information about the contents of the secret known_string.
Source: https://www.php.net/manual/en/function.hash-equals.php

Batsirai-muchareva (approved):

PR #31177, opened 2025-05-14
Tweak: Unite url prop types [ED-19047]

murzin-nikita:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #


baghdasarovelementor (approved):

modules/atomic-widgets/prop-types/link-control-url-prop-type.php
```diff
@@ -1,25 +0,0 @@
...
-	}
-
-	protected function validate_value( $value ): bool {
-		return true;
-	}
-
-	protected function sanitize_value( $value ) {
-		$url = trim( $value );
-
-		return htmlspecialchars( $url, ENT_QUOTES, 'UTF-8' );
```
ronros-elementor, line 23: I noticed the sanitization here is slightly different (probably that contributes to the many changes you have in tests)

so this is just to clarify - do we want the sanitization to remain as-is in the `LinkPropType` or should sanitization change as well in case we set the `skip_validations` setting to true?

ronros-elementor (changes requested):

modules/atomic-widgets/prop-types/link-control-url-prop-type.php
mykytamurzin: From my point of view sanitisation trough native wp-functions is better than custom. It's a bit different, but this difference for me looks acceptable. 

@baghdasarovelementor what do you think? do we need to port sanitisation from `link-control-url-prop-type.php`.

baghdasarovelementor: I think your solution is ok. Creating this prop type for allow "google" as a URL, and keep XSS and injection away. Your solution is also keeping same. Maybe we can ask this to Doron is this ok for product side or not.

ronros-elementor (approved):

PR #31194, opened 2025-05-14
Internal: Marketing Bday WP notice [ED-19228]

ronenboxer:


core/admin/pointers/manager.php
```diff
@@ -0,0 +1,56 @@
...
+	const DISMISS_ACTION_KEY = 'elementor-dismissed-pointers';
+
+	private $notices = [];
+
+	public function __construct() {
+		add_action( 'wp_ajax_' . self::DISMISSED_POINTERS_META_KEY, [ $this, 'dismiss_pointers' ] );
+	}
+
+	public function register_pointer( $slug, $notice_constructor ) {
+		$this->notices[ $slug ] = new $notice_constructor();
```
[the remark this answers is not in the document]
mykytamurzin: this will allow to Manager do their job = manage things

core/admin/pointers/manager.php
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	public function dismiss_pointers() {
+		if ( empty( $_POST['nonce'] ) || ! wp_verify_nonce( sanitize_text_field( wp_unslash( $_POST['nonce'] ) ), self::DISMISS_ACTION_KEY ) ) {
```
mykytamurzin, line 24: use `Utils::get_super_global_value`. This function is doing sanitisation for you

core/admin/pointers/notices/birthday.php
```diff
@@ -0,0 +1,132 @@
...
+				} );
+			} );
+		</script>
+		<?php
+	}
+
+	public function should_display_notice(): bool {
+		return $this->is_dashboard_page() &&
+			$this->is_user_allowed() &&
+			! Manager::is_dismissed( self::SLUG ) &&
```
mykytamurzin, line 91: I would suggest to make the notice independent of the manager.

core/admin/pointers/notices/birthday.php
```diff
@@ -0,0 +1,132 @@
...
+							pointer: '<?php echo esc_attr( self::SLUG ); ?>'
+						},
+						nonce: '<?php echo esc_attr( wp_create_nonce( Manager::DISMISS_ACTION_KEY ) ); ?>'
+					}
+				} );
+			};
+
+			jQuery( document ).ready( function( $ ) {
+				$( "#<?php echo esc_attr( self::ELEMENTOR_POINTER_ID ); ?>" ).pointer( {
+					content: '<?php echo wp_kses( $pointer_content, $allowed_tags ); ?>',
```
mykytamurzin, line 66: I think we don't need wp_kses here. This content is fully managed by us, so should be 100% secure

core/admin/pointers/notices/birthday.php
```diff
@@ -0,0 +1,132 @@
...
+					onClose().promise().done( () => {
+						location = $( this ).attr( "href" );
+					} );
+				} );
+			} );
+		</script>
+		<?php
+	}
+
+	public function should_display_notice(): bool {
```
mykytamurzin, line 88: Let's make it static, just to initialise and call the constructor(with add_action) only if it's fully valid

core/admin/pointers/manager.php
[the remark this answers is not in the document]
ronenelementor: I removed the manager as you suggested, so this is irrelevant :)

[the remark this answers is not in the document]
ronenelementor: I removed the manager as you suggested, so this is irrelevant :)

core/admin/pointers/manager.php
ronenelementor: Perfect, done

core/admin/pointers/notices/birthday.php
ronenelementor: Same

core/admin/pointers/notices/birthday.php
ronenelementor: Originally I just printed it, and I got a lint error telling to use it, this is also Yakir's reference from the image optimizer
https://github.com/elementor/elementor/actions/runs/15020186359/job/42207334003?pr=31194#step:5:43

core/admin/pointers/notices/birthday.php
ronenelementor: I tried your suggestion, there's a timing issue with using it, it can't run too early because it returns null when using `get_current_page`, and combined with the `wp_ajax` hook, the endpoint misses it's the registration window

core/admin/pointer-notices/birthday.php
```diff
@@ -0,0 +1,153 @@
...
+	private static function is_already_seen_today() {
+		return get_transient( self::get_user_transient_id() );
+	}
+
+	private function set_seen_today() {
+		$now = time();
+		$midnight = strtotime( 'tomorrow midnight' );
+		$seconds_until_midnight = $midnight - $now;
+
+		set_transient( self::get_user_transient_id(), $now, $seconds_until_midnight );
```
davseve, line 122: 🥇

davseve (approved):

PR #31335, opened 2025-05-28
Internal: [V4] free text control for custom [ED-18637]

ronenboxer:


modules/atomic-widgets/prop-types/size-prop-type.php
```diff
@@ -21,17 +21,36 @@ protected function validate_value( $value ): bool {
...
+	}
+
+	private function validate_string_size( $value ): bool {
+		return ( 'custom' === $value['unit'] && null !== $value['size'] ) ||
+			( 'auto' === $value['unit'] && empty( $value['size'] ) );
+	}
+
+	private function validate_numeric_size( $value ): bool {
+		return ( ! in_array( $value['unit'], [ 'auto', 'custom' ], true ) && ( ! empty( $value['size'] ) || 0 === $value['size'] ) && is_numeric( $value['size'] ) );
+	}
```
murzin-nikita:
```suggestion
private function validate_size( $value ): bool {
	if ( ! is_array( $value ) || ! isset( $value['unit'], $value['size'] ) ) {
		return false;
	}

	switch ( $value['unit'] ) {
		case 'custom':
			return null !== $value['size'];
		case 'auto':
			return empty( $value['size'] );
		default:
			return (
				in_array( $value['unit'], self::SUPPORTED_UNITS, true ) &&
				( ! empty( $value['size'] ) || 0 === $value['size'] ) &&
				is_numeric( $value['size'] )
			);
	}
}
```

murzin-nikita: it can even replace `validate_value` function

ronenboxer: Nice! Done!

murzin-nikita (dismissed):

murzin-nikita (approved):

PR #31350, opened 2025-05-29
Internal: Add Youtube Atom icon to eicons [ED-19374]

nami-p:


hein-obox (approved):

assets/lib/eicons/css/elementor-icons.css
```diff
@@ -507,6 +507,10 @@
   content: "\e876";
 }
 
+.eicon-youtube-atom:before {
```
KingYes, line 510: @nami-p @hein-obox Why is the name ‘youtube-atom’ instead of just ‘youtube’?

KingYes (changes requested):

assets/lib/eicons/css/elementor-icons.css
KingYes: I see we have another youtube class, but maybe we can merge it?

KingYes (changes requested):

assets/lib/eicons/css/elementor-icons.css
nami-p: may be e-youtube like e-heading ?

Danitsh: we should use e-youtube

PR #31367, opened 2025-06-02
Tweak: Conditionally show Cloud Kits tab [ED-19355]

louiswol94:


modules/cloud-library/connect/cloud-library.php
```diff
@@ -341,6 +341,15 @@ public function get_quota() {
 		] );
 	}
 
+	/**
+	 * @return array|\WP_Error
+	 */
+	public function get_kits_quota() {
+		return $this->http_request( 'GET', 'quota/kits', [], [
```
ManorHazaz: Based on HLD...
```suggestion
		return $this->http_request( 'GET', 'kits/quota', [], [
```

app/modules/kit-library/module.php
```diff
@@ -103,6 +104,7 @@ public function __construct() {
 		Plugin::$instance->data_manager_v2->register_controller( new Kits_Controller() );
 		Plugin::$instance->data_manager_v2->register_controller( new Taxonomies_Controller() );
 		Plugin::$instance->data_manager_v2->register_controller( new Kits_Cloud_Controller() );
+		Plugin::$instance->data_manager_v2->register_controller( new Kits_Cloud_Eligibility_Controller() );
```
ManorHazaz: Why do we need to create another controller?
Why not use Kits_Cloud_Controller?

app/modules/kit-library/assets/js/hooks/use-cloud-kits-eligibility.js
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Fetch cloud kits availability from WordPress backend
+ *
+ * @return {Promise<boolean>} Whether cloud kits should be available
+ */
+async function fetchCloudKitsEligibility() {
+	const response = await $e.data.get( 'kits-cloud-eligibility/index', {}, { refresh: true } );
+
+	const isCloudKitsExperimentActive = elementorCommon?.config?.experimentalFeatures?.e_cloud_library_kits;
```
ManorHazaz: Early return to reduce number of calls if experiment is not active?
```suggestion
	const isCloudKitsExperimentActive = elementorCommon?.config?.experimentalFeatures?.e_cloud_library_kits;
	 if( ! isCloudKitsExperimentActive ) {
	 return false;
	 }
	
	const response = await $e.data.get( 'kits-cloud-eligibility/index', {}, { refresh: true } );
```

app/modules/kit-library/assets/js/module.js
```diff
@@ -14,6 +15,7 @@ export default class KitLibrary {
 		$e.components.register( new TaxonomiesComponent() );
 		$e.components.register( new KitLibraryComponent() );
 		$e.components.register( new KitsCloudComponent() );
+		$e.components.register( new KitsCloudEligibilityComponent() );
```
ManorHazaz: Why register a new component? why not use the same component?

ManorHazaz (changes requested):

modules/cloud-library/connect/cloud-library.php
louiswol94: @ManorHazaz - @max-zu ended up using `quota/kits`

app/modules/kit-library/assets/js/hooks/use-cloud-kits-eligibility.js
louiswol94: YEs, good catch!

app/modules/kit-library/module.php
louiswol94: For this, and for your other question about the component, mainly because I wanted to use `$e.data.get()`.

The existing component is for `$e.data.get( 'kits-cloud/index' )` and it maps to `Kits_Cloud_Controller()`

I could not create a command under the same component `$e.data.get( 'kits-cloud/eligibility' )` for example, because if you look at the nature of `Elementor\Data\V2\Base\Controller` it only supports one root GET route per controller. And any other endpoint that belongs to this controller maps to `something/{id}` and `$e.data.get( 'kits-cloud/eligibility/{what_id}' )` would not make sense.

So I had to create a separate controller/component.

louiswol94: @ManorHazaz Please disregard the above comment, I made a mistake. I updated the code and am now using the existing component.

app/modules/kit-library/assets/js/module.js
louiswol94: Updated

ManorHazaz (dismissed):

app/modules/kit-library/module.php
```diff
@@ -113,6 +113,7 @@ public function __construct() {
 
 		add_action( 'elementor/connect/apps/register', function ( ConnectModule $connect_module ) {
 			$connect_module->register_app( 'kit-library', Kit_Library::get_class_name() );
+			$connect_module->register_app( 'cloud-kits', Cloud_Kits::get_class_name() );
```
ManorHazaz: only when the experiment on?

ManorHazaz (changes requested):

ManorHazaz (approved):

PR #31370, opened 2025-06-03
Internal: Create centralized place for rendering styles [ED-19440]

Omerisra6:
### ✨ PR Description
Purpose: Revamps Elementor's CSS handling for atomic widgets, introducing a new Atomic_Styles_Manager to manage CSS file generation and application more efficiently.

Main changes:
- Added new `elementor/post/render` action to track post rendering for post-specific style application
- Created a memoization system to optimize style generation by caching computed styles
- Replaced monolithic CSS file generation with breakpoint-specific CSS files to improve performance
- Migrated Global Classes CSS handling from file-based system to the new Atomic_Styles_Manager architecture


modules/atomic-widgets/elements/atomic-heading/atomic-heading.php
```diff
@@ -121,6 +122,7 @@ protected function define_base_styles(): array {
 				->add_variant(
 					Style_Variant::make()
 						->add_prop( 'margin', $margin_value )
+						->add_prop( 'color', Color_Prop_Type::generate( 'yellow' ) )
```
ronros-elementor: Is it intentional? 🙃

modules/atomic-widgets/module.php
```diff
@@ -102,6 +103,10 @@ public function __construct() {
 			$this->register_experimental_features();
 		}
 
+		Atomic_Styles_Manager::instance()->register_hooks();
+
+		( new Opt_In() )->init();
```
ronros-elementor: Why is it here?

modules/atomic-widgets/elements/atomic-heading/atomic-heading.php
Omerisra6: removed

modules/atomic-widgets/module.php
Omerisra6: removed

modules/atomic-widgets/styles/atomic-styles-manager.php
```diff
@@ -0,0 +1,126 @@
...
+		$styles_cache = [];
+
+		$breakpoints = $this->get_breakpoints();
+		foreach ( $breakpoints as $breakpoint_key ) {
+			foreach ( $this->registered_styles_by_key as $style_key => $get_styles ) {
+				$render_css = function () use ( $get_styles, &$styles_cache, $style_key, $breakpoint_key ) {
+					if ( ! isset( $styles_cache[ $style_key ] ) ) {
+						$styles_cache[ $style_key ] = $get_styles();
+					}
+
```
Nevoss:
```suggestion
$cache = Cache::make();

$styles_by_key = Collection::make( $this->registered_styles_by_key )
	->map( fn( $cb, $key ) => fn () => $cache->get( $key, $cb ) )

```

modules/atomic-widgets/styles/atomic-styles-manager.php
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	private function enqueue_styles() {
+		$post_ids = apply_filters( 'elementor/atomic-widgets/styles/posts', [] );
+
+		if ( ! is_array( $post_ids ) || empty( $post_ids ) ) {
+			return;
+		}
+
+		do_action( 'elementor/atomic-widgets/styles/register', $this, $post_ids );
```
ronros-elementor: Do we _want_ to give access to the `AtomicStylesManager` instance upon registration from the subscribers (i.e style providers) side? 

If we only care for the registration itself, maybe it would be better to pass the register function specifically?

** It would also help with future debugging, as with the current flow you need to track back all subscribers of `'elementor/atomic-widgets/styles/register'` and see how they use `$this`

*** Assuming context binding handling is not a nightmare 🙂

tests/phpunit/elementor/modules/atomic-widgets/styles/test-atomic-widgets-base-styles.php
```diff
@@ -63,56 +68,40 @@ public function test_it__inject_elements_base_styles() {
...
-							->add_prop( 'color', Color_Prop_Type::generate( 'red' ) )
-							->add_prop( 'font-family', String_Prop_Type::generate( 'Poppins' ) )
-					)
-					->add_variant( Style_Variant::make()
-						->set_breakpoint( 'mobile' )
-						->add_prop( 'font-family', String_Prop_Type::generate( 'Inter' ) )
-					)
-					->add_variant( Style_Variant::make()
-						->set_breakpoint( 'tablet' )
-						->add_prop( 'font-family', String_Prop_Type::generate( 'Inter' ) )
```
ronros-elementor, line 98: We don't need to make sure the font families in variants are enqueued?

ronros-elementor (changes requested):

modules/global-classes/atomic-global-styles.php
```diff
@@ -0,0 +1,56 @@
...
+			$context = is_preview() ? Global_Classes_Repository::CONTEXT_PREVIEW : Global_Classes_Repository::CONTEXT_FRONTEND;
+
+			return Global_Classes_Repository::make()->context( $context )->all()->get_items()->map( function( $item ) {
+				$item['id'] = $item['label'];
+				return $item;
+			})->all();
+		};
+
+		$styles_manager->register(
+			self::CSS_FILE_KEY,
```
ronros-elementor: shouldn't it be with the context as well?

modules/atomic-widgets/styles/atomic-styles-manager.php
Omerisra6: It is the only public function, so I don't see why it is a problem. Am I missing something?

tests/phpunit/elementor/modules/atomic-widgets/styles/test-atomic-widgets-base-styles.php
Omerisra6: It is handled by the styles manager, I added a test there, thanks

modules/global-classes/atomic-global-styles.php
Omerisra6: For now, I am rendering all globals

modules/atomic-widgets/styles/atomic-styles-manager.php
```diff
@@ -0,0 +1,144 @@
...
+			}
+
+			Plugin::instance()->frontend->enqueue_font( $value );
+		} )->render( $styles );
+
+		$breakpoint_instance = Plugin::$instance->breakpoints->get_breakpoints( $breakpoint_key );
+
+		$media = $breakpoint_instance
+			? 'screen and (max-width: ' . $breakpoint_instance->get_value() . 'px)'
+			: 'all';
```
ronros-elementor: This is a bit wrong, as widescreen for example have a min-width instead, as it goes up from desktop

You could use [styles-renderer.php](https://github.com/elementor/elementor/blob/23e8d39ccca62803a1fad8d3921b38e961f6100e/modules/atomic-widgets/styles/styles-renderer.php#L147-L167) as a reference maybe as a reference

modules/global-classes/atomic-global-styles.php
ronros-elementor: Yeah but there's a difference between preview/draft to the normal frontend

Omerisra6: You are right, added

modules/atomic-widgets/styles/atomic-styles-manager.php
Omerisra6: Added the logic, thanks!

ronros-elementor (dismissed):

modules/atomic-widgets/styles/atomic-styles-manager.php
```diff
@@ -0,0 +1,132 @@
...
+		if ( empty( $this->post_ids ) ) {
+			return;
+		}
+
+		do_action( 'elementor/atomic-widgets/styles/register', $this, $this->post_ids );
+
+		$get_styles_cache = new Cache();
+		$styles_by_key = Collection::make( $this->registered_styles_by_key )->map_with_keys( function( $get_styles, $style_key ) use ( $get_styles_cache ) {
+			return [
+				$style_key => $get_styles_cache->cache( $style_key, $get_styles ),
```
Nevoss, line 52:
```suggestion
				$style_key => $get_styles_cache->get( $key, $get_styles ),
```

modules/atomic-widgets/styles/atomic-styles-manager.php
```diff
@@ -0,0 +1,132 @@
...
+			];
+		} )->all();
```
Nevoss:
```suggestion
		$styles_by_key = Collection::make( $this->registered_styles_by_key )
			->map_with_keys( fn( $get_styles, $key ) => [ $style_key => $cache->get( $style_key, $get_styles ) ] )
			->all();
```

modules/atomic-widgets/styles/atomic-styles-manager.php
```diff
@@ -0,0 +1,132 @@
...
+		$breakpoints = $this->get_breakpoints();
+		foreach ( $breakpoints as $breakpoint_key ) {
+			foreach ( $styles_by_key as $style_key => $get_styles ) {
+				$render_css = function () use ( $get_styles, $style_key, $breakpoint_key, $group_by_breakpoint_cache ) {
+					$cache_key = $style_key . '-' . $breakpoint_key;
+					$get_grouped_styles = $group_by_breakpoint_cache->cache( $cache_key, fn() => $this->group_by_breakpoint( $get_styles() ) );
+					$grouped_styles = $get_grouped_styles();
+
+					return $this->render_css( $grouped_styles[ $breakpoint_key ] ?? [], $breakpoint_key );
+				};
```
Nevoss: I think it should be extracted into function

modules/atomic-widgets/styles/atomic-styles-manager.php
```diff
@@ -0,0 +1,132 @@
...
+
+				$style_file = ( new CSS_Files_Manager() )->get( $style_key . '-' . $breakpoint_key, $render_css );
```
Nevoss: Create variable for `$style_key . '-' . $breakpoint_key`
it is created twice

modules/atomic-widgets/styles/atomic-styles-manager.php
```diff
@@ -0,0 +1,132 @@
...
+		add_action( 'elementor/post/render', function( $post_id ) {
+			$this->post_ids[] = $post_id;
+		} );
+	}
+
+	public function register( string $key, callable $get_style_defs ) {
+		$this->registered_styles_by_key[ $key ] = $get_style_defs;
+	}
+
+	private function enqueue_styles() {
```
Nevoss, line 48: I think it will be more clear If you will prepare the data before in some structure like
```ts
{
	'desktop-base-style': () => {},
	'mobile-base-style': () => {},
	'desktop-elements-200': () => {},
	...
}
```

Then you will have only one cache and not multiple.

Omerisra6: that way the methods won't run only once

PR #31445, opened 2025-06-09
Fix: Classes not applied on some elements [ED-19524]

Nevoss:


modules/atomic-widgets/assets/js/editor/div-block-view.js
```diff
@@ -398,7 +398,13 @@ const DivBlockView = BaseElementView.extend( {
 	},
 
 	getClasses() {
-		return this.options?.model?.getSetting( 'classes' )?.value || [];
+		const transformer = window?.elementorV2?.editorCanvas?.settingsTransformersRegistry?.get?.( 'classes' );
```
Omerisra6, line 401: What about atomic svg?

modules/atomic-widgets/assets/js/editor/div-block-view.js
Omerisra6, line 1: can it be tested?

Omerisra6 (changes requested):

Omerisra6 (approved):

PR #31489, opened 2025-06-12
Internal: [V4] prop types infra [ED-19535]

ronenboxer:


IshayMaya (dismissed):

modules/atomic-widgets/prop-dependencies/manager.php
```diff
@@ -0,0 +1,86 @@
...
+
+	/**
+	 * @param $config array{
+	 *  operator: string,
+	 *  path: array<string>,
+	 *  value?: mixed,
+	 * }
+	 * @return self
+	 */
+	public function where( array $config, int $index = -1 ): self {
```
IshayMaya, line 47: what if someone passes index?

ronenboxer: Line 59 should solve it
However, I will add the if statement another check to see that the index exists in the dependencies,
WDYT?

IshayMaya (dismissed):

IshayMaya (approved):

PR #31557, opened 2025-06-18
Internal: [V4] Disable HTML tag control when there's a link [ED-18550]

ronenboxer:
### ✨ PR Description
Purpose: Adds dependency management to widget props, conditionally displaying the 'tag' property based on link destination existence.

Main changes:
- Adds PropDependencies Manager to conditionally show/hide widget properties
- Modifies Plain_Prop_Type to support dependencies through new dependencies() method
- Improves Manager class to handle dependency configurations properly


modules/atomic-widgets/elements/div-block/div-block.php
```diff
@@ -41,12 +42,19 @@ public function get_icon() {
 	}
 
 	protected static function define_props_schema(): array {
+		$tag_dependencies = Dependency_Manager::make()
+			->where( [
```
IshayMaya, line 46: what about the effect?

ronenboxer: default to disable, and default to have an or relation

IshayMaya (dismissed):

IshayMaya (dismissed):

IshayMaya (approved):

PR #31568, opened 2025-06-19
Internal: Replace SASS direction-based styling with native CSS solution [ED-19656]

rami-elementor:
### ✨ PR Description
Purpose: Modernizes RTL styling by replacing deprecated getValueByDirection() function with CSS variables for improved direction handling across elements.

Main changes:
- Introduces --direction-multiplier CSS variable for RTL responsive scaling instead of hard-coded direction values
- Replaces hard-coded transform rotation values with calc() expressions using the direction multiplier
- Replaces icon content switching with scale transformations that flip elements in RTL mode
- Adds explicit icons in templates (eicon-chevron-left) instead of conditionally changing content values


assets/dev/scss/editor/_flex-icons.scss
```diff
@@ -6,20 +6,20 @@
...
 
-	--is-ltr: 1;
-	--is-rtl: 0;
+	@if 'rtl' == $direction {
+		--is-ltr: 0;
+		--is-rtl: 1;
+	} @else {
+		--is-ltr: 1;
+		--is-rtl: 0;
+	}
```
nuritsha, line 15: @rami-elementor can these variables be initialized in the same place & way you are initializing `--direction-multiplier`? 
This way you will be able to avoid using `$direction`

rami-elementor: @nuritsha 

Using `--direction-multiplier` is a global solution. It's used in the frontend, editor, preview, etc. While this is a very specific workaround used specifically in container controls. When you change the container direction, the "justified content" and "align items" controls have their icons rotating. I prefer not to mix the two.

nuritsha (approved):

PR #31590, opened 2025-06-22
Fix: The CSS ID is missing from the Twig of some of our atomic widgets [ED-19635]

davseve:


davseve:
Eventually the cssid was missing in the youtube widget only

YaelAvitan (dismissed):

ronenboxer (dismissed):

murzin-nikita (approved):

PR #31650, opened 2025-06-26
Internal: [V4] Variables - Inputs validations for duplicated values [ED-19647]

nami-p:


modules/variables/storage/repository.php
```diff
@@ -44,6 +45,25 @@ private function assert_if_variables_limit_reached( array $db_record ) {
...
+			if ( isset( $variable['deleted'] ) && $variable['deleted'] ) {
+				continue;
+			}
+
+			if ( isset( $new_variable['id'] ) && $new_variable['id'] === $id ) {
+				continue;
+			}
+
+			if ( isset( $variable['label'] ) && $variable['label'] === $new_variable['label'] ) {
+				throw new DuplicatedLabel( 'Variable name already exists' );
```
mike-elementor:
```suggestion
				throw new DuplicatedLabel( 'Variable label already exists' );
```
this message is for developers, and we revolve in "label" space (all records, properties etc.)

only in translations visible to the end-user -> we call them "names" (guess for marketing purposes)

modules/variables/storage/repository.php
```diff
@@ -44,6 +45,25 @@ private function assert_if_variables_limit_reached( array $db_record ) {
 		}
 	}
 
+	/**
+	 * @throws DuplicatedLabel
+	 */
+	private function assert_if_variable_label_is_duplicate( array $db_record, array $new_variable = [] ) {
```
mike-elementor:
```suggestion
	private function assert_if_variable_label_is_duplicated( array $db_record, array $variable ) {
```

tests/phpunit/elementor/modules/variables/classes/test-variable-repository.php
```diff
@@ -435,9 +542,79 @@ public function test_restore_variable() {
 
 		// Assert.
 		$this->assertArrayNotHasKey( 'deleted', $result['variable'] );
-		$this->assertArrayNotHasKey( 'deleted_at', $result );
```
mike-elementor, line 438: nice one... 👍

modules/variables/storage/repository.php
mike-elementor: variable is required (no need for default `[]` value)
and let's re-name for more clarity

mike-elementor (approved):

PR #31681, opened 2025-07-02
Fix: Make variable label duplicate validation case-insensitive [ED-19647]

nami-p:


modules/variables/storage/repository.php
```diff
@@ -58,7 +58,8 @@ private function assert_if_variable_label_is_duplicated( array $db_record, array
 				continue;
 			}
 
-			if ( isset( $existing_variable['label'] ) && $existing_variable['label'] === $variable['label'] ) {
+			if ( isset( $existing_variable['label'] ) && isset( $variable['label'] ) &&
+				 strtolower( $existing_variable['label'] ) === strtolower( $variable['label'] ) ) {
```
mike-elementor: when the `if ( { condition } )` becomes a multi-line, it's better move it out for readability.
so in our case, we could split into 2 different..
```
if ( ! isset ... ) continue;
if ( strtolower !== ... ) continue;
throw
```

alternative
```
$isLabelDuplicated = isset ....
 && strtolower ...;

if ( $isLabelDuplicated ) throw
```

mike-elementor (approved):

PR #31784, opened 2025-07-09
Internal: Create centralized place for rendering styles - take 2 [ED-19458]

ronros-elementor:
### ✨ PR Description
Purpose: Add centralized rendering system for atomic widget styles with caching and breakpoint support

Main changes:
- Created Atomic_Styles_Manager to handle style rendering across multiple breakpoints with efficient caching
- Added Cache_Validity system to manage invalidation of cached styles by component and breakpoint
- Implemented post-specific style rendering through new action hooks in the rendering pipeline
- Moved global classes styles from CSS file injection to the new centralized rendering system


modules/atomic-widgets/styles/atomic-styles-manager.php
```diff
@@ -108,22 +115,83 @@ private function enqueue_styles() {
...
 				);
+
+				$this->add_fonts_to_enqueue( $style_key );
+			}
+		}
+
+		$this->enqueue_fonts();
+	}
+
+	private function initialize_styles( array $styles_by_key ) {
```
IshayMaya: why is it called `initialize_styles` and `initialize_style` ?
do we have some more stuff we're going to add here other than the fonts?

modules/atomic-widgets/styles/atomic-styles-manager.php
```diff
@@ -108,22 +115,83 @@ private function enqueue_styles() {
 					[],
 					$style_file->get_media()
 				);
+
+				$this->add_fonts_to_enqueue( $style_key );
```
IshayMaya: can't we merge it with what we're doing in `initialize_styles` ?
it feels a bit decoupled when we first initialize and then running this here

tests/phpunit/elementor/modules/atomic-widgets/styles/__snapshots__/Test_Styles_Renderer__test_render_atomic_widget_styles__append_css_of_styles_with_flex_partial_values__1.txt
```diff
@@ -0,0 +1 @@
+.test-style-grow-only{flex:2;}.test-style-grow-shrink{flex:1 2;}.test-style-basis-only{flex:0 1 100px;}
```
IshayMaya, line 1: is it ok?

modules/atomic-widgets/styles/atomic-styles-manager.php
ronros-elementor: The initialize part is actually to _clear_ all collected fonts under a given style key, in case cache is invalid.

The latter guarantees we're enqueuing all fonts _after_ we executed all the relevant "render_css" callbacks of all breakpoints, because _before_ the css file generations - we have no clue which fonts are used in which style definition, and if we'd clean out the fonts _after_ we ran over all breakpoints - we'll reset the fonts after them being added as part of the css rendering execution

So the steps are:
If cache is invalid -
1. Empty recorded fonts
2. Populate fonts as part of `render_css`
3. Enqueue all recorded fonts

If cache is valid -
1. Enqueue all recorded fonts

tests/phpunit/elementor/modules/atomic-widgets/styles/__snapshots__/Test_Styles_Renderer__test_render_atomic_widget_styles__append_css_of_styles_with_flex_partial_values__1.txt
ronros-elementor: I assume so, as it corresponds to the prior `tests/phpunit/elementor/modules/atomic-widgets/styles/__snapshots__/Test_Atomic_Styles__test_parse_atomic_widget_styles__append_css_of_styles_with_flex_partial_values__1.txt`, but with one difference due to the test structure (without the .elementor prefix)

IshayMaya (approved):

PR #31807, opened 2025-07-10
Internal: [V4] Prop dependencies - remove effect and meta from logic

ronenboxer:
### ✨ PR Description

Purpose: Refactor prop dependencies system to use direct property fields instead of metadata for improved type safety and simplified implementation.
Main changes:
- Replaced `meta('dependencies')` with explicit `set_dependencies()` and `get_dependencies()` methods
- Removed effect-based dependency logic (hide/disable) in favor of a single dependency mechanism
- Fixed operator logic from inverted conditions (eq/ne, exists/not_exists) to consistent behavior
- Updated dependency resolution to handle null dependencies and better type safety


modules/atomic-widgets/prop-dependencies/manager.php
```diff
@@ -92,53 +89,29 @@ public function where( array $config ): self {
 		];
 
 		if ( empty( $this->dependencies ) ) {
-			$this->new_dependency();
+			$this->new();
 		}
 
-		$last_index = array_key_last( $this->dependencies );
-		$this->dependencies[ $last_index ]['terms'][] = $term;
+		$this->dependencies['terms'][] = $term;
```
ronenboxer, line 95: Since there are no "model"ed effects, no need to have an array of dependencies. We can have nested multiple dependencies under the "terms" property

packages/packages/core/editor-editing-panel/src/controls-registry/styles-field.tsx
```diff
@@ -28,7 +27,7 @@ export const StylesField = ( { bind, placeholder, propDisplayName, children }: S
 
 	const { values, setValues, canEdit } = useStylesFields( [ ...depList, bind ] );
```
IshayMaya: we are now no longer using the other deps
```suggestion
	const { values, setValues, canEdit } = useStylesFields( [ bind ] );
```

packages/packages/libs/editor-props/src/utils/prop-dependency-utils.ts
```diff
@@ -11,16 +12,20 @@ type ParsedTerm = DependencyTerm;
 
 type Relation = Dependency[ 'relation' ];
 
-export function shouldApplyEffect( { relation, terms }: Dependency, values: PropValue ): boolean {
+export function isPropDependencyMet( propType: PropType | null, values: PropValue ): boolean {
```
IshayMaya: why the extra function? isn't that already checked in line 20

packages/packages/libs/editor-props/src/utils/__tests__/prop-dependency-utils.test.ts
```diff
@@ -745,7 +728,7 @@ describe( 'prop-dependency-utils', () => {
 					],
 				};
 
-				expect( shouldApplyEffect( dependency, values ) ).toBe( true );
+				expect( isPropDependencyMet( { dependencies: dependency } as PropType, values ) ).toBe( true );
```
IshayMaya: i think we can still pass just the dependencies like before and then all these type casting will not be needed

modules/atomic-widgets/styles/style-schema.php
```diff
@@ -125,18 +125,18 @@ private static function get_typography_props() {
...
+					Dependency_Manager::make( Dependency_Manager::RELATION_AND )
 					->where( [
-						'operator' => 'lt',
+						'operator' => 'gte',
 						'path' => [ 'column-count' ],
 						'value' => 1,
 					] )
 					->where( [
-						'operator' => 'not_exist',
+						'operator' => 'exists',
```
IshayMaya: gte should cover it no?

modules/atomic-widgets/prop-types/union-prop-type.php
```diff
@@ -112,17 +114,17 @@ public function jsonSerialize(): array {
...
-			$this->meta( 'dependencies', $dependencies );
-		}
+	public function set_dependencies( ?array $dependencies ): self {
+		$this->dependencies = empty( $dependencies ) ? null : $dependencies;
 
 		return $this;
 	}
+
+	public function get_dependencies(): ?array {
+		return empty( $this->dependencies['terms'] ?? [] ) ? null : $this->dependencies;
```
IshayMaya: why do we need to check it?

ronenboxer: Discussed in private, decided we don't need it

modules/atomic-widgets/styles/style-schema.php
ronenboxer: Right, done

packages/packages/libs/editor-props/src/utils/__tests__/prop-dependency-utils.test.ts
ronenboxer: Done

packages/packages/libs/editor-props/src/utils/prop-dependency-utils.ts
ronenboxer: Nmv, removed

packages/packages/core/editor-editing-panel/src/controls-registry/styles-field.tsx
ronenboxer: Switched to `useStylesFiled`

IshayMaya (approved):

PR #31844, opened 2025-07-14
Internal: Prevent the indications popover from opening in the Repeaters controls [ED-20023]

elchugreeva:
### ✨ PR Description
Purpose: Prevent indications popover from opening in specific Repeaters controls by adding conditional filtering logic.
Main changes:
- Added skipControls array containing control types to exclude from the inheritance indicator
- Moved isUsingNestedProps constant outside component for better performance
- Enhanced conditional check to filter out specified control types from inheritance chain


mike-elementor (dismissed):

packages/packages/core/editor-editing-panel/src/styles-inheritance/components/styles-inheritance-indicator.tsx
```diff
@@ -20,12 +20,18 @@ import { StylesInheritanceInfotip } from './styles-inheritance-infotip';
 export const StylesInheritanceIndicator = () => {
 	const { path, propType } = useBoundProp();
 
+	const skipControls = [ 'box-shadow', 'background-overlay', 'filter', 'backdrop-filter', 'transform' ];
```
Ntnelbaba: Const that is not going to change no matter what is the input should be move to file level (after imports)
The reason is that whenever you are calling StylesInheritanceIndicator, it will recreate the array each time, which is memory-consuming, and we prefer to avoid

This is the same as you are importing the EXPERIMENTAL_FEATURES, it is a block list that is declared once and always used the same

packages/packages/core/editor-editing-panel/src/styles-inheritance/components/styles-inheritance-indicator.tsx
```diff
@@ -20,12 +20,18 @@ import { StylesInheritanceInfotip } from './styles-inheritance-infotip';
...
+
 	const isUsingNestedProps = isExperimentActive( EXPERIMENTAL_FEATURES.V_3_30 );
 
 	const finalPath = isUsingNestedProps ? path : path.slice( 0, 1 );
 
 	const inheritanceChain = useStylesInheritanceChain( finalPath );
 
+	if ( path.some( ( pathItem ) => skipControls.includes( pathItem ) ) ) {
```
Ntnelbaba: You can unify the checks with line 35 as they are returning the same (also, don't you want to validate that the path is not undefined?)

Ntnelbaba (approved):

packages/packages/core/editor-editing-panel/src/styles-inheritance/components/styles-inheritance-indicator.tsx
```diff
@@ -17,16 +17,18 @@ import { type SnapshotPropValue } from '../types';
...
 	const { path, propType } = useBoundProp();
 
-	const isUsingNestedProps = isExperimentActive( EXPERIMENTAL_FEATURES.V_3_30 );
-
 	const finalPath = isUsingNestedProps ? path : path.slice( 0, 1 );
 
 	const inheritanceChain = useStylesInheritanceChain( finalPath );
 
-	if ( ! inheritanceChain.length ) {
+	if ( ! path || path.some( ( pathItem ) => skipControls.includes( pathItem ) ) || ! inheritanceChain.length ) {
```
mike-elementor, line 31: this is a bit hard to read/digest (support in the future)
can we do a slight split:

```
if ( noInheritance() || nonSupportedControl() ) {
   return null; 
}
```

so that maybe, it will be easier to read/update later on...
wdyt?

mike-elementor (approved):

PR #31853, opened 2025-07-14
Tweak: Add settings customization to the export process [ED-19808]

ManorHazaz:
### ✨ PR Description
Purpose: Add settings customization dialog to enable granular control of exported settings in the Import/Export module

Main changes:
- Implemented settings customization UI with toggleable options for theme, global colors/fonts, and experiments
- Created data model to store and retrieve customization preferences during export process
- Added compatibility layer to support both old (v2.0) and new (v3.0) manifest formats
- Upgraded site-settings export runner to selectively include settings based on user choices


app/modules/import-export-customization/assets/js/export/components/kit-content.js
```diff
@@ -1,10 +1,17 @@
...
 import { useExportContext } from '../context/export-context';
 
 export default function KitContent() {
 	const { data, dispatch } = useExportContext();
+	const [ dialogOpenItems, setDialogOpenItems ] = useState( () => {
+		return kitContentData.reduce( ( acc, item ) => {
+			acc[ item.type ] = false;
+			return acc;
+		}, {} );
+	} );
```
max-zu: if only 1 section can be expanded/opened at the moment - let's simplify it
	const [ expandedSection, setSection ] = useState(null);
	
	and use some sort of simple map in the code
	
	const KIT_CUSTOMIATION_DIALOG = {
	 'site-settings': <KitSettingsCustomizationDialog />
	}

app/modules/import-export-customization/assets/js/export/components/kit-settings-customization-dialog.js
```diff
@@ -0,0 +1,213 @@
...
+			} }
+		>
+			<DialogHeader onClose={ handleClose }>
+				<DialogTitle>
+					{ __( 'Edit settings & configurations', 'elementor' ) }
+				</DialogTitle>
+			</DialogHeader>
+
+			<DialogContent dividers sx={ { p: 3 } }>
+				<Stack spacing={ 0 }>
```
max-zu: isn't 0 the default value ?

app/modules/import-export-customization/runners/export/kit-data.json
```diff
@@ -0,0 +1,767 @@
+{
+  "content": [],
+  "settings": {
+    "template": "default",
+    "body_background_background": "classic",
```
max-zu: do we need this file ?

app/modules/import-export-customization/runners/export/site-settings.php
```diff
@@ -48,7 +60,7 @@ public function export( array $data ) {
 			$kit_data['experiments'] = $experiments_data;
 		}
 
-		$manifest_data['site-settings'] = $kit_tabs;
+		$manifest_data['site-settings'] = array_fill_keys( self::allowed_settings, true );
```
max-zu: does import work correctly with this change ?

app/modules/import-export-customization/runners/export/kit-data.json
ManorHazaz: No, pushed by mistake

app/modules/import-export-customization/runners/export/site-settings.php
ManorHazaz: Maybe not, but I added adaptor so maybe we will need to fix it
I don't want to it in that PR as well :|

app/modules/import-export-customization/assets/js/export/components/kit-settings-customization-dialog.js
ManorHazaz: Removed

app/modules/import-export-customization/assets/js/export/components/kit-content.js
ManorHazaz: Good catch
Simplified

max-zu (dismissed):

max-zu (approved):

PR #31869, opened 2025-07-15
Internal: Fix filters would not update and empty values break render [ED-20053]

RobiFerentz:


packages/packages/libs/editor-controls/src/controls/filter-repeater-control.tsx
```diff
@@ -226,26 +227,38 @@ const ItemContent = ( {
...
 
 const PropContent = ( { item, anchorEl }: { item: FilterItemPropValue; anchorEl?: HTMLElement | null } ) => {
 	const propContext = useBoundProp( cssFilterFunctionPropUtil );
-	const [ value, setValue ] = useState( item.value );
-	const handleChange = ( val: string | null ) => {
-		const newValue = { ...filterConfig[ val ?? '' ].defaultValue.value };
-		setValue( structuredClone( newValue ) );
+
+	const handleValueChange = (
+		nVal: FilterItemPropValue[ 'value' ],
```
davseve: maybe?
```suggestion
		newValueParam: FilterItemPropValue[ 'value' ],
```

packages/packages/libs/editor-controls/src/controls/filter-repeater-control.tsx
```diff
@@ -226,26 +227,38 @@ const ItemContent = ( {
...
+		options?: CreateOptions,
+		meta?: { bind?: PropKey }
+	) => {
+		let newValue = structuredClone( nVal );
+		const newFuncName = newValue?.func.value ?? '';
+		if ( meta?.bind === 'func' ) {
+			newValue = structuredClone( filterConfig[ newFuncName ].defaultValue.value );
```
murzin-nikita, line 251: what if newFuncName will be `''` in this case?

packages/packages/libs/editor-controls/src/controls/filter-repeater-control.tsx
RobiFerentz: Changed

packages/packages/libs/editor-controls/src/controls/filter-repeater-control.tsx
murzin-nikita: it could be 2 times structuredClone with rewriting `newValue`

RobiFerentz: That's for typescript only, there option does not actually exists in the options drop down. It's only valid names, no empty strings or nulls etc.

murzin-nikita (approved):

davseve (approved):

PR #31878, opened 2025-07-15
Fix: Slow first load due to local Google Fonts [ED-19938]

KingYes:
### ✨ PR Description
Purpose: Fix slow first page load by optimizing Google Fonts loading in Elementor when in preview mode.

Main changes:
- Added force_enqueue_from_cdn parameter to Google_Font::enqueue method to bypass local font fetching
- Modified AJAX request for Google Fonts to use non-blocking mode with true parameter
- Added detection for preview mode to use CDN directly when in editor preview


assets/dev/js/editor/utils/helpers.js
```diff
@@ -299,7 +299,7 @@ module.exports = {
 					elementorCommon.ajax.addRequest( 'enqueue_google_fonts', {
 						data: { font_name: font },
 						unique_id: 'enqueue_google_fonts_' + font,
-					} );
+					}, true );
```
eyalmrejen-elementor, line 302: What does `true` mean? Do I need to step into the actual function call, or can you offer a meaningful variable name here - call for clarity, not optimization

core/files/fonts/google-font.php
```diff
@@ -28,12 +28,12 @@ class Google_Font {
...
 	private static array $folders = [];
 
-	public static function enqueue( string $font_name, string $font_type = self::TYPE_DEFAULT ): bool {
+	public static function enqueue( string $font_name, string $font_type = self::TYPE_DEFAULT, $force_enqueue_from_cdn = false ): bool {
 		if ( static::enqueue_style( $font_name ) ) {
 			return true;
 		}
 
-		if ( ! static::fetch_font_data( $font_name, $font_type ) ) {
+		if ( $force_enqueue_from_cdn || ! static::fetch_font_data( $font_name, $font_type ) ) {
```
eyalmrejen-elementor, line 36: `IF` something `OR` `NOT` something else - if this is `true` then.... - a little hard to read - would you?

assets/dev/js/editor/utils/helpers.js
louiswol94: true = `immediately` - but it's all over the code base like this 🙈

eyalmrejen-elementor: I understand - this is why I want to up our game. Using an immediately variable with the value of true is much more readable. I can look into the function and understand, but why should I, right?

core/files/fonts/google-font.php
```diff
@@ -28,12 +28,12 @@ class Google_Font {
...
 			static::enqueue_from_cdn( $font_name, $font_type );
 			return false;
 		}
```
louiswol94, line 39: Not sure I agree, if you don't have the file, and the CDN failed, then there's nothing to enqueue.

louiswol94 (approved):

asafdl (approved):

PR #31905, opened 2025-07-20
Internal: Variables - Variables Mixpanel events [ED-19265]

elchugreeva:


modules/editor-events/assets/js/editor/events-config.js
```diff
@@ -129,6 +131,12 @@ const eventsConfig = {
 				checklistHide: 'checklist_userpreferences_toggle_hide',
 			},
 		},
+		variables: {
+            open: 'open_variables_popover',
```
Ntnelbaba, line 135: Prettier please 😄

packages/packages/core/editor-variables/src/components/color-variables-selection.tsx
```diff
@@ -44,7 +50,18 @@ export const ColorVariablesSelection = ( { closePopover, onAdd, onEdit, onSettin
 
 	if ( onAdd ) {
 		actions.push(
-			<IconButton key="add" size={ SIZE } onClick={ onAdd }>
+			<IconButton
+				key="add"
+				size={ SIZE }
+				onClick={ () => {
+					trackVariableEvent( {
```
Ntnelbaba: Wrap this in an arrow function named `onAddAndTrack`

Also, onAdd should be done first (we don't want to track something that failed/ failed to track doesn't need to block the onAdd)
Do it for all occurrences

packages/packages/core/editor-variables/src/components/font-variables-selection.tsx
```diff
@@ -43,7 +49,18 @@ export const FontVariablesSelection = ( { closePopover, onAdd, onEdit, onSetting
 
 	if ( onAdd ) {
 		actions.push(
-			<IconButton key="add" size={ SIZE } onClick={ onAdd }>
+			<IconButton
+				key="add"
+				size={ SIZE }
+				onClick={ () => {
+					trackVariableEvent( {
```
Ntnelbaba: @mike-elementor Did we create a Tech debt to delete that duplicate code?

packages/packages/core/editor-variables/src/hooks/use-prop-font-variable-action.tsx
```diff
@@ -17,6 +18,12 @@ export const usePropFontVariableAction = (): PopoverActionProps => {
 		icon: ColorFilterIcon,
 		title: __( 'Variables', 'elementor' ),
 		content: ( { close: closePopover } ) => {
+			trackVariableEvent( {
```
Ntnelbaba, line 21: @mike-elementor As well as those actions duplication

packages/packages/core/editor-variables/src/utils/tracking.ts
```diff
@@ -0,0 +1,39 @@
+type VariableEventData = {
+	varType: string;
+	controlPath: string;
+	action: 'open' | 'add' | 'connect' | 'save';
+};
+
+export const trackVariableEvent = ( { varType, controlPath, action }: VariableEventData ) => {
+	const extendedWindow = window as unknown as Window & {
```
Ntnelbaba, line 8: All `Record<string, string>` should be `Record<string, unknown>` as you can pass to Mixpanel more types other than string (better to see the documentation to find out what is expected)

modules/editor-events/assets/js/editor/events-config.js
elchugreeva: Screenshot 2025-07-20 at 15 23 33
names are according to product request, was especially discussed

packages/packages/core/editor-variables/src/components/color-variables-selection.tsx
elchugreeva: @Danitsh, Netamel is right, in the task it is described to track click event, not successful variable creation or connection. Do you want to track user intent to do this actions or only when action was successful?

Ntnelbaba (approved):

PR #31936, opened 2025-07-22
Internal: Sanitize file upload [ED-19588]

asafdl:
### ✨ PR Description
Purpose: Add security validation for uploaded files to prevent potential arbitrary file access vulnerabilities.

Main changes:
- Added is_uploaded_file() validation to verify that tmp_name is a legitimate uploaded file
- Modified file content loading to only occur after security validation passes
- Set file_content to false by default as fallback if validation fails

### DIFF
--- a/includes/template-library/classes/class-import-images.php
+++ b/includes/template-library/classes/class-import-images.php
@@ -111,8 +111,12 @@ class Import_Images {
 		if ( isset( $attachment['tmp_name'] ) ) {
 			// Used when called to import a directly-uploaded file.
 			$filename = $attachment['name'];
+			$file_content = false;
 
-			$file_content = Utils::file_get_contents( $attachment['tmp_name'] );
+			// security validation in case the tmp_name has been tampered with
+			if ( is_uploaded_file( $attachment['tmp_name'] ) ) {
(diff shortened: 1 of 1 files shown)


Ntnelbaba (approved):

includes/template-library/classes/class-import-images.php
```diff
@@ -111,8 +111,12 @@ public function import( $attachment, $parent_post_id = null ) {
 		if ( isset( $attachment['tmp_name'] ) ) {
 			// Used when called to import a directly-uploaded file.
 			$filename = $attachment['name'];
+			$file_content = false;
 
-			$file_content = Utils::file_get_contents( $attachment['tmp_name'] );
+			// security validation in case the tmp_name has been tampered with
+			if ( is_uploaded_file( $attachment['tmp_name'] ) ) {
```
eyalmrejen-elementor, line 117: ugly comment - why not remove it, and have a simple variable called `wasTheFileNotTemepered` and then have the if f
`if(wasTheFileNotTemepered){}`
Cleaner code and no ugly comment that will live forever and nobody would update it when the next like is - becuase devs never do...

PR #32009, opened 2025-07-28
Tweak: Add plugins customization to the import process [ED-20043]

louiswol94:
### ✨ PR Description
Purpose: Add plugins customization to the import process for selective plugin import during website template customization

Main changes:
- Added customization settings to selectively enable/disable specific plugins during import
- Refactored plugin data handling to use consistent plugin key format across import/export
- Updated UI components to display customized plugin selections in import/export flows
- Fixed navigation with React Router to replace direct window.location manipulation


app/modules/import-export-customization/assets/js/import/pages/import-complete.js
```diff
@@ -111,40 +116,46 @@ export default function ImportComplete() {
...
-								{ __( 'Global Colors | Global Fonts | Typography | Buttons | Images | Form Fields | Previousground | Layout | Lightbox | Page Transitions | Custom CSS', 'elementor' ) }
-							</Typography>
-						</Box>
-						<Box>
-							<Stack direction="row" alignItems="center" spacing={ 1 }>
-								<Typography variant="body2" color="text.primary" >
-									{ __( 'Content', 'elementor' ) }
+							</Box>
+						) }
+						{ data.includes.includes( 'content' ) && (
```
louiswol94, line 133: @ManorHazaz @max-zu I think for each customization task going forward we can double check the summaries. I handled the plugins only.

app/modules/import-export-customization/assets/js/import/context/import-context.js
```diff
@@ -48,8 +69,9 @@ const initialState = {
 	importedData: null,
 	kitUploadParams: null,
 	plugins: [],
-	includes: [ 'content', 'templates', 'settings', 'plugins' ], // All items selected by default
+	includes: [ 'plugins' ],
```
louiswol94, line 72: `plugins` are required. The other includes would be added during the import process.

app/modules/import-export-customization/assets/js/export/components/export-complete-summary.js
```diff
@@ -15,19 +23,49 @@ export default function ExportCompleteSummary( { kitInfo, includes } ) {
...
+						<Stack direction="row" alignItems="center" spacing={ 1 }>
+							<Typography variant="body2" color="text.primary" >
+								{ __( 'Site settings', 'elementor' ) }
+							</Typography>
+							<ExternalLinkIcon />
+						</Stack>
+						<Typography variant="caption" color="text.secondary">
+							{ __( 'Global Colors | Global Fonts | Typography | Buttons | Images | Form Fields | Previousground | Layout | Lightbox | Page Transitions | Custom CSS', 'elementor' ) }
+						</Typography>
+					</Box>
```
max-zu: I would mobve it to a separate Component. smth like 
function SummarySection( { title, subTitle } ) {
 return (
   <Box>
						<Stack direction="row" alignItems="center" spacing={ 1 }>
							<Typography variant="body2" color="text.primary" >
								{ title }
							</Typography>
							<ExternalLinkIcon />
						</Stack>
						<Typography variant="caption" color="text.secondary">
							{  subTitle }
						</Typography>
					</Box>
 )
}

also I would create smth like sections_titles_map = {
 content: {
   title: __(.....),
   subtitle: __(....)
 }
}

and i would iterate through includes:

includes.map(section => (
  <SummarySection  key={section} title={sections_titles_map[section].title} subTitle={sections_titles_map[section].subTitle} />
))

app/modules/import-export-customization/assets/js/import/hooks/use-import-kit.js
```diff
@@ -32,11 +32,9 @@ export const IMPORT_PROCESSING_STATUS = {
 	DONE: 'DONE',
 };
 
-export function useImportKit() {
+export function useImportKit( { data, includes, customization, isProcessing, dispatch } ) {
 	const [ status, setImportStatus ] = useState( IMPORT_PROCESSING_STATUS.PENDING );
 	const [ error, setError ] = useState( null );
-	const { data, isProcessing, dispatch } = useImportContext();
-	const [ runnersState, setRunnersState ] = useState( {} );
```
max-zu, line 39: Why did you decide to move this out of the hook? We're already passing the parameters from the context into the hook (useImport) where it's used anyway. So why not encapsulate this logic inside right away?
Also, there's no need to pass includes separately — it's already part of data.

app/modules/import-export-customization/assets/js/shared/components/kit-plugins-customization-dialog.js
```diff
@@ -43,7 +43,24 @@ const ExternalLinkIcon = ( props ) => {
 };
 
 export function KitPluginsCustomizationDialog( { open, handleClose, handleSaveChanges, data } ) {
-	const { pluginsList, isLoading } = useKitPlugins( { open } );
+	const isImport = data.hasOwnProperty( 'uploadedData' );
```
max-zu, line 46: data contains `uploadedData` always and it's default value is null, so I think hasOwnProperty should be replaced with some more simple check, or I don't understand how it should work

tests/jest/unit/app/modules/import-export-customization/assets/js/export/components/export-error.test.js
```diff
@@ -12,11 +20,10 @@ describe( 'ExportError Component', () => {
 		};
 		global.elementorAppConfig = mockElementorAppConfig;
 
-		delete window.location;
-		window.location = { href: '' };
-
 		mockWindowOpen = jest.fn();
 		global.window.open = mockWindowOpen;
+
+		mockDispatch.mockClear();
```
max-zu: i think you can move jest.clearAllMocks(); to the beginning of this section and this should be enough

app/modules/import-export-customization/assets/js/import/hooks/use-import-kit.js
louiswol94: @max-zu I need to know which plugins were installed from the runner (backend), in order to display it in the Export complete summary, so I decided to make it part of the broader context.

The only other way I could see to know this is...

1. check `uploadedData.manifest.plugins` (all the plugins form the manifest)
2. check `customization.plugins`
3. and build an array to display here.

But we're already doing this in the backend, so why not just use the result?

app/modules/import-export-customization/assets/js/shared/components/kit-plugins-customization-dialog.js
louiswol94: `uploadedData` is only here during the IMPORT flow.

tests/jest/unit/app/modules/import-export-customization/assets/js/export/components/export-error.test.js
louiswol94: Done

app/modules/import-export-customization/assets/js/export/components/export-complete-summary.js
louiswol94: Done

max-zu (approved):

PR #32015, opened 2025-07-28
Internal: Add atomic components' styles to relevant posts [ED-20200]

ronros-elementor:
### To test manually
1. Create a new page, with some content
2. **Style the content**
3. Save the ID aside
4. Create another page
5. Go to its corresponding row in the `wp_postmeta` DB table
6. Edit it's meta `_elementor_data`
7. Under the first `elements` array, add the following (change `<COMPONENT_ID>` with the post ID saved in step 3) -
```
{
   "id":"a1d2cb6",
   "elType":"widget",
   "settings":{
      "component_id":{
         "$$type":"string",
         "value": <COMPOENNT_ID>
      }
   },
   "elements":[
      
   ],
   "widgetType":"e-component",
   "styles":[
      
   ],
   "editor_settings":[
      
   ],
   "version":"0.0"
}
```

### ✨ PR Description
Purpose: Add metadata storage capability to Cache_Validity system and implement atomic component styles processing for relevant posts.
Main changes:
- Enhanced Cache_Validity class to store and retrieve metadata alongside validity state
- Created Atomic_Component_Styles class to track and render component styles across posts
- Added new experiment flag "e_atomic_components" to enable atomic components functionality


modules/atomic-widgets/styles/atomic-component-styles.php
```diff
@@ -0,0 +1,91 @@
...
+
+			$this->add_post_ids_to_styles_manager( $component_ids );
+		} );
+
+		add_action( 'elementor/document/after_save', fn( Document $document ) => $this->invalidate_cache(
+			[ $document->get_main_post()->ID ]
+		), 20, 2 );
+	}
+
+	private function add_post_ids_to_styles_manager( array $post_ids ) {
```
Omerisra6: Because the render hook is global and not specific for the styles manager I think you should use a more generic name
```suggestion
	private function declare_components_rendered( array $post_ids ) {
```

Maybe you have a better idea

modules/atomic-widgets/styles/atomic-component-styles.php
```diff
@@ -0,0 +1,91 @@
...
+
+			$components = $this->get_components_from_post( $post_id );
+			$component_ids = array_unique(
+				array_map(
+					fn( $component ) => $component['settings']['component_id']['value'],
+					array_filter(
+						$components,
+						fn ( $component ) => isset( $component['settings']['component_id']['value'] ),
+					)
+				)
```
Omerisra6: [Suggestion] Maybe you can use the collection util to make it cleaner

modules/atomic-widgets/styles/atomic-component-styles.php
```diff
@@ -0,0 +1,91 @@
...
+use Elementor\Modules\AtomicWidgets\Cache_Validity;
+
+/**
+ * Atomic Component styles fetching for render
+ */
+class Atomic_Component_Styles {
+	const CACHE_KEY_PREFIX = 'atomic-component-styles-related-posts';
+
+	public function register_hooks() {
+		add_action( 'elementor/post/render', function( $post_id ) {
```
Omerisra6: [Suggestion] Extract to a function

modules/atomic-widgets/styles/atomic-component-styles.php
```diff
@@ -0,0 +1,91 @@
...
+
+				$this->add_post_ids_to_styles_manager( $related_posts );
+
+				return;
+			}
+
+			$components = $this->get_components_from_post( $post_id );
+			$component_ids = array_unique(
+				array_map(
+					fn( $component ) => $component['settings']['component_id']['value'],
```
Omerisra6: [Question] Don't you need to first check that the component ID is set?
Maybe you can use the same approach we used for classes and create a new prop type for the component ID setting so that it could exist in different props?
It could be helpful if it can exist in more than one prop (two components in one widget)

modules/atomic-widgets/styles/atomic-component-styles.php
```diff
@@ -0,0 +1,91 @@
...
+			);
+
+			$cache_validity->validate( [ self::CACHE_KEY_PREFIX, $post_id ], $component_ids );
+
+			$this->add_post_ids_to_styles_manager( $component_ids );
+		} );
+
+		add_action( 'elementor/document/after_save', fn( Document $document ) => $this->invalidate_cache(
+			[ $document->get_main_post()->ID ]
+		), 20, 2 );
```
Omerisra6: [Question] Why is it handled here?
I think it should be in another place because it is not related directly to the components

modules/atomic-widgets/styles/atomic-component-styles.php
```diff
@@ -0,0 +1,91 @@
...
+
+		$components = [];
+
+		Plugin::$instance->db->iterate_data( $elements_data, function( $element_data ) use ( &$components ) {
+			if ( isset( $element_data['widgetType'] ) && 'e-component' === $element_data['widgetType'] ) {
+				$components[] = $element_data;
+			}
+		} );
+
+		return $components;
```
Omerisra6: [Suggestion] I think we should create a util (something like iterate document elements) just because we do that a lot

modules/atomic-widgets/styles/atomic-component-styles.php
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	private function invalidate_cache( ?array $post_ids = null ) {
+		$cache_validity = new Cache_Validity();
+
+		if ( empty( $post_ids ) ) {
+			$cache_validity->invalidate( [ self::CACHE_KEY_PREFIX ] );
+
+			return;
+		}
```
Omerisra6: [Question] Are we using this functionality?

tests/phpunit/elementor/modules/atomic-widgets/styles/test-atomic-component-styles.php
```diff
@@ -0,0 +1,168 @@
...
+                    ]
+                ]
+            ]
+        ];
+
+		// Store the elements data in post meta (this is how Elementor stores page data)
+		update_post_meta( $post_id, '_elementor_data', wp_json_encode( $test_elements_data ) );
+		update_post_meta( $post_id, '_elementor_edit_mode', 'builder' );
+
+		return $post_id;
```
Omerisra6: [Suggestion] You can use the factory like in`test-global-classes-cleanup`:
```
private function make_mock_post_with_elements( $elements_data ) {
		$document = $this->factory()->documents->create_and_get( [
			'post_title' => 'Test Post',
			'post_status' => 'publish',
		] );

		$document->update_json_meta( '_elementor_data', $elements_data );

		return $document->get_main_id();
	}
```

tests/phpunit/elementor/modules/atomic-widgets/styles/test-atomic-component-styles.php
```diff
@@ -0,0 +1,168 @@
...
+		// Act
+		do_action( 'elementor/post/render', $post_id );
+
+		// Assert
+		$this->assertEquals( 1, $this->hook_call_count, 'Hook should be called once' );
+
+		$expected_post_ids = [ $post_id ];
+		$this->assertEquals( $expected_post_ids, $this->rendered_post_ids, 'Should render original post only' );
+	}
+}
```
Omerisra6: [Test] Add tests for the invalidation

tests/phpunit/elementor/modules/atomic-widgets/test-cache-validity.php
```diff
@@ -10,6 +10,43 @@
...
+			);
+			
+			// Delete each option individually using WordPress functions
+			foreach ( $options as $option_name ) {
+				delete_option( $option_name );
+			}
+		}
+		
+		// Clear the object cache to ensure options are truly cleared
+		wp_cache_flush();
```
Omerisra6: [Suggestion] Try to avoid accessing the database to get all option names and just use static ones @cursor

Omerisra6 (changes requested):

modules/atomic-widgets/styles/atomic-component-styles.php
ronros-elementor: But I'm hooking up on the save action to clear associated *component* ids cached under this post id, upon changes to a post

Also it's exactly the same flow as we're doing with our [local styles](https://github.com/elementor/elementor/blob/main/modules/atomic-widgets/styles/atomic-widget-styles.php#L19-L21)

modules/atomic-widgets/styles/atomic-component-styles.php
ronros-elementor: Good catch, added now also the action `'elementor/core/files/clear_cache'`, (invoked via the "clear all cache" button on the wp admin side), and it's the same use as of the, also quite like the [local styles](declare_components_rendered) )

modules/atomic-widgets/styles/atomic-component-styles.php
ronros-elementor: Since I'm filtering first I know it would exist there, but indeed changing to a collection would make it much clearer (this language is one of the most confusing ones... filter expects first an array, map expects first the iterator 😖)

modules/atomic-widgets/styles/atomic-component-styles.php
ronros-elementor: Will go with your suggestion in the absence of a better one 😁

tests/phpunit/elementor/modules/atomic-widgets/test-cache-validity.php
ronros-elementor: Extracted the prefix, made the 'test' root key const, and now will simply use the standard wordpress' flow to clear the responsive option instead of directly doing it via the db

modules/atomic-widgets/styles/atomic-component-styles.php
```diff
@@ -0,0 +1,83 @@
+<?php
```
Omerisra6, line 1: Should the components feature be part of the system or a separate module?

Does it make sense to make it a pro feature?

If so create a different module for it.

modules/atomic-widgets/styles/atomic-component-styles.php
Omerisra6: Oh, I missed the invalidate cache

modules/atomic-widgets/utils.php
```diff
@@ -28,4 +29,22 @@ public static function generate_id( string $prefix = '', $existing_ids = [] ): s
 
 		return $id;
 	}
+
+	public static function foreach_post_element( string $post_id, callable $iterate_callback ): void {
```
Omerisra6:
```suggestion
	public static function process_post_elements( string $post_id, callable $iterate_callback ): void {
```

Omerisra6 (dismissed):

Omerisra6 (approved):

PR #32059, opened 2025-07-30
Fix: Placeholders are missing in controls - Margin and Padding [ED-20154]

nami-p:


mike-elementor:
@nami-p a question, I did not followed the placeholders, is it possible to add some test-coverage for this?

elchugreeva (approved):

PR #32122, opened 2025-08-03
Internal: Add an initial package for atomic components for its style provider [ED-20255]

ronros-elementor:


modules/atomic-widgets/module.php
```diff
@@ -98,9 +99,11 @@ class Module extends BaseModule {
 	const EXPERIMENT_CUSTOM_CSS = 'atomic_custom_css';
 	const TRANSITION_EXPERIMENT = 'atomic_widgets_should_use_transition';
 	const EXPERIMENT_UNSTABLE_REPEATER = 'e_unstable_repeater';
+	const EXPERIMENT_COMPONENTS = 'e_atomic_components';
 
 	const PACKAGES = [
 		'editor-canvas',
+		'editor-components',
```
ronros-elementor, line 106: This is the only relevant change for the current PR

PR #32181, opened 2025-08-07
Tweak: Change IE Customization experiment [ED-20224]

louiswol94:
### ✨ PR Description
Purpose: Change the Import/Export Customization experiment from inactive to active and update related UI elements and navigation paths.

Main changes:
- Changed experiment default state from STATE_INACTIVE to STATE_ACTIVE and removed hidden flag
- Updated import/export URL paths to point to new customization routes
- Modified button text from "Open the Library" to "Import from library"
- Added conditional registration of settings tab based on experiment state

### DIFF
--- a/app/modules/import-export-customization/module.php
+++ b/app/modules/import-export-customization/module.php
@@ -130,13 +130,13 @@ class Module extends BaseModule {
 	 * Render the import/export tab content.
 	 */
 	private function render_import_export_tab_content() {
-		$is_cloud_kits_available = Plugin::$instance->experiments->is_feature_active( 'cloud-library' ) && CloudKitLibrary::get_app()->is_eligible();
+		$is_cloud_kits_available = Plugin::$instance->experiments->is_feature_active( 'cloud-library' ) && CloudKitLibrary::get_app()->check_eligibility()['is_eligible'];
 
 		$content_data = [
 			'export' => [
 				'title' => esc_html__( 'Export this website', 'elementor' ),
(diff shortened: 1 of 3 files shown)


app/modules/import-export/module.php
```diff
@@ -483,9 +483,11 @@ private function register_actions() {
 
 		add_action( 'admin_enqueue_scripts', [ $this, 'enqueue_scripts' ] );
 
-		$page_id = Tools::PAGE_ID;
+		if ( ! Plugin::$instance->experiments->is_feature_active( 'import-export-customization' ) ) {
```
louiswol94, line 486: We want to keep this module active, and only stop it from registering a settings tab.

app/modules/import-export-customization/module.php
```diff
@@ -156,7 +156,7 @@ private function render_import_export_tab_content() {
 		if ( $is_cloud_kits_available ) {
 			$content_data['import']['button_secondary'] = [
 				'url' => Plugin::$instance->app->get_base_url() . '#/kit-library/cloud',
-				'text' => esc_html__( 'Open the Library', 'elementor' ),
+				'text' => esc_html__( 'Import from library', 'elementor' ),
```
louiswol94, line 159: We made some changes afterwards but never realized it wasn't updated here as well since the old app's settings screen always loaded regardless of the experiment status (fixed in app/modules/import-export/module.php).

max-zu (approved):

PR #32237, opened 2025-08-12
Internal: Mixpanel events for tansitions effects [ED-19790]

marcin-el:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #


RobiFerentz (dismissed):

RobiFerentz (dismissed):

packages/packages/libs/editor-controls/src/controls/transition-control/transition-repeater-control.tsx
```diff
@@ -49,6 +50,24 @@ function getChildControlConfig( recentlyUsedList: string[] ) {
 }
 
 export const TransitionRepeaterControl = createControl( ( props: { recentlyUsedList: string[] } ) => {
+	const onAddItem = () => {
+		const extendedWindow = window as unknown as ExtendedWindow;
+		const config = extendedWindow?.elementorCommon?.eventsManager?.config;
+		const selectedElements = getSelectedElements();
+		const widgetType = selectedElements[ 0 ]?.type ?? null;
```
asafdl: so 'transition'?

packages/packages/libs/editor-controls/src/components/unstable-repeater/context/repeater-context.tsx
```diff
@@ -92,6 +97,8 @@ export const RepeaterContextProvider = < T extends RepeatablePropValue = Repeata
 
 		setOpenItemIndex( newIndex );
 		popoverState.open( rowRef ?? ev );
+
+		repeaterEventBus.emit( 'item-added', { repeaterType: propTypeUtil.key } );
```
asafdl: event names should move to a const / enum

packages/packages/libs/editor-controls/src/components/unstable-repeater/context/repeater-context.tsx
```diff
@@ -92,6 +97,8 @@ export const RepeaterContextProvider = < T extends RepeatablePropValue = Repeata
...
 	};
 
 	const removeItem = ( index: number ) => {
```
asafdl, line 106: should probably go ahead and add events to all the current actions

packages/packages/libs/editor-controls/src/components/unstable-repeater/context/repeater-context.tsx
asafdl: don't we want to send additional data here?

packages/packages/libs/editor-controls/src/controls/transition-control/transition-repeater-control.tsx
```diff
@@ -78,6 +79,19 @@ export const TransitionRepeaterControl = createControl(
 		currentStyleState: StyleDefinitionState;
 	} ) => {
 		const currentStyleIsNormal = currentStyleState === null;
+
+		React.useEffect( () => {
+			const unsubscribe = repeaterEventBus.subscribe( 'item-added', ( data ) => {
+				if ( data?.repeaterType === 'transition' ) {
+					repeaterEventBus.emit( 'transition-item-added', {
```
asafdl: why do we catch this event, then emit a new event

packages/packages/libs/editor-controls/src/services/repeater-event-bus.ts
```diff
@@ -0,0 +1,50 @@
...
+
+	emit( eventType: string, data?: TransitionItemData ) {
+		this.listeners.get( eventType )?.forEach( ( callback ) => callback( data ) );
+
+		if ( eventType === 'transition-item-added' ) {
+			this.trackTransitionItemAdded( data );
+		}
+	}
+
+	private trackTransitionItemAdded( data?: TransitionItemData ) {
```
asafdl: we should have a mixpanel wrapper inside the project, and it should not be here

packages/packages/libs/editor-controls/src/services/repeater-event-bus.ts
```diff
@@ -0,0 +1,50 @@
...
+
+		return () => {
+			this.listeners.get( eventType )?.delete( callback );
+		};
+	}
+
+	emit( eventType: string, data?: TransitionItemData ) {
+		this.listeners.get( eventType )?.forEach( ( callback ) => callback( data ) );
+
+		if ( eventType === 'transition-item-added' ) {
```
asafdl: this is definitely not the way we want to do this

the repeater emits events, does not track anything. tracking should be done somewhere else, maybe in a transitions-events.ts file

packages/packages/libs/editor-controls/src/services/repeater-event-bus.ts
```diff
@@ -0,0 +1,50 @@
+import { type ExtendedWindow, getSelectedElements } from '@elementor/editor-elements';
+
+interface TransitionItemData {
+	transition_type?: string;
+	widget_type?: string;
+	repeaterType?: string;
+}
+
+export class RepeaterEventBus {
```
asafdl: we should expose unsubscribe

packages/packages/libs/editor-controls/src/services/repeater-event-bus.ts
```diff
@@ -0,0 +1,50 @@
+import { type ExtendedWindow, getSelectedElements } from '@elementor/editor-elements';
+
+interface TransitionItemData {
```
asafdl: this shouldn't be here

packages/packages/libs/editor-controls/src/controls/transition-control/transition-repeater-control.tsx
asafdl: i don't think this is how we want to do this.
item-added event should expose the args we need then all we need to do is listen to event, and send request

packages/packages/libs/editor-controls/src/services/repeater-event-bus.ts
```diff
@@ -0,0 +1,50 @@
...
+			this.listeners.set( eventType, new Set() );
+		}
+		this.listeners.get( eventType )?.add( callback );
+
+		return () => {
+			this.listeners.get( eventType )?.delete( callback );
+		};
+	}
+
+	emit( eventType: string, data?: TransitionItemData ) {
```
asafdl: maybe this is something to think about, will the repeater-event-bus emit a general event with "repeater.key"
or will it emit event per repeater:
'trainsition-item-added'
'transform-item-added' etc

packages/packages/libs/editor-controls/src/services/repeater-event-bus.ts
marcin-el: subscribe returns unsubscribe function, don't think we would want to unsubscribe from somewhere else than the component itself that uses it

asafdl: events are async things, sometimes you don't want to subscribe or unsubscribe in the same flow

packages/packages/libs/editor-controls/src/services/repeater-event-bus.ts
```diff
@@ -0,0 +1,26 @@
...
+			this.listeners.set( eventName, new Set() );
+		}
+		this.listeners.get( eventName )?.add( callback );
+
+		return () => {
+			this.listeners.get( eventName )?.delete( callback );
+		};
+	}
+
+	emit( eventName: RepeaterEventType ) {
```
asafdl: i feel like repeater events should be emitted with payload, and this should be inserted to callback

packages/packages/libs/editor-controls/src/utils/event-tracking.ts
```diff
@@ -0,0 +1,22 @@
+import { type ExtendedWindow, getSelectedElements } from '@elementor/editor-elements';
```
asafdl: my idea is to have a wrapper for sendMixpanelEvent( mixpanelProps ) {
}

then in transitions control, just subscribe and use callback to send event with the correct props

packages/packages/libs/editor-controls/src/utils/event-tracking.ts
```diff
@@ -0,0 +1,22 @@
...
+export const sendAddTransitionControlEvent = () => {
+	const extendedWindow: ExtendedWindow = window;
+	const config = extendedWindow.elementorCommon?.eventsManager?.config;
+	const eventName = config?.names.elementorEditor?.transitions?.clickAddedTransition;
+
+	const selectedElements = getSelectedElements();
+	const widgetType = selectedElements[ 0 ].type ?? null;
+
+	if ( config && eventName && extendedWindow.elementorCommon?.eventsManager ) {
+		extendedWindow.elementorCommon.eventsManager.dispatchEvent( eventName, {
```
asafdl: we should wrap this in try catch, don't want to throw exception in case event tracking fails

packages/packages/libs/editor-controls/src/components/unstable-repeater/context/repeater-context.tsx
asafdl: this wasn't addressed

packages/packages/libs/editor-controls/src/components/unstable-repeater/context/repeater-context.tsx
```diff
@@ -5,8 +5,24 @@ import { type PopupState, usePopupState } from '@elementor/ui';
...
+const getNestedValue = ( obj: Record< string, unknown >, path: string ) => {
+	return path.split( '.' ).reduce( ( current: Record< string, unknown >, key ) => {
+		if ( current && typeof current === 'object' ) {
+			return current[ key ] as Record< string, unknown >;
+		}
+		return {};
+	}, obj );
+};
+
+const getSelectionValue = ( value: unknown ): unknown => {
```
asafdl: this isn't a generic repeater function, and shouldn't be here

packages/packages/libs/editor-elements/src/index.ts
```diff
@@ -1,6 +1,7 @@
 // types
 export * from './types';
 export type { V1Element, V1ElementModelProps, V1ElementSettingsProps, V1ElementConfig } from './sync/types';
+export type { ExtendedWindow } from './sync/types';
```
asafdl: we don't export the extended window, each package has its own

packages/packages/libs/editor-elements/src/sync/types.ts
```diff
@@ -20,6 +20,30 @@ export type ExtendedWindow = Window & {
 		getContainer?: ( id: string ) => V1Element | undefined;
 	};
 	elementorCommon?: {
+		eventsManager?: {
```
asafdl: is. this used anymore?

packages/packages/libs/editor-controls/src/services/repeater-event-bus.ts
```diff
@@ -0,0 +1,19 @@
+export class RepeaterEventBus {
```
asafdl: this shouldn't be exported, the event bus needs to be a singleton

packages/packages/libs/editor-controls/src/__tests__/repeater-event-bus.test.ts
```diff
@@ -0,0 +1,78 @@
+import { RepeaterEventBus } from '../services/repeater-event-bus';
```
asafdl: the event bus needs to be a singleton, please test it as such

packages/packages/libs/editor-controls/src/controls/transition-control/trainsition-events.ts
```diff
@@ -0,0 +1,38 @@
+import { getSelectedElements } from '@elementor/editor-elements';
+import { sendMixpanelEvent } from '@elementor/utils';
+
+import { repeaterEventBus } from '../../services/repeater-event-bus';
+
+interface TransitionItemValue {
```
asafdl: this should be a type, also maybe its possible to use the transition prop schema and not create a duplicate type

packages/packages/libs/utils/src/mixpanel-tracking.ts
```diff
@@ -0,0 +1,26 @@
...
+			dispatchEvent: ( name: string, data: Record< string, unknown > ) => void;
+		};
+	};
+};
+
+export type MixpanelEvent = {
+	location: string;
+	secondaryLocation: string;
+	trigger: string;
+	transition_type: string;
```
asafdl: this is probably a specific key that can be done like:
```ts
type s = {
  a: string;
  b: string;
  c: string;
} & { [key: string]: unknown }
```
so that you can add any extra args in there if you need, this type should also be used for the eventsManager.dispatchEvent

murzin-nikita (dismissed):

asafdl (approved):

PR #32302, opened 2025-08-17
Internal: Create a way for rendering elements inside an element with default settings [ED-20245]

Omerisra6:
### ✨ PR Description
Purpose: Add default children rendering capability for Atomic Elements to support nested widget structures with pre-configured components.

Main changes:
- Added define_default_children() method in Atomic_Element_Base for specifying child elements
- Created Element_Builder and Widget_Builder classes for programmatic element generation
- Implemented Atomic Tabs component hierarchy demonstrating parent-child widget relationships
- Updated document filtering to exclude column and section elements from widget panel


core/base/document.php
```diff
@@ -729,9 +729,9 @@ protected function get_initial_config() {
 		if ( static::get_property( 'has_elements' ) ) {
 			$widgets_config = Plugin::$instance->widgets_manager->get_widget_types_config();
 
-			// defines the elements that will be shown in the panel
+			$excluded_elements = [ 'column', 'section' ];
```
IshayMaya, line 732: is it not longer related to the panel?

IshayMaya: also, is it the same as `isValidChild` in `atomic-element-model` ?

modules/atomic-widgets/assets/js/editor/atomic-element-model.js
```diff
@@ -9,4 +9,37 @@ export default class AtomicContainer extends elementor.modules.elements.models.E
 
 		return 'section' !== elType && 'column' !== elType;
 	}
+
+	initialize( attributes, options ) {
+		const elementType = this.get( 'elType' );
+		this.config = elementor.config.elements[ elementType ];
+
+		const isNewElementCreate = 0 === this.get( 'elements' ).length &&
+			$e.commands.currentTrace.includes( 'document/elements/create' );
```
IshayMaya, line 18: 😨

modules/atomic-widgets/assets/js/editor/atomic-element-model.js
```diff
@@ -9,4 +9,37 @@ export default class AtomicContainer extends elementor.modules.elements.models.E
...
+
+	getDefaultChildren() {
+		const { default_children: defaultChildren } = this.config;
+
+		return defaultChildren.map( ( element ) => {
+			return {
+				elType: element.elType,
+				widgetType: element.widgetType,
+				id: elementorCommon.helpers.getUniqueId(),
+				settings: element.settings || {},
```
IshayMaya, line 35: not sure if its needed now but keep in mind we will need to have initial settings with some relations between the children. something like this:
```
tabPane: { id: '1' }
tabLink: { settings: { tab-pane-id: { $$type: string, value: '1' }  } }
```

modules/atomic-widgets/elements/atomic-tabs/atomic-tab-list.php
```diff
@@ -0,0 +1,177 @@
...
+		->where( [
+			'operator' => 'not_exist',
+			'path' => [ 'link', 'destination' ],
+		] )
+		->get();
+
+		return [
+			'classes' => Classes_Prop_Type::make()
+				->default( [] ),
+			'tag' => String_Prop_Type::make()
```
IshayMaya: not sure these props are relevant

modules/atomic-widgets/elements/atomic-tabs/atomic-tab-list.php
```diff
@@ -0,0 +1,177 @@
...
+		];
+	}
+
+	protected function define_atomic_controls(): array {
+		return [
+			Section::make()
+				->set_label( __( 'Settings', 'elementor' ) )
+				->set_id( 'settings' )
+				->set_items( [
+					Select_Control::bind_to( 'tag' )
```
IshayMaya: same

modules/atomic-widgets/elements/atomic-tabs/atomic-tab.php
```diff
@@ -0,0 +1,167 @@
...
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+class Atomic_Tab extends Atomic_Element_Base {
+	const BASE_STYLE_KEY = 'base';
+
+	public static function get_type() {
+		return 'e-tab';
```
IshayMaya: `e-tab-link` ?
also for the class and other places

modules/atomic-widgets/elements/atomic-tabs/atomic-tab.php
```diff
@@ -0,0 +1,167 @@
...
+			->where( [
+				'operator' => 'not_exist',
+				'path' => [ 'link', 'destination' ],
+			] )
+			->get();
+
+		return [
+			'classes' => Classes_Prop_Type::make()
+				->default( [] ),
+			'tag' => String_Prop_Type::make()
```
IshayMaya: here as well, im not sure its needed (or valid)

modules/atomic-widgets/elements/atomic-tabs/atomic-tab.php
```diff
@@ -0,0 +1,167 @@
...
+	protected static function define_props_schema(): array {
+		$tag_dependencies = Dependency_Manager::make()
+			->where( [
+				'operator' => 'not_exist',
+				'path' => [ 'link', 'destination' ],
+			] )
+			->get();
+
+		return [
+			'classes' => Classes_Prop_Type::make()
```
IshayMaya: here we will need a `tab-pabe-id` prop

modules/atomic-widgets/elements/atomic-tabs/atomic-tab-list.php
```diff
@@ -0,0 +1,177 @@
...
+					Style_Variant::make()
+						->add_prop( 'display', $display )
+						->add_prop( 'min-width', $min_width )
+						->add_prop( 'gap', $gap )
+						->add_prop( 'justify-content', $justify_content )
+				),
+		];
+	}
+
+	protected function add_render_attributes() {
```
IshayMaya, line 88: we need to add the aria attributes as well (some are dynamic and some are consistent)

IshayMaya:
tab pane and tab content?

core/base/document.php
Omerisra6: This function defines which elements will be included in the panel, and it also defines which element config will be included in the widgetsCache global variable (which is used across the system for several things).

That caused a problem because I filtered all elements that had the `show_in_panel` property, and they were also removed from the config.

Now I include all element types except the column and section (which is exactly like the previous state).

modules/atomic-widgets/elements/atomic-tabs/atomic-tab-list.php
Omerisra6: Why not?
Maybe we can remove the tag, but I think it should be set dynamically.

modules/atomic-widgets/elements/atomic-tabs/atomic-tab.php
Omerisra6: I will ask elli about the naming

modules/atomic-widgets/elements/atomic-tabs/atomic-tab-list.php
Omerisra6: We also need a defined product decision about it.

Let's create a list and schedule a meeting with Elli.

IshayMaya (approved):

PR #32317, opened 2025-08-18
Fix: Enable typing negative values in style fields that can support it [ED-19596]

MiryamOren:
image
image
image
image
image

### ✨ PR Description
Purpose: Fix style field behavior to support entering negative values in positioning and margin controls.

Main changes:
- Added min={-Number.MAX_SAFE_INTEGER} to position and margin SizeControl components
- Created new NumberInput component with validation for negative inputs
- Added custom validation to useBoundProp for validating style input values
- Fixed invalid input handling for numeric fields


packages/packages/libs/editor-controls/src/bound-prop-context/use-bound-prop.ts
```diff
@@ -79,9 +79,13 @@ const useValidation = ( propType: PropType ) => {
 
 	// If the value does not pass the prop type validation, set the isValid state to false.
 	// This will prevent the value from being set in the model, and its fallback will be used instead.
-	const validate = ( value: PropValue | null ) => {
+	const validate = ( value: PropValue | null, validation?: ( value: PropValue ) => boolean ) => {
```
IshayMaya, line 82: make sure to add test that make sure we dont override the built-in validations

packages/packages/libs/editor-controls/src/components/number-input.tsx
```diff
@@ -0,0 +1,57 @@
...
+		props.onKeyDown?.( event );
+	};
+
+	const handleBlur = ( event: React.FocusEvent< HTMLInputElement > ) => {
+		props.onBlur?.( event );
+
+		const { valid } = event.target.validity;
+
+		// HTML number input quirk: invalid input (e.g. "-9-") returns value="" but displays "-9-" to user,
+		// so when we revert to last valid value we must re-render the component to actually display it.
```
IshayMaya:
```suggestion
		// so when we revert to last valid value we must re-mount the component to actually display it.
```

packages/packages/libs/editor-controls/src/components/number-input.tsx
```diff
@@ -0,0 +1,57 @@
...
+		if ( ! valid ) {
+			setKey( String( Math.random() ) );
```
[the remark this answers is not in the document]
IshayMaya: @MiryamOren might be safer

packages/packages/libs/editor-controls/src/controls/number-control.tsx
```diff
@@ -61,11 +65,7 @@ export const NumberControl = createControl(
 							</InputAdornment>
 						) : undefined,
 					} }
-					onKeyDown={ ( event: KeyboardEvent ) => {
-						if ( RESTRICTED_INPUT_KEYS.includes( event.key ) ) {
-							event.preventDefault();
-						}
-					} }
+					allowNegative={ min < 0 }
```
IshayMaya: why is it different from size control?
can we rely on the min in all the places instead of `allowNegative` boolean?

packages/packages/libs/editor-controls/src/controls/size-control.tsx
```diff
@@ -102,9 +104,14 @@ export const SizeControl = createControl(
 		const actualExtendedOptions = useSizeExtendedOptions( extendedOptions || [], disableCustom ?? false );
 		const popupState = usePopupState( { variant: 'popover' } );
 
+		const isSizeInputValid = useRef< boolean >( true );
```
IshayMaya: i understand the issue here but it feel a bit risky, can we pass it through the setState?

packages/packages/libs/editor-controls/src/bound-prop-context/use-bound-prop.ts
MiryamOren: 🫡

packages/packages/libs/editor-controls/src/controls/number-control.tsx
```diff
@@ -27,45 +26,45 @@ export const NumberControl = createControl(
...
 		return (
 			<ControlActions>
-				<TextField
+				<NumberInput
 					size="tiny"
 					type="number"
 					fullWidth
 					disabled={ disabled }
 					value={ isEmptyOrNaN( value ) ? '' : value }
 					onChange={ handleChange }
```
IshayMaya: we need to use the `onInput` in both cases so it will run in every input event (for example - type negative value and then deleting it will keep the value instead of removing it)

IshayMaya (approved):

PR #32320, opened 2025-08-19
Internal: Create nested elements actions with history [ED-20442]

IshayMaya:
### ✨ PR Description
Purpose: Add hooks and utilities for managing nested elements with history support in the editor to enable more complex hierarchical element structures.

Main changes:
- Implemented useNestedElements hook with undoable create/remove operations for nested widgets
- Enhanced useElementChildren to recursively find children at multiple levels
- Added createElementId utility function for generating unique element identifiers
- Modified createElement and deleteElement to support element positioning with 'at' parameter


packages/packages/libs/editor-elements/src/hooks/use-element-children.ts
```diff
@@ -24,19 +24,22 @@ export function useElementChildren< T extends ElementChildren >(
...
-			return childrenTypes.reduce( ( acc, type ) => {
-				const childElements = children
-					.filter( ( child ) => child.model?.get( 'widgetType' ) === type )
-					.map( ( child ) => ( { id: child.id } ) );
+				return acc;
+			}, {} as ElementChildren );
+
+			container?.children?.forEachRecursive?.( ( child ) => {
+				const widgetType = child.model.get( 'widgetType' );
+				const id = child.id;
```
eyalmrejen-elementor: I like better
`const {id} = child`

packages/packages/libs/editor-elements/src/hooks/use-nested-elements.ts
```diff
@@ -0,0 +1,208 @@
...
+									...createParams,
+									id: elementId,
+								},
+							} );
+						} );
+
+						return { elementsData };
+					},
+					undo: ( _: { elements: CreateElementParams[] }, { elementsData }: CreatedElementsResult ) => {
+						// Delete elements in reverse order to avoid dependency issues
```
eyalmrejen-elementor: remove this cursor comment - you can see in the next line that you are deleting, so why have this?

IshayMaya: i kept this one for the dependency issues clarification, but some can be removed

packages/packages/libs/editor-elements/src/hooks/use-nested-elements.ts
```diff
@@ -0,0 +1,208 @@
...
+						[ ...elementsData ].reverse().forEach( ( { elementId } ) => {
```
eyalmrejen-elementor: 1. Why do u reverse?
  2. I would use `toReversed ` since this does not mutate the origianl array, removing your spread operator - always better to work with no side effects

packages/packages/libs/editor-elements/src/hooks/use-nested-elements.ts
```diff
@@ -0,0 +1,208 @@
...
+						elementsData.forEach( ( { createParams } ) => {
+							// Recreate element with original ID and settings
+							const element = createElement( {
+								...createParams,
+								options: { ...createParams.options, useHistory: false },
+							} );
+
+							const elementId = element.id;
+
+							// Store element data again for potential future operations
```
eyalmrejen-elementor: remove comment

packages/packages/libs/editor-elements/src/hooks/use-nested-elements.ts
```diff
@@ -0,0 +1,208 @@
...
+						} );
+					},
+					redo: (
+						_: { elements: CreateElementParams[] },
+						{ elementsData }: CreatedElementsResult
+					): CreatedElementsResult => {
+						const newElementsData: CreatedElementData[] = [];
+
+						elementsData.forEach( ( { createParams } ) => {
+							// Recreate element with original ID and settings
```
eyalmrejen-elementor: aaaarrrgggrrrrr - why cursor add these stupid comments?

packages/packages/libs/editor-elements/src/hooks/use-nested-elements.ts
```diff
@@ -0,0 +1,208 @@
...
+								elementId,
+								options: { useHistory: false },
+							} );
+						} );
+					},
+					redo: (
+						_: { elements: CreateElementParams[] },
+						{ elementsData }: CreatedElementsResult
+					): CreatedElementsResult => {
+						const newElementsData: CreatedElementData[] = [];
```
eyalmrejen-elementor: I don't like this name Data at the end - when this is not data? What is the differance between CreatedElement, and CreatedElementData? - MetaData is better if you feel it is needed
Naming, naming, naming

IshayMaya: agree, renamed to created and removed elements

packages/packages/libs/editor-elements/src/hooks/use-nested-elements.ts
IshayMaya: missed them in the last iteration

packages/packages/libs/editor-elements/src/hooks/use-element-children.ts
IshayMaya: can just destruct both tbh

packages/package-lock.json
Omerisra6, line 1: revert

packages/packages/libs/editor-elements/src/hooks/use-element-children.ts
```diff
@@ -24,19 +24,21 @@ export function useElementChildren< T extends ElementChildren >(
...
+				acc[ type ] = [];
 
-			return childrenTypes.reduce( ( acc, type ) => {
-				const childElements = children
-					.filter( ( child ) => child.model?.get( 'widgetType' ) === type )
-					.map( ( child ) => ( { id: child.id } ) );
+				return acc;
+			}, {} as ElementChildren );
+
+			container?.children?.forEachRecursive?.( ( { model, id } ) => {
```
Omerisra6, line 33: Where does the `forEachRecursive` comes from?

packages/packages/libs/editor-elements/src/hooks/use-element-children.ts
```diff
@@ -24,19 +24,21 @@ export function useElementChildren< T extends ElementChildren >(
...
-			const children = container?.children || [];
+			const elementChildren = childrenTypes.reduce( ( acc, type ) => {
+				acc[ type ] = [];
 
-			return childrenTypes.reduce( ( acc, type ) => {
-				const childElements = children
-					.filter( ( child ) => child.model?.get( 'widgetType' ) === type )
-					.map( ( child ) => ( { id: child.id } ) );
+				return acc;
+			}, {} as ElementChildren );
```
Omerisra6, line 31: Can't you do that inside the other loop?

Just to ensure we are not iterating over the children twice

packages/packages/libs/editor-elements/src/hooks/use-nested-elements.ts
```diff
@@ -0,0 +1,201 @@
...
+	parent: V1Element | null;
+	at: number;
+};
+
+type RemovedElementsResult = {
+	elementIds: string[];
+	removedElements: RemovedElement[];
+};
+
+export const useNestedElements = () => {
```
Omerisra6: Can you split each action (create and remove) into a separate file, and maybe extract the logic into methods?

It makes it hard to revirew

packages/packages/libs/editor-elements/src/sync/create-element-id.ts
```diff
@@ -0,0 +1,7 @@
+import { type ExtendedWindow } from './types';
+
+export const createElementId = () => {
```
Omerisra6:
```suggestion
export const generateElementId = () => {
```
?

Omerisra6 (changes requested):

packages/packages/libs/editor-elements/src/hooks/use-element-children.ts
IshayMaya: `elementor/assets/dev/js/editor/container/model/children-array.js`

packages/packages/libs/editor-elements/src/sync/create-elements.ts
```diff
@@ -0,0 +1,101 @@
...
+				_: { elements: CreateElementParams[] },
+				{ createdElements }: CreatedElementsResult
+			): CreatedElementsResult => {
+				const newElements: CreatedElement[] = [];
+
+				createdElements.forEach( ( { createParams } ) => {
+					const element = createElement( {
+						...createParams,
+						options: { ...createParams.options, useHistory: false },
+					} );
```
Omerisra6, line 77: Don't you need to reverse it again so that you will create the elements at the right order?

packages/packages/libs/editor-elements/src/sync/remove-elements.ts
```diff
@@ -0,0 +1,101 @@
...
+				const removedElements: RemovedElement[] = [];
+
+				elementIdsParam.forEach( ( elementId ) => {
+					const container = getContainer( elementId );
+
+					if ( container ) {
+						const model = container.model.toJSON();
+						const parent = container.parent;
+
+						const at = container.view?._index ?? 0;
```
Omerisra6, line 46: How does it work?
 
The `_index_` holds the index of the child?

packages/packages/libs/editor-elements/package.json
```diff
@@ -43,7 +43,8 @@
     "@elementor/editor-props": "3.32.0",
     "@elementor/editor-styles": "3.32.0",
     "@elementor/editor-v1-adapters": "3.32.0",
-    "@elementor/utils": "3.32.0"
+    "@elementor/utils": "3.32.0",
+    "@wordpress/i18n": "^3.1.0"
```
Omerisra6, line 47: where do you use it?

packages/packages/libs/editor-elements/src/sync/create-elements.ts
IshayMaya: i only remove it from last to first, order stays the same

packages/packages/libs/editor-elements/src/sync/remove-elements.ts
IshayMaya: yes, so we can add it back to where it was

packages/packages/libs/editor-elements/package.json
IshayMaya: undoable subtitle

Omerisra6 (approved):

PR #32345, opened 2025-08-20
Internal: Add mixpanel events for top bar 'connect' flow [ED-20293]

RobiFerentz:


packages/packages/core/editor-app-bar/src/extensions/connect/hooks/use-connect-link-config.tsx
```diff
@@ -31,9 +50,15 @@ export default function useConnectLinkConfig() {
 					?.css( 'display', 'none' )
 					?.appendTo( 'body' );
 
-				$tempButton.elementorConnect();
+				$tempButton.elementorConnect( {
+					success: () => {
+						dispatchConnectClickEvent( 'accountConnected' );
```
marcin-el, line 55: could: maybe just use `config.names.topBar.accountConnected` instead hardcoded string? same for line 61?

marcin-elementor, line 55: could: maybe just use `config.names.topBar.accountConnected` instead hardcoded string? same for line 61?

marcin-el (approved):

packages/packages/core/editor-app-bar/src/extensions/connect/hooks/use-connect-link-config.tsx
RobiFerentz: That would mean another checking of the entire config object

PR #32384, opened 2025-08-21
Internal: Save changes in the Variable Manager - BE [ED-20398]

elchugreeva:


modules/variables/storage/repository.php
```diff
@@ -228,6 +229,213 @@ public function restore( string $id, $overrides = [] ) {
...
+			return [
+				'error' => [
+					'status' => $this->get_error_status_code( $e ),
+					'message' => $e->getMessage(),
+				],
+			];
+		}
+	}
+
+	private function process_create_operation( array &$db_record, array $operation ): array {
```
Ntnelbaba, line 301: I might don't know but don't we already have a create operation function? Or any other operation?

modules/variables/classes/rest-api.php
```diff
@@ -393,4 +412,80 @@ private function prepare_error_response( $status_code, $error, $message ) {
...
+	public function process_batch( WP_REST_Request $request ) {
+		try {
+			$watermark = $request->get_param( 'watermark' );
+			$operations = $request->get_param( 'operations' );
+
+			$result = $this->variables_repository->process_atomic_batch( $operations, $watermark );
+
+			$this->clear_cache();
+
+			return $this->success_response( $result );
```
mike-elementor: for the sake of readability, can we extract the bodies of the try { ... } to a sep. functions?

modules/variables/storage/repository.php
elchugreeva: @Ntnelbaba, yes, I wanted to reuse them in the beginning, but each individual function ends with saving data to db and increasing watermark. In the batch we save in the very end if everything is successful.

modules/variables/storage/repository.php
```diff
@@ -228,6 +229,213 @@ public function restore( string $id, $overrides = [] ) {
...
+			'temp_id' => $temp_id,
+		];
+	}
+
+	private function process_update_operation( array &$db_record, array $operation ): array {
+		$id = $operation['id'];
+		$variable_data = $operation['variable'];
+
+		if ( ! isset( $db_record['data'][ $id ] ) ) {
+			throw new \Elementor\Modules\Variables\Storage\Exceptions\RecordNotFound( 'Variable not found' );
```
mike-elementor, line 331: aren't they listed in "use" on the top?

modules/variables/storage/repository.php
mike-elementor: @elchugreeva so, essentially we can do some refactoring, something similar to:
```
function update(...) { ... update_variable ... save }
function process_create_operation( ... ) { ... update_variable ... }
```
and there is part of common logic -> `update_variable`, and batch/atomic updates/creates etc.

p.s. because as I see it, we introduced the new `created_at` / `updated_at` timestamps, but they'll be added only in case variables are managed through the manager, and will not be there if variables are created from the panel.

mike-elementor (dismissed):

Ntnelbaba (dismissed):

mike-elementor (approved):

PR #32405, opened 2025-08-24
New: Create REST API for components [ED-20334]

MiryamOren:
### ✨ PR Description
Purpose: Add REST API for components to enable creation and retrieval of reusable UI components with associated styles.
Main changes:
- Implemented Components REST API with GET and POST endpoints for managing components
- Created components repository to handle database interactions and component storage
- Added component parser with validation for component names and content structure
- Refactored utility classes into Core\Utils\Api namespace for shared use


modules/components/documents/component.php
```diff
@@ -0,0 +1,56 @@
...
+
+		$labels = [
+			'name' => $plural_label,
+			'singular_name' => $singular_label,
+		];
+
+		return $labels;
+	}
+
+	public static function get_supported_features(): array {
```
Omerisra6, line 46: Is that a `Document` method?
What is it used for?

modules/components/components-parser.php
```diff
@@ -0,0 +1,38 @@
...
+
+class Components_Parser {
+	const MIN_NAME_LENGTH = 2;
+	const MAX_NAME_LENGTH = 50;
+	
+	public static function make() {
+		return new static();
+	}
+
+	public function parse_name( $name, $existing_components_names ): Parse_Result {
```
Omerisra6, line 19: What is the data structure of the component? Are there any other parts we should consider parsing?

modules/components/components-repository.php
```diff
@@ -0,0 +1,95 @@
...
+	public static function make(): Components_Repository {
+		return new self();
+	}
+
+	public function all() {
+		$posts = get_posts( [
+			'post_type' => Component_Document::TYPE,
+			'post_status' => 'publish',
+			'posts_per_page' => -1,
+		] );
```
Omerisra6: Consider bulk selecting (look at what we did at the global classes usage class), there could be a situation when there are too many posts and a memory issue.

modules/components/components-repository.php
```diff
@@ -0,0 +1,95 @@
...
+		$styles = [];
+		foreach ( $posts as $post ) {
+			$doc = Plugin::$instance->documents->get( $post->ID );
+
+			if ( ! $doc ) {
+				continue;
+			}
+
+			$components[] = $doc;
+			$styles[ $doc->get_main_id() ] = $this->extract_styles( $doc->get_elements_data() );
```
Omerisra6: Why do we need the styles to be included in the components repository?

Maybe we can create a helper function to extract styles from a component by ID and delegate the responsibility to the repository user.

modules/components/components-repository.php
```diff
@@ -0,0 +1,95 @@
...
+					'component_id' => null,
+					'error' => Error_Builder::make( 'content_validation_failed' )
+											->set_status( 400 )
+											->set_message( $error_message )
+											->build(),
+				];
+			}
+
+			throw $e;
+		}
```
Omerisra6: The repository shouldn't be responsible for errors; it should contain the minimal logic and throw exceptions if something happens.

let the api handle errors and their formatting.

modules/components/components-rest-api.php
```diff
@@ -0,0 +1,136 @@
...
+				'callback' => fn() => $this->route_wrapper( fn() => $this->get_styles() ),
+				'permission_callback' => fn() => is_user_logged_in(),
+			],
+		] );
+
+		register_rest_route( self::API_NAMESPACE, '/' . self::API_BASE, [
+			[
+				'methods' => 'POST',
+				'callback' => fn( $request ) => $this->route_wrapper( fn() => $this->create_component( $request ) ),
+				'permission_callback' => fn() => current_user_can( 'manage_options' ),
```
Omerisra6, line 53: Will the permissions be handled in a different PR?

modules/components/components-rest-api.php
```diff
@@ -0,0 +1,136 @@
...
+		] );
+	}
+
+	private function get_components() {
+		$components = $this->get_repository()->all();
+
+		$components_list = $components->get_components()->map( fn( $component ) => [
+			'id' => $component->get_id(),
+			'name' => $component->get_post()->post_title,
+		])->all();
```
Omerisra6: I think the normalization should be in the repo level

tests/phpunit/elementor/modules/components/mocks/component-1.json
Omerisra6, line 1: I think it would be better to create a test util that generates a component and set some defaults.

modules/components/documents/component.php
MiryamOren: It's not a `Document` method, it's for the cpt registration, we use it in `modules/components/module.php:56`:
```
register_post_type( Component_Document::TYPE, [
			'label'                 => Component_Document::get_title(),
			'labels' => Component_Document::get_labels(),
			'public' => false,
			'supports' => Component_Document::get_supported_features(),
		] );
```

modules/components/components-parser.php
MiryamOren: The create EP gets `name` & `content`, the content is parsed in the save document process, so I don't parse it here

modules/components/components-repository.php
MiryamOren: We need it to display the components styles in the editor canvas. Let's talk about where to put it

modules/components/components-rest-api.php
MiryamOren: Yes, since we don't yet have a product definition of the permission for this feature. 
I opened a task for it so we won't forget

modules/components/components-repository.php
MiryamOren: I'll move to the API file

Omerisra6 (approved):

PR #32408, opened 2025-08-24
Internal: [V4] transform base controls to affect indicator [ED-20527]

ronenboxer:


modules/atomic-widgets/prop-types/transform/transform-origin-prop-type.php
```diff
@@ -17,12 +17,9 @@ public static function get_key(): string {
...
 	}
 
-	protected function get_default_value_by_bind( $bind ): array {
+	protected function get_default_value_by_bind( $bind ): ?array {
 		return 'z' === $bind
-			? [
-				'size' => null,
-				'unit' => Size_Constants::UNIT_PX,
-			]
+			? null
```
ronenboxer, line 22: Fixes the validation error when publishing, since this is the correct way

packages/packages/core/editor-editing-panel/src/reset-style-props.tsx
```diff
@@ -18,8 +18,10 @@ export function useResetStyleValueProps() {
 	const isStyle = useIsStyle();
 	const { value, setValue, path } = useBoundProp();
 
+	const isInRepeater = path?.some( ( key ) => ! isNaN( Number( key ) ) );
```
ronenboxer: Resolves 'reset' indicator not showing inside the base transform popover

asafdl (approved):

PR #32469, opened 2025-08-27
Internal: Add move elements function [ED-20560]

IshayMaya:
### ✨ PR Description
Purpose: Add moveElement and moveElements functions to the editor-elements library for manipulating element positions with history support.
Main changes:
- Implemented moveElement function to relocate single elements between containers
- Created moveElements utility to handle batch movement of multiple elements with history tracking
- Added comprehensive test coverage for move operations including undo/redo functionality


packages/packages/libs/editor-elements/src/sync/move-element.ts
```diff
@@ -0,0 +1,43 @@
...
+	deleteElement( {
+		elementId,
+		options,
+	} );
+
+	const newContainer = createElement( {
+		containerId: targetContainerId,
+		model: modelToRecreate,
+		options: { edit: false, ...options },
+	} );
```
Omerisra6: How can we ensure that if the delete operation fails, the create operation won't start?

packages/packages/libs/editor-elements/src/sync/move-element.ts
```diff
@@ -0,0 +1,43 @@
...
+
+	if ( ! target ) {
+		throw new Error( `Target container with ID "${ targetContainerId }" not found` );
+	}
+
+	const modelToRecreate = container.model.toJSON();
+
+	deleteElement( {
+		elementId,
+		options,
```
Omerisra6: Don't you need to ensure the `edit: false` is set?

packages/packages/libs/editor-elements/src/sync/move-element.ts
```diff
@@ -0,0 +1,43 @@
...
+	} );
+
+	const newContainer = createElement( {
+		containerId: targetContainerId,
+		model: modelToRecreate,
+		options: { edit: false, ...options },
```
Omerisra6: Maybe add a comment explaining that the edit false is for not creating a history item?

packages/packages/libs/editor-elements/src/sync/move-elements.ts
```diff
@@ -0,0 +1,119 @@
...
+export type { MoveElementsParams, MovedElement, MovedElementsResult };
+
+export const moveElements = ( {
+	moves,
+	title,
+	subtitle = __( 'Elements moved', 'elementor' ),
+}: MoveElementsParams ): MovedElementsResult => {
+	const undoableMove = undoable(
+		{
+			do: ( { moves: movesParam }: { moves: MoveElementParams[] } ): MovedElementsResult => {
```
Omerisra6: We need to somehow avoid the Param suffix

packages/packages/libs/editor-elements/src/sync/move-element.ts
IshayMaya: im keeping the current functionality as the `move` command

packages/packages/libs/editor-elements/src/sync/move-element.ts
IshayMaya: delete does not have edit mode

IshayMaya: added the `useHistory` though

Omerisra6 (approved):

PR #32490, opened 2025-08-28
Internal: Disable already used transitions in selector [ED-19785]

RobiFerentz:


asafdl (dismissed):

packages/packages/libs/editor-controls/src/hooks/use-filtered-items-list.ts
```diff
@@ -15,7 +16,7 @@ export const useFilteredItemsList = ( itemsList: Category[], searchValue: string
 			acc.push( { type: 'category', value: category.label } );
 
 			filteredItems.forEach( ( item ) => {
-				acc.push( { type: 'item', value: item } );
+				acc.push( { type: 'item', value: item, disabled: disabledItems?.includes( item ) } );
```
murzin-nikita: let's use `diabled: false` as default value. This will remove the case when disabled is undefined

murzin-nikita:
```
disabled: disabledItems?.includes(item) ?? false
```

packages/packages/libs/editor-controls/src/controls/__tests__/item-selector.test.ts
```diff
@@ -59,11 +59,11 @@ describe( 'ItemSelector', () => {
 		// Assert.
 		const expected = [
 			{ type: 'category', value: 'System' },
-			{ type: 'item', value: 'Font1' },
-			{ type: 'item', value: 'Font2' },
+			{ type: 'item', value: 'Font1', disabled: undefined },
```
murzin-nikita: shouldn't this test be in `packages/packages/libs/editor-controls/src/components/__tests__/` folder, near `item-selector.tsx` component

RobiFerentz: Maybe, but I'm not the one who created this file

packages/packages/libs/editor-controls/src/hooks/use-filtered-items-list.ts
RobiFerentz: Done.

murzin-nikita (approved):

PR #32543, opened 2025-08-31
Internal: Support default html tags for atomic elements [ED-20609]

Omerisra6:
### ✨ PR Description
Purpose: Add support for configurable default HTML tags in atomic elements to improve component flexibility and semantic markup.

Main changes:
- Added define_default_html_tag() method to Atomic_Element_Base with 'div' as default value
- Refactored get_html_tag() to use the new default tag method instead of hardcoded 'div'
- Renamed Atomic_Tab_Link class to Atomic_Tab with default tag set to 'button'
- Updated JS editor view to use element's configured default_html_tag property


modules/atomic-widgets/assets/js/editor/atomic-element-view.js
```diff
@@ -15,8 +15,9 @@ export default function createAtomicElementView( type ) {
 
 			const tagControl = this.model.getSetting( 'tag' );
 			const tagControlValue = tagControl?.value || tagControl;
+			const defaultTag = this.model.config.default_html_tag;
```
IshayMaya, line 18: is `default_html_tag` required?

modules/atomic-widgets/elements/atomic-element-base.php
```diff
@@ -92,8 +97,9 @@ abstract protected static function define_props_schema(): array;
 	 */
 	protected function get_html_tag(): string {
 		$settings = $this->get_atomic_settings();
+		$default_html_tag = $this->define_default_html_tag();
 
-		return ! empty( $settings['link']['href'] ) ? 'a' : ( $settings['tag'] ?? 'div' );
+		return ! empty( $settings['link']['href'] ) ? 'a' : ( $settings['tag'] ?? $default_html_tag );
```
IshayMaya, line 102: maybe break this row down for better readibility?

modules/atomic-widgets/assets/js/editor/atomic-element-view.js
Omerisra6: Yes, I added a default value for it just in case

modules/atomic-widgets/elements/atomic-element-base.php
Omerisra6: I think it is readable...

IshayMaya (approved):

PR #32555, opened 2025-08-31
New: Components tab: add component [ED-20424]

MiryamOren:
### ✨ PR Description
Purpose: Add component insertion feature to Components tab with support for drag-and-drop placement in the editor.
Main changes:
- Implemented ComponentsTab UI with component listing and insertion functionality
- Created getContainerForNewElement utility to determine optimal element placement
- Modified base-container.js to support optional history tracking during element addition
- Added dropElement API to handle component insertion programmatically


assets/dev/scss/editor/panel/_elements.scss
```diff
@@ -43,6 +43,10 @@
 	z-index: $first-layer;
 }
 
+#elementor-panel-elements-search-area:empty {
```
IshayMaya, line 46: we don't have any selector that tells us we're in components?

packages/packages/core/editor-components/src/components/components-tab.tsx
```diff
@@ -1,6 +1,55 @@
...
+	const { data: components } = useComponents();
+
+	return (
+		<List sx={ { display: 'flex', flexDirection: 'column', gap: 0.5, px: 2 } }>
+			{ components?.map( ( component ) => <ComponentItem key={ component.id } component={ component } /> ) }
+		</List>
+	);
 }
+
+const ComponentItem = ( { component }: { component: Component } ) => {
```
IshayMaya, line 20: breaking it to another component is redundant here imo

packages/packages/core/editor-components/src/components/create-component-form/utils/replace-element-with-component.ts
```diff
@@ -4,13 +4,17 @@ import { numberPropTypeUtil } from '@elementor/editor-props';
...
-			settings: {
-				component_id: numberPropTypeUtil.create( componentId ),
-			},
-		},
+		newElement: getComponentModel( componentId ),
 		withHistory: false,
 	} );
 };
+
+export const getComponentModel = ( componentId: number ) => {
```
IshayMaya:
```suggestion
export const createComponentModel = ( componentId: number ) => {
```

packages/packages/core/editor-components/src/components/components-tab.tsx
```diff
@@ -1,6 +1,55 @@
...
+	const handleClick = () => {
+		addComponentToPage( getComponentModel( component.id ) );
+	};
+
+	return (
+		<ListItem disablePadding>
+			<ListItemButton
+				sx={ { border: '1px solid', borderColor: 'divider', py: 0.5, px: 1 } }
```
IshayMaya, line 28: don't we have a variant for that?

assets/dev/js/editor/views/base-container.js
```diff
@@ -140,6 +146,7 @@ module.exports = Marionette.CompositeView.extend( {
 				options: {
 					at: options.at,
 					scrollIntoView: options.scrollIntoView,
+					useHistory,
```
IshayMaya, line 149: this might break some existing functionalities no?

packages/packages/libs/editor-elements/src/sync/get-container-for-new-element.ts
```diff
@@ -0,0 +1,54 @@
...
+			break;
+		}
+		default: {
+			container = selectedEl;
+			break;
+		}
+	}
+
+	return { container, options };
+};
```
IshayMaya: maybe something like this?
also, maybe for now it should be in the components package? 
```suggestion
export const getContainerForNewElement = (): { container: V1Element | null; options?: { at: number } } => {
	const currentDocumentContainer = getCurrentDocumentContainer();
	const selectedElement = getSelectedElementContainer();

	let container, options;

	if ( selectedElement ) {
		switch ( selectedElement.model.get( 'elType' ) ) {
			case 'widget': {
				container = selectedElement?.parent;

				const selectedElIndex = selectedElement.view?._index ?? -1;

				if ( selectedElIndex > -1 ) {
					options = { at: selectedElIndex + 1 };
				}

				break;
			}
			case 'section': {
				container = selectedElement?.children?.[ 0 ];
				break;
			}
			default: {
				container = selectedElement;
				break;
			}
		}
	}

	return { container: container ?? currentDocumentContainer, options };
};

function getSelectedElementContainer() {
	const selectedElements = getSelectedElements();

	if ( selectedElements.length !== 1 ) {
		return undefined;
	}

	return getContainer( selectedElements[ 0 ].id );
}

```

assets/dev/scss/editor/panel/_elements.scss
MiryamOren: No, because it's not inside the tab content, it's above

packages/packages/core/editor-components/src/components/components-tab.tsx
MiryamOren: No, `ListItemButton` doesn't have variants

packages/packages/libs/editor-elements/src/sync/get-container-for-new-element.ts
MiryamOren: Nice! will do

packages/packages/libs/editor-elements/src/index.ts
```diff
@@ -39,7 +39,7 @@ export { getWidgetsCache } from './sync/get-widgets-cache';
 export { updateElementSettings, type UpdateElementSettingsArgs } from './sync/update-element-settings';
 export { generateElementId } from './sync/generate-element-id';
 export { replaceElement } from './sync/replace-element';
-export { getContainerForNewElement } from './sync/get-container-for-new-element';
+export { default as getCurrentDocumentContainer } from './sync/get-current-document-container';
```
IshayMaya: can you change the original function export to be named instead of default? it's our preferred convention anyway

IshayMaya (dismissed):

IshayMaya (approved):

PR #32612, opened 2025-09-03
Fix: Handle components created from top-level elements [ED-20695]

MiryamOren:
Add wrapper container to component instance

### ✨ PR Description
Purpose: Fix component creation process to properly handle top-level elements by wrapping widgets with containers when needed.
Main changes:
- Added logic to detect and handle top-level elements in the document structure
- Implemented container wrapper creation for widgets at document root level
- Refactored element location determination into separate helper functions
- Added test coverage for top-level element component creation scenario


packages/packages/libs/editor-elements/src/sync/__tests__/replace-element.test.ts
```diff
@@ -57,4 +56,49 @@ describe( 'replaceElement', () => {
 			options: { useHistory: true },
 		} );
 	} );
+
+	it( 'should wrap widget in container when replacing element at document top level', () => {
```
IshayMaya, line 60: maybe just keep the arrang/act/assert and remove other comments

packages/packages/libs/editor-elements/src/sync/replace-element.ts
```diff
@@ -21,11 +42,22 @@ export const replaceElement = ( { currentElement, newElement, withHistory = true
 		throw new Error( `Element ${ currentElement.id } not found in parent container. Cannot replace element.` );
 	}
```
IshayMaya, line 44: maybe something like this?

```
	let containerId = parent.id;

	const elementIndex = currentElement.view?._index ?? 0;

	// If the element is at document top level and is a widget, wrap it with an empty container
	if ( parent.id === 'document' && newElement.elType === 'widget' ) {
		containerId = createWrapperForWidget( parent.id, elementIndex );
	}

	return { containerId, index: elementIndex };
}

function createWrapperForWidget( parentId: string, elementIndex: number ) {
	const container = createElement( {
		containerId: parentId,
		model: { elType: 'container' },
		options: { at: elementIndex, useHistory: false },
	} );

	return container.id;
}

```

IshayMaya (approved):

PR #32635, opened 2025-09-04
Internal: [V4] Add to cart dynamic tag [ED-20495]

ronenboxer:


RobiFerentz (dismissed):

modules/atomic-widgets/controls/types/link-control.php
```diff
@@ -3,34 +3,21 @@
...
-				'post_title' => 'label',
-				'post_type' => 'groupLabel',
-			],
-		] ) );
+	private ?string $placeholder = null;
 
-		return $instance;
+	public function __construct( ...$args ) {
+		parent::__construct( ...$args );
+		$this->set_placeholder( __( 'Type or paste your URL', 'elementor' ) );
```
murzin-nikita: this is unnecessary

modules/atomic-widgets/query/has-query-props.php
```diff
@@ -0,0 +1,45 @@
...
+		$this->minimum_input_length = $input_length;
+
+		return $this;
+	}
+
+	public function set_allow_custom_values( bool $allow_custom_values ): self {
+		$this->allow_custom_values = $allow_custom_values;
+
+		return $this;
+	}
```
murzin-nikita:
```suggestion
     protected $config;
     
     public function set_query_props_config( $config ) {
	      $this->config = $config; 
	 }
```

murzin-nikita: and config will be: 
```
class Query_Props_Config
{
     public function set_post_types()
     public function set_namespace()
     ...
}

murzin-nikita (approved):

PR #32687, opened 2025-09-09
Fix: Identify send events opt in [ED-20877]

asafdl:
### ✨ PR Description
Purpose: Fix event tracking implementation to respect user opt-in preferences when sending analytics data.
Main changes:
- Added tracking_opted_in flag to library connection data based on data_share_opted_in setting
- Modified connection callbacks to properly check user opt-in status before enabling analytics
- Restructured mixpanel initialization to always initialize but only track for opted-in users
- Added enableTracking method to properly separate initialization from user-permitted tracking


app/modules/kit-library/assets/js/pages/cloud/connect-screen.js
```diff
@@ -33,6 +33,10 @@ export default function ConnectScreen( {
 				elementorCommon.config.library_connect.current_access_tier = data.access_tier;
 				elementorCommon.config.library_connect.plan_type = data.plan_type;
 
+				if ( data.tracking_opted_in && elementorCommon.config.editor_events ) {
```
davseve: Maybe set it in a function to increase readability?
E.g
```suggestion
				if ( this.isTrackingActive() ) {
					elementorCommon.config.editor_events.can_send_events = true;
```

I would also set this entire condition in a function, I see you use it more than once, this way we won't need to repeat ourself

RobiFerentz (approved):

PR #32701, opened 2025-09-10
Fix: Remove extra item actions in filter repeater control [ED-20941]

RobiFerentz:


IshayMaya (approved):

packages/packages/libs/editor-controls/src/controls/filter-control/filter-repeater-control.tsx
```diff
@@ -92,9 +92,6 @@ const Repeater = ( { propTypeUtil, label, filterPropName }: RepeaterProps ) => {
 						</>
 					}
 				/>
-				<DuplicateItemAction />
-				<DisableItemAction />
-				<RemoveItemAction />
```
asafdl, line 97: we should always have remove, no?

RobiFerentz: There is one, all 3 of these are leftover duplicates

asafdl (approved):

PR #32715, opened 2025-09-11
Internal(Components): Load styles in editor [ED-20255]

ronros-elementor:
### ✨ PR Description
Purpose: Add component styles loading functionality to editor to properly render component styles when used in documents

Main changes:
- Implemented styles repository provider for component styles with extraction and caching logic
- Added styles loading system to fetch component configurations when documents are loaded
- Refactored store structure with styles management and component configuration fetching
- Fixed style renderer to properly output component styles in the editor


packages/packages/core/editor-canvas/src/components/style-renderer.tsx
```diff
@@ -18,10 +18,8 @@ export function StyleRenderer() {
 
 	return (
 		<Portal container={ container }>
-			{ styleItems.map( ( item ) => (
-				<style data-e-style-id={ item.id } key={ `${ item.id }-${ item.breakpoint }` }>
-					{ item.value }
-				</style>
+			{ styleItems.map( ( item, i ) => (
+				<style key={ `${ item.id }-${ i }-${ item.breakpoint }` }>{ item.value }</style>
```
ronros-elementor, line 22: As we're not really using the data attribute `data-e-style-id`, nor should we (as react should ideally take care of the rendering natively with no side effect interference required) - removing it

packages/packages/core/editor-documents/src/types.ts
```diff
@@ -65,6 +65,7 @@ export type V1Document = {
 			main_dashboard: string;
 			all_post_type: string;
 		};
+		elements: unknown;
```
MiryamOren: maybe 
```suggestion
		elements: unknown[];
```

packages/packages/core/editor-components/src/init.ts
```diff
@@ -16,4 +32,39 @@ export function init() {
...
+
+			return ids;
+		} );
+	};
+
+	listenTo( commandStartEvent( 'editor/documents/attach-preview' ), () => {
+		const id = getCurrentDocumentId();
+
+		if ( id ) {
+			removeInitialDocumentStyles( id );
```
MiryamOren: Please unify into the event handler in line 36

MiryamOren (changes requested):
We should handle nested components

packages/packages/core/editor-components/src/init.ts
ronros-elementor: After looking into it closer, the separation actually is intended, one callback runs upon the execution of the attach-preview command, and the second runs _after_, so left it as as, but rearranged the code so both callbacks would be next to each other

lmk if that makes sense :)

ronros-elementor:
To elaborate a bit on the [changes](https://github.com/elementor/elementor/pull/32715/files/b281719a47eb413e5aa7b653457ee6377a014d17..f1775a5dbfdb82a90acef362c56138de2cde2e12) -

- Moved the components styles provider out of the editor-styles-repository, and into the editor-components package
  - Renamed the store from initial-document-styles-provider to components-styles-provider
  - Rearranged store-related files to be under 'store' folder within the package
- Added support of nested components by adding a recursion upon the traversal over data retrieved from 'load' in `.../editor-components/src/store/component-config`
- Added tests to the `add-compoennt-styles` module

packages/packages/core/editor-components/src/store/add-component-styles.ts
```diff
@@ -0,0 +1,39 @@
+import { type V1ElementData } from '@elementor/editor-elements';
+import { type StyleDefinition } from '@elementor/editor-styles';
+import { __dispatch as dispatch, __getState as getState } from '@elementor/store';
+
+import { getComponentIds } from '../utils/get-component-ids';
+import { load } from './component-config';
+import { selectData, slice } from './components-styles-store';
+
+type InitialDocumentId = number;
```
MiryamOren:
```suggestion
type ComponentId = number;
```

packages/packages/core/editor-components/src/store/__tests__/add-component-styles.test.ts
```diff
@@ -0,0 +1,198 @@
...
+		expect( load ).toHaveBeenCalledTimes( 2 );
+		expect( load ).toHaveBeenNthCalledWith( 1, DOC_ID );
+		expect( load ).toHaveBeenNthCalledWith( 2, COMP_ID );
+
+		expect( dispatch ).toHaveBeenCalledTimes( 2 );
+		expect( dispatch ).toHaveBeenNthCalledWith( 1, slice.actions.add( { [ DOC_ID ]: [] } ) );
+		expect( dispatch ).toHaveBeenNthCalledWith( 2, slice.actions.add( { [ COMP_ID ]: [ STYLE_1 ] } ) );
+
+		expect( selectData( getState() ) ).toEqual( {
+			[ DOC_ID ]: [],
```
MiryamOren: Don't we want to store here the components only? Why do we need the doc too?

packages/packages/core/editor-components/src/store/component-config.ts
```diff
@@ -0,0 +1,22 @@
...
+type ComponentParams = {
+	id: number;
+};
+
+export async function load( id: number ) {
+	return ajax.load< ComponentParams, V1ElementData >( getParams( id ) );
+}
+
+export async function invalidateCache( id: number ) {
+	ajax.invalidateCache< ComponentParams >( getParams( id ) );
```
MiryamOren: What does this function do?

packages/packages/core/editor-components/src/store/__tests__/add-component-styles.test.ts
MiryamOren: Actually, the function gets only component ids, not other docs. Let's update the tests so it will be clear

packages/packages/core/editor-components/src/store/add-component-styles.ts
```diff
@@ -0,0 +1,39 @@
...
+
+	const newDocumentsConfigs = await Promise.all(
+		ids.filter( ( id ) => ! data[ id ] ).map( async ( id ) => [ id, await load( id ) ] as const )
+	);
+
+	const styles = Object.fromEntries(
+		newDocumentsConfigs.map( ( [ id, config ] ) => [ id, extractStyles( config ) ] )
+	);
+
+	dispatch( slice.actions.add( styles ) );
```
MiryamOren: I suggest separating the styles extraction from the store update.
I think it will make it easier to read, and will definitely make the tests simpler

packages/packages/core/editor-components/src/store/component-config.ts
ronros-elementor: It clears the cached response from the V3's ajax manager, but my code only exposes it, it's accessible also simply via `elementorCommon.ajax.invalidateCache`, and its code can be found [here](https://github.com/elementor/elementor/blob/f2d0f6f927ad927e4a4ea9dc96839930193c5f8a/core/common/modules/ajax/assets/js/ajax.js#L241-L244)

MiryamOren (dismissed):

packages/packages/core/editor-components/src/api.ts
```diff
@@ -25,4 +32,6 @@ export const apiClient = {
 		httpService()
 			.post< HttpResponse< CreateComponentResponse > >( `${ BASE_URL }`, payload )
 			.then( ( res ) => res.data.data ),
+	getConfig: ( id: number ) => ajax.load< { id: number }, V1ElementData >( getParams( id ) ),
+	invalidateCache: ( id: number ) => ajax.invalidateCache< { id: number } >( getParams( id ) ),
```
MiryamOren:
```suggestion
	invalidateCache: ( id: number ) => ajax.invalidateComponentCofigCache< { id: number } >( getParams( id ) ),
```

packages/packages/core/editor-components/src/api.ts
```diff
@@ -25,4 +32,6 @@ export const apiClient = {
 		httpService()
 			.post< HttpResponse< CreateComponentResponse > >( `${ BASE_URL }`, payload )
 			.then( ( res ) => res.data.data ),
+	getConfig: ( id: number ) => ajax.load< { id: number }, V1ElementData >( getParams( id ) ),
```
MiryamOren:
```suggestion
	getComponentConfig: ( id: number ) => ajax.load< { id: number }, V1ElementData >( getParams( id ) ),
```

packages/packages/core/editor-components/src/store/__tests__/components-styles-provider.test.ts
```diff
@@ -0,0 +1,56 @@
...
+
+describe( 'componentsStylesProvider', () => {
+	const DOC_ID_1 = 1;
+	const DOC_ID_2 = 2;
+
+	const STYLE_1 = createMockStyleDefinition( { id: 's-1' } );
+	const STYLE_2 = createMockStyleDefinition( { id: 's-2' } );
+	const STYLE_3 = createMockStyleDefinition( { id: 's-3' } );
+
+	beforeEach( () => {
```
MiryamOren, line 24: Maybe add a test for get all?

packages/packages/core/editor-components/src/store/components-styles-store.ts
```diff
@@ -0,0 +1,32 @@
+// import { type StyleDefinition } from '@elementor/editor-styles';
```
MiryamOren: Delete file

packages/packages/core/editor-components/src/store/store.ts
```diff
@@ -15,25 +15,37 @@ export type ComponentsState = {
...
 	initialState,
 	reducers: {
 		add: ( state, { payload } ) => {
-			state.data.push( payload );
+			state.data = { ...payload };
 		},
 		load: ( state, { payload } ) => {
 			state.data = payload;
 		},
+		remove( state, { payload }: PayloadAction< { id: ComponentId } > ) {
```
MiryamOren:
```suggestion
		removeStyles( state, { payload }: PayloadAction< { id: ComponentId } > ) {
```

packages/packages/core/editor-components/src/store/store.ts
```diff
@@ -15,25 +15,37 @@ export type ComponentsState = {
 	data: Component[];
 	loadStatus: Status;
 	createStatus: Status;
+	styles: StylesDefinition;
 };
 
+export type StateWithInitialDocumentsStyles = SliceState< typeof slice >;
```
MiryamOren:
```suggestion
export type ComponentsSlice = SliceState< typeof slice >;
```

MiryamOren: Also, looks like we only use it in this file, so we shouldn't export it

packages/packages/core/editor-components/src/store/thunks.ts
```diff
@@ -15,4 +16,8 @@ const loadComponents = createAsyncThunk( 'components/load', async () => {
 	return response;
 } );
 
-export { createComponent, loadComponents };
+async function loadStyles( ids: number[] ): Promise< [ number, V1ElementData ][] > {
```
MiryamOren: This is not a thunk, let's move it to `load-components-styles.ts`

MiryamOren (dismissed):

MiryamOren (approved):

PR #32738, opened 2025-09-14
Internal: [V4] apply new tag value when link is conflicting [ED-19977]

ronenboxer:
### ✨ PR Description
```
Purpose: Add HTML tag control and dependency management system to automatically change tag to 'a' when link is set on atomic widgets.

Main changes:
- Created new Html_Tag_Control to handle tag selection with fallback label support for conflicting dependencies
- Implemented prop dependency system that saves previous state and applies new value when dependencies conflict
- Added ConditionalControlInfotip component to display tooltips explaining why certain controls are disabled
- Extended dependency evaluation to track failing dependencies and support automatic value changes
```


ronenboxer:
PW tests
https://github.com/elementor/elementor/actions/runs/17722424328
https://github.com/elementor/elementor/actions/runs/17722915719

packages/packages/libs/editor-props/src/utils/prop-dependency-utils.ts
```diff
@@ -11,19 +11,32 @@ type ParsedTerm = DependencyTerm;
...
-			: evaluateTerm( term, extractValue( term.path, values )?.value )
-	);
+	const failingDependencies: DependencyTerm[] = [];
+	const isMet = terms[ method ]( ( term: ParsedTerm | Dependency ) => {
+		const isNestedDependency = isDependency( term );
+		const result = isNestedDependency
+			? isDependencyMet( term, values ).isMet
+			: evaluateTerm( term, extractValue( term.path, values )?.value );
+
+		if ( ! result && ! isNestedDependency ) {
```
RobiFerentz, line 32: maybe 
```ts
! (result || isNestedDependency)
```
?

packages/packages/libs/editor-controls/src/controls/html-tag-control.tsx
```diff
@@ -0,0 +1,88 @@
...
+	disabled?: boolean;
+};
+
+type Props = {
+	options: SelectOption[];
+	onChange?: ( newValue: string | null, previousValue: string | null | undefined ) => void;
+	fallbackLabels?: Record< string, string >;
+};
+
+const StyledSelect = styled( Select )( () => ( { '.MuiSelect-select.Mui-disabled': { cursor: 'not-allowed' } } ) );
```
RobiFerentz, line 25: This is just a nitpick, but I've seen lately that styled components are causing performance problems in many react projects. Perhaps we should avoid using them? Maybe phase them out? 
Just something to think about if we want some performance gains.

ronenboxer: I think it's best to sort that out with Rotem

packages/packages/libs/editor-props/src/utils/prop-dependency-utils.ts
ronenboxer: This will require more brackets and I personally find the && more straight forward

RobiFerentz (dismissed):

packages/packages/libs/editor-controls/src/controls/html-tag-control.tsx
```diff
@@ -0,0 +1,88 @@
...
+									<Typography component="span" variant="caption" color="text.tertiary">
+										{ displayText }
+									</Typography>
+								);
+							}
+							return '';
+						}
+						const option = findOptionByValue( selectedValue );
+						return option?.label || fallbackLabels[ selectedValue ] || selectedValue;
+					} }
```
davseve: Maybe we can extract this entire function to increase the readability of this component.
There is also a nested condition that we can maybe clean up a bit.

packages/packages/libs/editor-props/src/utils/prop-dependency-utils.ts
davseve: Have to agree with Ronen on this one :)

davseve (dismissed):

packages/packages/libs/editor-controls/src/controls/html-tag-control.tsx
ronenboxer: Done, also done the same for 2 function in the utils file under `editor-editing-panel` package

davseve (approved):

RobiFerentz (approved):

PR #32742, opened 2025-09-15
Internal(Cache-Validity): Reduce stored data size [ED-20908]

ronros-elementor:
This PR introduces changes the `Cache_Validity` class to reduce the size it uses in the DB.

## Changing the data stored 

```
$cache_validity->validate( [ 'cache-key', 'nested' ] )
```

### Before:
```
'elementor_atomic_cache_validity-cache-key` => [
  'state' => false,
  'meta' => null,
  'children' => [
    'nested' => [
      'state' => true,
      'children' => []
    ]
  ]
]
```

### After
```
'elementor_atomic_cache_validity__cache-key` =>  [
  'state' => false,
  'children' => [
    'nested' => true
  ]
]
```
Since 'nested' has no children nor meta - will store only the boolean value

## Clean data upon invalidation
Now invalidation will remove the item from the data, and not just set its state as false and reset its meta (as anyway the children validation state is useless as all are set to `false`)

## Comments
- The option prefix was changed to `elementor_atomic_cache_validity__` to avoid breakage upon a version rollback (as the data structure would be different of what was expected prior to this change)
- Added a small cleanup to the atomic styles fonts options - `elementor_atomic_styles_fonts-...` - if no fonts used for a given style - remove the options entirely
- All used options will *not* be autoloaded

### ✨ PR Description
Purpose: Refactor cache validity implementation to reduce stored data size and optimize memory usage in atomic widgets system.
Main changes:
- Moved Cache_Validity to a dedicated namespace with improved tree-based storage structure
- Added option to store cache data with autoload=false to prevent unnecessary memory usage
- Implemented data normalization to remove obsolete nodes from cache tree
- Added post deletion hook to automatically invalidate associated caches


modules/atomic-widgets/styles/atomic-widget-styles.php
```diff
@@ -24,6 +24,11 @@ public function register_hooks() {
 			'elementor/core/files/clear_cache',
 			fn() => $this->invalidate_cache(),
 		);
+
+		add_action(
+			'deleted_post',
```
Omerisra6, line 29: add test

modules/atomic-widgets/cache-validity/cache-validity-item.php
```diff
@@ -0,0 +1,254 @@
...
+	 */
+	private function get_stored_data() {
+		return get_option( self::CACHE_KEY_PREFIX . $this->root, [ 'state' => false ] );
+	}
+
+	/**
+	 * @param array{state: boolean, meta: array<string, mixed> | null, children: array<string, self>} $data
+	 */
+	private function update_stored_data( $data ) {
+		update_option( self::CACHE_KEY_PREFIX . $this->root, $data, false );
```
Omerisra6, line 269: Add a comment explaining the autoload logic

modules/atomic-widgets/cache-validity/cache-validity-item.php
```diff
@@ -0,0 +1,254 @@
...
+
+	public function invalidate( array $keys ) {
+		return $this->wrap_exception( function() use ( $keys ) {
+			$data = $this->get_stored_data();
+
+			if ( empty( $keys ) ) {
+				$this->delete_stored_data();
+
+				return;
+			}
```
Omerisra6, line 58: move before the `$data` decleration

modules/atomic-widgets/cache-validity/cache-validity-item.php
```diff
@@ -0,0 +1,254 @@
...
+			$data = $this->get_stored_data();
+
+			if ( empty( $keys ) ) {
+				$data['state'] = true;
+				$data['meta'] = $meta;
+
+				$this->update_stored_data( $data );
+
+				return;
+			}
```
Omerisra6, line 46: same

modules/atomic-widgets/cache-validity/cache-validity-item.php
```diff
@@ -0,0 +1,254 @@
...
+
+		return $data;
+	}
+
+	/**
+	 * @param array{state: boolean, meta: array<string, mixed> | null, children: array<string, self>} | boolean $data
+	 * @param array<string>                                                                                     $keys
+	 * @return array<array{state: boolean, meta: array<string, mixed> | null, children: array<string, self>} | boolean | null>
+	 */
+	private function get_data_with_placeholders( array $data, array $keys ): ?array {
```
Omerisra6: Try to avoid changing by reference if you don't have to

modules/atomic-widgets/cache-validity/cache-validity-item.php
```diff
@@ -0,0 +1,254 @@
...
+	 * @param array<string>                                                                                     $keys
+	 * @param mixed | null                                                                                      $meta
+	 */
+	private function validate_nested_node( array $data, array $keys, $meta = null ) {
+		$data = $this->get_data_with_placeholders( $data, $keys );
+
+		$last_key = array_pop( $keys );
+		$parent = &$this->get_node( $data, $keys );
+
+		if ( ! isset( $parent['children'] ) ) {
```
Omerisra6: Check if it works when the parent is null, and add a check if it doesn't

modules/atomic-widgets/cache-validity/cache-validity-item.php
```diff
@@ -0,0 +1,254 @@
...
+			}
+
+			$node['state'] = true;
+
+			if ( null !== $meta ) {
+				$node['meta'] = $meta;
+			}
+		} else {
+			$parent['children'][ $last_key ] = true;
+		}
```
Omerisra6: try to simplify

modules/atomic-widgets/cache-validity/cache-validity-item.php
```diff
@@ -0,0 +1,254 @@
...
+		$parent = &$this->get_node( $data, $keys );
+
+		if ( ! is_array( $parent ) || ! isset( $parent['children'][ $last_key ] ) ) {
+			// node doesn't exist - no need to do anything
+			return;
+		}
+
+		if ( isset( $parent['children'] ) && count( $parent['children'] ) > 1 ) {
+			// if parent has more children - simply unset the invalidated node from the parent's children.
+			unset( $parent['children'][ $last_key ] );
```
Omerisra6: why can't you always do that?

modules/atomic-widgets/cache-validity/cache-validity-item.php
```diff
@@ -0,0 +1,254 @@
...
+			$this->update_stored_data( $data );
+			return;
+		}
+
+		$data = $this->get_data_without_placeholders( $data, $keys, $last_key );
+
+		$this->update_stored_data( $data );
+	}
+
+	private function get_data_without_placeholders( array $data, array $keys, string $last_key ) {
```
Omerisra6: I'm not sure how, but I think it would be great if you could add the return structure here and in the `get_data_with_placeholders`.

modules/atomic-widgets/cache-validity/cache-validity-item.php
```diff
@@ -0,0 +1,254 @@
...
+				$current = &$current['children'][ $key ];
+			} else {
+				$current = null;
+			}
+		}
+
+		return $current;
+	}
+
+	private function is_placeholder_node( $data ): bool {
```
Omerisra6: we need to come up with a better name

modules/atomic-widgets/cache-validity/cache-validity-item.php
```diff
@@ -0,0 +1,254 @@
...
+					'key' => $key,
+					'node' => &$parent,
+				];
+			} elseif ( is_array( $current ) && ! $this->is_placeholder_node( $current ) ) {
+				$remove_node = [
+					'key' => null,
+					'node' => null,
+				];
+			}
+		}
```
Omerisra6, line 192: Maybe we could extract this logic into some kind of helper function (find node or something)

modules/atomic-widgets/cache-validity/cache-validity-item.php
```diff
@@ -0,0 +1,254 @@
...
+		];
+
+		if ( ! isset( $data['children'] ) || $this->is_placeholder_node( $data ) ) {
+			$remove_node['node'] = &$data;
+		}
+
+		$current = &$data;
+		$parent = &$current;
+
+		while ( ! empty( $keys ) ) {
```
Omerisra6, line 176: Can you explain the logic here?

Omerisra6 (changes requested):

modules/atomic-widgets/cache-validity/cache-validity-item.php
ronros-elementor: Here I cannot just early return, as I _am_ relying on the previous data, otherwise I'd loose its children (if set in advance)

modules/atomic-widgets/cache-validity/cache-validity-item.php
ronros-elementor: After going over the code - it was intentional to pass by ref in order for the `$current` mutation to apply for the general data object that is returned, as long as the param itself is not passed by reference - the original value passed to `$data` is not mutated

modules/atomic-widgets/cache-validity/cache-validity-item.php
ronros-elementor: The first condition is indeed redundant, due to the check to see the invalidated node even exists
The latter though is needed, as I don't need to search and remove placeholders as long as the invalidated node is not a single child node (i.e. I can just unset, and not clear anything in addition)

modules/atomic-widgets/cache-validity/cache-validity-item.php
ronros-elementor: Changed the implementation a bit to make it more readable, also based on your prior comments, lmk if it's clearer now

modules/atomic-widgets/cache-validity/cache-validity-item.php
ronros-elementor: Went with `normalize` instead of `without_placeholders`, `get_data_with_path` instead of `with_placeholders`, and `is_rendundant_node` instead of `is_placeholder_node`

Also named an extracted logic function to `find_redundant_path_root`, which would return the key and node reference to the root node of a path full with these "placeholder"/"redundant" nodes

modules/atomic-widgets/cache-validity/cache-validity-item.php
```diff
@@ -0,0 +1,281 @@
...
+
+		return $root_node;
+	}
+
+	/**
+	 * @param array{state: boolean, meta: array<string, mixed> | null, children: array<string, self>} | boolean $data
+	 * @param array<string>                                                                                     $keys
+	 * @return array{state: boolean, meta: array<string, mixed> | null, children: array<string, self>}
+	 */
+	private function get_data_with_path( array $data, array $keys ): ?array {
```
Omerisra6:
```suggestion
	private function get_data_by_path( array $data, array $keys ): ?array {
```

modules/atomic-widgets/cache-validity/cache-validity-item.php
```diff
@@ -0,0 +1,281 @@
...
+
+		$current = &$data;
+		$parent = &$current;
+
+		while ( ! empty( $keys ) ) {
+			$key = array_shift( $keys );
+			$parent = &$current;
+			$current = &$current['children'][ $key ];
+
+			if ( isset( $current['children'] ) && $this->is_redundant_node( $current ) && empty( $root_node['node'] ) ) {
```
Omerisra6: "Check if we really need the isset."

Omerisra6 (dismissed):

modules/atomic-widgets/cache-validity/cache-validity-item.php
```diff
@@ -0,0 +1,283 @@
...
+	}
+
+	/**
+	 * Retrieves the stored tree, guaranteed to have a path representation based on $keys
+	 *
+	 * @param array{state: boolean, meta: array<string, mixed> | null, children: array<string, self>} | boolean $data
+	 * @param array<string>                                                                                     $keys
+	 * @return array{state: boolean, meta: array<string, mixed> | null, children: array<string, self>}
+	 */
+	private function get_data_with_enforced_path( array $data, array $keys ): ?array {
```
IshayMaya: maybe `ensure_data_path` ?

modules/atomic-widgets/cache-validity/cache-validity-item.php
```diff
@@ -0,0 +1,283 @@
...
+
+		if ( count( $parent['children'] ) > 1 ) {
+			// if the parent has more children - simply unset the invalidated node
+			unset( $parent['children'][ $last_key ] );
+
+			$this->update_stored_data( $data );
+			return;
+		}
+
+		$data = $this->get_normalized_data( $data, $keys, $last_key );
```
IshayMaya: do we get here with a single children or can it be 0? if not maybe the condition should be reversed?

modules/atomic-widgets/cache-validity/cache-validity-item.php
```diff
@@ -0,0 +1,283 @@
...
+	}
+
+	/**
+	 * @param array{state: boolean, meta: array<string, mixed> | null, children: array<string, self>} $data
+	 * @param array<string>                                                                           $keys
+	 * @param string                                                                                  $last_key
+	 * @return array{state: boolean, meta: array<string, mixed> | null, children: array<string, self>}
+	 */
+	private function get_normalized_data( array $data, array $keys, string $last_key ) {
+		$redundant_root = &$this->find_empty_parents_path_root( $data, $keys, $last_key );
```
IshayMaya: i don't have other word in mind but im not sure about redundant

IshayMaya (approved):

PR #32849, opened 2025-09-22
Internal: WP-Rest to have more querying types, terms and users[ED-20487]

ronenboxer:


RobiFerentz (dismissed):

modules/atomic-widgets/dynamic-tags/dynamic-tags-editor-config.php
```diff
@@ -229,30 +233,55 @@ private function convert_textarea_control_to_atomic( $control ) {
...
-		} else {
-			$post_types = isset( $control['autocomplete']['query']['post_type'] ) ? $control['autocomplete']['query']['post_type'] : [];
-			$post_types = ! empty( $post_types ) && 'any' !== $post_types ? $post_types : null;
+		$query_type = Post_Query::ENDPOINT;
+
+		switch ( true ) {
+			case $this->is_control_term_query( $control ):
+				$query_type = Term_Query::ENDPOINT;
+				$included_types = null;
+				$excluded_types = [];
```
murzin-nikita: can we get rid of null ? 
Like use only arrays and empty array in case of null. This will reduce overall complexity by using the same typing/interface everywhere

modules/wp-rest/classes/term-query.php
```diff
@@ -0,0 +1,200 @@
...
+
+		$terms = new Collection( get_terms( $query_args ) );
+		$term_by_id = is_numeric( $term ) ? get_term( $term ) : null;
+
+		if ( $term_by_id && ! $terms->find( fn( $t ) => (int) $t->term_id === (int) $term ) ) {
+			$terms->push( $term_by_id );
+		}
+
+		$terms = $terms->filter( function ( $term ) use ( $included_types, $excluded_types ) {
+			return ( empty( $included_types ) || in_array( $term->taxonomy, $included_types, true ) )
```
murzin-nikita: why we don't add `'taxonomy' => 'posts'` to get_terms() request?

modules/wp-rest/classes/user-query.php
```diff
@@ -0,0 +1,231 @@
...
+			'fields' => $fields,
+		];
+
+		if ( ! empty( $params[ self::META_QUERY_KEY ] ) && is_array( $params[ self::META_QUERY_KEY ] ) ) {
+			$query_args['meta_query'] = $params[ self::META_QUERY_KEY ];
+		}
+
+		$this->add_filter_to_customize_query();
+
+		$users = Collection::make( get_users( $query_args ) )
```
murzin-nikita: we can do `get_users` by role
```
	'role'         => '',
	'role__in'     => array(),
	'role__not_in' => array(),
```

modules/wp-rest/classes/user-query.php
```diff
@@ -0,0 +1,231 @@
...
+			->filter( function ( $user ) use ( $included_roles, $excluded_roles ) {
+				if ( ! ( $user instanceof \WP_User ) ) {
```
murzin-nikita: it looks like if `'fields'` won't be `'all'` - we will always get false here

tests/phpunit/elementor/modules/wp-rest/providers/term-query.php
```diff
@@ -0,0 +1,193 @@
...
+		foreach ( $this->terms as $term ) {
+			wp_delete_term( $term->term_id, $term->taxonomy, $this->force_delete );
+		}
+
+		foreach ( get_terms( [
+			'exclude' => [ $this->uncategorized_term_id ],
+			'number' => $this->count,
+			'hide_empty' => false,
+		] ) as $term ) {
+			var_dump( $term->name );
```
murzin-nikita: var_dump

modules/atomic-widgets/dynamic-tags/dynamic-tags-editor-config.php
ronenboxer: The thing is, we have a default setter for some of the params, and having `[ ]` as value is a statement meaning to override the default setter

modules/wp-rest/classes/term-query.php
ronenboxer: `taxonomy' should be array/string of taxonomies, such as 'category', 'post_tag', custom taxonomies etc...

modules/wp-rest/classes/user-query.php
ronenboxer: Can't, firstly, what we are interested in is capabilities, which can be customized per user, ever if the user's role doesn't include them. Secondly, this doesn't seem to propagate roles, meaning searching for editor should also bring up administrator, but it doesn't...

modules/wp-rest/classes/user-query.php
ronenboxer: You are right!

ronenboxer: I removed the fields, it makes no sense to use that

tests/phpunit/elementor/modules/wp-rest/providers/term-query.php
ronenboxer: Got me

modules/wp-rest/classes/user-query.php
ronenboxer: I have an idea how to implement it better without the filter and reduce.
I'll ping you

modules/wp-rest/base/query.php
```diff
@@ -0,0 +1,155 @@
...
+			$raw = sanitize_text_field( $input );
+			$decoded = json_decode( $raw, true );
+			if ( is_array( $decoded ) ) {
+				$input = $decoded;
+			} else {
+				$input = false !== strpos( $raw, ',' ) ? explode( ',', $raw ) : ( '' !== $raw ? [ $raw ] : [] );
+			}
+		}
+
+		$array = new Collection( json_decode( json_encode( $input ), true ) );
```
murzin-nikita: json manipulations isn't necessary here, $input will be array in all cases

modules/atomic-widgets/dynamic-tags/dynamic-tags-editor-config.php
```diff
@@ -263,41 +267,66 @@ private function convert_number_control_to_atomic( $control ) {
...
-			$post_types = [ Source_Local::CPT ];
-			$query_config[ Post_Query::EXCLUDED_POST_TYPE_KEYS ] = [];
-		} else {
-			$post_types = isset( $control['autocomplete']['query']['post_type'] ) ? $control['autocomplete']['query']['post_type'] : [];
-			$post_types = ! empty( $post_types ) && 'any' !== $post_types ? $post_types : null;
+		$query_type = Post_Query::ENDPOINT;
+
+		switch ( true ) {
+			case $this->is_control_term_query( $control ):
+				$query_type = Term_Query::ENDPOINT;
```
murzin-nikita, line 286: can we move all these query logic to `modules/atomic-widgets/dynamic-tags/dynamic-tags-converter.php` . 

And here just do: 
```
private function convert_query_control_to_atomic( $control ) {
    return Dynamic_Tags_Converter::convert_query_control_to_atomic( $control )
}

File: modules/atomic-widgets/dynamic-tags/dynamic-tags-editor-config.php [hunk 8]
```diff
@@ -263,41 +267,66 @@ private function convert_number_control_to_atomic( $control ) {
...
-			->set_label( $control['label'] );
+			->set_label( html_entity_decode( $control['label'] ) );
 	}
 
 	private function convert_textarea_control_to_atomic( $control ) {
 		return Textarea_Control::bind_to( $control['name'] )
-			->set_placeholder( $control['placeholder'] ?? '' )
-			->set_label( $control['label'] );
+			->set_placeholder( html_entity_decode( $control['placeholder'] ?? '' ) )
+			->set_label( html_entity_decode( $control['label'] ) );
```
murzin-nikita: can we add `html_entity_decode` inside set_label / set_placeholder function ?

modules/wp-rest/classes/user-query.php
```diff
@@ -0,0 +1,291 @@
...
+			],
+		];
+	}
+
+	private function get_propagated_role_list( $roles = [] ) {
+		if ( empty( $roles ) ) {
+			return [];
+		}
+
+		$this->arrange_roles_by_capabilities();
```
murzin-nikita: 1. Let's rework `arrange_roles_by_capabilities` to be a regular function, which will return sorted roles list with caching
2. We can keep  `add_action( 'add_role', fn () => $this->arrange_roles_by_capabilities( true ), 10, 2 )` in order to have this list up-to-date 
3. Idea to sort roles by capabilities attached to temp users looks like overkill. 
3.1 We can get capabilities of each role and test it against each other, something like: 
```
    $role_a = get_role( $role_a_slug );
    $role_b = get_role( $role_b_slug );

    if ( ! $role_a || ! $role_b ) {
        return 0;
    }

    // Get active capabilities for each role
    $caps_a = array_keys( array_filter( $role_a->capabilities ) );
    $caps_b = array_keys( array_filter( $role_b->capabilities ) );

    // Check if role A has all capabilities of role B
    $a_has_all_b = empty( array_diff( $caps_b, $caps_a ) );
    
    // Check if role B has all capabilities of role A
    $b_has_all_a = empty( array_diff( $caps_a, $caps_b ) );

    if ( $a_has_all_b && ! $b_has_all_a ) {
        return -1; // A is stronger (should come first)
    } elseif ( $b_has_all_a && ! $a_has_all_b ) {
        return 1;  // B is stronger  
    } else {
        // Equal capabilities or incomparable - fall back to capability count
        return count( $caps_b ) - count( $caps_a );
    }
```
_(i'm not sure if we need get_role here, or we can get capabilities even from `global $wp_roles;`)_

3.2 if this won't work we can do just a direct sql request on wp_usermeta(with distinct maybe)

modules/atomic-widgets/dynamic-tags/dynamic-tags-editor-config.php
ronenboxer: Done

modules/atomic-widgets/dynamic-tags/dynamic-tags-editor-config.php
ronenboxer: I think you might have confused this with the`convert_control_to_prop_type`, and this is irrelevant there

modules/wp-rest/base/query.php
ronenboxer: Removed

modules/atomic-widgets/query/query-builder.php
```diff
@@ -2,26 +2,44 @@
...
 
 if ( ! defined( 'ABSPATH' ) ) {
 	exit; // Exit if accessed directly.
 }
 
 class Query_Builder {
+	const QUERY_TYPE_KEY = 'query_type';
+
 	public static function build( ?array $config = [] ): array {
+		$map = [
```
asafdl: so the whole notion of this class seems off to me. we have a general builder class, that maps different queries and "builds" the required query from my understanding.
So this is basically a factory pattern, with a query factory and 3 different types of queries to create
However it's done in a bit of a bizzare way.
this is how i would expect a factory pattern to look like:
```php
abstract class Query_Builder {
    abstract public function build(array $config): array;
}

class Post_Query_Builder extends Query_Builder {
    public function build(array $config): array {
        // Post-specific building logic
    }
}

class Term_Query_Builder extends Query_Builder {
    public function build(array $config): array {
        // Term-specific building logic  
    }
}

class Query_Builder_Factory {
    private const BUILDERS = [
        Post_Query::ENDPOINT => Post_Query_Builder::class,
        Term_Query::ENDPOINT => Term_Query_Builder::class,
        User_Query::ENDPOINT => User_Query_Builder::class,
    ];
    
    public static function create(string $endpoint): Query_Builder {
        $class = self::BUILDERS[$endpoint] ?? throw new \Exception('Unsupported query type');
        return new $class();
    }
}```

File: modules/wp-rest/base/query.php [hunk 11]
```diff
@@ -0,0 +1,153 @@
...
+		$nonce = $request->get_header( self::NONCE_KEY );
+
+		return current_user_can( 'edit_posts' ) && wp_verify_nonce( $nonce, 'wp_rest' );
+	}
+
+	/**
+	 * @param callable $cb The route callback.
+	 * @return \WP_REST_Response | \WP_Error
+	 */
+	private function route_wrapper( callable $cb ) {
```
asafdl: this can be send( callable $query )

tests/phpunit/elementor/modules/wp-rest/test-post-query.php
```diff
@@ -37,9 +37,19 @@ public function test_post_query_results() {
 	private function execute( $params, $expected ) {
 		// Arrange
 		$request = new \WP_REST_Request( 'GET', self::URL );
-		$request->set_param( Post_Query::EXCLUDED_POST_TYPE_KEYS, $params[ Post_Query::EXCLUDED_POST_TYPE_KEYS ] );
+		$request->set_param( Post_Query::EXCLUDED_TYPE_KEY, $params[ Post_Query::EXCLUDED_TYPE_KEY ] ?? null );
```
asafdl, line 40: can we add more flows here? i assume that there are dozens of flows with the amount of if else this code possess

tests/phpunit/elementor/modules/wp-rest/providers/user-query.php
```diff
@@ -0,0 +1,112 @@
...
+		return $this->factory()->user->create_and_get( [
+			'user_login' => $name,
+			'display_name' => implode( ' ', explode( '_', $name ) ),
+			'user_email' => $name . '@test.com',
+			'user_pass' => 'password',
+			'role' => $role,
+		] );
+	}
+
+	public function data_provider_user_query() {
```
asafdl, line 65: can we add more flows here? maybe negative flows

asafdl: this goes for all the tests, im sure that there are more flows to test given the amount of if / else the code has

modules/atomic-widgets/query/query-builder.php
ronenboxer: Done

modules/wp-rest/base/query.php
ronenboxer: Done

tests/phpunit/elementor/modules/wp-rest/providers/user-query.php
ronenboxer: User has no more if else other than name like and id, so there's no need,
as for the rest, the test cases should cover it, made sure to go through each if else at least once

asafdl (dismissed):

murzin-nikita (approved):

PR #32925, opened 2025-09-30
Fix: Autocomplete results should not show categories if there is only one. [ED-21046]

RobiFerentz:
### ✨ PR Description
Purpose: Fix autocomplete dropdown to hide category labels when all items belong to a single category.

Main changes:
- Updated isCategorizedOptionPool to only group options when multiple categories exist
- Added test case to verify single category options are not grouped
- Added test for category visibility when filtering autocomplete results

### DIFF
--- a/packages/packages/libs/editor-controls/src/components/autocomplete.tsx
+++ b/packages/packages/libs/editor-controls/src/components/autocomplete.tsx
@@ -173,7 +173,13 @@ export function findMatchingOption(
 }
 
 export function isCategorizedOptionPool( options: FlatOption[] | CategorizedOption[] ): options is CategorizedOption[] {
-	return options.every( ( option ) => 'groupLabel' in option );
+	if ( options.length <= 1 ) {
+		return false;
+	}
+
+	const uniqueGroupLabels = new Set( options.map( ( option ) => option.groupLabel ) );
+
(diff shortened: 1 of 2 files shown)


packages/packages/libs/editor-controls/src/components/autocomplete.tsx
```diff
@@ -173,7 +173,9 @@ export function findMatchingOption(
 }
 
 export function isCategorizedOptionPool( options: FlatOption[] | CategorizedOption[] ): options is CategorizedOption[] {
-	return options.every( ( option ) => 'groupLabel' in option );
+	const uniqueGroupLabels = new Set( options.map( ( option ) => option.groupLabel ) );
```
marcin-el, line 180: what about 
```suggestion
	const [first] = options;
    const hasMultipleGroups = options.some(o => o.groupLabel !== first.groupLabel);
```

RobiFerentz: It does the same thing in a different way. There's only a tiny optimization that could be made for either method, but it's not gonna change much.

marcin-el (dismissed):

marcin-el (approved):

YaelAvitan (approved):

PR #32938, opened 2025-09-30
Internal: Improve tabs initial render in frontend [ED-20988]

Omerisra6:
- need to handle nested nested tabs
-

### ✨ PR Description
Purpose: Add render context mechanism to improve tabs initial rendering by setting correct active tab on page load.
Main changes:
- Created Render_Context class to maintain hierarchical context across widget rendering
- Extended Atomic_Element_Base with print_content and define_render_context methods
- Added initial visibility state to tab panels with proper hidden attributes
- Implemented get_atomic_setting helper method for simplified prop access


modules/atomic-widgets/elements/atomic-tabs/atomic-tab-panel.php
```diff
@@ -112,6 +112,10 @@ protected function add_render_attributes() {
 		$base_style_class = $this->get_base_styles_dictionary()[ static::BASE_STYLE_KEY ];
 		$initial_attributes = $this->define_initial_attributes();
 
+		$active_tab = $this->get_context( 'e-tabs' )['activeTab'] ?? null;
```
Nevoss:
```suggestion
		$active_tab = $this->get_context( Atomic_Tabs::CONTEXT )['activeTab'] ?? null;
```

OR

```suggestion
		$active_tab = $this->get_context( Atomic_Tabs::class )['activeTab'] ?? null;
```

maybe? + maybe it will be helpful as collection?

modules/atomic-widgets/elements/atomic-tabs/atomic-tabs.php
```diff
@@ -143,6 +143,14 @@ public function get_script_depends() {
 		return [ 'elementor-tabs-handler' ];
 	}
 
+	protected function define_children_context() {
```
Nevoss:
```suggestion
	protected function define_context() {
```

modules/atomic-widgets/elements/atomic-tabs/atomic-tabs.php
```diff
@@ -143,6 +143,14 @@ public function get_script_depends() {
...
+		$default_active_tab = $this->get_atomic_settings()['default-active-tab'];
```
Nevoss: Does it all the transformation all over again? Maybe it should be cached?

modules/atomic-widgets/elements/atomic-tabs/atomic-tabs.php
```diff
@@ -143,6 +143,14 @@ public function get_script_depends() {
...
+
+		return [
+			'activeTab' => $default_active_tab,
+		];
```
Nevoss, line 151: Each element has only a single context?

modules/atomic-widgets/elements/atomic-element-base.php
```diff
@@ -71,6 +73,10 @@ protected function define_initial_attributes() {
 		return [];
 	}
 
+	protected function define_children_context() {
```
Nevoss:
```suggestion
	protected function define_render_context(): array {
```

modules/atomic-widgets/elements/atomic-element-base.php
```diff
@@ -142,6 +148,46 @@ protected function _get_default_child_type( array $element_data ) {
...
+					$contexts[ $this->get_type() ] = [];
+				}
+				$contexts[ $this->get_type() ][] = $element_context;
+			}
+
+			return $contexts;
+		};
+
+		$this->context_filter_stack[] = $filter;
+		add_filter( 'elementor/atomic/contexts', $filter );
```
Nevoss: Why not use prop on the child element? Or static prop on the parent element?

modules/atomic-widgets/elements/atomic-element-base.php
```diff
@@ -142,6 +148,46 @@ protected function _get_default_child_type( array $element_data ) {
...
+
+		parent::print_content();
+
+		remove_filter( 'elementor/atomic/contexts', array_pop( $this->context_filter_stack ) );
```
Nevoss: Something not clear here

Nevoss (changes requested):

tests/phpunit/elementor/modules/atomic-widgets/test-atomic-tabs.php
```diff
@@ -0,0 +1,225 @@
...
+use Elementor\Modules\AtomicWidgets\Elements\Atomic_Tabs\Atomic_Tabs;
+use Elementor\Modules\AtomicWidgets\Elements\Atomic_Tabs\Atomic_Tab;
+use Elementor\Modules\AtomicWidgets\Elements\Atomic_Tabs\Atomic_Tab_Panel;
+use Elementor\Modules\AtomicWidgets\Elements\Atomic_Tabs\Atomic_Tabs_List;
+use Elementor\Modules\AtomicWidgets\Elements\Atomic_Tabs\Atomic_Tabs_Content;
+use Elementor\Plugin;
+use ElementorEditorTesting\Elementor_Test_Base;
+use Spatie\Snapshots\MatchesSnapshots;
+
+class Test_Atomic_Tabs extends Elementor_Test_Base {
```
Nevoss, line 12: Check https://developer.wordpress.org/reference/classes/wp_html_tag_processor/

modules/atomic-widgets/elements/atomic-tabs/atomic-tab-panel.php
Omerisra6: Will use the class as key, I don't think we should always return a collection, an array is ok.

modules/atomic-widgets/elements/atomic-tabs/atomic-tabs.php
Omerisra6: Will go with define render context

modules/atomic-widgets/elements/atomic-tabs/atomic-tabs.php
Omerisra6: Will create a get_atomic_settings that will trigger only the necessary transformation.

modules/atomic-widgets/elements/atomic-element-base.php
Omerisra6: Here I will create a global `Render_Context` class and move the logic there (the elements will use it directly)

modules/atomic-widgets/elements/atomic-element-base.php
```diff
@@ -142,6 +148,24 @@ protected function _get_default_child_type( array $element_data ) {
 		return Plugin::$instance->widgets_manager->get_widget_types( $element_data['widgetType'] );
 	}
 
+	public function print_content() {
+		$element_context = $this->define_render_context();
+
+		if ( ! empty( $element_context ) ) {
```
Nevoss: Move to variable `has_context` or something like that

modules/atomic-widgets/elements/atomic-tabs/atomic-tabs.php
```diff
@@ -143,6 +143,14 @@ public function get_script_depends() {
 		return [ 'elementor-tabs-handler' ];
 	}
 
+	protected function define_render_context(): array {
+		$default_active_tab = $this->get_atomic_setting( 'default-active-tab' );
+
+		return [
+			'activeTab' => $default_active_tab,
```
Nevoss:
```suggestion
			'default_active_tab' => $default_active_tab,
```

modules/atomic-widgets/elements/atomic-tabs/atomic-tab-panel.php
```diff
@@ -112,6 +112,9 @@ protected function add_render_attributes() {
 		$base_style_class = $this->get_base_styles_dictionary()[ static::BASE_STYLE_KEY ];
 		$initial_attributes = $this->define_initial_attributes();
 
+		$active_tab = $this->get_context( Atomic_Tabs::class )['activeTab'] ?? null;
```
Nevoss:
```suggestion
		$active_tab = Render_Context::get(Atomic_Tabs::class)['activeTab'] ?? null;
```

modules/atomic-widgets/render-context.php
```diff
@@ -0,0 +1,38 @@
...
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+class Render_Context {
+	private static $context_stack = [];
+
+	public static function push( string $key, array $context ): void {
+		if ( ! isset( self::$context_stack[ $key ] ) ) {
```
Nevoss:
```suggestion
		if ( ! self::has[ $key ] ) ) {
```

modules/atomic-widgets/render-context.php
```diff
@@ -0,0 +1,38 @@
...
+			self::$context_stack[ $key ] = [];
+		}
+
+		self::$context_stack[ $key ][] = $context;
+	}
+
+	public static function pop( string $key ): void {
+		if ( isset( self::$context_stack[ $key ] ) && ! empty( self::$context_stack[ $key ] ) ) {
+			array_pop( self::$context_stack[ $key ] );
```
Nevoss, line 22: Try avoid using function that change the reference

modules/atomic-widgets/render-context.php
```diff
@@ -0,0 +1,38 @@
...
+		}
+	}
+
+	public static function get( string $key ): array {
+		if ( ! isset( self::$context_stack[ $key ] ) || empty( self::$context_stack[ $key ] ) ) {
+			return [];
+		}
+
+		return end( self::$context_stack[ $key ] );
```
Nevoss: Same as above

Nevoss (changes requested):

tests/phpunit/elementor/modules/atomic-widgets/test-atomic-tabs.php
Nevoss: Whay about that?

modules/atomic-widgets/elements/atomic-tabs/atomic-tab-panel.php
Omerisra6: Do you think it's fine that the set context comes from the base class and the get will come directly from the Render Context class?

modules/atomic-widgets/render-context.php
Omerisra6: Do you mean get?

tests/phpunit/elementor/modules/atomic-widgets/test-atomic-tabs.php
Omerisra6: will get to it in a different PR (it isn't that intuitive)

Nevoss (approved):

PR #32961, opened 2025-10-05
Internal: Transition Restrictions for Non-Pro Users [ED-21125]

YaelAvitan:
### ✨ PR Description
Purpose: Add transition restrictions for non-Pro users by disabling transitions in non-normal style states and limiting available properties.
Main changes:
- Disabled Add button in transition controls when non-normal style state or all properties used
- Added validation to filter out invalid transitions not present in allowed properties list
- Updated tests to validate transition control restrictions behavior


packages/packages/libs/editor-controls/src/controls/transition-control/transition-repeater-control.tsx
```diff
@@ -118,7 +140,7 @@ export const TransitionRepeaterControl = createControl(
 				childControlConfig={ getChildControlConfig( recentlyUsedList, disabledItems ) }
 				propKey="transition"
 				addItemTooltipProps={ {
-					disabled: ! currentStyleIsNormal,
+					disabled: ! currentStyleIsNormal || allPropertiesUsed,
```
RobiFerentz: Just a nitpick: Can we not make this a single boolean above the return statement? It just looks better in the JSX.

RobiFerentz (dismissed):
tiny nitpick, but approved.

packages/packages/libs/editor-controls/src/controls/transition-control/transition-repeater-control.tsx
```diff
@@ -99,13 +99,36 @@ export const TransitionRepeaterControl = createControl(
 			[]
 		);
 
-		const { value } = useBoundProp( childArrayPropTypeUtil );
+		const { value, setValue } = useBoundProp( childArrayPropTypeUtil );
 		const disabledItems = useMemo( () => getDisabledItems( value ), [ value ] );
 
+		const allowedTransitionValues = useMemo(
+			() => transitionProperties.flatMap( ( category ) => category.properties.map( ( prop ) => prop.value ) ),
```
asafdl: its a small improvement, but this would be more efficient if we created a map (record) from this, so instead of iterating over it constantly we can just ask if prop exists in map or not

packages/packages/libs/editor-controls/src/controls/transition-control/transition-repeater-control.tsx
```diff
@@ -99,13 +99,36 @@ export const TransitionRepeaterControl = createControl(
...
+			[]
+		);
+
+		useEffect( () => {
+			if ( ! value || value.length === 0 ) {
+				return;
+			}
+			const sanitized = value.filter( ( item ) => {
+				const selectionValue =
+					( item?.value?.selection?.value as { value?: { value?: string } } )?.value?.value ?? '';
```
asafdl, line 120: i guess just clean up this code here? do we need the casting? it looks ugly

packages/packages/libs/editor-controls/src/controls/transition-control/transition-repeater-control.tsx
```diff
@@ -99,13 +99,36 @@ export const TransitionRepeaterControl = createControl(
...
+				return allowedTransitionValues.includes( selectionValue );
+			} );
+
+			if ( sanitized.length !== value.length ) {
+				setValue( sanitized );
+			}
+		}, [ value, setValue, allowedTransitionValues ] );
```
asafdl: this effect will recurse, we monitor value, and we setvalue inside it, so it will change value

packages/packages/libs/editor-controls/src/controls/transition-control/transition-repeater-control.tsx
YaelAvitan: okay, used 'set' it is also o(1) and more simple

packages/packages/libs/editor-controls/src/controls/transition-control/transition-repeater-control.tsx
YaelAvitan: It will happen one more time at most, Only if we really needed to drop values
There can't be an infinite loop or anything
I couldn't easily prevent the extra time, is that necessary?

packages/packages/libs/editor-controls/src/controls/transition-control/transition-repeater-control.tsx
YaelAvitan: do need this casting

RobiFerentz (approved):

PR #32964, opened 2025-10-05
Internal: Display admin notice about disabled local google fonts. [ED-21154]

RobiFerentz:


core/admin/admin-notices.php
```diff
@@ -512,6 +513,49 @@ private function notice_send_app_promotion() {
 		return true;
 	}
 
+	private function notice_local_google_fonts_disabled() {
+		$notice_id = 'local_google_fonts_disabled';
```
davseve: I see that you have this string in two places, can we create only one var with that string?

core/admin/admin-notices.php
```diff
@@ -512,6 +513,49 @@ private function notice_send_app_promotion() {
...
+			return false;
+		}
+
+		$is_local_gf_enabled = (bool) get_option( 'elementor_local_google_fonts', '0' );
+
+		if ( $is_local_gf_enabled ) {
+			return false;
+		}
+
+		$title = sprintf( esc_html__( 'Important: Local Google Fonts Settings in Elementor', 'elementor' ) );
```
davseve: This var looks redundant as we use it only once.

core/admin/admin-notices.php
```diff
@@ -512,6 +513,49 @@ private function notice_send_app_promotion() {
...
+
+		if ( User::is_user_notice_viewed( $notice_id ) ) {
+			return false;
+		}
+
+		$is_local_gf_enabled = (bool) get_option( 'elementor_local_google_fonts', '0' );
+
+		if ( $is_local_gf_enabled ) {
+			return false;
+		}
```
davseve: Aren't these conditions the same?

RobiFerentz: no, one checks if a notice has been viewed, one checks if an setting is enabled or not.

core/admin/admin-notices.php
RobiFerentz: Fixed.

core/admin/admin-notices.php
RobiFerentz: Fixed

core/admin/admin-notices.php
```diff
@@ -512,6 +514,46 @@ private function notice_send_app_promotion() {
...
+
+		$options = [
+			'title' => sprintf( esc_html__( 'Important: Local Google Fonts Settings in Elementor', 'elementor' ) ),
```
davseve: Why do we need `sprintf` here?

RobiFerentz: Fixed

davseve (approved):

PR #33054, opened 2025-10-09
Internal: Components Save Endpoint [ED-21032]

Nevoss:
### ✨ PR Description
Purpose: Add API endpoint for bulk component saving to improve performance and reduce network requests when saving multiple components simultaneously.

Main changes:
- Implemented batch saving endpoint that accepts multiple components in a single request
- Added validation service for component collections with name uniqueness and limit checks
- Replaced single component creation with bulk creation in REST API and client
- Updated permission callbacks to require edit_posts capability for component endpoints


packages/packages/core/editor-components/src/utils/before-save.ts
```diff
@@ -45,17 +45,22 @@ async function createComponents(
...
-	const promises = components.map( ( component ) => {
-		return apiClient.create( { name: component.name, content: component.content, status } ).then( ( response ) => {
-			tempIdToComponentId.set( component.id, response.component_id );
-		} );
+	const response = await apiClient.create( {
+		status,
+		items: components.map( ( component ) => ( {
+			temp_id: component.id,
+			title: component.name,
+			elements: component.content,
```
MiryamOren: Since we renamed `content` to `elements`, I suggest changing it in FE too

MiryamOren (dismissed):
LGTM 🤝

packages/packages/core/editor-components/src/utils/before-save.ts
Nevoss: Done

MiryamOren (approved):
🤝

PR #33055, opened 2025-10-09
Fix: Cached CSS files not busted properly [ED-21016]

ronros-elementor:
Set a version for each styles key upon validating its cache (only if invalid initially)
Pass the version to `wp_enqueue_style` responsively

Additional - fixed the `media` value returned from `Style_File`, to not include `@media `, thus only using the (max-width=...) in the `<style />` tag, for its `media` attribute

### ✨ PR Description
Purpose: Fix CSS cache busting by adding version parameter to enqueued style files and properly handling media query strings.
Main changes:
- Added version parameter to wp_enqueue_style using cache validity meta data
- Fixed media query string handling in Style_File::get_media() to properly strip '@media' prefix
- Refactored cache validation timing to maintain consistent cache state
- Enhanced tests to verify correct versioning and media query formatting


modules/atomic-widgets/styles/style-file.php
```diff
@@ -36,6 +36,10 @@ public function get_url(): string {
 	}
 
 	public function get_media(): string {
+		if ( str_starts_with( $this->media, '@media' ) ) {
+			return substr( $this->media, 6 );
+		}
```
Omerisra6, line 41: Why do we need to support both options?

Omerisra6 (approved):

modules/atomic-widgets/styles/style-file.php
ronros-elementor: could just be 'all' (value is set [here](https://github.com/elementor/elementor/pull/33055/files/7019916c7d1f416b66713e38f6b906a0d362e8d0) ) which is a valid string to use

PR #33140, opened 2025-10-19
Fix: Repeaters indication popovers [ED-20064]

nami-p:


packages/packages/core/editor-editing-panel/src/styles-inheritance/init-styles-inheritance-transformers.ts
```diff
@@ -29,12 +31,31 @@ export function initStylesInheritanceTransformers() {
 }
 
 function registerCustomTransformers() {
+	const originalStyleTransformers = styleTransformersRegistry.all();
+
```
elchugreeva: do we need this const 2 times

mike-elementor (approved):

elchugreeva (approved):

PR #33168, opened 2025-10-20
Internal: Update kits library and dashboard flows [ED-21265]

nicoladj77:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #

### ✨ PR Description
Purpose: Update tracking system for kits library and WordPress dashboard to improve session management and event tracking capabilities.

Main changes:
- Removed deprecated KITDEMO events and simplified KIT_IMPORT_STATUS tracking
- Added PluginActions class for tracking core/pro plugin deactivation and deletion events 
- Implemented session storage for persistence across page navigation
- Refactored navigation tracking with simplified menu click handling


hein-obox (dismissed):

.gitignore
ManorHazaz, line 1: Why?

app/assets/js/event-track/apps-event-tracking.js
```diff
@@ -10,8 +10,6 @@ const EVENTS_MAP = {
 	KIT_CLOUD_LIBRARY_DELETE: 'kit_cloud_library_delete',
 	IMPORT_EXPORT_ADMIN_ACTION: 'ie_admin_action',
 	KIT_IMPORT_UPLOAD_FILE: 'kit_import_upload_file',
-	KITDEMO_APPLY_COMPLETED: 'kitdemo_apply_completed',
-	KITDEMO_APPLY_FAILED: 'kitdemo_apply_failed',
```
ManorHazaz, line 14: kitdemo?

app/assets/js/event-track/dashboard/plugin-actions.js
```diff
@@ -0,0 +1,110 @@
...
+				}
+
+				const pluginAction = this.getPluginAction( link, href );
+				if ( ! pluginAction ) {
+					return;
+				}
+
+				this.trackControl( link, pluginAction );
+			},
+			{ capture: true },
```
ManorHazaz: What? Why are we guessing all of it instead just adding the event at the right place?

app/assets/js/event-track/dashboard/plugin-actions.js
ManorHazaz, line 1: All of this file feels weird

app/assets/js/event-track/wp-dashboard-tracking.js
```diff
@@ -42,27 +43,73 @@ export default class WpDashboardTracking {
...
+			this.sessionStartTime = storedSession.sessionStartTime;
+			this.navItemsVisited = new Set( storedSession.navItemsVisited );
+			this.lastActivityTime = Date.now();
+			this.sessionEnded = false;
+		} else {
+			this.sessionStartTime = Date.now();
+			this.lastActivityTime = Date.now();
+			this.sessionEnded = false;
+			this.navItemsVisited = new Set();
+		}
```
ManorHazaz, line 75: Not sure about this session storage part..

ManorHazaz (changes requested):

app/assets/js/event-track/wp-dashboard-tracking.js
nicoladj77: @ManorHazaz Yotam asked me to track user sessions through multiple page loads, as long as the user is on an Elementor page. So I store data in sessionStorage

app/assets/js/event-track/apps-event-tracking.js
nicoladj77: @ManorHazaz this was the old implementation which duplicated what was already existing. After discussing the duplication Yotam asked me to remove it.

app/assets/js/event-track/dashboard/plugin-actions.js
nicoladj77: @ManorHazaz this is the last thing Yotam asked me to track plugin Uninstall.

app/assets/js/event-track/dashboard/plugin-actions.js
nicoladj77: @ManorHazaz For this Dashboard tracking I prefer delegated events tracking because otherwise you have   too many listene, you don't lose anything AJAX and you can track React apps without hardcoding things into React. So this is how I implemented everything basically, if you prefer I can switch approach here ( but if Yotam wants more complex tracking I would still keep it delegated ), but for other parts I think it makes a lot more sense.

app/assets/js/event-track/dashboard/plugin-actions.js
nicoladj77: I have kept it in a separate file as I have one file for each tracking type

.gitignore
nicoladj77: @ManorHazaz because it happend to me that I committed it after creating a build for the patchstack stuff, and I had to remove it from the history of git, which is annoying, so I thought it made sense to ignore it for the future. Can remove it of course.

nicoladj77: or maybe ignore all zip files in the root folder, because in the past this has happened to me more than once

app/assets/js/event-track/dashboard/plugin-actions.js
```diff
@@ -0,0 +1,191 @@
...
+	ELEMENTOR_PRO: 'pro',
+};
+
+class PluginActions extends BaseTracking {
+	static selectedReason = null;
+
+	static init() {
+		this.attachCoreDeactivationTracking();
+		this.attachProDeactivationTracking();
+		this.attachProDeletionTracking();
```
ManorHazaz, line 15: Mmm, usually we don't want the Core to know Pro..
Do we have other options?

ManorHazaz (changes requested):

app/assets/js/event-track/dashboard/plugin-actions.js
nicoladj77: @ManorHazaz We could techincally have pro deactivation detection in pro, but of course not deletion, because core might be active when pro gets deleted. In general you cannot have pro active and core inactive, so it is basically impossible to be in a situation where pro gets deactivated by a user and core is not active. And if that was the case, Pro could not track as all tracking logic is in core and it is not duplicated.
I think that while functionally it make sense that core doesn't know pro, from a tracking perspective, given how small the scope is, this is ok. 
One possibility would be to generalize this to "other plugins deactivation tracking" and use a filter to get the slugs that should be tracked, because maybe you would also want to track hello-plus or something else, but it looks a bit overkill

ManorHazaz: Agree

ManorHazaz (dismissed):

ManorHazaz (approved):

PR #33183, opened 2025-10-21
Internal: Dynamic Tags Contact URL > Google Calendar [ED-20606]

YaelAvitan:


packages/packages/libs/editor-props/src/prop-types/date-time.ts
```diff
@@ -0,0 +1,14 @@
+import { z } from '@elementor/schema';
+
+import { createPropUtils } from '../utils/create-prop-utils';
+import { unknownChildrenSchema } from './utils';
+
+export const DateTimePropTypeUtil = createPropUtils(
+	'date-time',
+	z.strictObject( {
+		date: unknownChildrenSchema,
```
RobiFerentz, line 9: Can't this be a string prop type util here?

YaelAvitan: I see everywhere using with the `unknownChildrenSchema`
`stringPropTypeUtil`throws some errors

RobiFerentz (dismissed):

packages/packages/core/editor-canvas/src/transformers/settings/date-time-transformer.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const dateTimeTransformer = createTransformer( ( values: { date?: string; time?: string }[] ) => {
+	return values
+		.map( ( value ) => {
+			const date = ( value.date || '' ).trim();
+			const time = ( value.time || '' ).trim();
+			if ( ! date && ! time ) {
+				return '';
+			}
+			return `${ date } ${ time }`.trim();
```
ronenboxer:
```suggestion
			const time = ( value.time || '' ).trim();

			return ! date && ! time ? '' : `${ date } ${ time }`.trim();
```

packages/packages/libs/editor-controls/src/controls/date-time-control.tsx
```diff
@@ -0,0 +1,106 @@
...
+export const DateTimeControl = createControl( ( { inputDisabled }: { inputDisabled?: boolean } ) => {
+	const { value, setValue, ...propContext } = useBoundProp( DateTimePropTypeUtil );
+
+	const handleChange = ( newValue: Props, meta: { bind: 'date' | 'time' } ) => {
+		const field = meta.bind;
+		const fieldValue = newValue[ field as 'date' | 'time' ];
+
+		if ( isTransformable( fieldValue ) ) {
+			setValue( { ...value, [ field ]: fieldValue } );
+			return;
```
ronenboxer: I would personally either add an empty line or just return at the same line as the set value `return setValue...`

packages/packages/libs/editor-controls/src/controls/date-time-control.tsx
```diff
@@ -0,0 +1,106 @@
...
+		}
+
+		let formattedValue = '';
+
+		if ( fieldValue ) {
+			const dayjsValue = fieldValue as Dayjs;
+			formattedValue = field === 'date' ? dayjsValue.format( 'YYYY-MM-DD' ) : dayjsValue.format( 'HH:mm' );
```
ronenboxer: Please create a `const DATE_FORMAT = 'YYYY-MM-DD';` and `const TIME_FORMAT = 'HH:mm';` at the top of the file

packages/packages/libs/editor-controls/src/controls/date-time-control.tsx
```diff
@@ -0,0 +1,106 @@
...
+		}
+
+		setValue( {
+			...value,
+			[ field ]: {
+				$$type: 'string',
```
ronenboxer, line 36: I might be mistaken, but since you created a prop type that you use as an argument in `useBoundProp`, the `$$type` isnt necessary and you should just use the value, no?
If not, shouldn't have the `$$type` been `date-time`?

packages/packages/libs/editor-controls/src/controls/date-time-control.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+	const parseDateValue = ( dateStr?: string | null ): Dayjs | null => {
+		if ( ! dateStr ) {
+			return null;
+		}
+
+		const d = ( dayjs as unknown as { default: ( s?: string | number | Date ) => Dayjs } ).default( dateStr );
+
+		return d && typeof d.isValid === 'function' && d.isValid() ? d : null;
+	};
```
ronenboxer, line 50: are the type casting necessary?

packages/packages/libs/editor-controls/src/controls/date-time-control.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+		const base = ( dayjs as unknown as { default: () => Dayjs } ).default();
+		return base.hour( h ).minute( m ).second( 0 ).millisecond( 0 );
+	};
+
+	return (
+		<PropProvider { ...propContext } value={ value } setValue={ setValue }>
+			<ControlActions>
+				<LocalizationProvider>
+					<div style={ { display: 'flex', gap: '8px', alignItems: 'center' } }>
```
ronenboxer: Maybe use Box with flex and `gap: 1`?

packages/packages/libs/editor-controls/src/controls/date-time-control.tsx
YaelAvitan: yes its necessary

packages/packages/libs/editor-controls/src/controls/date-time-control.tsx
YaelAvitan: date-time includes date and time
here we set date or time 
so we need to write explicitly and the type is string

ronenboxer (approved):

PR #33189, opened 2025-10-21
Internal: Save editable value on blur [ED-21327]

IshayMaya:
### ✨ PR Description
Purpose: Update editable field component to save values on blur instead of Enter keypress to improve user experience and match standard input behavior.
Main changes:
- Modified useEditable hook to trigger submission on blur instead of Enter keypress
- Updated tests to use fireEvent.blur instead of fireEvent.keyDown
- Added component-based tests replacing renderHook implementation for better test coverage


packages/packages/libs/editor-ui/src/hooks/use-editable.ts
```diff
@@ -62,7 +61,10 @@ export const useEditable = ( { value, onSubmit, validation, onClick, onError }:
 
 		if ( [ 'Enter' ].includes( event.key ) ) {
 			event.preventDefault();
-			return submit( ( event.target as HTMLElement ).innerText );
+			// trigger the blur event to submit the value to avoid submitting twice.
```
ronros-elementor: Maybe -

```suggestion
			// submission is invoked only on blur, to avoid issues with double-submission in certain cases
```

(nothing dramatic, just seems a bit clearer to anyone who'd stumble upon it in the future without any context)

ronros-elementor (dismissed):

ronstovsky (approved):

ronros-elementor (approved):

PR #33240, opened 2025-10-27
Internal: Render display conditions for atomic widgets. [ED-21370]

RobiFerentz:


modules/atomic-widgets/controls/types/display-conditions-control.php
```diff
@@ -0,0 +1,19 @@
...
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+class Display_Conditions_Control extends Atomic_Control_Base {
+	public function get_type(): string {
+		return 'display-conditions';
+	}
+
+
```
davseve:
```suggestion

```

modules/atomic-widgets/elements/has-atomic-base.php
```diff
@@ -12,7 +12,9 @@
 use Elementor\Modules\AtomicWidgets\Parsers\Props_Parser;
 use Elementor\Modules\AtomicWidgets\Parsers\Style_Parser;
 use Elementor\Modules\AtomicWidgets\PropTypes\Primitives\String_Prop_Type;
+use Elementor\Modules\AtomicWidgets\PropTypes\Display_Conditions_Prop_Type;
 use Elementor\Utils;
+use Stringable;
```
davseve: Where do we use this?

modules/atomic-widgets/controls/types/display-conditions-control.php
asafdl: the display control should be in pro repo, unless theres a good reason its not, no?

modules/atomic-widgets/elements/has-atomic-base.php
RobiFerentz: we don't, just weird autocomplete error, I think. I'll remove it

modules/atomic-widgets/controls/types/display-conditions-control.php
RobiFerentz: No, this is the control to show when there is no pro installed or active, it will be replaced by pro with a real one when that is done.

RobiFerentz: Fixed the extra line

modules/atomic-widgets/elements/has-atomic-base.php
RobiFerentz: Removed.

packages/packages/core/editor-editing-panel/src/components/settings-tab.tsx
```diff
@@ -85,6 +78,35 @@ const Control = ( { control: { value, type } }: { control: Control | ElementCont
 	);
 };
 
+const SectionControls = ( { items, element }: { items: ControlItem[]; element: Element } ) => {
+	const dcControlIndex = React.useMemo(
+		() => items.findIndex( ( item ) => ( item.value as Control[ 'value' ] )?.bind === 'display-conditions' ),
+		[ items ]
+	);
+
+	const sortedItems = React.useMemo( () => {
```
murzin-nikita, line 87: Controls sorting looks a bit fragile/hard to understand. 

I think it would be better to create some kinda of slots for items, to render them in specific order. And if there are no slot for item - just render it in the end. WDYT?

PR #33281, opened 2025-10-29
Internal: Block editing interactions for components children [ED-21330]

Nevoss:
### ✨ PR Description
Purpose: Block editing interactions for component children to prevent unintended modifications in nested elements.
Main changes:
- Updated Container.isEditable() to check for inactive status in element's editSettings
- Modified component rendering to recursively set inactive flag on all nested elements
- Added interaction blocking in BaseElementView for non-editable containers
- Refactored component template rendering process to properly handle async operations


packages/packages/core/editor-canvas/src/legacy/create-templated-element-type.ts
```diff
@@ -91,14 +79,24 @@ export function createTemplatedElementView( {
...
 			this.#abortController = new AbortController();
 
 			const process = signalizedProcess( this.#abortController.signal )
+				.then( () => this.#beforeRender() )
+				.then( () => this._renderTemplate() )
+				.then( () => this._renderChildren() ) // This will render children and should be async as well.
+				.then( () => this.#afterRender() );
+
+			return process.execute();
+		}
```
MiryamOren, line 97: I know the full explanation is too long for a comment, but maybe just add something like
```
// We override the render function to support async _renderTemplate
// Note that `_renderChildren` asynchronity is still NOT supported, so only the parent element rendering can be async
```

Maybe we should also remove the separate `then` so it will be clear
```suggestion
		render() {
			this.#abortController?.abort();
			this.#abortController = new AbortController();

			const process = signalizedProcess( this.#abortController.signal )
				.then( () => this.#beforeRender() )
				.then( () => this._renderTemplate() )
				.then( () => {
				    this._renderChildren();
				    this.#afterRender();
				 });

			return process.execute();
		}
```

packages/packages/core/editor-canvas/src/legacy/create-templated-element-type.ts
```diff
@@ -91,14 +79,24 @@ export function createTemplatedElementView( {
...
+
+		// Overriding Marionette original render method to inject our renderer.
```
MiryamOren:
```suggestion
		// Overriding Marionette original _renderTemplate method to inject our renderer.
```

packages/packages/core/editor-canvas/src/legacy/create-templated-element-type.ts
Nevoss: Done

packages/packages/core/editor-canvas/src/legacy/create-templated-element-type.ts
Nevoss: Done

MiryamOren (approved):

PR #33295, opened 2025-10-30
Internal(Components): Bug with components lock mechanism [ED-21559]

MuliElementor:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #

### ✨ PR Description
Purpose: Fix component lock mechanism to prevent concurrent editing of the same component by multiple users.
Main changes:
- Renamed Lock_Component_Manager to Component_Lock_Manager for consistent naming
- Added validation to ensure operations only run on component post types
- Implemented get_updated_status method to correctly handle expired locks
- Added comprehensive unit tests for the lock manager functionality


modules/components/document-lock-manager.php
```diff
@@ -18,6 +18,7 @@ class Document_Lock_Manager {
 	private $lock_duration;
 	private const LOCK_USER_META = '_lock_user';
 	private const LOCK_TIME_META = '_lock_time';
+	private const LOCK_EDIT_LOCK_META = '_edit_lock';
```
ronros-elementor, line 21: do you still need that?

modules/components/document-lock-manager.php
```diff
@@ -86,13 +87,14 @@ public function lock( $document_id ) {
...
-		// Remove WordPress post lock
-		delete_post_meta( $document_id, '_edit_lock' );
 
 		// Remove custom document lock metadata
 		delete_post_meta( $document_id, self::LOCK_USER_META );
 		delete_post_meta( $document_id, self::LOCK_TIME_META );
 
+		// Remove edit lock metadata
+		delete_post_meta( $document_id, self::LOCK_EDIT_LOCK_META );
+
```
ronros-elementor, line 97: same

tests/phpunit/elementor/modules/components/test-component-lock-manager.php
```diff
@@ -0,0 +1,340 @@
...
+		// Register post type for components
+		register_post_type( Component_Document::TYPE, [
+			'label' => Component_Document::get_title(),
+			'labels' => Component_Document::get_labels(),
+			'public' => false,
+			'supports' => Component_Document::get_supported_features(),
+		] );
+
+		// Create test documents of different types
+		$this->test_document_ids = $this->create_test_documents();
```
ronros-elementor: this naming is a bit strange, because it doesn't hold on document ids, one is of a document, the other is for component.

I'd suggest splitting the two, so that the create test document would just create a document, just like `create_test_component` only creates a component :)

```suggestion
		$this->test_document_id = $this->create_test_document();
		$this->test_component_id = $this->create_test_component();
```

tests/phpunit/elementor/modules/components/test-component-lock-manager.php
ronros-elementor, line 1: I think there's another thing that this test reveals - the components lock manager doesn't seem to care about the type of the post, so you could run it also on documents, which is wrong.
I assume there is a reason to this inheritance, even if theoretical - you have a documents lock manager, and a component one, so I'd assume you shouldn't use it for documents...
so you should check for the post type and handle it properly (we wouldn't want to lock a document with a component's constraints, nor vice-versa)

also, once changed - there's a lot of variable names and assertion descriptions that refer to documents, where it should instead refer to components.

+ we'll need a test to show it would NOT lock a document, using the components lock manager

ronros-elementor (changes requested):

modules/components/document-lock-manager.php
MuliElementor: yes

modules/components/document-lock-manager.php
MuliElementor: yes

modules/components/component-lock-manager.php
```diff
@@ -61,4 +65,38 @@ public function heartbeat_received( $response, $data ) {
...
+		return $lock_data;
+	}
+
+	public function lock( $document_id ) {
+		if ( ! $this->is_component_post( $document_id ) ) {
+			throw new \Exception( 'Post is not a component type' );
+		}
+
+		return parent::lock( $document_id );
+	}
```
ronros-elementor, line 101: If it doesn't add any functionality that is relevant to components in particular - why do we need to override it? Even if theoretically it won't be an id of a component, but of a document - it would anyway go to the document-lock-manager (i.e. `parent::lock(...)`)

modules/components/components-rest-api.php
```diff
@@ -243,12 +251,26 @@ private function get_lock_status( \WP_REST_Request $request ) {
...
 
+	private function is_valid_component( $post_id ) {
+		$post = get_post( $post_id );
+
+		if ( ! $post ) {
+			return false;
+		}
+
+		return get_post_type( $post_id ) === Component_Document::TYPE;
+	}
```
ronros-elementor: Why do we need to validate if a component, if anyway we have this validation in the lock manager on every function call? I'd say - either remove it from there, or from here... Also I think that by only using get_post_type - we no longer need to verify the post even exists (but please make sure I'm right... but anyway in the lock manager's implementation - the check is only against the post type)

ronros-elementor (changes requested):

ronros-elementor:
Also - the corresponding ticket marked as "Done", so I'll open a ticket and change the PR title to reflect that

modules/components/component-lock-manager.php
MuliElementor: ill keep this one
remove from the api

modules/components/components-rest-api.php
MuliElementor: removed

tests/phpunit/elementor/modules/components/test-components-rest-api.php
```diff
@@ -795,22 +795,6 @@ public function test_get_lock_status__returns_allowed_when_locked_by_current_use
 		$this->assertNotNull( $data['locked_by'], 'Should show who locked the component' );
 	}
 
-	public function test_get_lock_status__fails_when_component_not_found() {
```
ronros-elementor: I don't think this test should be removed,  I think that since we removed the original validation from the REST API, which used `Response_Builder` to return an error - we should now handle the invalid exception thrown  from the lock manager similarly, but with a try/catch to then pass forward the error message as catched, to that same Response_Builder

ronros-elementor (approved):

PR #33400, opened 2025-11-10
Internal: Add component uid meta [ED-21541]

MiryamOren:
### ✨ PR Description
Purpose: Add component unique identifier metadata to enable stable component references and better tracking across the system.
Main changes:
- Added component_uid meta field for component documents with unique identifier generation
- Updated component creation flow to use UIDs instead of temporary IDs
- Modified REST API to support UID-based component identification
- Enhanced validation for preventing UID duplication


modules/components/save-components-validator.php
```diff
@@ -51,10 +51,11 @@ private function validate_count( Collection $data ): array {
 		return [];
 	}
 
-	private function validate_duplicated_names( Collection $data ): array {
+	private function validate_duplicated_values( Collection $data ): array {
```
ronros-elementor, line 54: I wonder, as now we check for two things, and this function is expected to return an array - would it make sense to collect the error messages, and return all at the end of the function, instead of the early returns approach? Or is it intentional to provide an array of only 1 or 0 items?

packages/packages/core/editor-components/src/component-id-transformer.ts
```diff
@@ -4,15 +4,23 @@ import { __getState as getState } from '@elementor/store';
...
+type PublishedComponentId = number;
+type UnpublishedComponentId = string;
+
+type ComponentId = PublishedComponentId | UnpublishedComponentId;
 
-	const unpublishedComponent = unpublishedComponents.find( ( component ) => component.id === id );
+export const componentIdTransformer = createTransformer( async ( id: ComponentId ) => {
+	const unpublishedComponents = selectUnpublishedComponents( getState() );
 
+	const unpublishedComponent = unpublishedComponents.find( ( component ) => component.uid === id );
```
ronros-elementor: While at it, maybe this would be nicer -
```suggestion
	const unpublishedComponent = unpublishedComponents.find( ( { uid } ) => uid === id );
```
?

modules/components/module.php
```diff
@@ -48,25 +50,26 @@ public function get_widgets() {
...
-			Component_Document::TYPE,
-			Component_Document::get_class_full_name()
-		);
-
+	private function register_component_post_type() {
 		register_post_type( Component_Document::TYPE, [
 			'label'    => Component_Document::get_title(),
 			'labels'   => Component_Document::get_labels(),
-			'public'   => false,
+			'public'   => true,
```
ronros-elementor: nope, I think.. right? 😄

modules/components/save-components-validator.php
```diff
@@ -53,16 +53,20 @@ private function validate_count( Collection $data ): array {
...
-			->map( function ( $component ) {
+			->map( function ( $component ) use ( $data ) {
+				$errors = [];
+
 				$title = $component['title'];
 				$uid = $component['uid'];
 
 				$is_name_exists = $this->components->some(
 					fn ( $component ) => $component['name'] === $title
+				) || $data->some(
```
ronros-elementor: wouldn't it mean it will always find a duplicate, as it searches for duplications of $data with $this->components, but if it would search against itself - it would always find duplication errors, wouldn't it?

ronros-elementor (changes requested):

modules/components/module.php
MiryamOren: 🙈

modules/components/save-components-validator.php
MiryamOren: You're right, fixed

ronros-elementor (approved):

PR #33476, opened 2025-11-16
Internal: Link / Button accessibility POC [ED-21599]

davseve:


modules/atomic-widgets/elements/has-action-link-support.php
```diff
@@ -0,0 +1,59 @@
...
+		$dynamic_tag_settings = $destination['value']['settings'] ?? [];
+
+		// We can use this approach for all action based tags, the data comes from the tag definition in pro.
+		// The issue is that contact-url dynamic tag is action based, so we need to change its group.
+		// if ( isset( $dynamic_tag_settings['action'] ) ) {
+		// 	return true;
+		// }
+
+		// This is the list approach, we can use one or the other.
+		// This way we wont need to change the contact-url dynamic tag group.
```
davseve: As said in the comments we cam commit this with two approaches. 
The tags are [defined by groups in the tag class ](https://github.com/elementor/elementor-pro/blob/main/modules/dynamic-tags/tags/contact-url.php#L23)
We need to investigate if changing this group value may cause issues.

modules/atomic-widgets/elements/has-action-link-support.php
```diff
@@ -0,0 +1,71 @@
...
+			return false;
+		}
+
+		$dynamic_tag_name = $destination['value']['name'] ?? '';
+
+		if ( ! $dynamic_tag_name ) {
+			return false;
+		}
+
+		$tag_info = Plugin::$instance->dynamic_tags->get_tag_info( $dynamic_tag_name );
```
davseve: Not sure about this as this is a v3 dynamic tags info instead of `Elementor\Modules\AtomicWidgets\DynamicTags\Dynamic_Tags_Module`

packages/packages/core/editor-canvas/src/legacy/create-templated-element-type.ts
```diff
@@ -136,6 +138,30 @@ export function createTemplatedElementView( {
...
+				return false;
+			}
+
+			const dynamicTagName = destination?.value?.name || '';
+
+			const legacyWindow = window as any;
+			const atomicDynamicTags =
+				legacyWindow.elementorCommon?.config?.atomicDynamicTags?.tags ||
+				legacyWindow.elementor?.config?.atomicDynamicTags?.tags ||
+				legacyWindow.elementorCommon?.config?.settings?.atomicDynamicTags?.tags ||
```
davseve, line 156: Didn't really liked this implementation but this is the only way we found in which we can receive all the registered dynamic tags with their groups.

modules/atomic-widgets/dynamic-tags/dynamic-tags-editor-config.php
```diff
@@ -108,6 +108,16 @@ private function convert_dynamic_tag_to_atomic( $tag ) {
 		return $converted_tag;
 	}
 
+	private function set_atomic_group( $tag ): string {
```
ronenboxer, line 111:
```suggestion
	private function get_atomic_group( $tag ): string {
```

modules/atomic-widgets/elements/has-action-link-support.php
```diff
@@ -0,0 +1,69 @@
...
+
+use Elementor\Plugin;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+trait Has_Action_Link_Support {
+
+	protected function has_action_in_link(): bool {
```
ronenboxer, line 13: maybe is_link_action_dynamic_tag

modules/atomic-widgets/elements/has-action-link-support.php
ronenboxer: Put this in `Dynamic_Tags_Module` under `Atomic_Widgets` module


	```public function get_v4_tag_info ( $tag ) {
		return $this->schemas[ $tag ];
	}```

Might work

packages/packages/core/editor-canvas/src/legacy/create-templated-element-type.ts
```diff
@@ -112,17 +112,19 @@ export function createTemplatedElementView( {
...
-					return renderer.render( templateKey, context );
-				} )
+			.then( async ( settings ) => {
+				// Same as the Backend.
+				const context = {
+					id: this.model.get( 'id' ),
+					type,
+					settings,
+					base_styles: baseStylesDictionary,
+					interactions: [],
```
ronenboxer, line 122: oopsie

packages/packages/core/editor-canvas/src/legacy/create-templated-element-type.ts
ronenboxer: `elementor.config.atomicDynamicTags.tags` this should suffice

PR #33542, opened 2025-11-19
Internal(Components): Modal is not shown when user tries to edit a locked component [ED-21738]

MuliElementor:
- Updated `unlock` method in `Document_Lock_Manager` to include error handling and ensure proper removal of WordPress post locks.
- Improved `is_locked` method to check both custom and WordPress locks, updating lock metadata accordingly.
- Refactored `extend_lock` method to accept optional lock data, reducing redundant calls.
- Enhanced `Component_Lock_Manager` to allow unlocking by users who own the WordPress lock.
- Updated unit tests to reflect changes in locking behavior and ensure correct functionality when locks expire.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #

### ✨ PR Description
Purpose: Fix component locking mechanism to properly display warning modal when users attempt to edit locked components.
Main changes:
- Refactored lock management APIs to use consistent data structure with locked_by/locked_at fields
- Improved expired lock detection with separate is_lock_expired method
- Fixed get_lock_status REST endpoint to correctly identify lock ownership and permissions

[ED-21738]: https://elementor.atlassian.net/browse/ED-21738?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/components/component-lock-manager.php
```diff
@@ -34,16 +34,28 @@ public function unlock( $post_id ) {
 		$lock_data = $this->is_locked( $post_id );
 		$current_user_id = get_current_user_id();
 
-		// Only allow unlocking if the current user owns the lock
-		if ( ! $lock_data['is_locked'] || (int) $lock_data['lock_user'] !== (int) $current_user_id ) {
+		if ( ! $lock_data['is_locked'] ) {
+			return parent::unlock( $post_id );
```
ronros-elementor: why do you need to unlock if it's not locked in the first place?

modules/components/document-lock-manager.php
```diff
@@ -113,12 +126,31 @@ public function is_locked( $document_id ) {
...
+					$wp_edit_lock = get_post_meta( $document_id, self::LOCK_EDIT_LOCK_META, true );
+					if ( $wp_edit_lock ) {
+						$lock_parts = explode( ':', $wp_edit_lock );
+						if ( isset( $lock_parts[1] ) ) {
+							$lock_time = (int) $lock_parts[1];
+							update_post_meta( $document_id, self::LOCK_TIME_META, $lock_time );
+						}
+					}
+				}
+			}
```
ronros-elementor:
```suggestion
			return [
				'is_locked'  => true,
				'lock_user'  => $lock_user ? (int) $lock_user : null,
				'lock_time'  => $lock_time ? (int) $lock_time : null,
			];
		}
		
		$wp_lock_user = $this->get_wp_lock_user( $document_id );
		
		if ( $wp_lock_user ) {
			$is_locked = true;
			if ( ! $lock_user ) {
				$lock_user = $wp_lock_user;
				update_post_meta( $document_id, self::LOCK_USER_META, $wp_lock_user );
			}
			
			if ( ! $lock_time ) {
				$wp_edit_lock = get_post_meta( $document_id, self::LOCK_EDIT_LOCK_META, true );
				if ( $wp_edit_lock ) {
					$lock_parts = explode( ':', $wp_edit_lock );
					if ( isset( $lock_parts[1] ) ) {
						$lock_time = (int) $lock_parts[1];
						update_post_meta( $document_id, self::LOCK_TIME_META, $lock_time );
					}
				}
			}
```

modules/components/document-lock-manager.php
```diff
@@ -87,21 +87,34 @@ public function lock( $document_id ) {
...
+			// Remove WordPress post lock
+			if ( ! function_exists( 'wp_clear_post_lock' ) ) {
+				require_once ABSPATH . 'wp-admin/includes/post.php';
+			}
 
-		// Remove edit lock metadata
-		delete_post_meta( $document_id, self::LOCK_EDIT_LOCK_META );
+			if ( function_exists( 'wp_clear_post_lock' ) ) {
+				call_user_func( 'wp_clear_post_lock', $document_id );
+			} else {
```
ronros-elementor: did you check that it's even a thing? we never used in in our codebase, and I couldn't find any wp (or any other) documentation of this function - `wp_clear_post_lock`...

modules/components/document-lock-manager.php
```diff
@@ -87,21 +87,34 @@ public function lock( $document_id ) {
...
 	}
 
 	/**
 	 * Check if a document is currently locked.
 	 *
 	 * Also handles automatic lock expiration based on lock duration.
+	 * Checks both custom lock metadata and WordPress's built-in post lock.
 	 *
 	 * @param int $document_id The document ID to check
 	 * @return array|false Lock data array if locked, false if not locked
```
ronros-elementor: side note - could you consider a method name change? it shouldn't really be is_ something, as it doesn't return a boolean value... I'd suggest renaming it to get_lock_data (as it always being assigned to `$lock_data` anyway, so goes with that flow)

especially as I see you placed back some data-mutating logic in this function (it now updates the post meta)

ronros-elementor (changes requested):

modules/components/component-lock-manager.php
MuliElementor: when user enter "edit-mode" , after an hour the lock released, but the user still in "edit-mode", when he exit ->
it will unlock when it was already unlocked

modules/components/component-lock-manager.php
```diff
@@ -9,7 +9,7 @@
 }
 
 class Component_Lock_Manager extends Document_Lock_Manager {
-	const ONE_HOUR = 60 * 60;
+	const ONE_HOUR = 10;
```
ronros-elementor: I don't think this is the right value

modules/components/component-lock-manager.php
```diff
@@ -57,46 +37,97 @@ public function heartbeat_received( $response, $data ) {
...
+	}
+
+	/**
+	 * Lock a component.
+	 *
+	 * @param int $post_id The component ID to lock
+	 * @return bool|null True if lock was successful, null if locked by another user, false otherwise
+	 * @throws \Exception If post is not a component type.
+	 */
+	public function lock_component( $post_id ) {
```
ronros-elementor: I still think the `get_lock_data` should not invoke `lock`, this is why you needed to come up with this new `lock_component`, which makes the code quite messy and non-OOP in nature

`get_lock_data` should simply provide the data in an expected structure, to handle default empty states mainly, not to also modify the lock data along the way

the consumer of `get_lock_data` should use this data externally to invoke the `unlock` in the context of components, probably the components REST API should orchestrate this logic, or alternatively, as suggested before - add a method to get a "mutated" lock data, something like `get_updated_lock_data` which we would expect to also trigger the `unlock` method (but `unlock` would use the standard `get_lock_data` which doesn't trigger any other methods)

ronros-elementor (changes requested):

modules/components/document-lock-manager.php
```diff
@@ -98,28 +102,39 @@ public function get_lock_data( $document_id ) {
...
+	}
+
 	/**
 	 * Extend the lock for a document.
 	 *
 	 * @param int $document_id The document ID
 	 * @return bool True if extended successfully, false if not locked or locked by another user
 	 */
 	public function extend_lock( $document_id ) {
+		if ( $this->is_lock_expired( $document_id ) ) {
```
ronros-elementor: if it's expired, why do we want to unlock it? wouldn't we instead prefer to actually *lock* it with the next expiration timestamp? why do we care its lock is expired at this point? (assuming it's expired, and not locked by another user)

modules/components/components-rest-api.php
```diff
@@ -302,7 +302,12 @@ private function unlock_component( \WP_REST_Request $request ) {
 	private function get_lock_status( \WP_REST_Request $request ) {
 		$component_id = (int) $request->get_param( 'componentId' );
 		try {
-			$lock_status = $this->get_component_lock_manager()->get_lock_data( $component_id );
+			$lock_manager = $this->get_component_lock_manager();
+			if ( $lock_manager->is_lock_expired( $component_id ) ) {
+				$lock_manager->unlock( $component_id );
+			}
+
+			$lock_status = $lock_manager->get_lock_data( $component_id );
```
ronros-elementor:
```suggestion
			$lock_data = $lock_manager->get_lock_data( $component_id );
```

tests/phpunit/elementor/modules/components/test-component-lock-manager.php
```diff
@@ -268,7 +321,7 @@ public function test_get_updated_status__fails_for_regular_post() {
 		// Act & Assert
 		$this->expectException( \Exception::class );
 		$this->expectExceptionMessage( 'Post is not a component type' );
-		$this->lock_manager->get_updated_status( $post_id );
+		$this->lock_manager->get_lock_data( $post_id );
```
ronros-elementor, line 324: not sure we need it here

ronros-elementor (changes requested):

modules/components/component-lock-manager.php
```diff
@@ -26,77 +26,117 @@ public function register_hooks() {
...
-		if ( ! $lock_data['is_locked'] ) {
-			if ( null !== $lock_data['lock_time'] ) {
-				parent::unlock( $post_id );
-			}
-			return [
-				'is_locked' => false,
-				'lock_user' => null,
-				'lock_time' => null,
-			];
+		if ( $this->is_lock_expired( $post_id ) ) {
```
ronros-elementor: also not sure why do we unlock here (similar to the extend_lock on the document lock manager)

modules/components/document-lock-manager.php
MuliElementor: removed

modules/components/component-lock-manager.php
MuliElementor: revmoved

ronros-elementor (approved):

PR #33543, opened 2025-11-19
Internal: Get nested children by parent [ED-21742]

IshayMaya:
### ✨ PR Description
```
Purpose: Refactor element children retrieval to support nested parent-child relationships in the editor by filtering elements based on parent type.

Main changes:
- Changed useElementChildren API to accept a Record<parentType, childType> map instead of a string array
- Implemented hierarchical element filtering based on parent container types
- Fixed tabs control to use the new element children retrieval pattern
- Added test cases for nested element hierarchy navigation
```

### DIFF
--- a/packages/packages/libs/editor-elements/src/hooks/__tests__/use-element-children.test.ts
+++ b/packages/packages/libs/editor-elements/src/hooks/__tests__/use-element-children.test.ts
@@ -18,7 +18,9 @@ describe( 'useElementChildren', () => {
 		mockGetContainer.mockReturnValue( null );
 
 		// Act.
-		const { result } = renderHook( () => useElementChildren( 'element-1', [ 'tab', 'accordion' ] ) );
+		const { result } = renderHook( () =>
+			useElementChildren( 'element-1', { tabs: 'tab', accordions: 'accordion' } )
+		);
 
 		// Assert.
 		expect( result.current ).toEqual< ElementChildren >( {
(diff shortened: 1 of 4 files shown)


packages/packages/libs/editor-elements/src/hooks/use-element-children.ts
```diff
@@ -24,20 +24,20 @@ export function useElementChildren< T extends ElementChildren >(
 		() => {
 			const container = getContainer( elementId );
 
-			const elementChildren = childrenTypes.reduce( ( acc, type ) => {
-				acc[ type ] = [];
+			const elementChildren = Object.entries( childrenTypes ).reduce( ( acc, [ parentType, childType ] ) => {
+				const parent = container?.children?.findRecursive?.(
+					( { model } ) => model.get( 'elType' ) === parentType
```
Omerisra6, line 29: What if I have tabs element inside my tabs-menu ?

I think it would also include it

IshayMaya: find will get the first one

packages/packages/libs/editor-elements/src/hooks/__tests__/use-element-children.test.ts
```diff
@@ -33,7 +35,9 @@ describe( 'useElementChildren', () => {
 		mockGetContainer.mockReturnValue( mockContainer );
 
 		// Act.
-		const { result } = renderHook( () => useElementChildren( 'element-1', [ 'tab', 'accordion' ] ) );
+		const { result } = renderHook( () =>
+			useElementChildren( 'element-1', { tabs: 'tab', accordions: 'accordion' } )
+		);
```
Omerisra6, line 40: potato, potahto

Omerisra6 (approved):

PR #33565, opened 2025-11-20
Internal: Canvas - Render Tiptap inline editor instead (or on top of) widget in canvas [ED-21527]

YaelAvitan:
### ✨ PR Description
Purpose: Add inline text editing capability directly on the canvas by rendering Tiptap editor on top of widget elements.
Main changes:
- Added InlineEditorOverlay component to render editable text directly on canvas elements
- Implemented registry-based overlay system with configurable rendering conditions
- Added sync between canvas edits and element settings with debounced updates
- Added tests for inline editing functionality on the canvas


packages/packages/core/editor-canvas/src/utils/inline-editing-utils.ts
```diff
@@ -0,0 +1,16 @@
+import { type V1Element } from '@elementor/editor-elements';
```
YaelAvitan: ** for me: this cause that it will be render only after add some content from the panel

RobiFerentz (dismissed):
LGTM

packages/packages/core/editor-canvas/src/components/elements-overlays.tsx
```diff
@@ -1,13 +1,35 @@
...
+	{
+		component: OutlineOverlay,
+		filter: () => true,
+	},
+	{
+		component: InlineEditorOverlay,
+		filter: ( { id, isSelected } ) => {
+			if ( ! isSelected ) {
+				return false;
+			}
```
asafdl: newline

packages/packages/core/editor-canvas/src/components/elements-overlays.tsx
```diff
@@ -21,14 +43,26 @@ export function ElementsOverlays() {
 
 	return (
 		isActive &&
-		elements.map( ( [ id, element ] ) => (
-			<ElementOverlay key={ id } id={ id } element={ element } isSelected={ selected.element?.id === id } />
-		) )
+		elements.map( ( [ id, element ] ) => {
+			const isSelected = selected.element?.id === id;
+
+			return overlayRegistry
```
asafdl: i think this should be a function.
also im not sure i like the way it's written, we iterate twice over the overlay registry (which isn't the end of the world) but feels wrong as a dev.

so i guess something like

```
.map( ( overlay, index ) => {
if ( ! overlay.filter(...) ) return null; 
```

packages/packages/core/editor-canvas/src/components/inline-editor-overlay.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { htmlPropTypeUtil } from '@elementor/editor-props';
+import { Box } from '@elementor/ui';
+import { FloatingPortal } from '@floating-ui/react';
+
+import { useFloatingOnElement } from '../hooks/use-floating-on-element';
+import type { ElementOverlayProps } from '../types/element-overlay';
+import { getInlineEditablePropertyName } from '../utils/inline-editing-utils';
+import { CANVAS_WRAPPER_ID } from './outline-overlay';
+
+export function InlineEditorOverlay( { element, isSelected, id }: ElementOverlayProps ) {
```
asafdl: add return type on function? but also we usually set react components as const () =>

packages/packages/core/editor-canvas/src/components/inline-editor-overlay.tsx
```diff
@@ -0,0 +1,55 @@
...
+		return null;
+	}
+
+	return (
+		<FloatingPortal id={ CANVAS_WRAPPER_ID }>
+			<Box
+				ref={ floating.setRef }
+				style={ {
+					...floating.styles,
+					zIndex: 1000,
```
asafdl: i think you should add it as a const
```suggestion
					zIndex: ${ OVERLAY_Z_INDEX },
```

packages/packages/core/editor-canvas/src/components/outline-overlay.tsx
```diff
@@ -23,7 +21,7 @@ const OverlayBox = styled( Box, {
 	pointerEvents: 'none',
 } ) );
 
-export function ElementOverlay( { element, isSelected, id }: Props ) {
+export function OutlineOverlay( { element, isSelected, id }: Props ) {
```
asafdl: No return type, and usually components are defined as const

packages/packages/core/editor-canvas/src/utils/inline-editing-utils.ts
```diff
@@ -0,0 +1,26 @@
+import { getElementType, type V1Element } from '@elementor/editor-elements';
+
+function getHtmlPropertyName( container: V1Element | null ): string {
```
asafdl: im not sure about this implementation, should all widgets with html prop automatically get the inline editor?

packages/packages/libs/editor-controls/src/components/inline-editor.tsx
```diff
@@ -41,6 +41,18 @@ export const InlineEditor = React.forwardRef(
 			onUpdate: ( { editor: updatedEditor } ) => setValue( updatedEditor.getHTML() ),
 		} );
 
+		React.useEffect( () => {
```
asafdl: probably need a useOnUpdate here

```ts
function useOnUpdate( callback: () => void, dependencies: DependencyList ) {
	const hasMounted = useRef( false );

	useEffect( () => {
		if ( hasMounted.current ) {
			callback();
		} else {
			hasMounted.current = true;
		}
		// eslint-disable-next-line react-hooks/exhaustive-deps
	}, dependencies );
}
```

packages/packages/core/editor-canvas/src/components/elements-overlays.tsx
```diff
@@ -1,13 +1,35 @@
...
+const ELEMENTS_DATA_ATTR = 'atomic';
+
+const overlayRegistry: ElementOverlayConfig[] = [
+	{
+		component: OutlineOverlay,
+		filter: () => true,
+	},
+	{
+		component: InlineEditorOverlay,
+		filter: ( { id, isSelected } ) => {
```
ronenboxer: I think that `hasInlineEditableProperty` should get containerId as an argument, get the container and do its checks. It seems to make more sense semantically IMO.

This way, the filter function will be simpler and actually doing its simple checks
filter: ( { id, isSelected } ) => isSelected && hasInlineEditableProperty( id )

packages/packages/core/editor-canvas/src/components/elements-overlays.tsx
```diff
@@ -21,14 +43,26 @@ export function ElementsOverlays() {
...
+					const OverlayComponent = overlay.component;
+					return (
+						<OverlayComponent
+							key={ `${ id }-${ index }` }
+							id={ id }
+							element={ element }
+							isSelected={ isSelected }
+						/>
+					);
+				} );
```
ronenboxer: map method can return undefined in React render context
and since filter is also an array method, I would name it `shouldRender` to avoid confusion
```suggestion
				.map( ( overlay, index ) => {
					const OverlayComponent = overlay.component;

					return ( overlay.filter( { id, element, isSelected } ) ) &&
						<OverlayComponent
							key={ `${ id }-${ index }` }
							id={ id }
							element={ element }
							isSelected={ isSelected }
						/>
					);
				} );
```

packages/packages/core/editor-canvas/src/components/inline-editor-overlay.tsx
ronenboxer: This is a react component, no need to add return type, but we do use const when declaring

packages/packages/core/editor-canvas/src/types/element-overlay.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export type OverlayFilterArgs = {
+	id: string;
+	element: HTMLElement;
+	isSelected: boolean;
+};
+
+export type ElementOverlayConfig = {
+	component: React.ComponentType< ElementOverlayProps >;
+	filter: ( args: OverlayFilterArgs ) => boolean;
```
ronenboxer: This name prevents confusion IMO
```suggestion
	shouldRender: ( args: OverlayFilterArgs ) => boolean;
```

packages/packages/core/editor-canvas/src/utils/inline-editing-utils.ts
```diff
@@ -0,0 +1,26 @@
...
+	}
+
+	const propsSchema = getElementType( widgetType )?.propsSchema;
+
+	if ( ! propsSchema ) {
+		return '';
+	}
+
+	const entry = Object.entries( propsSchema ).find( ( [ , propType ] ) => propType.key === 'html' );
+	return entry?.[ 0 ] ?? '';
```
ronenboxer, line 27:
```suggestion
	const entry = Object.entries( propsSchema ).find( ( [ , propType ] ) => propType.key === 'html' );

	return entry?.[ 0 ] ?? '';
```

tests/playwright/sanity/modules/v4-tests/inline-text-editing/canvas-editing.test.ts
```diff
@@ -0,0 +1,70 @@
...
+
+		editor = await wpAdminPage.openNewPage();
+	} );
+
+	test.afterAll( async () => {
+		await wpAdminPage?.resetExperiments();
+		await context.close();
+	} );
+
+	test( 'Edit the heading-title from the canvas and check the value in the panel and in the front', async () => {
```
ronenboxer, line 29: Add a test case for deleting entire value then entering a new one

packages/packages/core/editor-canvas/src/utils/inline-editing-utils.ts
YaelAvitan: I wasn't sure about that either,
but what other widgets might use this?
Is it better to manually give it the names of the widgets it will apply to?

packages/packages/libs/editor-controls/src/components/inline-editor.tsx
ronenboxer: whats the purpose of this?

packages/packages/core/editor-canvas/src/utils/inline-editing-utils.ts
asafdl: unless there is some smarter way to detect what we need here i would write a hardcoded list of widget types.
you should take a minute to talk with doron, we may want to enable third parties to have inline editing, (as we did in v3)

so maybe we should think about it a bit

in any case, speak with doron about it, and lets change it for now to hardcoded list

packages/packages/libs/editor-controls/src/components/inline-editor.tsx
```diff
@@ -10,6 +10,19 @@ import Text from '@tiptap/extension-text';
 import Underline from '@tiptap/extension-underline';
 import { EditorContent, useEditor } from '@tiptap/react';
 
+const useOnUpdate = ( callback: () => void, dependencies: DependencyList ): void => {
```
asafdl, line 20: functions should be below

imports
types
consts
funcs

tests/playwright/sanity/modules/v4-tests/inline-text-editing/canvas-editing.test.ts
```diff
@@ -0,0 +1,70 @@
+import { BrowserContext, Page, expect } from '@playwright/test';
```
asafdl, line 1: why do we have  PW tests, but no unit tests?
is any reason we can't test most of the logic in unit tests?

YaelAvitan: Do you think that pw tests are not necessary?
Or should unit tests be added in addition to them?

packages/packages/core/editor-canvas/src/components/elements-overlays.tsx
```diff
@@ -16,18 +32,31 @@ export function ElementsOverlays() {
...
-			<ElementOverlay key={ id } id={ id } element={ element } isSelected={ selected.element?.id === id } />
-		) )
-	);
-}
+	if ( ! isActive ) {
+		return null;
+	}
 
-const ELEMENTS_DATA_ATTR = 'atomic';
+	return elements.map( ( [ id, element ] ) => {
```
asafdl, line 43: comments here not addressed

packages/packages/core/editor-canvas/src/components/elements-overlays.tsx
```diff
@@ -16,18 +32,31 @@ export function ElementsOverlays() {
...
+
+				return (
+					<OverlayComponent
+						key={ `${ id }-${ index }` }
+						id={ id }
+						element={ element }
+						isSelected={ isSelected }
+					/>
+				);
+			} );
```
ronenboxer:
```suggestion
		return overlayRegistry
			.map( ( { shouldRender, component: Overlay }, index ) =>
				shouldRender( { id, element, isSelected } ) && (
					<Overlay
						key={ `${ id }-${ index }` }
						id={ id }
						element={ element }
						isSelected={ isSelected }
					/>
				)
			);
```

packages/packages/core/editor-canvas/src/components/inline-editor-overlay.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+	const propertyName = React.useMemo( () => {
+		const container = getContainer( id );
+		return getInlineEditablePropertyName( container );
+	}, [ id ] );
+
+	const contentProp = useElementSetting( id, propertyName );
+	const value = React.useMemo( () => htmlPropTypeUtil.extract( contentProp ) || '', [ contentProp ] );
+
+	const handleValueChange = React.useCallback(
```
asafdl: do we want a debouncer here to ensure that not on every single key stroke this function is called?

packages/packages/core/frontend-handlers/src/lifecycle-events.ts
```diff
@@ -13,6 +13,10 @@ export const onElementRender = ( {
 	elementType: string;
 	elementId: string;
 } ) => {
+	if ( ! element ) {
```
asafdl: ?

tests/playwright/sanity/modules/v4-tests/inline-text-editing/canvas-editing.test.ts
asafdl: i think that the first objective is to have as many unit tests as possible, then if you feel like they don't test enough, add PW tests

ronenboxer: PW are mandatory, and in this case for a good reason, we want to make sure that it works as expected in the entire context

tests/playwright/sanity/modules/v4-tests/inline-text-editing/canvas-editing.test.ts
```diff
@@ -0,0 +1,102 @@
...
+		const headingElement = previewFrame.locator( `.elementor-element-${ headingId }` );
+
+		await expect( headingElement ).toBeVisible();
+
+		// Act
+		await headingElement.click();
+		const inlineEditor = page.locator( INLINE_EDITING_SELECTORS.canvasInlineEditor );
+
+		await expect( inlineEditor ).toBeVisible();
+
```
ronenboxer, line 87: Just make sure there is content before deleting
type something, assert it is typed and shown, then delete and so on

tests/playwright/sanity/modules/v4-tests/inline-text-editing/canvas-editing.test.ts
asafdl: this does not absolve the need for unit tests though

ronenboxer (dismissed):

ronenboxer (dismissed):

asafdl (dismissed):

ronenboxer (approved):

PR #33568, opened 2025-11-20
Internal: Handle DOM manipulation in Alpine [ ED-21593]

Omerisra6:
### ✨ PR Description
```
Purpose: Add parent-child relation model to frontend handlers enabling element communication and DOM synchronization for nested atomic elements.

Main changes:
- Implemented `listenToChildren` API for parent elements to subscribe to child element events
- Added DOM event bubbling mechanism to propagate child element render events to parent handlers
- Removed manual initialization in tabs components in favor of new relation-based approach
- Added refreshTree utility for Alpine.js to synchronize DOM changes with reactive state
```


packages/packages/core/frontend-handlers/src/handlers-registry.ts
```diff
@@ -1,19 +1,33 @@
...
+export const handlers: Map< string, Map< string, HandlerRegistration > > = new Map();
 
 export const register = ( { elementType, id, callback }: { elementType: string; id: string; callback: Handler } ) => {
 	if ( ! handlers.has( elementType ) ) {
 		handlers.set( elementType, new Map() );
 	}
 
 	if ( ! handlers.get( elementType )?.has( id ) ) {
-		handlers.get( elementType )?.set( id, callback );
+		handlers.get( elementType )?.set( id, { callback } );
```
IshayMaya: why?

packages/packages/core/alpinejs/src/index.ts
```diff
@@ -1,2 +1,11 @@
+import { Alpine } from '@alpinejs/csp';
+
 export { Alpine } from '@alpinejs/csp';
 export { init } from './init';
+
+export const reinitTree = ( element: Element ) => {
```
IshayMaya: name is a bit weird, maybe `refreshTree` ?
also, i'm not sure it should be part of the alpine package, feels like the frontend-handlers implementation

packages/packages/core/frontend-handlers/src/lifecycle-events.ts
```diff
@@ -14,17 +14,49 @@ export const onElementRender = ( {
 	const controller = new AbortController();
 	const manualUnmount: ( () => void )[] = [];
 
+	element.dispatchEvent(
+		new CustomEvent( 'elementor/element/rendered', {
```
IshayMaya: extract the event name to a const
notice that we also have `elementor/element/render` it might be confusing

IshayMaya (approved):

PR #33577, opened 2025-11-22
Internal: Update dashboard action flows [ED-21621]

nicoladj77:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #

### ✨ PR Description
```
Purpose: Add data-id attributes to dashboard UI controls for accurate action tracking and event reporting in Elementor admin pages.
Main changes:
- Added new ActionControlTracking class to handle different control types (buttons, links, toggles, etc.)
- Added data-id attributes to form buttons, system info downloads, and element manager controls
- Extended tracking capability to filter buttons and form submissions across admin interfaces
```


core/common/modules/events-manager/module.php
```diff
@@ -32,7 +32,7 @@ public static function get_editor_events_config() {
 		$session_replays = self::get_remote_mixpanel_config()[0]['sessionReplays'] ?? [];
 
 		$settings = [
-			'can_send_events' => $can_send_events,
+			'can_send_events' => true,
```
ManorHazaz: Intended?

modules/element-manager/assets/js/app.js
```diff
@@ -356,6 +357,8 @@ export const App = () => {
 										size={ '__unstable-large' }
 										__nextHasNoMarginBottom={ true }
 										options={ plugins }
+										name="elementor-element-manager-select-filter-by-plugin"
```
ManorHazaz, line 360: is name working?

ManorHazaz (changes requested):

core/common/modules/events-manager/module.php
nicoladj77: no, sorry, it's the usual fix to send events in any case. could we define a constant similar to  `ELEMENTOR_SHOW_HIDDEN_EXPERIMENTS` that forces sending events? This would make life easier both for dev and testing.

ManorHazaz: It still has to be based on user consent, so I think you have to have it with logic..

modules/element-manager/assets/js/app.js
nicoladj77: @ManorHazaz yes this works because here they do the sensible thing and [spread the other props on the base select](https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/select-control/index.tsx), which is a simple <select> element. here a 'data-id' property would appear. 
But ToggleControl [doesn't do that](https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/toggle-control/index.tsx#L91) so you would be forced to set the data-id on the ref imperatively, which is really a bad practice in react.

ManorHazaz (approved):

core/common/modules/events-manager/module.php
nicoladj77: @ManorHazaz yes it would be still be based on user consent, but we could have specific builds for testing where this could be enforced. And we could have also the constant defined in wp-config for dev. It would not change any flow, just make our life easier.

PR #33599, opened 2025-11-25
Fix: Nested tabs arrow navigation should still work when using custom ids for tabs. [ED-21516]

RobiFerentz:


modules/nested-tabs/widgets/nested-tabs.php
```diff
@@ -1260,9 +1262,10 @@ private function content_template_single_item( $tab_index, $item, $element_uid,
 		?>
 		<#
 		const tabCount = tabIndex + 1,
+			tabTitleId = 'e-n-tab-title-' + elementUid + ( tabIndex + 1 ),
```
davseve:
```suggestion
			tabTitleId = 'e-n-tab-title-' + elementUid + ( tabCount ),
```

RobiFerentz: Fixed

murzin-nikita (approved):

davseve:
Tests to this PR are added in https://elementor.atlassian.net/browse/ED-21797

PR #33604, opened 2025-11-25
Fix: Nested tabs arrow navigation should still work when using custom ids for tabs. [ED-21516]

RobiFerentz:


davseve:
Tests to this PR are added in https://elementor.atlassian.net/browse/ED-21797

davseve (approved):

PR #33614, opened 2025-11-26
Internal: Add test for tabs arrow navigation after id change. [ED-21797]

RobiFerentz:


tests/playwright/sanity/modules/nested-tabs/nested-tabs-2.test.ts
```diff
@@ -58,6 +58,67 @@ test.describe( 'Nested Tabs tests @nested-tabs', () => {
...
+			await frame.locator( '.elementor-widget-n-tabs' ).hover();
+			await frame.locator( '.elementor-widget-n-tabs .elementor-editor-element-edit' ).first().click();
+			const secondRepeaterField = page.locator( '.elementor-control-tabs .elementor-repeater-fields:nth-child( 2 )' );
+			await secondRepeaterField.waitFor( { state: 'visible' } );
+			await secondRepeaterField.click();
+			const elementIdInput = page.locator( '.elementor-repeater-fields:nth-child( 2 ) .elementor-control-element_id input' );
+			await elementIdInput.fill( customCssId );
+
+			// Assert
+			await expect( elementIdInput ).toHaveValue( customCssId );
```
davseve: I think the check should be that the element has the id( in our case tabTitleThree), not the repeater.
WDYT?

tests/playwright/sanity/modules/nested-tabs/nested-tabs-2.test.ts
```diff
@@ -58,6 +58,67 @@ test.describe( 'Nested Tabs tests @nested-tabs', () => {
...
+		await editor.loadTemplate( filePath, false );
+		await frame.waitForSelector( '.e-n-tabs' );
+
+		const customCssId = 'custom-tab-id';
+
+		// Act
+		await test.step( 'Change CSS ID of second tab in content tab', async () => {
+			await frame.locator( '.elementor-widget-n-tabs' ).hover();
+			await frame.locator( '.elementor-widget-n-tabs .elementor-editor-element-edit' ).first().click();
+			const secondRepeaterField = page.locator( '.elementor-control-tabs .elementor-repeater-fields:nth-child( 2 )' );
```
davseve: Here you need to:
1.  Click the tabs so the repeater will be visible 
2. Click the desired repeater item 
3. Change the css value

As toy wrote it now you don't change the css value

davseve (commented):
WDYT?

davseve (approved):

PR #33618, opened 2025-11-26
Internal: Component overridable prop type [ED-21778] 

MiryamOren:
### ✨ PR Description
Purpose: Add Component Overridable prop type to enable overriding component properties with custom values while maintaining component schema integrity.
Main changes:
- Created Component_Overridable_Prop_Type class with validation, sanitization, and origin prop type reference
- Implemented Component_Overridable_Schema_Extender for automatically extending prop schemas
- Added transformers for processing component overridable values at runtime
- Fixed Dynamic_Prop_Types_Mapping to use new schema extension pattern

[ED-21778]: https://elementor.atlassian.net/browse/ED-21778?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/atomic-widgets/dynamic-tags/dynamic-prop-type.php
+++ b/modules/atomic-widgets/dynamic-tags/dynamic-prop-type.php
@@ -15,7 +15,7 @@ class Dynamic_Prop_Type extends Plain_Prop_Type {
 
 	/**
 	 * Return a tuple that lets the developer ignore the dynamic prop type in the props schema
-	 * using `Prop_Type::add_meta()`, e.g. `String_Prop_Type::make()->add_meta( Dynamic_Prop_Type::ignore() )`.
+	 * using `Prop_Type::meta()`, e.g. `String_Prop_Type::make()->meta( Dynamic_Prop_Type::ignore() )`.
 	 */
 	public static function ignore(): array {
 		return [ static::META_KEY, false ];
--- a/modules/atomic-widgets/dynamic-tags/dynamic-prop-types-mapping.php
+++ b/modules/atomic-widgets/dynamic-tags/dynamic-prop-types-mapping.php
@@ -2,14 +2,12 @@
 
 namespace Elementor\Modules\AtomicWidgets\DynamicTags;
 
-use Elementor\Modules\AtomicWidgets\PropTypes\Base\Array_Prop_Type;
-use Elementor\Modules\AtomicWidgets\PropTypes\Base\Object_Prop_Type;
+use Elementor\Modules\AtomicWidgets\PropTypes\Utils\Prop_Types_Schema_Extender;
 use Elementor\Modules\AtomicWidgets\PropTypes\Contracts\Transformable_Prop_Type;
 use Elementor\Modules\AtomicWidgets\PropTypes\Contracts\Prop_Type;
 use Elementor\Modules\AtomicWidgets\PropTypes\Image_Src_Prop_Type;
 use Elementor\Modules\AtomicWidgets\PropTypes\Primitives\Number_Prop_Type;
 use Elementor\Modules\AtomicWidgets\PropTypes\Primitives\String_Prop_Type;
-use Elementor\Modules\AtomicWidgets\PropTypes\Union_Prop_Type;
 use Elementor\Modules\AtomicWidgets\PropTypes\Url_Prop_Type;
 use Elementor\Modules\DynamicTags\Module as V1_Dynamic_Tags_Module;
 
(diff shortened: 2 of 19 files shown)


modules/atomic-widgets/dynamic-tags/dynamic-prop-type.php
```diff
@@ -15,7 +15,7 @@ class Dynamic_Prop_Type extends Plain_Prop_Type {
 
 	/**
 	 * Return a tuple that lets the developer ignore the dynamic prop type in the props schema
-	 * using `Prop_Type::add_meta()`, e.g. `String_Prop_Type::make()->add_meta( Dynamic_Prop_Type::ignore() )`.
```
Omerisra6, line 18: why?

modules/atomic-widgets/dynamic-tags/dynamic-prop-types-mapping.php
```diff
@@ -2,94 +2,39 @@
...
-	 * @param Prop_Type $prop_type
-	 *
-	 * @return Prop_Type|Union_Prop_Type
-	 */
-	private function get_modified_prop_type( Prop_Type $prop_type ) {
-		$transformable_prop_types = $prop_type instanceof Union_Prop_Type ?
-			$prop_type->get_prop_types() :
-			[ $prop_type ];
-
+	protected function get_prop_types_to_add( Prop_Type $prop_type, array $nested_prop_types ): array {
```
Omerisra6: What is the `$nested_prop_types`?

modules/atomic-widgets/prop-types/utils/prop-types-schema-extender.php
```diff
@@ -0,0 +1,81 @@
...
+		}
+
+		foreach ( $prop_types_to_add as $added_prop_type ) {
+			$union_prop_type->add_prop_type( $added_prop_type );
+		}
+
+		return $union_prop_type;
+	}
+
+	abstract protected function get_prop_types_to_add( Prop_Type $prop_type, array $nested_prop_types ): array;
```
Omerisra6: Maybe we should just let the developer return the modified prop type.

And expose a method like: `modify_prop_type`

modules/components/prop-types/component-overridable-prop-type.php
```diff
@@ -0,0 +1,73 @@
...
+
+		['origin_value' => $origin_value] = $value;
+
+		return $origin_prop_type->validate( $origin_value );
+	}
+
+	protected function sanitize_value( $value ): array {
+		['override_key' => $override_key, 'origin_value' => $origin_value] = $value;
+
+		$origin_prop_type = $this->get_origin_prop_type();
```
Omerisra6, line 53: what if it returns null (I think you should return false/null) same for the validate

modules/components/prop-types/component-overridable-prop-type.php
```diff
@@ -0,0 +1,73 @@
...
+namespace Elementor\Modules\Components\PropTypes;
+
+use Elementor\Modules\AtomicWidgets\PropTypes\Base\Plain_Prop_Type;
+use Elementor\Modules\AtomicWidgets\PropTypes\Contracts\Prop_Type;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+class Component_Overridable_Prop_Type extends Plain_Prop_Type {
```
Omerisra6, line 12: Maybe we should not call it a component overridable prop type and jsut overridable prop type as it not realy related to components

modules/components/transformers/component-overridable-transformer.php
```diff
@@ -0,0 +1,17 @@
...
+use Elementor\Modules\AtomicWidgets\PropsResolver\Transformer_Base;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+class Component_Overridable_Transformer extends Transformer_Base {
+	public function transform( $value, Props_Resolver_Context $context ) {
+		// todo: render component overrides
+		return $value['origin_value'];
```
Omerisra6: Why can't you trigger the transformer?

Or maybe return null?

packages/packages/core/editor-components/src/component-overridable-transformer.ts
```diff
@@ -0,0 +1,8 @@
+import { createTransformer } from '@elementor/editor-canvas';
```
Omerisra6: don't you need to also add a prop type (ts) for that?

Omerisra6 (changes requested):

modules/atomic-widgets/dynamic-tags/dynamic-prop-type.php
MiryamOren: There is no `add_meta` method in `Prop_Type` - it's a mistake. I was a bit confused by it, so I thought I'll fix the comment. 
You can see we already use `String_Prop_Type::make()->meta( Dynamic_Prop_Type::ignore() )` both in 
`tests/phpunit/elementor/modules/atomic-widgets/dynamic-tags/test-dynamic-tags-module.php:362`
and
`modules/atomic-widgets/elements/atomic-youtube/atomic-youtube.php:58`

modules/atomic-widgets/dynamic-tags/dynamic-prop-types-mapping.php
MiryamOren: If the prop type we want to add dynamic to is either an object or an array, we pass to this function a list of the object's inner prop types or the array item prop type. 
It's the same logic we had before, I just refactored it a bit, so we can have different implementations of `get_prop_types_to_add`

modules/components/prop-types/component-overridable-prop-type.php
MiryamOren: Generally, I assume that in `sanitize_value` we get a value that has already been validated. Is it wrong?

modules/atomic-widgets/prop-types/utils/prop-types-schema-extender.php
MiryamOren: Adding prop type is the only thing we do here, so:
1) I think `get_prop_types_to_add` is clearer - it's more specific, and for now we don't change the schema in any other way
2) I'm not really sure how `modify_prop_type` would look

modules/components/transformers/component-overridable-transformer.php
MiryamOren: I don't understand? WDYM by "trigger the transformer"? 
Generally, `$value['origin_value']` is what this transformer will return if it hasn't given `overrides` context

packages/packages/core/editor-components/src/component-overridable-transformer.ts
MiryamOren: I didn't find any place where we register prop types in `ts`, but maybe I missed something?

modules/components/prop-types/component-overridable-prop-type.php
MiryamOren: Hmmm, IDK, do you think it will be helpful in other features? If so we'll have to change some other things to make it generic

modules/atomic-widgets/dynamic-tags/dynamic-prop-types-mapping.php
Omerisra6: Maybe add a phpdoc to explain it

modules/components/prop-types/component-overridable-prop-type.php
Omerisra6: Yes, we need to handle it as a separate flows.

modules/components/prop-types/component-overridable-prop-type.php
Omerisra6: I think so, you can also just change the key to ensure BC and keep all other code as it is

modules/components/transformers/component-overridable-transformer.php
Omerisra6: Don't you need to trigger the "origin transformer" as you did for the validation?

packages/packages/core/editor-components/src/component-overridable-transformer.ts
Omerisra6: We have the `editor-props` package (it is used for type utils and not for validations).


Look at a control it usually passes the propTypeUtil to the useBoundProp

modules/components/transformers/component-overridable-transformer.php
MiryamOren: Oh, got it. You're right. 
I thought we don't have to do it, as in the transforming process we go down -> up, but it's only true for Union, Array and Object, not for Plain prop type. 
I'll fix it

packages/packages/core/editor-components/src/component-overridable-transformer.ts
MiryamOren: @ronros-elementor is adding the propTypeUtil in the PR where he's using it

modules/components/prop-types/component-overridable-prop-type.php
MiryamOren: Done

modules/components/prop-types/component-overridable-prop-type.php
MiryamOren: Done

modules/atomic-widgets/dynamic-tags/dynamic-prop-types-mapping.php
MiryamOren: Fixed

modules/components/transformers/component-overridable-transformer.php
MiryamOren: Done

Omerisra6 (dismissed):

Omerisra6 (approved):

PR #33631, opened 2025-11-26
Internal: Add overridable prop indicator in setting fields [ED-21251]

ronros-elementor:
Screenshot 2025-11-25 at 8 20 09

Adding a new indicator to expose properties of component's nested widgets to be overridable on the instance level,

In this PR I'm adding 
- the indicator itself, with all states as described in the [figma](https://www.figma.com/design/c9toOa40958fiqatsEXnIY/Components?node-id=6519-115611&t=SH7511XmJpz8btri-0) - non-overridable, hover and overridable
- a simple block list of some common fields we do not wish to allow making overridable - _cssid (labeled as "ID"), and attributes
- a corresponding `propType` for validation purposes - `componentOverridablePropTypeUtil`

### ✨ PR Description
```
Purpose: Add visual indicator in component setting fields to show and manage overridable properties in the Elementor editor.
Main changes:
- Implemented indicator icon component that toggles between plus/check states for overridable properties
- Created OverridablePropIndicator component with popover UI for property override management
- Registered field indicator in editor panel to display on setting fields when editing components
```

### DIFF
--- a/packages/packages/core/editor-components/src/components/overridable-props/__tests__/overridable-prop-indicator.test.tsx
+++ b/packages/packages/core/editor-components/src/components/overridable-props/__tests__/overridable-prop-indicator.test.tsx
@@ -1,0 +1,157 @@
+import * as React from 'react';
+import { createMockDocumentData, createMockPropType } from 'test-utils';
+import { useBoundProp } from '@elementor/editor-controls';
+import { getV1CurrentDocument } from '@elementor/editor-documents';
+import { useElement } from '@elementor/editor-editing-panel';
+import { type PropValue } from '@elementor/editor-props';
+import { __createStore, __registerSlice } from '@elementor/store';
+import { ThemeProvider } from '@elementor/ui';
+import { generateUniqueId } from '@elementor/utils';
+import { render, screen } from '@testing-library/react';
+
+import { componentOverridablePropTypeUtil } from '../../../prop-types/component-overridable-prop-type';
+import { selectOverridableProps, slice } from '../../../store/store';
+import { COMPONENT_DOCUMENT_TYPE } from '../../consts';
+import { FORBIDDEN_KEYS, OverridablePropIndicator } from '../overridable-prop-indicator';
+
+jest.mock( '@elementor/editor-controls', () => ( {
+	...jest.requireActual( '@elementor/editor-controls' ),
+	useBoundProp: jest.fn(),
+} ) );
+jest.mock( '@elementor/editor-documents' );
+jest.mock( '@elementor/editor-editing-panel', () => ( {
+	...jest.requireActual( '@elementor/editor-editing-panel' ),
+	useElement: jest.fn(),
+} ) );
+jest.mock( '../../../store/store', () => ( {
+	...jest.requireActual( '../../../store/store' ),
+	selectOverridableProps: jest.fn(),
+} ) );
+
+const MOCK_ELEMENT_ID = 'test-element-123';
+const MOCK_COMPONENT_ID = 456;
+const MOCK_WIDGET_TYPE = 'e-heading';
+
+describe( 'OverridablePropIndicator', () => {
+	beforeEach( () => {
+		__registerSlice( slice );
+		__createStore();
+
+		jest.mocked( useElement ).mockReturnValue( {
+			element: { id: MOCK_ELEMENT_ID, type: MOCK_WIDGET_TYPE },
+			elementType: { key: MOCK_WIDGET_TYPE, propsSchema: {}, controls: [], title: 'Test Element' },
+		} );
+
+		jest.mocked( selectOverridableProps ).mockReturnValue( {
+			props: {},
+			groups: {
+				items: {},
+				order: [],
+			},
+		} );
+	} );
+
+	afterEach( () => {
+		jest.resetAllMocks();
+	} );
+
+	it.each( [
+		{
+			should: 'not show indicator on fields when not editing a component',
+			bind: 'title',
+			isOverridable: false,
+			isComponent: false,
+		},
+		{
+			should: 'be a plus icon if prop is not overridable',
+			bind: 'title',
+			isOverridable: false,
+			isComponent: true,
+		},
+		{
+			should: 'be a check icon if prop is overridable',
+			bind: 'title',
+			isOverridable: true,
+			isComponent: true,
+		},
+		{
+			should: 'not show indicator on fields bound to forbidden keys',
+			bind: '_cssid',
+			isOverridable: true,
+			isComponent: true,
+		},
+	] )( 'should $should', ( { bind, isOverridable, isComponent } ) => {
+		// Arrange
+		const mockDocument = createMockDocumentData( {
+			id: MOCK_COMPONENT_ID,
+			type: isComponent ? COMPONENT_DOCUMENT_TYPE : 'wp-page',
+		} );
+
+		jest.mocked( getV1CurrentDocument ).mockReturnValue( mockDocument );
+		jest.mocked( useBoundProp ).mockReturnValue(
+			mockPropType(
+				{
+					bind,
+					value: { $$type: 'string', value: 'Test' },
+				},
+				isOverridable
+			)
+		);
+
+		// Act
+		renderFieldWithIndicator();
+
+		// Assert
+		if ( ! isComponent ) {
+			expect( screen.queryByRole( 'button' ) ).not.toBeInTheDocument();
+
+			return;
+		}
+
+		if ( FORBIDDEN_KEYS.includes( bind ) ) {
+			expect( screen.queryByRole( 'button' ) ).not.toBeInTheDocument();
+
+			return;
+		}
+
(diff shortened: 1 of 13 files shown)


packages/packages/core/editor-components/src/components/overridable-props/overridable-prop-indicator.tsx
```diff
@@ -0,0 +1,94 @@
...
+export const FORBIDDEN_KEYS = [ '_cssid', 'attributes' ];
+
+export function OverridablePropIndicator() {
+	const { bind, value } = useBoundProp();
+	const currentDocument = getV1CurrentDocument();
+	const componentOverrides = selectOverridableProps( getState(), currentDocument.id );
+
+	if ( currentDocument.config.type !== COMPONENT_DOCUMENT_TYPE || ! currentDocument.id ) {
+		return null;
+	}
```
MiryamOren:
```suggestion
	if ( currentDocument.config.type !== COMPONENT_DOCUMENT_TYPE || ! currentDocument.id ) {
		return null;
	}
	
	const componentOverridableProps = selectOverridableProps( getState(), currentDocument.id );
```

packages/packages/core/editor-components/src/components/overridable-props/indicator.tsx
```diff
@@ -0,0 +1,78 @@
...
+	}
+`;
+
+type Props = {
+	isOverridable: boolean;
+	triggerProps: ReturnType< typeof bindTrigger >;
+	isOpen: boolean;
+};
+export const Indicator = ( { triggerProps, isOpen, isOverridable }: Props ) => (
+	<IconWrapper { ...triggerProps } className={ isOpen || isOverridable ? 'enlarged' : '' }>
```
MiryamOren:
```suggestion
export const Indicator = React.forwardRef<HTMLDivElement, Props>(( { triggerProps, isOpen, isOverridable }, ref ) => (
	<IconWrapper ref={ ref } { ...triggerProps } className={ isOpen || isOverridable ? 'enlarged' : '' }>
```

packages/packages/core/editor-components/src/components/overridable-props/overridable-prop-indicator.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+	if ( ! isPropAllowed( bind ) ) {
+		return null;
+	}
+
+	const isOverridable = componentOverridablePropTypeUtil.isValid( value );
```
MiryamOren, line 30:
```suggestion
	const isOverridable = componentOverridablePropTypeUtil.isValid( value );
	
```

packages/packages/core/editor-components/src/components/overridable-props/indicator.tsx
```diff
@@ -0,0 +1,78 @@
...
+		<IconContainer
+			className="icon"
+			aria-label={
+				isOverridable ? __( 'Overridable property', 'elementor' ) : __( 'Make prop overridable', 'elementor' )
+			}
+		>
+			{ isOverridable ? <CheckIcon fontSize={ SIZE } /> : <PlusIcon fontSize={ SIZE } /> }
```
MiryamOren, line 78: Can we add new icons from Figma? These look a bit different

packages/packages/core/editor-components/src/components/overridable-props/overridable-prop-indicator.tsx
```diff
@@ -0,0 +1,94 @@
...
+	const { bind } = useBoundProp();
+
+	const popupState = usePopupState( {
+		variant: 'popover',
+	} );
+
+	const triggerProps = bindTrigger( popupState );
+	const popoverProps = bindPopover( popupState );
+
+	const currentValue = Object.values( overridables?.props ?? {} ).find(
```
MiryamOren: IMO `currentValue` is a bit confusing - at first I thought it was the prop value (something like `{ '$$type': 'string', 'value': 'text' }`, how about `overridableConfig` or something like that?

packages/packages/core/editor-components/src/prop-types/component-overridable-prop-type.ts
```diff
@@ -0,0 +1,10 @@
+import { createPropUtils } from '@elementor/editor-props';
+import { z } from '@elementor/schema';
+
+export const componentOverridablePropTypeUtil = createPropUtils(
+	'component-overridable',
+	z.object( {
+		override_key: z.string(),
+		default_value: z.any().nullable(),
```
MiryamOren:
```suggestion
		default_value: z.object({
            $$type: z.string(),
            value: z.unknown()
        }),
```

packages/packages/core/editor-components/src/types.ts
```diff
@@ -20,9 +20,34 @@ export type UnpublishedComponent = BaseComponent & {
...
+	groups: {
+		items: Record< string, OverridablePropGroup >;
+		order: string[];
+	};
+};
+
 type BaseComponent = {
 	uid: string;
 	name: string;
+	overrides?: OverridableProps;
```
MiryamOren: Should be named `overridables` or `overridableProps`

packages/packages/core/editor-components/src/types.ts
```diff
@@ -20,9 +20,34 @@ export type UnpublishedComponent = BaseComponent & {
 	elements: V1ElementData[];
 };
 
+export type OverridableProp = {
+	'override-key': string;
```
MiryamOren: My bad, fixed in HLD
```suggestion
	'overrideKey': string;
```

packages/packages/core/editor-components/src/types.ts
```diff
@@ -20,9 +20,34 @@ export type UnpublishedComponent = BaseComponent & {
...
+	label: string;
+	elementId: string;
+	propKey: string;
+	widgetType: string;
```
MiryamOren, line 29: Again, my bad, we need the elType too (can be element, not widget)
```suggestion
    elType: string
	widgetType: string;
```

packages/packages/core/editor-components/src/components/overridable-props/__tests__/overridable-prop-indicator.test.tsx
```diff
@@ -0,0 +1,157 @@
...
+				order: [],
+			},
+		} );
+	} );
+
+	afterEach( () => {
+		jest.resetAllMocks();
+	} );
+
+	it.each( [
```
MiryamOren, line 58: Can we move the assertions to the parameters? So, for example, we'll have something like
```
[
		{
			should: 'not show indicator on fields when not editing a component',
			....
			expect: {
                showIndicator: false,
            }
		},
		{
			should: 'be a plus icon if prop is not overridable',
			....
			expect: {
                showIndicator: true,
                indicatorLabel: 'Make prop overridable',
            }
		},
		{
			should: 'be a check icon if prop is overridable',
			...
			expect: {
                showIndicator: true,
                indicatorLabel: 'Overridable property',
            }
		},
		{
			should: 'not show indicator on fields bound to forbidden keys',
			...
			expect: {
                showIndicator: false,
            }
		},
]
```

packages/packages/core/editor-components/src/components/overridable-props/indicator.tsx
```diff
@@ -0,0 +1,78 @@
+import * as React from 'react';
+import { CheckIcon, PlusIcon } from '@elementor/icons';
+import { type bindTrigger, Box, styled } from '@elementor/ui';
+import { __ } from '@wordpress/i18n';
+
+const SIZE = 'tiny';
+
+const IconWrapper = styled( Box )`
```
MiryamOren: Can we name it `IndicatorWrapper` / `IndicatorContainer` so the hierarchy between it and `IconContainer` will be clearer?

packages/packages/core/editor-components/src/components/overridable-props/indicator.tsx
ronros-elementor: I've made the icons slightly boldקר, with a strike width of 2px (on the `<svg />` node)

packages/packages/core/editor-components/src/prop-types/component-overridable-prop-type.ts
ronros-elementor: but it should still be nullable though, as it could have no value when being set as overridable (i.e. no default value) - updated in my PR

packages/packages/core/editor-components/src/prop-types/component-overridable-prop-type.ts
```diff
@@ -5,6 +5,11 @@ export const componentOverridablePropTypeUtil = createPropUtils(
 	'component-overridable',
```
MiryamOren: (Have been changed in PHP 🥲 )
```suggestion
	'overridable',
```

MiryamOren (changes requested):
Looks good! 
Please note the change in overridable prop type key

MiryamOren (approved):
:zap:

PR #33671, opened 2025-12-01
Internal: Add control replacement for overridable props [ED-21457]

ronros-elementor:
### ✨ PR Description
Purpose: Add control replacement functionality for overridable props to manage and update component property values with their original values.
Main changes:
- Created OverridablePropControl component that wraps original controls while preserving prop values
- Enhanced useControlReplacement hook to support passing original control to replacements
- Implemented updateOverridablePropOriginValue utility to update component props while maintaining override keys


packages/packages/core/editor-components/src/prop-types/component-overridable-prop-type.ts
```diff
@@ -2,14 +2,16 @@ import { createPropUtils } from '@elementor/editor-props';
 import { z } from '@elementor/schema';
 
 export const componentOverridablePropTypeUtil = createPropUtils(
-	'overridable',
+	'component-overridable',
```
MiryamOren: It's actually `overridable`

packages/packages/core/editor-components/src/store/update-overridable-prop-default-value.ts
```diff
@@ -0,0 +1,39 @@
+import { __dispatch as dispatch, __getState as getState } from '@elementor/store';
+
+import { type ComponentOverridablePropValue } from '../prop-types/component-overridable-prop-type';
+import { type OverridableProps } from '../types';
+import { selectOverridableProps, slice } from './store';
+
+export function updateOverridablePropDefaultValue( componentId: number, propValue: ComponentOverridablePropValue ) {
```
MiryamOren:
```suggestion
export function updateOverridablePropOriginValue( componentId: number, propValue: ComponentOverridablePropValue ) {
```

packages/packages/core/editor-components/src/store/update-overridable-prop-default-value.ts
```diff
@@ -0,0 +1,39 @@
...
+	const overridableProps = selectOverridableProps( getState(), componentId );
+
+	if ( ! overridableProps ) {
+		return;
+	}
+
+	const existingOverridableProp = Object.values( overridableProps.props ).find(
```
MiryamOren: `overridableProps.props` is an object `<overrideKey>: <propConfig>`, so we can just do
`overridableProps.props[ propValue.override_key ]`, no need to loop over the array

packages/packages/core/editor-components/src/store/update-overridable-prop-default-value.ts
MiryamOren: Just for naming consistency - `originValue` instead of `defaultValue`

packages/packages/core/frontend-handlers/src/lifecycle-events.ts
```diff
@@ -9,10 +9,14 @@ export const onElementRender = ( {
...
 }: {
-	element: Element;
+	element: Element | undefined;
 	elementType: string;
 	elementId: string;
 } ) => {
+	if ( ! element ) {
+		return;
+	}
+
```
MiryamOren: How is this change related to the overridable prop control replacement?

packages/packages/core/editor-components/src/components/overridable-props/overridable-prop-control.tsx
```diff
@@ -0,0 +1,91 @@
...
+	);
+}
+
+function useAllControls(): Control[] {
+	const { elementType } = useElement();
+
+	return iterateControls( elementType.controls );
+}
+
+function iterateControls( controls: ControlItem[] ): Control[] {
```
MiryamOren: Instead of this, we can pass the original control to replacement controls like that:
```
// packages/packages/libs/editor-controls/src/control-replacements.tsx
// refactor useControlReplacement
export const useControlReplacement = ( OriginalComponent: ComponentType ) => {
	const { value, placeholder } = useBoundProp();
	const replacements = useContext( ControlReplacementContext );

	try {
		const replacement = replacements.find( ( r ) => r.condition( { value, placeholder } ) );

		return { 
			ControlToRender: replacement?.component ?? OriginalComponent,
			OriginalControl: OriginalComponent,
			isReplaced: !!replacement,
		};
	} catch {
		return { ControlToRender: OriginalComponent, OriginalControl: OriginalComponent };
	}
};
```
and
```
// packages/packages/libs/editor-controls/src/create-control.tsx
export function createControl< T extends AnyComponentType >( Control: T ) {
	return ( ( props: ComponentProps< T > ) => {
		const {
			ControlToRender,
			OriginalControl,
			isReplaced,
		} = useControlReplacement( Control );

		const controlProps = isReplaced ? { ...props, OriginalControl } : props;

		return (
			<ErrorBoundary fallback={ null }>
				<ControlToRender { ...controlProps } />
			</ErrorBoundary>
		);
	} ) as ControlComponent< T >;
}
```

packages/packages/core/editor-components/src/components/overridable-props/__tests__/overridable-prop-control.test.tsx
```diff
@@ -0,0 +1,225 @@
+import * as React from 'react';
```
MiryamOren, line 1: Generally in this file, please change `default` to `origin`

packages/packages/core/editor-components/src/components/overridable-props/__tests__/overridable-prop-control.test.tsx
```diff
@@ -0,0 +1,225 @@
...
+				origin_value: stringPropTypeUtil.create( 'Default Title' ),
+			} ),
+		},
+		{
+			render: 'a text control with no content',
+			bind: BIND,
+			expected: '',
+			value: componentOverridablePropTypeUtil.create( {
+				override_key: MOCK_OVERRIDE_KEY,
+				origin_value: null,
```
MiryamOren: how can `origin_value` be null? shouldn't it be `stringPropTypeUtil.create( null )`?

packages/packages/core/editor-components/src/components/overridable-props/__tests__/overridable-prop-control.test.tsx
```diff
@@ -0,0 +1,225 @@
...
+			} ),
+		},
+		{
+			render: "a text control with the original prop type's default value",
+			bind: BIND_WITH_DEFAULT,
+			expected: 'Default PropType Text',
+			value: componentOverridablePropTypeUtil.create( {
+				override_key: MOCK_OVERRIDE_KEY,
+				origin_value: null,
```
MiryamOren, line 138: Same here

packages/packages/core/editor-components/src/components/overridable-props/__tests__/overridable-prop-control.test.tsx
```diff
@@ -0,0 +1,225 @@
...
+
+			expect( screen.getByRole( 'textbox' ) ).toHaveValue( expected );
+			return;
+		}
+
+		expect( screen.queryByRole( 'spinbutton' ) ).not.toBeInTheDocument();
+		expect( screen.queryByRole( 'textbox' ) ).not.toBeInTheDocument();
+	} );
+
+	it( 'should auto-generate override_key and wrap setValue when value is not overridable first', () => {
```
MiryamOren: Should it? How can we get to this scenario?

packages/packages/core/editor-components/src/components/overridable-props/__tests__/overridable-prop-control.test.tsx
```diff
@@ -0,0 +1,225 @@
...
+		const props = { value, setValue, bind: BIND, propType: mockPropType };
+
+		// Act
+		renderOverridableControl( props );
+
+		const input = screen.getByRole( 'textbox' );
+		fireEvent.change( input, { target: { value: 'New Value' } } );
+
+		// Assert
+		expect( setValue ).toHaveBeenCalledWith( {
```
MiryamOren, line 203: Can we add assertion for `updateOverridablePropDefaultValue` too?

packages/packages/core/editor-components/src/components/overridable-props/__tests__/overridable-prop-control.test.tsx
ronros-elementor: the main idea here is that I don't mind about the origin value, I just expect to see it if it's valid, but I can keep one test with a create(null)

packages/packages/core/editor-components/src/components/overridable-props/__tests__/overridable-prop-control.test.tsx
ronros-elementor: You're right, as this control should be rendered only after meeting the criteria of being overridable in the first place
But if someone would enforce this control for any reason (as in - not via the control-replacement flow, or without any condition) - we should expect something... I could expect an error though if you would prefer such handling instead, but I would say it's still worth testing

packages/packages/core/editor-components/src/components/overridable-props/__tests__/overridable-prop-control.test.tsx
ronros-elementor: it was painful, but added it. 😄

** also changed some parts of the tests suite to address the changes with how we pass the OriginalComponent prop to the replaced control component

packages/packages/core/editor-components/src/components/overridable-props/overridable-prop-control.tsx
ronros-elementor: 🔥

packages/packages/core/frontend-handlers/src/lifecycle-events.ts
ronros-elementor: it's not, and it accidentally got somehow pushed to both of my extracted PRs, apologize!

packages/packages/core/editor-components/src/components/overridable-props/overridable-prop-control.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+export function OverridablePropControl< T extends object >( {
+	OriginalControl,
+	...props
+}: T & { OriginalControl: ComponentType< T > } ) {
+	const { elementType } = useElement();
+
+	const { value, bind, setValue, placeholder, ...propContext } = useBoundProp();
+	const { extract, create } = componentOverridablePropTypeUtil;
+	const componentId = getCurrentDocumentId() ?? 0;
```
MiryamOren: I think we should throw error if `getCurrentDocumentId()` returns null / undefined

MiryamOren: IMO, in case `getCurrentDocumentId()` returns a nullish value we shouldn't replace it with 0, but throw an error

packages/packages/core/editor-components/src/components/overridable-props/overridable-prop-control.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+	const currentValue = extract( value ) ?? {
+		override_key: generateUniqueId(),
```
MiryamOren: We shouldn't just generate uid for the prop value only - if we don't add it to the component's overridable props meta, it will never be exposed

packages/packages/core/editor-components/src/components/overridable-props/__tests__/overridable-prop-control.test.tsx
MiryamOren: See
image
I would just throw error

packages/packages/core/editor-components/src/components/overridable-props/__tests__/overridable-prop-control.test.tsx
MiryamOren: ⚡

MiryamOren (commented):
Overall looks great, some little issues

packages/packages/core/editor-components/src/components/overridable-props/__tests__/overridable-prop-control.test.tsx
ronros-elementor: The only remaining use of `default` is for a prop type with a default value, which is indeed "default" - so is this considered resolved? :)

ronros-elementor: I even reverted a bit of the prior "default"->"origin" change, as it got a bit confusing at some point, since this _is_ intended now for specifically the `propType`'s default value (e.g. "This is a title")

packages/packages/core/editor-components/src/components/overridable-props/__tests__/overridable-prop-control.test.tsx
ronros-elementor: btw, though marked as resolved, for context (added a comment in the test as well) - there's actually a difference between the prop util's create(null) output, to a real null value, and it also behaves like that in our templates -
in our text control, for example, we would get the default prop type (e.g. "This is a title") as the actual value to display, but in reality - `settings.title` is actually undefined (null is equivalent for that matter).
if you'd type in anything, `settings.title` would be updated with the current value, but if you'd erase everything, you'd end up with 
```
settings.title = {
  $$type: 'string',
  value: '', // equivalent to null
}
```
and in such case we would _not_ fallback to the default value of the prop type - for our heading widget we would simply not display anything, as it would behave as if `title` is empty

packages/packages/core/editor-components/src/components/overridable-props/overridable-prop-control.tsx
ronros-elementor: will anyway change it to be using our store, and will add a throw

packages/packages/core/editor-components/src/components/overridable-props/overridable-prop-control.tsx
ronros-elementor: changing it to be using the componentOverridablePropUtil and as you rightfully suggested - will through an error if the value is not overridable to begin with

MiryamOren (approved):

PR #33686, opened 2025-12-02
Fix: [V4] Custom CSS field breaks from certain characters [ED-21876]

YaelAvitan:
### ✨ PR Description
Purpose: Fix encoding/decoding of custom CSS fields to properly handle Unicode characters and special symbols.
Main changes:
- Updated encodeString to use TextEncoder for proper Unicode handling before base64 conversion
- Improved decodeString to use TextDecoder for reliable Unicode character reconstruction
- Added tests to verify correct handling of mixed Unicode characters and invalid inputs

### DIFF
--- a/packages/packages/libs/utils/src/__tests__/encoding.test.ts
+++ b/packages/packages/libs/utils/src/__tests__/encoding.test.ts
@@ -1,21 +1,19 @@
 import { decodeString, encodeString } from '../encoding';
 
 describe( 'encoding', () => {
-	describe( 'encodeString and decodeString', () => {
-		it( 'should encode and decode mixed Unicode characters', () => {
-			const original = 'Hello שלום مرحبا 你好 🎉 /*...*/';
-			const encoded = encodeString( original );
-			const decoded = decodeString( encoded );
+	it( 'should encode and decode mixed Unicode characters', () => {
+		const original = 'Hello שלום مرحبا 你好 🎉 /*...*/';
+		const encoded = encodeString( original );
+		const decoded = decodeString( encoded );
 
-			expect( decoded ).toBe( original );
-		} );
+		expect( decoded ).toBe( original );
+	} );
 
-		it( 'should return fallback on decode error', () => {
-			const invalid = 'not-valid-base64!!!';
-			const fallback = 'fallback value';
-			const decoded = decodeString( invalid, fallback );
+	it( 'should return fallback on decode error', () => {
+		const invalid = 'not-valid-base64!!!';
+		const fallback = 'fallback value';
+		const decoded = decodeString( invalid, fallback );
 
-			expect( decoded ).toBe( fallback );
-		} );
+		expect( decoded ).toBe( fallback );
 	} );
 } );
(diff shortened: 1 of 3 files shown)


murzin-nikita (dismissed):

packages/packages/libs/utils/src/__tests__/encoding.test.ts
```diff
@@ -0,0 +1,21 @@
+import { decodeString, encodeString } from '../encoding';
+
+describe( 'encoding', () => {
+	describe( 'encodeString and decodeString', () => {
```
RobiFerentz: do we need a nested describe here?

RobiFerentz (dismissed):

murzin-nikita (dismissed):

packages/packages/libs/utils/src/encoding.ts
```diff
@@ -1,11 +1,18 @@
 export const encodeString = ( value: string ): string => {
-	return btoa( value );
+	const utf8Bytes = encodeURIComponent( value ).replace( /%([0-9A-F]{2})/g, ( _, hex ) =>
+		String.fromCharCode( parseInt( hex, 16 ) )
+	);
+	return btoa( utf8Bytes );
 };
 
-export const decodeString = < T = string >( value: string, fallback?: T ): T | string => {
+export const decodeString = < T = string >( value: string, fallback?: T ): string | T => {
```
asafdl, line 6: https://www.geeksforgeeks.org/web-tech/web-textencoder-api-textencoder-encode-method/
can we use text encoder api?

asafdl: with utf8

YaelAvitan: I tried to do it but
The problem with TextEncoder:
Not available in Node.js - The tests (Jest) run in Node.js which does not have TextEncoder by default, which causes Refe-failed in the ci.

RobiFerentz (dismissed):

packages/packages/libs/utils/src/encoding.ts
```diff
@@ -1,11 +1,22 @@
 export const encodeString = ( value: string ): string => {
-	return btoa( value );
+	const encoder = new TextEncoder();
```
asafdl: small style suggestion but i prefer:
```
const binary = Array.from(new TextEncoder().encode(value), b => String.fromCharCode(b)).join('');
```
same in decode

packages/packages/libs/utils/src/encoding.ts
```diff
@@ -1,11 +1,22 @@
...
+export const decodeString = < T = string >( value: string, fallback?: T ): string | T => {
 	try {
-		return atob( value );
+		const binary = atob( value );
+		const bytes = new Uint8Array( [ ...binary ].map( ( char ) => char.charCodeAt( 0 ) ) );
+		const decoder = new TextDecoder();
+		return decoder.decode( bytes );
 	} catch {
-		return fallback || '';
+		return fallback ?? '';
```
asafdl: fallback may be nullable (if T is nullable) so maybe
`fallback !== undefined `?

asafdl (approved):

PR #33692, opened 2025-12-02
Internal: Save component overridable props metadata [ED-21262]

MiryamOren:
### ✨ PR Description
Purpose: Add metadata persistence for component overridable props by implementing parsers and update hooks in the document save process

Main changes:
- Renamed 'defaultValue' to 'originValue' in Component_Overridable_Prop class for property origin representation
- Added Component_Overridable_Props_Parser to validate and sanitize overridable props metadata
- Implemented document update_overridable_props method with validation and sanitization
- Created save hook to store component overridable props during document save operations


assets/dev/js/editor/document/save/commands/internal/save.js
```diff
@@ -6,9 +6,14 @@ export class Save extends $e.modules.CommandInternalBase {
...
 		}
 
-		const container = document.container,
-			settings = container.settings.toJSON( { remove: [ 'default' ] } ),
-			oldStatus = container.settings.get( 'post_status' );
+		const container = document.container;
+
+		if ( elementorCommon.config.experimentalFeatures?.e_components ) {
+			await elementorCommon.__beforeSave?.( { container, status } );
+		}
```
Omerisra6, line 13: This command shouldn't be aware of the `e_components` experiment. I think there is a hooks system for this command (similar to the WordPress hooks), try to use it.

Omerisra6: I see now that you are not the one who added it, but why did you moved it here?

modules/components/documents/component-overridable-props-parser.php
```diff
@@ -0,0 +1,422 @@
...
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+class Component_Overridable_Props_Parser {
+	public static function make(): self {
+		return new static();
+	}
+
+	public function parse( $data ): Parse_Result {
```
Omerisra6: I would add a phpdoc for the expected structure of the incoming data

modules/components/documents/component-overridable-props-parser.php
Omerisra6, line 1: I think you should split it into files: 

`overridable_group_parser`
`overridable_prop_parser`
`overridable_props_parser`


I also think each file should have one `parse` method that runs `validate` and `sanitize` internally to make it more readable.

modules/components/documents/component.php
```diff
@@ -64,4 +65,20 @@ public function get_overridable_props(): Component_Overridable_Props {
 
 		return new Component_Overridable_Props( $meta );
 	}
+
+	public function set_overridable_props( $data ): Parse_Result {
```
Omerisra6:
```suggestion
	public function save_overridable_props( $data ): Parse_Result {
```

modules/components/documents/component-overridable-props.php
```diff
@@ -19,6 +19,7 @@ private function __construct( $overridable_props_meta ) {
 			$this->props = [];
 			$this->groups = [];
 
+			return;
```
Omerisra6, line 22: why?

assets/dev/js/editor/document/save/commands/internal/save.js
MiryamOren: Yes, it's a workaround we made - we can't use hooks, since async hook dependencies are not supported, so we added this call.
I moved it up, so we can push settings to the document container before the save hook is reading it (line 15)

modules/components/documents/component-overridable-props-parser.php
MiryamOren: 🫡

modules/components/documents/component.php
MiryamOren: hmm, maybe `update_overridable_props`? since what we do is calling `update_metadata`

modules/components/documents/component-overridable-props.php
MiryamOren: Not actually related to this PR, I just fixed this error by the way - if the meta is empty we want to return a class with empty groups and props, and not to continue and try to format these nonexistent fields content

modules/components/documents/component-overridable-props-parser.php
MiryamOren: Done 🤝

packages/packages/core/editor-components/src/sync/set-component-overridable-props-settings-before-save.ts
```diff
@@ -0,0 +1,19 @@
...
+	container: V1Element & { document: V1Document };
+} ) => {
+	const currentDocument = container.document;
+
+	if ( ! currentDocument || currentDocument.config.type !== COMPONENT_DOCUMENT_TYPE ) {
+		return;
+	}
+
+	// todo: get overridable props from redux store
+	// container.settings.set( 'overridable_props', overridableProps );
```
ronros-elementor: We have [selectOverridableProps](https://github.com/elementor/elementor/blob/main/packages/packages/core/editor-components/src/store/store.ts#L136) to get a component's overridable props, if you'd like to fully connect the dots 😃

MiryamOren: Done 🤝

Omerisra6 (dismissed):

Omerisra6 (dismissed):

Omerisra6 (approved):

PR #33704, opened 2025-12-03
Internal: [V4] Inline text editor - Link button (toolbar) [ED-21813]

YaelAvitan:
### ✨ PR Description
Purpose: Add link functionality to inline text editor toolbar enabling users to create hyperlinks with customizable target attributes in V4 atomic widgets.

Main changes:
- Implemented link button in inline editor toolbar with URL popover for href and target="_blank" configuration
- Added TipTap Link extension with openOnClick disabled for inline editor functionality
- Extended HTML sanitization to allow anchor tag target attribute in atomic widget prop types
- Created comprehensive test suite covering link creation, editing, removal, and partial text selection scenarios


modules/atomic-widgets/module.php
```diff
@@ -378,4 +379,9 @@ private function register_frontend_scripts() {
 		$loader = new Frontend_Assets_Loader();
 		$loader->register_scripts();
 	}
+
+	private function add_inline_styles() {
+		$inline_css = '.e-heading a, .e-paragraph a { all: unset; cursor: pointer; }';
+		wp_add_inline_style( 'elementor-frontend', $inline_css );
+	}
```
YaelAvitan, line 386: this is for prevent an outline when clicking on text with a link
not sure this is the best way to do it... ?

modules/atomic-widgets/elements/atomic-heading/atomic-heading.html.twig
```diff
@@ -2,7 +2,7 @@
 	{% set id_attribute = settings._cssid is not empty ? 'id=' ~ settings._cssid | e('html_attr') : '' %}
 	<{{ settings.tag | e('html_tag') }} 
 		data-interaction-id="{{ id }}" 
-		class="{{ settings.classes | merge( [ base_styles.base ] ) | join(' ') }}" 
+		class="{{ settings.classes | merge( [ base_styles.base, 'e-heading' ] ) | join(' ') }}" 
 		{{ id_attribute }}
```
YaelAvitan, line 6: this is for prevent the outline style as what I wrote in [modules/atomic-widgets/module.php‎]

packages/packages/libs/editor-controls/src/components/url-popover.tsx
```diff
@@ -0,0 +1,65 @@
...
+				paper: { sx: { borderRadius: '16px', width: anchorRef.current?.offsetWidth + 'px', marginTop: -1 } },
+			} }
+			{ ...bindPopover( popupState ) }
+			anchorOrigin={ { vertical: 'top', horizontal: 'left' } }
+			transformOrigin={ { vertical: 'top', horizontal: 'left' } }
+			onClose={ handleClose }
+		>
+			<Stack direction="row" alignItems="center" gap={ 1 } sx={ { p: 1.5 } }>
+				<TextField
+					value={ value }
```
RobiFerentz, line 51: Do we have some sort of validation here? Not sure of type=url will do what we need, but should we not ensure the link is a valid input? Does tiptap already do that when trying to input something invalid to the actual editor?

modules/atomic-widgets/elements/atomic-heading/atomic-heading.html.twig
ronenboxer: I'm not sure if this is the right place to add that

There is a const `BASE_STYLE_KEY` that is used by many atoms.
It should be removed from their classes and be put in `Atomic_Widget_Base` as well as `Atomic_Element_Base`

Then, you will be able to use it `trait Has_Base_Style`, something like this:
`$result[ self::BASE_STYLE_KEY ] ??= '';
$result[ self::BASE_STYLE_KEY ] .= ' ' . static::get_element_type();`

modules/atomic-widgets/module.php
ronenboxer: Why not use `.e-heading-base` which is already there?

ronenboxer: This way you won't need any of the previous changes

packages/packages/libs/editor-controls/src/components/__tests__/inline-editor-toolbar.test.tsx
```diff
@@ -99,4 +109,83 @@ describe( 'InlineEditorToolbar', () => {
...
 	} );
+
+	describe( 'Link functionality', () => {
+		it( 'should open URL popover when clicking link button', () => {
+			// Arrange.
+			const mockEditor = createMockEditor();
+
+			// Act.
+			renderWithTheme( <InlineEditorToolbar editor={ mockEditor } /> );
+			const linkButton = screen.getByLabelText( 'Link' );
```
ronenboxer, line 121:
```suggestion
			renderWithTheme( <InlineEditorToolbar editor={ mockEditor } /> );

			const linkButton = screen.getByLabelText( 'Link' );
```

packages/packages/libs/editor-controls/src/components/__tests__/url-popover.test.tsx
```diff
@@ -0,0 +1,184 @@
...
+			// Arrange.
+			const props = {
+				...defaultProps,
+				popupState: createMockPopupState( true, anchorElement ),
+				value: '',
+			};
+
+			// Act.
+			renderWithTheme( <UrlPopover { ...props } /> );
+			const input = screen.getByPlaceholderText( 'Type a URL' );
```
ronenboxer, line 106:
```suggestion
			renderWithTheme( <UrlPopover { ...props } /> );

			const input = screen.getByPlaceholderText( 'Type a URL' );
```

packages/packages/libs/editor-controls/src/components/inline-editor-toolbar.tsx
```diff
@@ -125,13 +168,26 @@ export const InlineEditorToolbar = ( { editor }: InlineEditorToolbarProps ) => {
...
+							onClick={ () =>
+								button.action === 'link' ? handleLinkClick() : button.method?.( editor )
+							}
 						>
 							{ button.icon }
 						</ToggleButton>
 					</Tooltip>
 				) ) }
 			</ToggleButtonGroup>
+			{ popupState.isOpen && (
```
ronenboxer: I think that the popover handles its displaying and hiding independently based on the popupState, please try to remove this condition

modules/atomic-widgets/module.php
YaelAvitan: you are right!
missed it... :(

packages/packages/libs/editor-controls/src/components/inline-editor-toolbar.tsx
YaelAvitan: Absolutely right!!

packages/packages/libs/editor-controls/src/components/url-popover.tsx
YaelAvitan: as what I see tiptap is not doing it
Im not sure about it
as what I see in the 'link-control' there is no validation too, so, I made it same to 
I will ask Doron about it

YaelAvitan: Doron said no need to validate (at least not now..)

RobiFerentz: ok

RobiFerentz (dismissed):

modules/atomic-widgets/module.php
ronenboxer: Shtuyot
As long as it's usable :)

ronenboxer (dismissed):

RobiFerentz (approved):

PR #33714, opened 2025-12-03
Internal: Save current component and path in store [ED-21250]

IshayMaya:
### ✨ PR Description
Purpose: Add current component and navigation path persistence in editor store to maintain component state across document switches.
Main changes:
- Created useCurrentComponent hook to access component data from store
- Replaced local state with centralized store management for component path tracking
- Added helper functions to handle component DOM element retrieval and navigation


packages/packages/core/editor-components/src/components/edit-component/edit-component.tsx
```diff
@@ -36,43 +31,39 @@ export function EditComponent() {
...
-			listenTo( commandEndEvent( 'editor/documents/attach-preview' ), () => {
-				const { component: currentComponent } = path.at( -1 ) ?? {};
-				const { id: currentComponentId } = currentComponent ?? {};
-				const nextDocument = documentsManager.getCurrent();
-
-				if ( nextDocument.id === currentComponentId ) {
-					return;
-				}
+	useEffect( () => {
+		return listenTo( commandEndEvent( 'editor/documents/attach-preview' ), () => {
```
IshayMaya, line 39: @ronros-elementor  wdyt about listening to `editor/documents/switch` here?
it works as expected for me but i might be missing something

ronros-elementor: what I remember (vaguely) about listening to this event instead of the switch is that the DOM was up to date at the time of invocation, in oppose to the latter
but in practice - if  you haven't experience any issues with the backdrop appearance (as in - not exactly around the component's bounding box), then it was probably a wrong assumption, and something else was affecting the calculation of the backdrop (and resolved somehow... it was a lot of code back then 😄), and you could certainly change it 🙏

IshayMaya: ok i'll do it in a [separated ticket](https://elementor.atlassian.net/browse/ED-21921) to avoid introducing a bug in this one

ronros-elementor (approved):

PR #33759, opened 2025-12-08
Internal: Refactor component-related prop types [ED-21894]

MiryamOren:
### ✨ PR Description
Purpose: Refactor component-related prop types to improve structure and maintainability by removing 'Component_' prefix from overridable types.

Main changes:
- Renamed Component_Overridable_Prop_Type to Overridable_Prop_Type and Component_Overridable_Schema_Extender to Overridable_Schema_Extender
- Refactored Component_Instance_Prop_Type from Plain_Prop_Type to Object_Prop_Type with nested structure for component_id and overrides
- Introduced Override_Prop_Type, Overrides_Prop_Type, and Component_Override_Parser classes to handle override validation and sanitization
- Updated sanitization to use sanitize_key instead of sanitize_text_field for override keys
- Modified component_id value structure from direct number to nested object with $type wrapper


modules/components/prop-types/component-override-parser.php
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	private ?Component_Overridable_Props $component_overridable_props = null;
+
+	public static function make(): self {
+		return new static();
+	}
+
+	public function validate_override( string $override_key, ?array $override_value, array $schema_source ): bool {
+		$component_id = $schema_source['id'];
```
Omerisra6, line 31: what if its not set?

modules/components/prop-types/component-override-parser.php
```diff
@@ -0,0 +1,122 @@
...
+			if ( ! $override_key_exists ) {
+				// If the override is not one of the component overridable props we'll remove it in sanitize_value method.
+				// This is a valid scenario, as the user can delete overridable props from the component after the override created.
+				return true;
+			}
+
+			if ( null === $override_value ) {
+				// Null override value is valid scenario -
+				// The user can create an override only to make it overridable, without providing override_value,
+				// so in case we won't get a higher level override for this prop, we'll use the origin_value of the prop.
```
Omerisra6: I think it's easy to understand without the comment

modules/components/prop-types/component-override-parser.php
```diff
@@ -0,0 +1,122 @@
...
+		}
+	}
+
+	private function get_component_overridable_prop( string $override_key, ?Component_Overridable_Props $component_overridable_props ): array {
+		if ( ! $component_overridable_props || ! isset( $component_overridable_props->props[ $override_key ] ) ) {
+			return [
+				'override_key_exists' => false,
+				'prop_type' => null,
+			];
+		}
```
Omerisra6, line 90: I think that extracting this into a different method will make the code much cleaner

modules/components/prop-types/overridable-prop-type.php
```diff
@@ -0,0 +1,77 @@
...
+	protected function sanitize_value( $value ): ?array {
+		['override_key' => $override_key, 'origin_value' => $origin_value] = $value;
+
+		$origin_prop_type = $this->get_origin_prop_type();
+
+		if ( ! $origin_prop_type ) {
+			return null;
+		}
+
+		$sanitized_override_key = sanitize_text_field( $override_key );
```
Omerisra6: Why do we need to sanitize the key? Are the keys being sanitized when saving?

If not, it can be an issue when the sanitized key is different than the original one.

modules/components/prop-types/override-parser.php
```diff
@@ -0,0 +1,39 @@
...
+use Elementor\Modules\Components\Documents\Component;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+abstract class Override_Parser {
+	public static function make(): self {
+		return new static();
+	}
```
Omerisra6, line 11:
```suggestion
	}

```

modules/components/prop-types/override-prop-type.php
```diff
@@ -0,0 +1,67 @@
...
+	protected function validate_value( $value ): bool {
+		if ( ! is_array( $value ) ) {
+			return false;
+		}
+
+		$required_fields = [
+			'override_key' => 'is_string',
+			'override_value' => fn( $value ) => is_null( $value ) || is_array( $value ),
+			'schema_source' => 'is_array',
+		];
```
Omerisra6, line 26: nice, can you apply this pattern in other places where you check structure?

modules/components/prop-types/overrides-prop-type.php
```diff
@@ -0,0 +1,28 @@
...
+	}
+
+	protected function define_item_type(): Prop_Type {
+		return Override_Prop_Type::make();
+	}
+
+	public function sanitize_value( $value ): array {
+		$sanitized = parent::sanitize_value( $value );
+
+		return array_filter( $sanitized, fn( $item ) => null !== $item['value'] );
```
Omerisra6, line 26: why this extra check is needed?

modules/components/prop-types/component-override-parser.php
MiryamOren: Fair 😬

modules/components/prop-types/component-override-parser.php
MiryamOren: Good catch! fixed
image

modules/components/prop-types/overridable-prop-type.php
MiryamOren: We do sanitize them, but it's a good point - it should be `sanitize_key`, not `sanitize_text_field` for consistency with the component overridable props sanitization

modules/components/prop-types/override-prop-type.php
MiryamOren: Done

modules/components/prop-types/overrides-prop-type.php
MiryamOren: A user can make a prop overridable, add overrides for it, and then make the prop non-overridable again. 
This is a valid case, so we don't fail the validation, but filter out these overrides.

Omerisra6: maybe add a comment

modules/components/styles/component-styles.php
```diff
@@ -39,8 +39,8 @@ private function render_post( string $post_id ) {
 
 		$components = $this->get_components_from_post( $post_id );
 		$component_ids = Collection::make( $components )
-			->filter( fn( $component ) => isset( $component['settings']['component_instance']['value']['component_id'] ) )
-			->map( fn( $component ) => $component['settings']['component_instance']['value']['component_id'] )
+			->filter( fn( $component ) => isset( $component['settings']['component_instance']['value']['component_id']['value'] ) )
```
Omerisra6, line 42: @ronros-elementor Miryam told me you did that, shame...

Omerisra6 (approved):

PR #33761, opened 2025-12-08
Fix: Proper handling for overridable prop indicator when layout is custom [ED-21973]

ronros-elementor:
## Problem
Most of our standard controls have their label rendered _outside_ of the control itself.
This work well in cases like this -
```
<Field>
  <PropProvider ...>
    <PropKeyProvider bind={ bind }>
      <ControlLabel />
      <ControlComponent />
    </PropKeyProvider>
  </PropProvider>
</Field>
```
because our control replacement mechanism is replacing `<ControlComponent />` with whatever is required per case.

For overridable props we would replace the`<ControlComponent />` as follows -
```
<Field>
  <PropProvider ...>
    <PropKeyProvider bind={ bind }>
      <ControlLabel />
      <OverridableControl>
        <AlteredPropProvider ...>
          <PropKeyProvider bind={ bind }>
            <ControlComponent />
          </PropKeyProvider>
        </AlteredPropProvider>
      </OverridableControl>
    </PropKeyProvider>
  </PropProvider>
</Field>
```
and the `<AlteredPropProvider />` would extract the `origin_value` from the overridable prop value, and pass it as the bound prop value itself, so that our control replacement acts as a HOC

Our indicators are injected to `<ControlLabel />` automatically, and are expected to use the standard hooks we provider (e.g. `useBoundProp`) for any contextual information they need.
Our overridable prop indicator uses the bound prop value to understand whether the value is compatible with `ComponentOverridablePropType`, and if it is - it marks the indicator accordingly.

When we render `<ControlLabel />` inside the control component, we would in fact render it under our `<AlteredPropProvider />`, hence the bound prop would be mistakenly considered as *non overridable*.

## Solution
Wrapping with a new custom context for overridable props specifically - wrapped only when we use the control replacement for overridable props' fields.

What that means is that we could always be able to simply rely on this piece of information to determine whether the prop is overridable or not

### ✨ PR Description
Purpose: Fix overridable prop indicator detection for custom layout controls by exposing override context when standard bound prop resolution fails.

Main changes:
- Added OverridablePropContext provider wrapping PropProvider to expose overridable value via useOverridablePropValue hook
- Updated indicator logic to fallback to context value when useBoundProp returns non-overridable value
- Replaced document-based component detection with selectCurrentComponentId selector for more reliable component identification


packages/packages/core/editor-components/src/components/overridable-props/__tests__/overridable-prop-indicator.test.tsx
Omerisra6, line 1: maybe use one of the complicated controls to test it

ronros-elementor: Added a test with a mock of such component, essentially I just care to see the `useOverridablePropValue` exposes the original prop value as intended, lmk if that's good enough 🙂

Omerisra6 (approved):

PR #33796, opened 2025-12-11
Internal: Add component instance edit panel [ED-21255]

ronros-elementor:
## Changes
- Added a new registry for editing panel content replacement
- Added a new component instance panel
  - Header
  - Collapsible groups
  - Temporary JSON output for the controls
    ** Controls will be rendered as part of another PR

Screenshot 2025-12-11 at 11 03 51
Screenshot 2025-12-11 at 11 26 06

### ✨ PR Description
Purpose: Add component instance editing panel to enable users to view and edit component properties directly from instances without switching documents.

Main changes:
- Implemented InstanceEditingPanel component with empty state, collapsible property groups, and edit component navigation
- Refactored document switching logic into switchToComponent utility function for consistent navigation behavior
- Registered editing panel replacement to conditionally render InstanceEditingPanel for component instance element types


packages/packages/core/editor-editing-panel/src/components/editing-panel.tsx
```diff
@@ -31,26 +32,34 @@ export const EditingPanel = () => {
...
-									<ElementProvider element={ element } elementType={ elementType }>
-										<EditingPanelTabs />
-									</ElementProvider>
-								</ControlReplacementsProvider>
-							</ControlActionsProvider>
-						</PanelBody>
-					</Panel>
+					<ControlActionsProvider items={ menuItems }>
+						<ControlReplacementsProvider replacements={ controlReplacements }>
+							<ElementProvider element={ element } elementType={ elementType }>
```
MiryamOren, line 59: I think we want to manipulate the `elementType`* for the instance editing panel, so maybe it should be part of the replacement?

*Since it contains the controls list

packages/packages/core/editor-editing-panel/src/components/editing-panel.tsx
```diff
@@ -31,26 +32,34 @@ export const EditingPanel = () => {
...
+								<Panel>
+									<PanelHeaderTopSlot />
+									{ ReplacementComponent ? (
+										<ReplacementComponent />
+									) : (
+										<>
```
MiryamOren: Maybe extract to a variable? something like `PanelContent` / `DefaultPanelContent`?

packages/packages/core/editor-editing-panel/src/editing-panel-replacement-registry.tsx
```diff
@@ -0,0 +1,30 @@
+import { type ComponentType } from 'react';
+import { type Element, type ElementType } from '@elementor/editor-elements';
+
+type EditingPanelReplacement = {
+	condition: ( element: Element, elementType: ElementType ) => boolean;
+	component: ComponentType;
+
+	// ordered from lowest to highest
```
MiryamOren, line 8: 🔥

MiryamOren: Thank you on behalf of the dyscalculic people

packages/packages/core/editor-components/src/store/store.ts
```diff
@@ -133,7 +133,7 @@ const selectUnpublishedData = ( state: ComponentsSlice ) => state[ SLICE_NAME ].
 const getCreatedThisSession = ( state: ComponentsSlice ) => state[ SLICE_NAME ].createdThisSession;
 const getPath = ( state: ComponentsSlice ) => state[ SLICE_NAME ].path;
 const getCurrentComponentId = ( state: ComponentsSlice ) => state[ SLICE_NAME ].currentComponentId;
-const selectComponent = ( state: ComponentsSlice, componentId: ComponentId ) =>
+export const selectComponent = ( state: ComponentsSlice, componentId: ComponentId ) =>
 	state[ SLICE_NAME ].data.find( ( component ) => component.id === componentId );
```
MiryamOren, line 137:
```suggestion
	[ ...state[ SLICE_NAME ].data, ...state[ SLICE_NAME ]unpublishedData ].find( ( component ) => component.id === componentId );
```

We should have (empty) editing panel for unpublished components too

packages/packages/core/editor-components/src/prop-types/component-instance-prop-type.ts
```diff
@@ -0,0 +1,11 @@
+import { createPropUtils } from '@elementor/editor-props';
+import { z } from '@elementor/schema';
+
+export const componentInstancePropTypeUtil = createPropUtils(
+	'component-instance',
+	z.object( {
+		component_id: z.unknown(),
+	} )
+);
```
MiryamOren, line 12:
```suggestion
const overridePropTypeUtil = createPropUtils('override', 
	z.object( {
		override_key: z.string(),
		override_value: z.unknown(),
		schema_source: z.object( { 
			type: z.literal('component'), 
			id: z.number(),
		} ),
	} )
);

export const componentInstancePropTypeUtil = createPropUtils(
	'component-instance',
	z.object( {
	  component_id: numberPropTypeUtil.schema,
	  overrides: z
		  .object( {
			  $$type: z.literal('overrides'),
			  value: z.array( overridePropTypeUtil.schema ),
		  } ),
    } )
);
```

packages/packages/core/editor-components/src/components/edit-instance-panel/edit-instance-panel.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { __ } from '@wordpress/i18n';
+
+import { componentInstancePropTypeUtil } from '../../prop-types/component-instance-prop-type';
+import { selectComponent, selectOverridableProps } from '../../store/store';
+import { type OverridablePropsGroup } from '../../types';
+import { switchToComponent } from '../../utils/switch-to-component';
+import { EmptyState } from './empty-state';
+import { OverridePropsGroup } from './override-props-group';
+
+export function EditInstancePanel() {
```
MiryamOren: Maybe `InstanceEditingPanel`? So it will be clear it replaces the `EditingPanel`

packages/packages/core/editor-components/src/components/instance-editing-panel/instance-editing-panel.tsx
```diff
@@ -0,0 +1,71 @@
...
+	const { element } = useElement();
+	const settings = useElementSetting( element.id, 'component_instance' );
+	const componentId = ( componentInstancePropTypeUtil.extract( settings )?.component_id as NumberPropValue ).value;
+
+	const component = componentId ? selectComponent( getState(), componentId ) : null;
+	const overridableProps = componentId ? selectOverridableProps( getState(), componentId ) : null;
+
+	const componentInstanceId = useSelectedElement()?.element?.id ?? null;
+
+	if ( ! componentId || ! overridableProps || ! component ) {
```
MiryamOren, line 28: No `overridableProps` is a valid state - we should show the empty state, not return null

packages/packages/core/editor-components/src/components/instance-editing-panel/instance-editing-panel.tsx
```diff
@@ -0,0 +1,71 @@
...
+						<PanelHeaderTitle>{ component.name }</PanelHeaderTitle>
+					</Stack>
+					<Tooltip title={ panelTitle }>
+						<IconButton size="tiny" onClick={ handleEditComponent } aria-label={ panelTitle }>
+							<PencilIcon fontSize="tiny" />
+						</IconButton>
+					</Tooltip>
+				</Stack>
+			</PanelHeader>
+			<PanelBody>
```
MiryamOren, line 60: I thought we'd use the `SettingsTab` and manipulate the `elementType.controls` in the element context to render different controls. 
Did you try this but it didn't work? Or do you think it's not a good idea?

packages/packages/core/editor-components/src/components/instance-editing-panel/instance-editing-panel.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { EmptyState } from './empty-state';
+import { OverridePropsGroup } from './override-props-group';
+
+export function EditInstancePanel() {
+	const { element } = useElement();
+	const settings = useElementSetting( element.id, 'component_instance' );
+	const componentId = ( componentInstancePropTypeUtil.extract( settings )?.component_id as NumberPropValue ).value;
+
+	const component = componentId ? selectComponent( getState(), componentId ) : null;
+	const overridableProps = componentId ? selectOverridableProps( getState(), componentId ) : null;
```
MiryamOren, line 24: Shouldn't we fetch the component's overridable props if this is the first time the user interacts with this component this session so we haven't yet fetched & stored them in redux?

ronros-elementor: We're loading it in parallel to all other components' required data points [here](https://github.com/elementor/elementor/blob/main/packages/packages/core/editor-components/src/store/actions/load-components-assets.ts#L14), so the instance editing panel + all props management indicators would be immediately available for our users

packages/packages/core/editor-components/src/components/instance-editing-panel/instance-editing-panel.tsx
ronros-elementor: Nope, if it's empty we'd still end up with an initial object of `props` and `groups`, to avoid the need to constantly check if a prop exists before appending any values to it (e.g. "if `groups` is missing, create `groups`, with `items` and `order`", when I simply want to create a new group)

in this selector we would only get an undefined value if no component exists

https://github.com/elementor/elementor/blob/main/packages/packages/core/editor-components/src/store/store.ts#L168-L177

ronros-elementor: (+ I even added a screenshot of our empty state 🙂)

packages/packages/core/editor-components/src/store/store.ts
ronros-elementor: But we don't have an ID in our unpublished components, so we cannot really do `component.id`...
Also - we immediately run an autosave upon creation, as anyway without having a valid id for the component, we cannot really enter its editing, nor expose any props from its nested widgets, so should we really support this?
(also - I'd need to add another selector, as this one is used in other selectors which are expecting only published components)

packages/packages/core/editor-components/src/prop-types/component-instance-prop-type.ts
ronros-elementor: I've originally added them on my next PR, but since you commented it, will move it to here... thanks for spoiling the surprise! 😄 (also - adding it with an additional intermediate prop type of the `overrides` (i.e. not hardcoded in the component-instance prop type)

MiryamOren (dismissed):

MiryamOren (approved):

packages/packages/core/editor-panels/src/components/external/panel-header.tsx
```diff
@@ -7,6 +7,7 @@ const Header = styled( Box )( ( { theme } ) => ( {
 	alignItems: 'center',
 	justifyContent: 'center',
 	gap: theme?.spacing( 0.5 ) || '4px',
+	paddingInline: theme?.spacing( 2 ) || '16px',
```
Ntnelbaba, line 10: @ronros-elementor This style broke Managers appearance, please revert it

PR #33800, opened 2025-12-11
Internal: Add the ability to expose overrides in component tool [ED-21958]

Omerisra6:
### ✨ PR Description
Purpose: Add overridable properties support to component creation tool enabling dynamic property customization in component instances through REST API validation and MCP integration.

Main changes:
- Added overridable_props parameter to component creation REST API with validation endpoint and schema parser
- Implemented document structure MCP resource exposing element hierarchy for AI-assisted component property selection
- Extended save-as-component MCP tool to enrich and validate overridable properties against element schemas

[ED-21958]: https://elementor.atlassian.net/browse/ED-21958?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/components/components-rest-api.php
```diff
@@ -435,13 +490,66 @@ private function archive_components( \WP_REST_Request $request ) {
 		}
 		return Response_Builder::make( $result )->build();
 	}
+
+	private function validate_components( \WP_REST_Request $request ) {
+		$items = Collection::make( $request->get_param( 'items' ) );
```
Omerisra6, line 501: Should we extract it and reuse it in the create and here

packages/packages/core/editor-canvas/src/mcp/resources/document-structure-resource.ts
```diff
@@ -0,0 +1,132 @@
...
+
+function getDocumentStructure() {
+	const extendedWindow = window as ExtendedWindow;
+	const document = extendedWindow.elementor?.documents?.getCurrent?.();
+
+	if ( ! document ) {
+		return { error: 'No active document found' };
+	}
+
+	function extractElementData( element: ElementorContainer ): Record< string, unknown > | null {
```
ronros-elementor: maybe extract to be declared outside the `getDocumentStructure`?

ronros-elementor (dismissed):

modules/components/components-rest-api.php
```diff
@@ -271,21 +316,32 @@ private function create_components( \WP_REST_Request $request ) {
...
+		foreach ( $items as $item ) {
 			$title = sanitize_text_field( $item['title'] );
 			$content = $item['elements'];
 			$uid = $item['uid'];
 
+			try {
+				$settings = isset( $item['settings'] ) ? $this->parse_settings( $item['settings'] ) : [];
+			} catch ( \Exception $e ) {
+				return Error_Builder::make( 'settings_validation_failed' )
+					->set_status( 400 )
```
ronros-elementor: can we maybe use a more semantic http response code for validation errors such as 422?

ronros-elementor: Also - do we want to immediately escape the process, or maybe instead - collect all failing components and return them all together?

By collecting all faulty components we could also keep the collection's `map_with_keys` as we don't need an early return in such case

packages/packages/core/editor-components/src/store/actions/create-unpublished-component.ts
```diff
@@ -5,16 +5,19 @@ import { generateUniqueId } from '@elementor/utils';
...
 import { trackComponentEvent } from '../../utils/tracking';
 import { slice } from '../store';
 
 export function createUnpublishedComponent(
 	name: string,
 	element: V1ElementData,
-	eventData: ComponentEventData | null
+	eventData: ComponentEventData | null,
+	overridableProps?: OverridableProps,
+	uid?: string | null
```
ronros-elementor, line 17: Maybe we could find a way to avoid the enforcement of an external uid, just to keep the flow consistent (e.g. our uids would have the 'compoent-' prefix and are formatted in a certain shape)

packages/packages/core/editor-components/src/api.ts
```diff
@@ -6,13 +6,18 @@ import { type DocumentSaveStatus, type OverridableProps, type PublishedComponent
 
 const BASE_URL = 'elementor/v1/components';
 
+export type ComponentItems = Array< {
+	uid: string;
+	title: string;
+	elements: V1ElementData[];
+	settings?: {
```
ronros-elementor, line 13: this shouldn't be in settings, as this configuration is stored as a post meta

packages/packages/core/editor-components/src/store/actions/create-unpublished-component.ts
Omerisra6: I dont think we should

packages/packages/core/editor-components/src/api.ts
Omerisra6: It should... : )

modules/components/components-rest-api.php
```diff
@@ -435,13 +496,72 @@ private function archive_components( \WP_REST_Request $request ) {
...
+		$result = Save_Components_Validator::make( $components )->validate( $items );
+
+		if ( ! $result['success'] ) {
+			return Error_Builder::make( 'components_validation_failed' )
+				->set_status( 422 )
+				->set_message( 'Validation failed: ' . implode( ', ', $result['messages'] ) )
+				->build();
+		}
+
+		$items->map_with_keys( function ( $item ) use ( &$validation_errors ) {
```
ronros-elementor: Do we need this `use`? looks like we're not really using `$validation_errors` here

packages/packages/core/editor-canvas/src/index.ts
```diff
@@ -14,6 +14,9 @@ export { settingsTransformersRegistry } from './settings-transformers-registry';
 export { styleTransformersRegistry } from './style-transformers-registry';
 export { endDragElementFromPanel, startDragElementFromPanel } from './sync/drag-element-from-panel';
 export { getCanvasIframeDocument } from './sync/get-canvas-iframe-document';
+export { DOCUMENT_STRUCTURE_URI } from './mcp/resources/document-structure-resource';
+export { WIDGET_SCHEMA_URI } from './mcp/resources/widgets-schema-resource';
```
ronros-elementor, line 18: purely out of curiosity, as it probably relates to stuff I'm not too familiar with - why is it exported? where will it be used outside of this package?

ronros-elementor (changes requested):

packages/packages/core/editor-canvas/src/index.ts
Omerisra6: We are using it in the components prompt to reference the uri of the resource and tell the agent to use it

ronros-elementor (approved):

PR #33815, opened 2025-12-14
Internal: [Prop BC] - Label Data with Version on save [ED-21937]

YaelAvitan:
### ✨ PR Description
Purpose: Add version tracking to prop type data structures by labeling all sanitized values and serialized schemas with version 1 for backward compatibility management.

Main changes:
- Added 'version' => 1 to sanitize() methods across all prop types (Array, Object, primitives)
- Injected version field into jsonSerialize() output for primitive prop types and Classes_Prop_Type
- Standardized version labeling in Array_Prop_Type and Object_Prop_Type schema definitions


modules/atomic-widgets/prop-types/base/array-prop-type.php
```diff
@@ -86,6 +86,7 @@ protected function validate_value( $value ): bool {
 
 	public function sanitize( $value ) {
 		$value['value'] = $this->sanitize_value( $value['value'] );
+		$value['version'] = 1;
```
asafdl: this hsouldn't sit here, we shouldn't add fields inside sanitize

modules/atomic-widgets/prop-types/primitives/boolean-prop-type.php
```diff
@@ -24,4 +24,17 @@ protected function validate_value( $value ): bool {
 	protected function sanitize_value( $value ) {
 		return (bool) $value;
 	}
+
+	public function sanitize( $value ) {
```
murzin-nikita: I think it should be part of some general infra, not per prop type definition

PR #33827, opened 2025-12-15
Internal: Support V4 dynamic tags in style color props [ED-21082]

RobiFerentz:
For VQA
Screenshot 2025-12-24 at 6 34 09

### ✨ PR Description
Purpose: Add support for V4 dynamic tags in style color properties by extending the dynamic tags mapping system to recognize and handle color prop types in style schemas.

Main changes:
- Added Color_Prop_Type recognition in Dynamic_Prop_Types_Mapping to map color properties to COLOR_CATEGORY dynamic tags
- Implemented get_extended_style_schema() method with recursive prop type inspection for nested color detection in unions and objects
- Registered dynamic tags module filter on atomic-widgets styles schema with priority 11 for proper filter chain execution
- Refactored update_font_family() to accept Union_Prop_Type in addition to String_Prop_Type for flexible property handling

### DIFF
--- a/modules/atomic-widgets/dynamic-tags/dynamic-tags-module.php
+++ b/modules/atomic-widgets/dynamic-tags/dynamic-tags-module.php
@@ -46,6 +46,13 @@ class Dynamic_Tags_Module {
 			fn( array $schema ) => Dynamic_Prop_Types_Mapping::make()->get_extended_schema( $schema )
 		);
 
+		add_filter(
+			'elementor/atomic-widgets/styles/schema',
+			fn( array $schema ) => Dynamic_Prop_Types_Mapping::make()->get_extended_style_schema( $schema ),
+			10,
+			2
+		);
+
 		add_action(
 			'elementor/atomic-widgets/settings/transformers/register',
 			fn ( $transformers, $prop_resolver ) => $this->register_transformers( $transformers, $prop_resolver ),
--- a/modules/atomic-widgets/prop-types/utils/prop-types-schema-extender.php
+++ b/modules/atomic-widgets/prop-types/utils/prop-types-schema-extender.php
@@ -31,20 +31,16 @@ abstract class Prop_Types_Schema_Extender {
 	}
 
 	public function get_extended_style_schema( array $schema ): array {
-		// right now, only colors support dynamic tags in styles, this list is root level keys that contain color props
-		// ED-21082
-		$allowed_keys = [
-			'backdrop-filter',
-			'background',
-			'border-color',
-			'box-shadow',
-			'color',
-			'filter',
-			'stroke',
-		];
+		$result = [];
 
 		foreach ( $schema as $key => $prop_type ) {
-			if ( ! in_array( $key, $allowed_keys ) ) {
+			if ( ! ( $prop_type instanceof Prop_Type ) ) {
+				$result[ $key ] = $prop_type;
+
+				continue;
+			}
+
+			if ( ! $this->contains_prop_type_instance( $prop_type, Color_Prop_Type::class ) ) {
 				$result[ $key ] = $prop_type;
 
 				continue;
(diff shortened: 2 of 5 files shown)


ronenboxer (dismissed):

modules/variables/hooks.php
```diff
@@ -72,7 +72,7 @@ private function register_styles_transformers() {
 	private function filter_for_style_schema() {
 		add_filter( 'elementor/atomic-widgets/styles/schema', function ( array $schema ) {
 			return ( new Style_Schema() )->augment( $schema );
-		} );
+		}, 11, 2 );
```
asafdl, line 75: probably change the calls to this filter instead of changing the overall filter from default, but why is the priority an issue?

modules/atomic-widgets/prop-types/utils/prop-types-schema-extender.php
```diff
@@ -29,6 +30,32 @@ public function get_extended_schema( array $schema ): array {
...
+			'background',
+			'border-color',
+			'box-shadow',
+			'color',
+			'filter',
+			'stroke',
+		];
+
+		foreach ( $schema as $key => $prop_type ) {
+			if ( ! in_array( $key, $allowed_keys ) ) {
```
asafdl: can we dynamicly investigate schema or prop_type to know if it has color? 
or at least add some variable to color prop type so it's identifieable

modules/variables/hooks.php
RobiFerentz: There are 'variables' related changes that that affect the dynamic tags changes if in the wrong order.

modules/atomic-widgets/prop-types/utils/prop-types-schema-extender.php
RobiFerentz: I tried doing that initially, but some color prop types are highly nested and do not seem get the expected prop type in php runtime, even when debugging (it did work for the non nested ones though).

modules/variables/classes/style-schema.php
```diff
@@ -53,9 +53,17 @@ private function update( $prop_type ) {
...
-					->add_prop_type( Font_Variable_Prop_Type::make() );
+	private function update_font_family( $prop_type ): Union_Prop_Type {
+		$result = null;
+		if ( $prop_type instanceof String_Prop_Type ) {
+			$result = Union_Prop_Type::create_from( $prop_type )
+				->add_prop_type( Font_Variable_Prop_Type::make() );
+		}
+		if ( $prop_type instanceof Union_Prop_Type ) {
+			$result = $prop_type->add_prop_type( Font_Variable_Prop_Type::make() );
+		}
```
ronenboxer, line 66: Empty lines between scopes dear

RobiFerentz: Fixed.

modules/atomic-widgets/prop-types/utils/prop-types-schema-extender.php
RobiFerentz: Tried another way, seems to be working.

asafdl (approved):

modules/variables/classes/style-schema.php
```diff
@@ -53,9 +53,19 @@ private function update( $prop_type ) {
...
+			$result = Union_Prop_Type::create_from( $prop_type )
+				->add_prop_type( Font_Variable_Prop_Type::make() );
+		}
+
+		if ( $prop_type instanceof Union_Prop_Type ) {
+			$result = $prop_type->add_prop_type( Font_Variable_Prop_Type::make() );
+		}
+
+		return $result;
 	}
```
mike-elementor, line 69:
```suggestion
if ( union ) {
  return union->add_prop_type;
}

if ( string ) {
  return new union;
}

return original;
```

modules/variables/hooks.php
mike-elementor, line 75: revert!

modules/atomic-widgets/dynamic-tags/dynamic-tags-module.php
```diff
@@ -46,6 +46,13 @@ public function register_hooks() {
 			fn( array $schema ) => Dynamic_Prop_Types_Mapping::make()->get_extended_schema( $schema )
 		);
 
+		add_filter(
+			'elementor/atomic-widgets/styles/schema',
+			fn( array $schema ) => Dynamic_Prop_Types_Mapping::make()->get_extended_style_schema( $schema ),
+			10,
```
mike-elementor, line 52: 6/8/9

PR #33922, opened 2025-12-21
Internal:  Component overridable props: change `propKey` to `path` [ED-22068]

MiryamOren:
### ✨ PR Description
Purpose: Refactor component overridable props data structure to support nested property paths, replacing single `propKey` with array-based `path` for hierarchical property access.

Main changes:
- Replaced `propKey` string field with `path` array containing key/type pairs in Component_Overridable_Prop class
- Added recursive path traversal logic in Parsing_Utils::get_prop_type_by_path() supporting Union, Object, and Array prop types
- Updated validation to check path structure and detect duplicate paths using concatenated element-path identifiers
- Modified all test mocks and fixtures to use path arrays instead of propKey strings


modules/components/overridable-props/overridable-props-parser.php
```diff
@@ -32,6 +32,7 @@ public function parse( array $props ): Parse_Result {
 		$validation_result = $this->validate( $parsed_props );
 
 		if ( ! $validation_result->is_valid() ) {
+			error_log( 'Validation result: ' . json_encode( $validation_result->errors() ) );
```
ronros-elementor, line 35: remove :)

modules/components/utils/parsing-utils.php
```diff
@@ -28,11 +31,39 @@ public static function get_prop_type( string $el_type, string $widget_type, stri
 
 		$props_schema = $element_instance->get_props_schema();
 
-		if ( ! isset( $props_schema[ $prop_key ] ) ) {
-			throw new \Exception( esc_html( "Prop key '$prop_key' does not exist in the schema of element '{$element_instance->get_element_type()}'." ) );
+		return self::get_prop_type_by_path( $props_schema, $path, $element_instance->get_element_type() );
+	}
+
+	private static function get_prop_type_by_path( array $schema, array $path, string $prev_path_items ): Prop_Type {
+		$first_path_item = $path[0];
```
ronros-elementor, line 38: Maybe a better name like `$current_segment`/`$current_item`

modules/components/utils/parsing-utils.php
```diff
@@ -28,11 +31,39 @@ public static function get_prop_type( string $el_type, string $widget_type, stri
 
 		$props_schema = $element_instance->get_props_schema();
 
-		if ( ! isset( $props_schema[ $prop_key ] ) ) {
-			throw new \Exception( esc_html( "Prop key '$prop_key' does not exist in the schema of element '{$element_instance->get_element_type()}'." ) );
+		return self::get_prop_type_by_path( $props_schema, $path, $element_instance->get_element_type() );
+	}
+
+	private static function get_prop_type_by_path( array $schema, array $path, string $prev_path_items ): Prop_Type {
```
ronros-elementor, line 37: We deal with `$prev_path_items` solely to print meaningful exceptions, so having it in the function args seems a bit too "noisy" for such purpose.

Maybe instead we could throw an exception, where we pass also the `$path` argument, catch it on `get_prop_type` level, and based on its length we could infer the path to print

e.g. if the original path is [1, 2, 3, 4], and the thrown path is [3, 4] - we'd print 1.2 + .3

Or to make it simpler, just pass the original path first, and after it the remaining path? (but it would probably be nullable, so we won't need to pass in the original path twice for the first invocation)

```suggestion
	private static function get_prop_type_by_path( array $schema, array $path, $current_path ): Prop_Type {
```

Bottom line - we're concatenating strings for something we may never use, and we are expected to have quite a lot of override props eventually (hopefully?)

ronros-elementor (changes requested):

PR #33926, opened 2025-12-21
Internal: [V4] inline editor marionette decoupling [ED-22037]

ronenboxer:


packages/packages/core/editor-canvas/src/legacy/replacements/base.ts
```diff
@@ -0,0 +1,44 @@
+import { type PropValue } from '@elementor/editor-props';
+
+export type ReplacementSettings = {
```
asafdl: move to replacements/types.ts

asafdl (dismissed):

asafdl (approved):

PR #33960, opened 2025-12-23
Internal: Prop-type migration interpreter [ED-21936]

murzin-nikita:
### ✨ PR Description
Purpose: Add PHP migration interpreter for prop-type transformations to enable runtime migration of atomic widget configuration schemas.

Main changes:
- Implemented Migration_Interpreter with set/delete/move operations supporting wildcard paths and conditional execution
- Added Path_Resolver utility for resolving wildcard patterns in nested data structures
- Created comprehensive test suite with 800+ lines covering operations, conditions, and wildcard correlation scenarios


asafdl:
Summary:
- Migration interpreter that interprets the [designed language](https://elementor.atlassian.net/wiki/spaces/UE/pages/1999110146/Prop+Types+Migration+Schema+Design+Investigation)
- path resolver helper
- tests

modules/atomic-widgets/prop-type-migrations/migration-interpreter.php
```diff
@@ -0,0 +1,523 @@
...
+			return $new_key;
+		}
+
+		$last_separator = max(
+			false === $last_dot ? -1 : $last_dot,
+			false === $last_bracket ? -1 : $last_bracket
+		);
+
+		if ( $last_separator === $last_bracket ) {
+			$close_bracket = strpos( $path, ']', $last_bracket );
```
marcin-el, line 337: probably an edge case but what if `$close_bracket` is `false`, maybe just a quick validation before using it in `strpos`?

modules/atomic-widgets/prop-type-migrations/migration-interpreter.php
```diff
@@ -0,0 +1,523 @@
...
+	exit;
+}
+
+class Migration_Interpreter {
+	public static function run( array $migration_schema, array $element_data, string $direction = 'up' ): array {
+		$operations = $migration_schema[ $direction ] ?? [];
+
+		if ( empty( $operations ) ) {
+			return $element_data;
+		}
```
davseve: Can this happen? can someone pass an empty string?
What are the available values the this function can receive in the `$direction` var? up and down? 
If so I would do something like:

```
    public static function run( array $migration_schema, array $element_data, string $direction = self::DIRECTION_UP ): array {

        if ( $direction !== self::DIRECTION_UP && $direction !== self::DIRECTION_DOWN ) {
			return $element_data;
		}
```

modules/atomic-widgets/prop-type-migrations/path-resolver.php
```diff
@@ -0,0 +1,214 @@
...
+			}
+		}
+
+		if ( '' !== $current ) {
+			$segments[] = self::create_segment( $current, 'key' );
+		}
+
+		return $segments;
+	}
+
```
davseve, line 82: I see this code block appears at least twice, can we extract it to a separate function? this way we also receive cleaner code by reducing nested conditions.

Maybe?

```suggestion
	private static function parse_path( string $path ): array {
	$segments = [];
	$current = '';
	$length = strlen( $path );

	for ( $i = 0; $i < $length; $i++ ) {
		$char = $path[ $i ];

		if ( '.' === $char ) {
			self::flush_current_segment( $segments, $current );
		} elseif ( '[' === $char ) {
			self::flush_current_segment( $segments, $current );

			$end = strpos( $path, ']', $i );

			if ( false === $end ) {
				break;
			}

			$index = substr( $path, $i + 1, $end - $i - 1 );
			$segments[] = self::create_segment( $index, 'index' );
			$i = $end;
		} else {
			$current .= $char;
		}
	}

	if ( '' !== $current ) {
		$segments[] = self::create_segment( $current, 'key' );
	}

	return $segments;
}

private static function flush_current_segment( array &$segments, string &$current ): void {
	if ( '' !== $current ) {
		$segments[] = self::create_segment( $current, 'key' );
		$current = '';
	}
}

```

modules/atomic-widgets/prop-type-migrations/path-resolver.php
```diff
@@ -0,0 +1,214 @@
...
+	private static function create_segment( string $value, string $type ): array {
+		return [
+			'value' => $value,
+			'type' => $type,
+			'is_wildcard' => '*' === $value,
+		];
+	}
+
+	private static function resolve_segments( array $segments, $data, array $wildcard_values, string $current_path = '', bool $allow_missing_leaf = true ): array {
```
davseve, line 95: I would clean this method a bit.
Maybe extract inner functionalities to separate methods.

modules/atomic-widgets/prop-type-migrations/migration-interpreter.php
```diff
@@ -0,0 +1,523 @@
...
+
+	private static function execute_operation( string $fn, array $op, string $path, array &$data ): void {
+		switch ( $fn ) {
+			case 'set':
+				self::execute_set( $op, $path, $data );
+				break;
+
+			case 'delete':
+				$clean = $op['clean'] ?? true;
+				self::delete_at_path( $data, $path, $clean );
```
davseve: Maybe
```suggestion
				self::execute_delete( $data, $path, $clean );
```
To maintain the consistency...

modules/atomic-widgets/prop-type-migrations/migration-interpreter.php
```diff
@@ -0,0 +1,523 @@
...
+				break;
+
+			case 'move':
+				self::execute_move( $op, $data );
+				break;
+		}
+	}
+
+	private static function execute_set( array $op, string $path, array &$data ): void {
+		$has_value = array_key_exists( 'value', $op );
```
davseve, line 99: We don't need to pass `$path` var in this method drill down. it is not used anywhere.
```suggestion
	private static function execute_set( array $op, array &$data ): void {
	    $path = $op['path'];
		$has_value = array_key_exists( 'value', $op );
```

modules/atomic-widgets/prop-type-migrations/migration-interpreter.php
```diff
@@ -0,0 +1,523 @@
...
+		$value = Path_Resolver::get( $resolved_path, $data );
+		$expected = $condition['value'] ?? null;
+
+		return self::evaluate_condition( $fn, $value, $expected );
+	}
+
+	private static function check_and_condition( array $condition, array $wildcard_values, array $data ): bool {
+		$conditions = $condition['conditions'] ?? [];
+
+		foreach ( $conditions as $cond ) {
```
davseve:
```suggestion
		foreach ( $conditions as $condition ) {
```
or

```suggestion
		foreach ( $conds as $cond ) {
```

tests/phpunit/elementor/modules/atomic-widgets/prop-type-migrations/test-migration-interpreter.php
```diff
@@ -0,0 +1,807 @@
...
+		// Act
+		$result = Migration_Interpreter::run( $migration, $data );
+
+		// Assert
+		$this->assertTrue( $result['settings']['link']['processed'] );
+		$this->assertTrue( $result['settings']['button']['processed'] );
+		$this->assertArrayNotHasKey( 'processed', $result['settings']['text'] );
+	}
+
+	public function test_interpret_down_direction() {
```
davseve, line 734: IS this the only test that checks the down direction?
Do we want to increase the downgrade test coverage?

modules/atomic-widgets/prop-type-migrations/migration-interpreter.php
```diff
@@ -0,0 +1,523 @@
...
+			}
+		}
+
+		return true;
+	}
+
+	private static function check_or_condition( array $condition, array $wildcard_values, array $data ): bool {
+		$conditions = $condition['conditions'] ?? [];
+
+		foreach ( $conditions as $cond ) {
```
davseve: Same as above

modules/atomic-widgets/prop-type-migrations/migration-interpreter.php
```diff
@@ -0,0 +1,523 @@
...
+		$segments = self::parse_path_segments( $path );
+
+		if ( empty( $segments ) ) {
+			return;
+		}
+
+		self::rename_key_recursive( $data, $segments, $new_key );
+	}
+
+	private static function parse_path_segments( string $path ): array {
```
davseve, line 379: I would break this method to smaller independent methods. 
I think it has to many nested conditions

modules/atomic-widgets/prop-type-migrations/migration-interpreter.php
asafdl: its just input validation, it can happen if someone misuses the function, but in practical sense shouldn't happen.
fixed direction input validation

modules/atomic-widgets/prop-type-migrations/migration-interpreter.php
asafdl: tiny optimization, we already extract path from before, id rather leave it as is, its a large refactor to change

modules/atomic-widgets/prop-type-migrations/migration-interpreter.php
asafdl: done

modules/atomic-widgets/prop-type-migrations/path-resolver.php
asafdl: done

modules/atomic-widgets/prop-type-migrations/path-resolver.php
asafdl: done

davseve (dismissed):

davseve (approved):

PR #33975, opened 2025-12-25
Fix: V4 Inline editor in panel should not lose focus when typing [ED-22125]

RobiFerentz:


ronenboxer (approved):

tests/playwright/sanity/modules/v4-tests/inline-text-editing/editor-in-panel.test.ts
```diff
@@ -0,0 +1,55 @@
...
+		const contentSection = page.getByLabel( INLINE_EDITING_SELECTORS.contentSection );
+		const panelInlineEditor = contentSection.locator( INLINE_EDITING_SELECTORS.panelInlineEditor );
+
+		await expect( panelInlineEditor ).toBeVisible();
+		await panelInlineEditor.clear();
+
+		// Act & Assert
+		for ( const letter of TEST_WORD ) {
+			await page.keyboard.type( letter );
+			// Wait for the letter to be typed and the focus to be maintained after value change propagation.
```
davseve, line 47: PLease remove the comment.
Have we tried to use playwright [delay](https://playwright.dev/docs/api/class-keyboard#keyboard-press-option-delay)

```await page.keyboard.type(TEST_WORD, { delay: 250 });```
instead of the loop above?

davseve (approved):

PR #34000, opened 2025-12-29
Fix: Site settings can't be changed [ED-22081]

Omerisra6:
### ✨ PR Description
Purpose: Fix null reference error in site settings save flow by adding null-safe property access to container model elements retrieval.

Main changes:
- Added optional chaining to container?.model.get() to prevent TypeError when container is undefined
- Added optional chaining to toJSON?.() method call to handle cases where elements may not have toJSON method
- Prevents site settings save failures caused by attempting to access properties on undefined container object

### DIFF
--- a/packages/packages/core/editor-components/src/sync/before-save.ts
+++ b/packages/packages/core/editor-components/src/sync/before-save.ts
@@ -20,7 +20,7 @@ type Options = {
 };
 
 export const beforeSave = ( { container, status }: Options ) => {
-	const elements = container?.model?.get?.( 'elements' )?.toJSON() ?? [];
+	const elements = container?.model.get( 'elements' ).toJSON?.() ?? [];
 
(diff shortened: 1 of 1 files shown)


packages/packages/core/editor-components/src/sync/before-save.ts
```diff
@@ -20,7 +20,7 @@ type Options = {
 };
 
 export const beforeSave = ( { container, status }: Options ) => {
-	const elements = container.model.get( 'elements' )?.toJSON() ?? [];
+	const elements = container?.model?.get?.( 'elements' )?.toJSON() ?? [];
```
IshayMaya:
```suggestion
	const elements = container?.model.get( 'elements' )?.toJSON() ?? [];
```

IshayMaya (approved):

PR #34017, opened 2025-12-30
Internal: Restrictions for non-pro users [ED-21221]

RobiFerentz:


packages/packages/core/editor-editing-panel/src/controls-registry/settings-field.tsx
```diff
@@ -46,7 +47,17 @@ export const SettingsField = ( { bind, children, propDisplayName }: SettingsFiel
 		return migratePropValues( elementSettingValues, propsSchema );
 	}, [ elementSettingValues, propsSchema ] );
 
-	const value = { [ bind ]: migratedValues?.[ bind ] ?? null };
+	const cleanedValue = useMemo( () => {
```
asafdl: i have a problem with the code being placed in settings-field, i need a bit of time to consider alternatives

RobiFerentz: Where would you move this? Another hook maybe?

modules/atomic-widgets/props-resolver/render-props-resolver.php
```diff
@@ -39,8 +57,10 @@ public function resolve( array $schema, array $props ): array {
 				continue;
 			}
 
+			$prop_value = $props[ $key ] ?? null;
+
 			$transformed = $this->resolve_item(
-				$props[ $key ] ?? $prop_type->get_default(),
+				$this->should_revert_to_default( $prop_value ) ? $prop_type->get_default() : $prop_value,
```
ronenboxer: How do you feel about this? Just to make a little bit more readable?

```			$prop_value = $props[ $key ] ?? null;
			$prop_value = $this->should_revert_to_default( $prop_value ) ? $prop_type->get_default() : $prop_value;

			$transformed = $this->resolve_item( $prop_value,```

File: packages/packages/core/editor-editing-panel/src/controls-registry/settings-field.tsx (hunk 1 above)
Comment (ronenboxer) [inline packages/packages/core/editor-editing-panel/src/controls-registry/settings-field.tsx]: normalizedValue?
cleaned suggest something different than supported/unsupported

Comment (ronenboxer) [reply packages/packages/core/editor-editing-panel/src/controls-registry/settings-field.tsx]: I agree with the hook suggestion

File: packages/packages/core/editor-editing-panel/src/dynamics/hooks/use-prop-dynamic-action.tsx [hunk 3]
```diff
@@ -10,7 +10,7 @@ import { supportsDynamic } from '../utils';
 export const usePropDynamicAction = (): PopoverActionProps => {
 	const { propType } = useBoundProp();
 
-	const visible = !! propType && supportsDynamic( propType );
+	const visible = !! window.elementorPro && !! propType && supportsDynamic( propType );
```
ronenboxer, line 13: I think this is problematic
IMO, this should be a filter hook, with !! window.elementorPro as  default value, used by the pro, and in 2 versions have the window.elementorPro removed and replaced with `false`

```window.elementor?.hooks?.filter('blahblah',!! window.elementorPro) ?? !! window.elementorPro```

File: packages/packages/libs/editor-props/src/utils/is-supported-dynamic-tag.ts [hunk 4]
```diff
@@ -0,0 +1,28 @@
...
+	const val = value as TransformablePropValue< 'dynamic', { name: string } >;
+	return !! extendedWindow.elementor?.config?.atomicDynamicTags?.tags[ val?.value?.name ];
+};
+
+export const isDynamicValueButUnsupportedTag = ( value: PropValue ): boolean => {
+	if ( ! value ) {
+		return false;
+	}
+	const val = value as TransformablePropValue< 'dynamic', { name: string } >;
+	return val.$$type === 'dynamic' && ! isSupportedDynamicTag( val );
```
ronenboxer: MT lines please

packages/packages/libs/editor-props/src/utils/is-supported-dynamic-tag.ts
```diff
@@ -0,0 +1,28 @@
...
+	}
+	const val = value as TransformablePropValue< 'dynamic', { name: string } >;
+	return !! extendedWindow.elementor?.config?.atomicDynamicTags?.tags[ val?.value?.name ];
+};
+
+export const isDynamicValueButUnsupportedTag = ( value: PropValue ): boolean => {
+	if ( ! value ) {
+		return false;
+	}
+	const val = value as TransformablePropValue< 'dynamic', { name: string } >;
```
ronenboxer: We generally avoid shortening of names, and val is meaningless in this context.
Since you're not changing the value, you can just do `value = value as....`

packages/packages/core/editor-editing-panel/src/dynamics/hooks/use-prop-dynamic-action.tsx
RobiFerentz: Not sure how this is problematic? If `elementorPro' does not exist, that means pro wasnt loaded, how is this different asking a hook that will never be called if pro isn't there?

packages/packages/libs/editor-props/src/utils/is-supported-dynamic-tag.ts
RobiFerentz: Fixed.

packages/packages/libs/editor-props/src/utils/is-supported-dynamic-tag.ts
RobiFerentz: Fixed.

packages/packages/core/editor-editing-panel/src/controls-registry/settings-field.tsx
RobiFerentz: Fixed.

modules/atomic-widgets/props-resolver/render-props-resolver.php
RobiFerentz: Fixed.

PR #34043, opened 2025-12-31
Internal: Update component error message texts [ED-22223]

Omerisra6:
## Summary

Updates the error message texts for component-related errors as specified in [ED-22223](https://elementor.atlassian.net/browse/ED-22223).

### 1. Circular Nesting Error
**Scenario**: User attempts to nest a component that would create circular reference (e.g., Comp A → Comp B → trying to add Comp A to Comp B)

- **Before**: "Cannot add this component here - it would create a circular reference."
- **After**: "Can't add this component - components that contain each other can't be nested."

### 2. Non-Atomic Element in Edit Mode
**Scenario**: User drops/adds a v3 (non-atomic) widget while editing a component

- **Before**: "Cannot add this element here - only atomic elements are allowed inside components."
- **After**: "This Widget isn't compatible with components. Use Atomic elements instead."

### 3. Composition with v3 Elements (Save as Component)
**Scenario**: User tries to save an element as a component when it contains non-atomic widgets

- **Before**: "Cannot save as component - contains non-supported elements. Only atomic elements are allowed inside components."
- **After**: "Components require Atomic elements only. Remove Widgets to create this component."

## Files Changed
- `packages/packages/core/editor-components/src/prevent-circular-nesting.ts`
- `packages/packages/core/editor-components/src/prevent-non-atomic-nesting.ts`
- `packages/packages/core/editor-components/src/components/create-component-form/create-component-form.tsx`

## Testing
- All existing tests pass ✅
- Text-only changes - no logic modifications

[ED-22223]: https://elementor.atlassian.net/browse/ED-22223?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### ✨ PR Description
Purpose: Update component validation error messages across REST API and editor to improve user clarity and provide structured error metadata for better debugging.

Main changes:
- Replaced technical concatenated error messages with user-friendly translatable text using __() function
- Added structured metadata fields (caused_by, non_atomic_elements) to API error responses for client consumption
- Standardized error messaging across PHP backend and TypeScript frontend validation handlers

### DIFF
--- a/packages/packages/core/editor-components/src/prevent-non-atomic-nesting.ts
+++ b/packages/packages/core/editor-components/src/prevent-non-atomic-nesting.ts
@@ -42,7 +42,7 @@ type StorageContent = {
 
 const NON_ATOMIC_ELEMENT_ALERT: NotificationData = {
 	type: 'default',
-	message: __( 'Cannot add this element here - only atomic elements are allowed inside components.', 'elementor' ),
+	message: __( "This Widget isn't compatible with components. Use Atomic elements instead.", 'elementor' ),
 	id: 'non-atomic-element-blocked',
 };
 
--- a/packages/packages/core/editor-components/src/prevent-non-atomic-nesting.ts
+++ b/packages/packages/core/editor-components/src/prevent-non-atomic-nesting.ts
@@ -42,7 +42,7 @@ type StorageContent = {
 
 const NON_ATOMIC_ELEMENT_ALERT: NotificationData = {
 	type: 'default',
-	message: __( "This Widget isn't compatible with components. Use Atomic elements instead.", 'elementor' ),
+	message: __( "This widget isn't compatible with components. Use atomic elements instead.", 'elementor' ),
 	id: 'non-atomic-element-blocked',
 };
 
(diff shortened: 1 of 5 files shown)


packages/packages/core/editor-components/src/prevent-non-atomic-nesting.ts
```diff
@@ -42,7 +42,7 @@ type StorageContent = {
 
 const NON_ATOMIC_ELEMENT_ALERT: NotificationData = {
 	type: 'default',
-	message: __( 'Cannot add this element here - only atomic elements are allowed inside components.', 'elementor' ),
+	message: __( "This Widget isn't compatible with components. Use Atomic elements instead.", 'elementor' ),
```
IshayMaya: i think it should be something like this, we don't use multiple uppercase in a sentence

This Widget -> This widget
Use Atomic -> Use atomic (maybe even use atomic idk)

IshayMaya (dismissed):

IshayMaya (approved):

PR #34050, opened 2025-12-31
Internal: PropType Migration Orchestrator [ED-21944]

asafdl:
### ✨ PR Description
Purpose: Add PropType Migration Orchestrator to automatically migrate atomic widget prop types and keys during document rendering to maintain backward compatibility with evolving widget schemas.

Main changes:
- Implemented Migrations_Orchestrator and Migrations_Loader to handle automatic prop type migrations using manifest-driven migration paths
- Added document-level migration hook in get_elements_raw_data() that executes before editor rendering when BC_MIGRATIONS experiment enabled
- Created Logger utility and comprehensive test coverage for migration interpreter error handling and orchestrator workflows


RobiFerentz (dismissed):

RobiFerentz (commented):
Multiple failing CI and lint, otherwise seems ok.

RobiFerentz (commented):
Multiple failing CI and lint, otherwise seems ok.

RobiFerentz (approved):
Is there some mechanism to avoid running migrations concurrently for the same document? For example: In case of two users accessing the same post in editor / frontend ?

Otherwise, LGTM

core/base/document.php
```diff
@@ -2091,4 +2097,30 @@ private function get_elements_iteration_actions() {
 
 		return $this->elements_iteration_actions;
 	}
+
+	private function migrate_atomic_widgets_data( array &$data ): void {
```
davseve, line 2101: Why do we use the reference `&`  I don't see that the $data is modified inside the function.

modules/atomic-widgets/prop-type-migrations/migrations-orchestrator.php
```diff
@@ -0,0 +1,313 @@
...
+
+	public static function destroy(): void {
+		self::$instance = null;
+	}
+
+	public function migrate_document(
+		array &$elements_data,
+		int $post_id,
+		callable $save_callback
+	): void {
```
davseve, line 40: We usually write this way:
```suggestion
	public function migrate_document( array &$elements_data, int $post_id, callable $save_callback ): void {
```

modules/atomic-widgets/prop-type-migrations/migrations-orchestrator.php
```diff
@@ -0,0 +1,313 @@
...
+		}
+
+		return self::$instance;
+	}
+
+	public static function destroy(): void {
+		self::$instance = null;
+	}
+
+	public function migrate_document(
```
davseve, line 36: I think that this method can be improved, it has 6 levels of nesting.
Maybe refactor it to smaller methods with names that describe what each method does.

core/base/document.php
```diff
@@ -1062,6 +1063,11 @@ public function get_elements_raw_data( $data = null, $with_html_content = false
 			$data = $this->get_elements_data();
 		}
 
+		// Migrate atomic widgets prop types
+		if ( ! $this->is_saving ) {
+			$this->migrate_atomic_widgets_data( $data );
```
davseve, line 1068: As we spoke before, you mentioned that the migration will be executed on page load.
Can we measure how much time this process will add to the page load?
This is some we need to inform product about.

core/base/document.php
```diff
@@ -1062,6 +1063,11 @@ public function get_elements_raw_data( $data = null, $with_html_content = false
 			$data = $this->get_elements_data();
 		}
 
+		// Migrate atomic widgets prop types
```
Omerisra6, line 1066: Maybe we can add a hook here and move that part to the atomic widget module, as its always good to have clear boundaries between the core codebase and other modules

modules/atomic-widgets/prop-type-migrations/migration-interpreter.php
Omerisra6, line 77: Do we have other security concerns that we need to keep in mind?
Like sanitizing the input somehow? (not sure if its relevant)

modules/atomic-widgets/prop-type-migrations/migrations-loader.php
```diff
@@ -0,0 +1,283 @@
+<?php
```
Omerisra6, line 1: Maybe we can add two separate classes for `Atomic_Widget_Migration` and `Prop_Type_Migration`?

modules/atomic-widgets/prop-type-migrations/migrations-orchestrator.php
```diff
@@ -0,0 +1,313 @@
...
+use Elementor\Modules\AtomicWidgets\Logger\Logger;
+use Elementor\Modules\AtomicWidgets\PropTypes\Contracts\Prop_Type;
+use Elementor\Modules\GlobalClasses\Utils\Atomic_Elements_Utils;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+class Migrations_Orchestrator {
+	private const MIGRATED_VERSION_META_KEY = '_elementor_data_version';
```
Omerisra6, line 14: Do we have one key per version?

What about multiple migrations on the same prop type/widget on the same version?

modules/atomic-widgets/prop-type-migrations/migrations-orchestrator.php
```diff
@@ -0,0 +1,313 @@
...
+			}
+
+			$has_changes = false;
+
+			foreach ( $elements_data as &$element ) {
+				if ( ! $this->is_atomic_widget( $element ) ) {
+					continue;
+				}
+
+				$widget_type = $element['widgetType'] ?? '';
```
Omerisra6, line 53: Shouldn't we support atomic elements as well? 

For example, div block has no widgetType set

tests/phpunit/elementor/modules/atomic-widgets/prop-type-migrations/fixtures/migrations/a-to-b.json
```diff
@@ -0,0 +1,20 @@
...
+  "down": [
+    {
+      "op": {
+        "fn": "set",
+        "path": "$$type",
+        "value": "type_a"
+      }
+    }
+  ]
+}
```
Omerisra6, line 20: Where will developers write the migrations?

In the codebase?

If not, how would they test it?

tests/phpunit/elementor/modules/atomic-widgets/prop-type-migrations/fixtures/migrations/a-to-b.json
Omerisra6, line 1: I think you should create a schema builder class, even if it's just for testing.

modules/atomic-widgets/prop-type-migrations/migrations-loader.php
asafdl: prop-type migration will be deleted next pr

modules/atomic-widgets/prop-type-migrations/migrations-orchestrator.php
asafdl: fixed in later pr, key does need to be more complex

tests/phpunit/elementor/modules/atomic-widgets/prop-type-migrations/fixtures/migrations/a-to-b.json
asafdl: they write it in codebase, next pr has first migrations written

modules/atomic-widgets/prop-type-migrations/migration-interpreter.php
asafdl: don't think so, developer creates a migration, we don't validate it other than code review, if it malforms the data... its malformed, just like a transformer can malform it

core/base/document.php
asafdl: we can

asafdl:
> Is there some mechanism to avoid running migrations concurrently for the same document? For example: In case of two users accessing the same post in editor / frontend ?
> 
> Otherwise, LGTM

we can't stop concurrent migrations (or at least it would be pretty complex to do so). however migrations are idempotent so running the same migration twice doesn't hurt anybody since they result in the same.
once migration has run and it's in cache, no more migrations will run and this problem solves itself

modules/atomic-widgets/prop-type-migrations/migrations-orchestrator.php
```diff
@@ -0,0 +1,313 @@
...
+			$this->mark_as_migrated( $post_id );
+		} catch ( \Exception $e ) {
+			Logger::warning( 'Document migration failed', [
+				'post_id' => $post_id,
+				'error' => $e->getMessage(),
+			] );
+		}
+	}
+
+	private function is_already_migrated( int $post_id ): bool {
```
davseve, line 100: Maybe
```suggestion
	private function is_migrated( int $post_id ): bool {
```

modules/atomic-widgets/prop-type-migrations/migrations-orchestrator.php
```diff
@@ -0,0 +1,313 @@
...
+			'has_changes' => $has_changes,
+		];
+	}
+
+	private function process_setting_key(
+		string $key,
+		$value,
+		array $schema,
+		string $widget_type,
+		array &$migrated_settings,
```
davseve, line 170: 1. I think that there are too many variable in this function
2. I am not sure about the extensive usage of references ini this PR. it may save us from making a new variable but eventually it can overload the memory management and cause side effects.

core/base/document.php
asafdl: since we are using cache, adding some latency to first page load, is unavoidable but probably not a big deal, since it only needs to happen once

core/base/document.php
asafdl: migrate document modifies the data

asafdl:
@davseve @Omerisra6 thanks for the comments, since i already branched out from this branch, they have all been addressed in next PR

PR #34052, opened 2025-12-31
Internal: Entering component edit mode improvements [ED-22208]

MiryamOren:
When entering component edit mode: 

1. The top-level container should be selected 

2. The navigator should be opened and all items expanded

### ✨ PR Description
Purpose: Enhance component edit mode navigation by adding auto-expand functionality and improving element selection when entering embedded document editing context.

Main changes:
- Added ExpandAll command to navigator with conditional logic for opening/expanding based on navigator state
- Enhanced navigator open() method to accept expandAllElements option for automatic tree expansion
- Modified switchToComponent() to auto-select top-level element and expand navigator when entering edit mode
- Added CSS rules to display element overlays in embedded documents with elementor-edit-area-active class

### DIFF
--- a/assets/dev/js/editor/regions/navigator/navigator.js
+++ b/assets/dev/js/editor/regions/navigator/navigator.js
@@ -96,7 +96,7 @@ export default class extends BaseRegion {
 		this.$el.resizable( this.getResizableOptions() );
 	}
 
-	open( model, options = { expandAllElements: false } ) {
+	open( model, options ) {
 		this.$el.show();
 
 		this.setSize();
@@ -109,7 +109,7 @@ export default class extends BaseRegion {
 			model.trigger( 'request:edit' );
 		}
 
-		if ( options.expandAllElements ) {
+		if ( options?.expandAllElements ) {
 			this.currentView.elements.currentView.recursiveChildInvoke( 'toggleList', true );
 		}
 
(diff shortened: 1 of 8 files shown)


packages/packages/core/editor-components/src/utils/switch-to-component.ts
```diff
@@ -1,19 +1,30 @@
...
 		mode: 'autosave',
 		setAsInitial: false,
 		shouldScroll: false,
 	} );
+
+	const currentDocumentContainer = getCurrentDocumentContainer();
+	const topLevelElement = currentDocumentContainer?.children?.[ 0 ];
+
+	if ( topLevelElement ) {
+		await selectElement( topLevelElement.id );
```
IshayMaya:
```suggestion
	selectElement( topLevelElement.id );
```
unless we do need to wait for it and then change the selectElement function

packages/packages/core/editor-components/src/utils/switch-to-component.ts
```diff
@@ -1,19 +1,30 @@
...
+		await expandNavigator();
```
IshayMaya: do we need the await?

assets/dev/js/editor/regions/navigator/navigator.js
```diff
@@ -96,7 +96,7 @@ export default class extends BaseRegion {
 		this.$el.resizable( this.getResizableOptions() );
 	}
 
-	open( model ) {
+	open( model, expandAllElements = false ) {
```
IshayMaya: maybe we should pass as an object? since we are not in ts it will be safer in case someone is passing any other param that will be interpreted as "true"

MiryamOren: I don't want to break the current signature, since I'm not sure who else is calling this function

IshayMaya (approved):

PR #34056, opened 2026-01-01
Fix: Overridable/override prop values stored in wrong docs [ED-22210]

ronros-elementor:
### ✨ PR Description
Purpose: Fix overridable and override prop values to retrieve from correct document version (autosave vs published) preventing data inconsistency in component editing workflow.

Main changes:
- Modified Components_Repository.get() to accept include_autosave parameter and retrieve autosave version when in preview mode
- Updated REST API to use get_json_meta() instead of get_meta() for retrieving overridable props metadata
- Refactored Component_Override_Parser and Component_Instance_Transformer to use Components_Repository with autosave support instead of direct document access
- Added conditional overridable prop updates in OverridablePropControl to prevent updates during component instance editing

[ED-22210]: https://elementor.atlassian.net/browse/ED-22210?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/components/components-repository.php
```diff
@@ -46,8 +46,12 @@ public function all() {
...
 	}
 
-	public function get( $id ) {
-		$doc = Plugin::$instance->documents->get( $id );
+	public function get( $id, bool $include_autosave = false ) {
+		$should_get_autosave = $include_autosave || is_preview();
+
+		$doc = $should_get_autosave
+			? Plugin::$instance->documents->get_doc_or_auto_save( $id, get_current_user_id() )
+			: Plugin::$instance->documents->get( $id );
```
Omerisra6, line 54: What does doc for autosave mean? Maybe find a better name?

does this returns the preview?

modules/components/prop-types/component-override-parser.php
```diff
@@ -106,15 +108,22 @@ private function get_component_overridable_props( int $component_id ) {
 			return $this->component_overridable_props;
 		}
 
-		$component = Plugin::$instance->documents->get( $component_id );
+		$component = $this->get_repository()->get( $component_id, true );
```
Omerisra6, line 111: why do we need the preview document here?

packages/packages/core/editor-components/src/components/overridable-props/overridable-prop-control.tsx
```diff
@@ -41,7 +41,11 @@ export function OverridablePropControl< T extends object >( {
 		} as ComponentOverridablePropValue;
 
 		setValue( propValue );
-		updateOverridableProp( componentId, propValue );
+
+		if ( ! isComponentInstance ) {
+			const existingProp = overridableProps?.props?.[ propValue.override_key ];
+			updateOverridableProp( componentId, propValue, existingProp?.originPropFields );
+		}
```
Omerisra6, line 48: is that related to the PR?

modules/components/components-repository.php
ronros-elementor: the get_doc_or_auto_save is something we always had in the documents manager, and basically it should retrieve the draft/autosave revision for the current user, is such exists and is newer of the publish revision

modules/components/prop-types/component-override-parser.php
ronros-elementor: because parsing occurs upon save, and when I save - I expect it to take into account the up to date state of all used components as well (for any used component instances)

part of the bug was that validation occured against the published version, which has changed in the current session - so I would add overrides to my nested component's draft version, then override from the post itself (which is of course in draft as well, sinec I'm in the editor), and then once publishing - it would validate my overrides against the published version of my component, which still didn't get updated - and then it would remove my override in sanitization, or fail upon validation... either way - it would be wrong

packages/packages/core/editor-components/src/components/overridable-props/overridable-prop-control.tsx
ronros-elementor: It was part of the mitigation, to avoid calling this twice, because this is also wrong (though eventually it didn't cause the actual bug)

What used to happen before that is upon each change from an overridable **override** (confusing much?) was that it invoked once this call, without the originPropFields, and then get called again externally from its wrapping OverridePropControl

modules/components/components-repository.php
Omerisra6: so it's the document's most updated version for the specific user?

Omerisra6 (approved):

PR #34078, opened 2026-01-05
Internal: [V4] Allow some action links in the editor [ED-22096]

ronenboxer:


includes/base/element-base.php
```diff
@@ -770,6 +770,7 @@ protected function add_render_attributes() {
 			],
 			'data-id' => $id,
 			'data-element_type' => $this->get_type(),
+			'data-e-type' => $this->get_type(),
```
asafdl, line 773: what?

modules/atomic-widgets/assets/js/frontend/handlers.js
```diff
@@ -1,38 +1,69 @@
...
+		: element.querySelector( ACTION_LINK_SELECTOR );
+	const url = actionLinkElement?.dataset.actionLink;
 
 	if ( ! url ) {
 		return;
 	}
 
 	const handler = ( event ) => {
-		if ( isEditorContext ) {
+		if ( actionLinkElement && actionLinkElement !== element && ! actionLinkElement.contains( event.target ) ) {
```
asafdl, line 25: can you reach this code if you are not actionLinkElement? don't we set this handler for action link elements?

modules/atomic-widgets/assets/js/frontend/handlers.js
```diff
@@ -1,38 +1,69 @@
...
+	url = decodeURIComponent( url );
+
+	const actionMatch = url.match( /action=([^&]+)/ );
+	const action = actionMatch?.[ 1 ] ?? null;
+
+	if ( ! action ) {
+		return false;
+	}
+
+	const whitelist = elementorFrontend?.hooks?.applyFilters( WHITELIST_FILTER, LINK_ACTIONS_EDITOR_WHITELIST ) ??
```
asafdl, line 61: we discussed that instead of adding complex whitelisting. we just allow internal links to work, and external links to not work.
is this not possible?

modules/atomic-widgets/assets/js/frontend/handlers.js
```diff
@@ -1,38 +1,69 @@
...
+
+function shouldFireLinkActionHandler( url ) {
+	if ( ! isEditorContext() ) {
+		return true;
+	}
+
+	url = decodeURI( url );
+	url = decodeURIComponent( url );
+
+	const actionMatch = url.match( /action=([^&]+)/ );
```
asafdl, line 54: you exit earlier when this isn't an actionLinkElement
then why do you check if it has action?

modules/atomic-widgets/assets/js/frontend/handlers.js
ronenboxer: Not all internal links should work
Only canvas and lightbox

modules/atomic-widgets/assets/js/frontend/handlers.js
ronenboxer: this is for title and paragraph widget handling, which don't have their tags replaced, but have the link element rendered within them

includes/base/element-base.php
ronenboxer: This is the attribute that 'frontend-handlers` package looks for on init

modules/atomic-widgets/assets/js/frontend/handlers.js
ronenboxer: `actionLinkElement` is an element with an attribute of `data-action-link`
Here, I'm checking what the action actually is

modules/atomic-widgets/assets/js/frontend/handlers.js
asafdl: which internal links shouldn't work? do we care if they work or not?

includes/base/element-base.php
asafdl: why not just use data-element-type?

modules/atomic-widgets/assets/js/frontend/handlers.js
asafdl: im still a bit confused about that. we have selector for data-link-action attribute, what exactly are we checking for here?

davseve (approved):

PR #34080, opened 2026-01-05
Internal: Enter edit mode after creating a new component

MiryamOren:
### ✨ PR Description
Purpose: Enable automatic edit mode entry after component creation to streamline component editing workflow and improve user experience.

Main changes:
- Converted component creation functions to async/await to handle edit mode transition after component save
- Added switchToComponent call with published component ID and instance ID after successful creation
- Updated replaceElement to return created component instance for accessing instance ID in workflow
- Modified CSS to remove unnecessary add-section overlay display rules in embedded document edit mode


assets/dev/scss/editor/preview/preview.scss
```diff
@@ -486,10 +486,9 @@ html.elementor-html {
 	:where(.elementor-edit-area-active) {
 		.elementor-element-overlay,
 		.elementor-empty,
-		.elementor-add-section,
-		.elementor-add-section-inline,
```
Omerisra6, line 490: Is that something new we added?

If not, are you sure it won't break current behavior?

packages/packages/core/editor-components/src/store/actions/create-unpublished-component.ts
```diff
@@ -37,7 +38,14 @@ export function createUnpublishedComponent(
...
 
-	runCommand( 'document/save/auto' );
+	await runCommand( 'document/save/auto' );
 
+	const publishedComponentId = ( selectComponentByUid( getState(), generatedUid ) as PublishedComponent )?.id;
+
+	if ( publishedComponentId ) {
+		switchToComponent( publishedComponentId, componentInstance.id );
+	} else {
+		throw new Error( 'Failed to find published component' );
```
Omerisra6: For me it makes more sense to make it as a separate functionality:
- create
- select

Then, if we have a situation where we will need to create without selecting, it will be easier.

assets/dev/scss/editor/preview/preview.scss
MiryamOren: It's something I added in the last task, and now I noticed these lines are incorrect

packages/packages/core/editor-components/src/store/actions/create-unpublished-component.ts
MiryamOren: Done

Omerisra6 (approved):

PR #34095, opened 2026-01-06
Internal: Prevent interactions with widgets inside non-edited components [ED-22199]

ronros-elementor:
### ✨ PR Description
Purpose: Prevent unintended interactions with widgets inside non-edited atomic components by conditionally disabling pointer events based on edit state.
Main changes:
- Added CSS rule for `.elementor-widget-e-component[data-atomic]` to disable pointer-events on child widgets when component is not in edit mode
- Enabled pointer-events restoration for nested elements only when `.elementor-edit-area-active` class is present on the component

### DIFF
--- a/assets/dev/scss/editor/preview/preview.scss
+++ b/assets/dev/scss/editor/preview/preview.scss
@@ -556,7 +556,7 @@ html.elementor-html {
 .elementor-widget-e-component[data-atomic] {
 	&.elementor-edit-area-active {
 		& > * > * {
-			pointer-events: revert;
+			pointer-events: initial;
 		}
 	}
 
--- a/assets/dev/scss/editor/preview/preview.scss
+++ b/assets/dev/scss/editor/preview/preview.scss
@@ -552,6 +552,21 @@ html.elementor-html {
 	pointer-events: none;
 }
 
+
+.elementor-widget-e-component[data-atomic] {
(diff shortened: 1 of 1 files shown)


assets/dev/scss/editor/preview/preview.scss
```diff
@@ -552,6 +552,21 @@ html.elementor-html {
 	pointer-events: none;
 }
 
+
+.elementor-widget-e-component[data-atomic] {
```
Omerisra6, line 556: Can we enqueue it from the module? we dont have to but if its not too much effort it makes more sense

assets/dev/scss/editor/preview/preview.scss
```diff
@@ -552,6 +552,21 @@ html.elementor-html {
...
+	&.elementor-edit-area-active {
+		& > * > * {
+			pointer-events: initial;
+		}
```
Omerisra6, line 560: What if we have nested components?

Will that enable the outer component pointer events if I'm editing the inner component?

ronros-elementor: outer contents are masked by the backdrop anyway, but checked and `elementor-edit-area-active` is applied on the current active element only, so from that element in - all elements would be clickable again, up to the next component wrapper

Omerisra6 (approved):

PR #34107, opened 2026-01-06
Internal: Prod ready and first Inline editing migrations  [ED-21950]

asafdl:


modules/atomic-widgets/prop-type-migrations/migrations-orchestrator.php
```diff
@@ -33,76 +38,207 @@ public static function destroy(): void {
...
+			'elementor/experiments/feature-state-change/' . Atomic_Widgets_Module::EXPERIMENT_INLINE_EDITING,
+			[ __CLASS__, 'clear_all_migration_caches' ],
+			10,
+			2
+		);
+
+		$registered = true;
+	}
+
+	public static function clear_all_migration_caches( $old_state = null, $new_state = null ): void {
```
murzin-nikita: should be try/catched as well

modules/atomic-widgets/prop-type-migrations/migrations-orchestrator.php
```diff
@@ -135,82 +271,96 @@ private function is_atomic_widget( array $element ): bool {
...
-	public function migrate_element( array $settings, array $schema, string $widget_type ): array {
-		$migrated_settings = [];
+	private function is_atomic_element( array $element ): bool {
+		$el_type = $element['elType'] ?? '';
+
+		if ( empty( $el_type ) || 'widget' === $el_type ) {
+			return false;
+		}
+
+		$element_instance = \Elementor\Plugin::$instance->elements_manager->get_element_types( $el_type );
```
murzin-nikita: let's use namespace imports

modules/atomic-widgets/module.php
```diff
@@ -395,4 +398,34 @@ private function add_inline_styles() {
...
+		}
+
+		$orchestrator = Migrations_Orchestrator::make( $this->get_migrations_base_path() );
+
+		$orchestrator->migrate_document(
+			$data,
+			$document->get_post()->ID,
+			function( $migrated_data ) use ( $document ) {
+				$document->update_json_meta(
+					\Elementor\Core\Base\Document::ELEMENTOR_DATA_META_KEY,
```
murzin-nikita: use Elementor\Core\Base\Document

modules/atomic-widgets/prop-type-migrations/migrations-orchestrator.php
asafdl: https://developer.wordpress.org/reference/classes/wpdb/query/
it seems like wpdb follows return false on error, so no need to try catch

modules/atomic-widgets/prop-type-migrations/migrations-orchestrator.php
asafdl: done

tests/phpunit/elementor/modules/atomic-widgets/prop-type-migrations/test-document-migration-integration.php
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * @group prop-type-migrations
+ * @group integration
+ */
+class Test_Document_Migration_Integration extends Elementor_Test_Base {
+	use MatchesSnapshots;
+
+	private string $fixtures_path = __DIR__ . '/fixtures/document-migrations/';
+
+	// public function setUp(): void {
```
murzin-nikita, line 25: can we delete this?

modules/atomic-widgets/prop-type-migrations/migrations-orchestrator.php
```diff
@@ -30,83 +36,225 @@ public static function make( string $migrations_base_path ): self {
...
+			'elementor/experiments/feature-state-change/' . Atomic_Widgets_Module::EXPERIMENT_INLINE_EDITING,
+			[ __CLASS__, 'clear_all_migration_caches' ],
+			10,
+			2
+		);
+
+		$registered = true;
+	}
+
+	public static function clear_all_migration_caches( $old_state = null, $new_state = null ): void {
```
murzin-nikita: `clear_migration_cache`

tests/phpunit/elementor/modules/atomic-widgets/prop-type-migrations/test-document-migration-integration.php
asafdl: id rather leave it, and open a ticket to fix, the test works and is valuable, but for some reason it interrupts other integration tests from running properly

murzin-nikita (approved):

PR #34116, opened 2026-01-07
Internal: Exposed property is still shown even after the element is deleted [ED-22296]

Omerisra6:
### ✨ PR Description
Purpose: Fix bug where overridable component properties remain visible in the UI after their associated elements are deleted from the document.

Main changes:
- Added cleanup-overridable-props-on-delete hook to automatically remove overridable props when elements are deleted
- Implemented collectDeletedElementIds to handle deletion of parent elements and all descendants
- Integrated hook initialization into editor-components init sequence with comprehensive test coverage

[ED-22296]: https://elementor.atlassian.net/browse/ED-22296?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/packages/core/editor-components/src/hooks/cleanup-overridable-props-on-delete.ts
```diff
@@ -0,0 +1,101 @@
...
+			return;
+		}
+
+		const currentComponentId = selectCurrentComponentId( state );
+
+		if ( ! currentComponentId ) {
+			return;
+		}
+
+		const overridableProps = selectOverridableProps( state, currentComponentId );
```
IshayMaya, line 27: i think this should all be 1 function from store

packages/packages/core/editor-components/src/hooks/cleanup-overridable-props-on-delete.ts
```diff
@@ -0,0 +1,101 @@
...
+		for ( const descendant of descendants ) {
+			const descendantId = descendant.model?.get?.( 'id' ) ?? descendant.id;
+
+			if ( descendantId ) {
+				elementIds.add( descendantId );
+			}
+		}
+	}
+
+	return elementIds;
```
IshayMaya:
```suggestion
	const elementIds = containers
		.filter( Boolean )
		.flatMap( ( container ) => [ container, ...getAllDescendants( container ) ] )
		.map( getElementId )
		.filter( ( id ): id is string => Boolean( id ) );

	return new Set( elementIds );
```

packages/packages/core/editor-components/src/store/actions/__tests__/delete-overridable-prop.test.ts
```diff
@@ -0,0 +1,240 @@
...
+	const MOCK_COMPONENT_ID = 1;
+	const MOCK_ELEMENT_ID = 'element-1';
+	const MOCK_PROP_KEY = 'prop-1';
+	const MOCK_SETTING_KEY = 'text';
+	const MOCK_GROUP_ID = 'group-1';
+	const MOCK_ORIGIN_VALUE = { $$type: 'string', value: 'Hello' };
+
+	let mockState: { data: PublishedComponent[] };
+
+	const createMockOverridableProps = (): OverridableProps => ( {
```
IshayMaya: move to utils? we have the same function in 2 other tests as well

packages/packages/core/editor-components/src/store/actions/delete-overridable-prop.ts
```diff
@@ -8,9 +8,10 @@ import { removePropFromAllGroups } from '../utils/groups-transformers';
 type DeletePropParams = {
 	componentId: ComponentId;
 	propKey: string;
+	skipRevert?: boolean;
 };
 
-export function deleteOverridableProp( { componentId, propKey }: DeletePropParams ): void {
+export function deleteOverridableProp( { componentId, propKey, skipRevert = false }: DeletePropParams ): void {
```
IshayMaya: where do we pass true

IshayMaya (approved):

PR #34134, opened 2026-01-07
Internal: Revert overridable props to overrides for component instances [ED-22232]

ronros-elementor:
### ✨ PR Description
Purpose: Refactor component instance overridable property deletion to properly revert overridable props to override format for component instances while maintaining backward compatibility with regular widgets.

Main changes:
- Extracted revertElementOverridableSetting utility to handle component instances differently, converting overridable props back to simple overrides instead of reverting to origin values
- Added comprehensive test suite with 473 lines covering widget reversion, component instance override conversion, and mixed override scenarios
- Updated test utilities to use createMockContainer helper with optional children parameter for improved test setup consistency

[ED-22232]: https://elementor.atlassian.net/browse/ED-22232?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/packages/core/editor-components/src/store/actions/delete-overridable-prop.ts
```diff
@@ -41,16 +55,83 @@ export function deleteOverridableProp( { componentId, propKey }: DeletePropParam
...
+	originValue: unknown,
+	overrideKey: string
+): void {
 	const container = getContainer( elementId );
 
 	if ( ! container ) {
 		return;
 	}
 
+	if ( container.model.get( 'widgetType' ) === 'e-component' ) {
```
annab1: 1. I think this should be called from `deleteOverridableProp`, not from this function
2. you can use `isComponentInstance` function

packages/packages/core/editor-components/src/store/actions/delete-overridable-prop.ts
```diff
@@ -41,16 +55,83 @@ export function deleteOverridableProp( { componentId, propKey }: DeletePropParam
 	);
 }
 
-function revertElementSetting( elementId: string, settingKey: string, originValue: unknown ): void {
+function revertElementSetting(
```
MiryamOren: Can we please move this function to a separate util file?

MiryamOren: Also, can we name it 
```suggestion
function revertElementOverridableSetting(
```
?

packages/packages/core/editor-components/src/store/actions/delete-overridable-prop.ts
```diff
@@ -41,16 +55,83 @@ export function deleteOverridableProp( { componentId, propKey }: DeletePropParam
...
 		id: elementId,
 		props: { [ settingKey ]: originValue ?? null },
 		withHistory: false,
 	} );
 }
+
+function revertComponentInstanceSetting( elementId: string, overrideKey: string ): void {
+	const setting = getElementSetting< ComponentInstanceProp >( elementId, 'component_instance' );
+
+	if ( ! setting?.value.overrides ) {
```
annab1: either do `if (!setttings)` or remove completely - you check overrides validity a few lines below

annab1 (changes requested):

packages/packages/core/editor-components/src/store/actions/delete-overridable-prop.ts
ronros-elementor: Regarding 1 - 
First of all - on the `deleteOverridableProp` scope I currently don't lookup for the container of the corresponding id, so having it there means I would need to do that on the delete overridable prop level
But moreover, now as I'm moving it out to be a util which we'd use for the detach flow as well - I find it more convenient to just provide it with the known attributes from the store without any extra prerequisites. As I suppose the logic would resemble in the sense of going over the overridables from the store, then reverting each of them in the corresponding widget/component-instance

wdyt?

(2 is done of course, and good catch, thank you!)

annab1 (approved):

PR #34168, opened 2026-01-11
Internal: Fix deleting overridable after element is deleted [ED-22409]

MiryamOren:
### ✨ PR Description
Purpose: Fix cleanup hook timing to prevent overridable props deletion errors when referenced elements are removed from the document.

Main changes:
- Changed registerDataHook from 'after' to 'dependency' phase to execute before element deletion completes
- Updated all early return statements to return true for proper hook chain continuation
- Added explicit return true after successful cleanup operation to signal dependency resolution

### DIFF
--- a/packages/packages/core/editor-components/src/hooks/cleanup-overridable-props-on-delete.ts
+++ b/packages/packages/core/editor-components/src/hooks/cleanup-overridable-props-on-delete.ts
@@ -11,35 +11,35 @@ type DeleteCommandArgs = {
 };
 
 export function initCleanupOverridablePropsOnDelete() {
-	registerDataHook( 'after', 'document/elements/delete', ( args: DeleteCommandArgs ) => {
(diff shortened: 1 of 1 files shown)


packages/packages/core/editor-components/src/hooks/cleanup-overridable-props-on-delete.ts
```diff
@@ -11,43 +11,43 @@ type DeleteCommandArgs = {
 };
 
 export function initCleanupOverridablePropsOnDelete() {
-	registerDataHook( 'after', 'document/elements/delete', ( args: DeleteCommandArgs ) => {
+	registerDataHook( 'dependency', 'document/elements/delete', ( args: DeleteCommandArgs ) => {
 		const state = getState() as ComponentsSlice | undefined;
 
 		if ( ! state ) {
-			return;
```
Omerisra6, line 18: why do we need to return boolean?

Omerisra6 (approved):

PR #34177, opened 2026-01-11
Internal: Restrictions for non-pro users [ED-21221]

ronenboxer:
### ✨ PR Description
Purpose: Restrict dynamic tag functionality for non-pro users by adding validation to prevent usage of unsupported dynamic tags and fallback to default values

Main changes:
- Added isDynamicTagSupported validation to check tag availability before rendering or transforming dynamic content
- Implemented fallback logic in dynamic transformer and props resolver to return default values for unsupported tags
- Modified DynamicSelectionControl to render original control component when dynamic tag is not supported for user tier


ronenboxer:
custom core pro tests
https://github.com/elementor/elementor-pro/actions/runs/20919307235/job/60101823631
https://github.com/elementor/elementor-pro/pull/6175

packages/packages/core/editor-editing-panel/src/dynamics/components/dynamic-selection-control.tsx
```diff
@@ -24,28 +30,44 @@ import { __ } from '@wordpress/i18n';
...
 
 	const [ propValueFromHistory ] = usePersistDynamicValue( bind );
 	const selectionPopoverState = usePopupState( { variant: 'popover' } );
 
 	const { name: tagName = '' } = value;
-
 	const dynamicTag = useDynamicTag( tagName );
 
+	if ( ! isDynamicTagSupported( tagName ) && OriginalControl ) {
+		console.log( 1232123121 );
```
asafdl: del

packages/packages/core/editor-editing-panel/src/dynamics/__tests__/dynamic-transformer.test.ts
```diff
@@ -53,6 +74,28 @@ describe( 'dynamicTransformer', () => {
...
+		const value = dynamicTransformer(
+			{ name: 'test-tag', settings: { settingKey: 'setting-value' } },
+			{ key: 'test' }
+		);
+
+		// Assert.
+		expect( value ).toBe( null );
+	} );
+
+	it( "should return default value if it exists when tag doesn't exist", async () => {
```
asafdl, line 89: also should return default value if value.name is null

packages/packages/core/editor-editing-panel/src/dynamics/components/dynamic-selection-control.tsx
```diff
@@ -24,28 +30,44 @@ import { __ } from '@wordpress/i18n';
...
+import { dynamicPropTypeUtil, isDynamicTagSupported } from '../utils';
 import { DynamicSelection } from './dynamic-selection';
 
 const SIZE = 'tiny';
 
 const tagsWithoutTabs = [ 'popup' ];
 
-export const DynamicSelectionControl = () => {
-	const { setValue: setAnyValue } = useBoundProp();
+export const DynamicSelectionControl = ( { OriginalControl, ...props }: { OriginalControl?: ControlComponent } ) => {
```
asafdl, line 45: why is original control optional? doesn't it return it every time?

packages/packages/core/editor-editing-panel/src/dynamics/__tests__/dynamic-transformer.test.ts
ronenboxer: added for protocol, null value doesn't get sent to this transformer because it only transforms dynamic values, if the value is null it won't even get sent to the transformer

packages/packages/core/editor-editing-panel/src/dynamics/components/dynamic-selection-control.tsx
ronenboxer: lo rotze

packages/packages/core/editor-editing-panel/src/dynamics/components/dynamic-selection-control.tsx
ronenboxer: It depends who renders it
The control replacement API does send an OriginalControl, but in general this can get rendered without it, like when the tests use it

asafdl (approved):

packages/packages/core/editor-editing-panel/src/sync/types.ts
Ntnelbaba, line 1: @ronenelementor @asafdl Next time, please, use the `global.d.ts` file then all of those ExtendedWindow balagan will not be required

PR #34200, opened 2026-01-12
Fix: Wait until component is render is done to open its editing panel [ED-22361]

MiryamOren:
### ✨ PR Description
Purpose: Fix race condition in component editing panel by ensuring panel only opens after component rendering is complete to prevent crashes when accessing inner elements.

Main changes:
- Extracted panel opening logic into `_openEditingPanel` method and added `_doAfterRender` helper to defer panel opening until render completion
- Added container validation check in override control to throw OverrideControlInnerElementNotFoundError when inner elements are not found
- Updated override control to use `getContainer` for element validation before accessing element type and properties


packages/packages/core/editor-canvas/src/legacy/create-templated-element-type.ts
```diff
@@ -230,5 +230,13 @@ export function createTemplatedElementView( {
 
 			this.triggerMethod( 'render', this );
 		}
+
+		_openEditingPanel( options?: { scrollIntoView: boolean } ) {
+			if ( this.isRendered ) {
+				super._openEditingPanel( options );
+			} else {
+				this.on( 'render', () => super._openEditingPanel( options ) );
```
IshayMaya:
```suggestion
				this.once( 'render', () => super._openEditingPanel( options ) );
```

IshayMaya: or maybe have some agnostic function in case we will need more stuff later?

```suggestion
		_doAfterRender( callback: () => void ) {
			if ( this.isRendered ) {
				callback();
			} else {
				this.once( 'render', callback );
			}
		}

		_openEditingPanel( options?: { scrollIntoView: boolean } ) {
			this._doAfterRender( () => super._openEditingPanel( options ) );
		}
```

MiryamOren: Nice! will do

IshayMaya (dismissed):

IshayMaya (approved):

PR #34217, opened 2026-01-13
Internal: MVP Input Atom. [ED-22391]

RobiFerentz:


includes/editor-templates/panel-elements.php
```diff
@@ -83,7 +83,7 @@
 		<# } else if ( false === obj.editable ) { #>
 			<i class="eicon-lock"></i>
 		<# } #>
-		<# if ( obj.categories.includes( 'v4-elements' ) ) { #>
+		<# if ( obj.categories.some( category => ['v4-elements', 'atomic-form'].includes( category ) ) ) { #>
```
ronenboxer: I'm not sure this logic is correct
Can you explain it please?

RobiFerentz: Think of this as checking an intersection of 2 arrays, I need to know that at least one category is shared by both arrays. I could use Set.intersection, but it's 'newly available' and not baseline support.

ronenboxer: got it, i found it a bit unreadble
maybe extract this `['v4-elements', 'atomic-form']` to a const?

ronenboxer (dismissed):

YaelAvitan (dismissed):

includes/editor-templates/panel-elements.php
asafdl: why is atomic form not under v4-elements?

doronwolf: Hi @asafdl look in the [figma design](https://www.figma.com/design/OmDDQqbg5neU8puZ3j7F4u/Atomic-Forms?node-id=281-6208&t=TJpZwEtbzL3Ip11C-4)

RobiFerentz: Fixed.

asafdl: im not sure how its related to figma, can you elaborate?

doronwolf: @asafdl If you look in the Figma link I've attached earlier, you can see the 'input' atom should be placed under the new "atomic form" group (instead of the 'v4-elements') in the widget panel. But maybe I misunderstood the context of this code? If so, please disregard my comment

murzin-nikita (approved):

PR #34233, opened 2026-01-13
Fix: Overridable tags break rendering [ED-22466]

Omerisra6:
### ✨ PR Description
Purpose: Fix rendering issues with overridable tag properties by implementing proper resolution logic and caching mechanism to handle dynamic tag transformations correctly.

Main changes:
- Implemented `_resolvePropValue()` method to recursively resolve overridable properties through transformer registry, supporting dynamic value overrides and nested resolution
- Refactored `tagName()` to use WeakMap caching and resolve overridable tag settings with render context, ensuring "a" tag returns when element has link
- Added tag cache invalidation in `_beforeRender()` and deferred link attribute application to `_afterRender()` to prevent premature DOM recreation with stale values


modules/atomic-widgets/assets/js/editor/create-atomic-element-base-view.js
```diff
@@ -69,12 +77,6 @@ export default function createAtomicElementBaseView( type ) {
 				local.id = cssId.value;
 			}
 
-			const link = this.getLink();
-
-			if ( link ) {
-				local[ link.attr ] = link.value;
-			}
```
Omerisra6, line 76: Don't we still need it?

modules/atomic-widgets/assets/js/editor/create-atomic-element-base-view.js
Omerisra6, line 130: maybe we can somehow use the `changedAttributes` to make the solution cleaner

Omerisra6:
## Responses to Review Comments

### Comment 1: "Don't we still need it?" (removing getLink() from attributes())

Yes, we still need the link in `attributes()`, but now it comes from the cache (`_resolvedLink`) instead of calling `getLink()` directly.

**The flow is:**
1. `attributes()` is called during element creation (before render context is available)
2. If `_resolvedLink` cache exists, it's included in the attributes
3. After render, `_applyResolvedAttributes()` resolves the link with render context, caches it, and triggers rerender
4. On rerender, `attributes()` now includes the cached link

This was changed because `getLink()` now uses render context for resolving overridable props, which isn't available when `attributes()` is first called during element construction.

See lines 91-95 where the cached link is included:
```javascript
const cachedLink = this.model.get( '_resolvedLink' );
if ( cachedLink ) {
    local[ cachedLink.attr ] = cachedLink.value;
}
```

---

### Comment 2: "maybe we can somehow use the changedAttributes to make the solution cleaner"

Using `changedAttributes` for a cleaner solution would be tricky because the cache invalidation needs to happen at different points:

1. **`_beforeRender()`** - Cache is cleared before every render cycle so fresh values are resolved
2. **`renderOnChange()` with tag/link changes** - Cache is cleared when the element's own settings change
3. **Parent component override changes** - The parent triggers a child rerender, which clears cache via `_beforeRender()`

The `changedAttributes` only tells us what changed on *this* element's model, but for overrides, the change happens in the *parent* component instance's settings. The child element's settings don't change - only the render context does.

The current approach with cache invalidation in `_beforeRender()` handles all these cases uniformly:
- Element's own settings change → triggers rerender → `_beforeRender()` clears cache
- Parent's overrides change → parent triggers child rerender → `_beforeRender()` clears cache

This makes the solution more robust against future changes in how overrides propagate.

modules/atomic-widgets/assets/js/editor/create-atomic-element-base-view.js
```diff
@@ -12,11 +12,19 @@ export default function createAtomicElementBaseView( type ) {
...
+			}
+
+			const renderContext = this.getResolverRenderContext?.();
+			const tag = this.model.getSetting( 'tag' );
+			const resolved = this._resolvePropValue( tag, renderContext );
+			const tagValue = resolved?.value ?? resolved;
+			const hasLink = this._hasLink( renderContext );
+
+			return hasLink ? 'a' : ( tagValue || this.model.config.default_html_tag || 'div' );
 		},
```
IshayMaya: maybe? for better readability
```suggestion
		tagName() {
			return resolvedTagCache.get( this.model ) ?? this._resolveTag();
		},

		_resolveTag() {
			const renderContext = this.getResolverRenderContext?.();
			const tagSetting = this.model.getSetting( 'tag' );
			const resolvedTag = this._resolvePropValue( tagSetting, renderContext );
			const tagValue = resolvedTag?.value ?? resolvedTag;

			if ( this._hasLink( renderContext ) ) {
				return 'a';
			}

			return tagValue || this.model.config.default_html_tag || 'div';
		},
```

modules/atomic-widgets/assets/js/editor/create-atomic-element-base-view.js
```diff
@@ -12,11 +12,19 @@ export default function createAtomicElementBaseView( type ) {
 		_childrenRenderPromises: [],
 
 		tagName() {
-			const tagControl = this.model.getSetting( 'tag' );
-			const tagControlValue = tagControl?.value || tagControl;
-			const defaultTag = this.model.config.default_html_tag;
+			const cachedTag = this.model.get( '_resolvedTag' );
```
IshayMaya: maybe we should create a weakMap for each type creation scope?
this way we wont have to make changed to the model

```
export default function createAtomicElementBaseView( type ) {
	const resolvedTagCache = new WeakMap();

	const AtomicElementView = BaseElementView.extend( {
....
```

modules/atomic-widgets/assets/js/editor/create-atomic-element-base-view.js
```diff
@@ -215,18 +218,47 @@ export default function createAtomicElementBaseView( type ) {
...
-			this._waitForChildrenToComplete().then( resolve );
+			this._waitForChildrenToComplete().then( () => {
+				this._applyResolvedAttributes();
+				resolve();
+			} );
 		},
 
 		_beforeRender() {
 			this._isRendering = true;
+			this.model.unset( '_resolvedTag' );
```
IshayMaya: extract to local methods

```
		_invalidateTagCache() {
			resolvedTagCache.delete( this.model );
		},

		_cacheResolvedTag( tag ) {
			resolvedTagCache.set( this.model, tag );
		},
```

modules/atomic-widgets/assets/js/editor/create-atomic-element-base-view.js
```diff
@@ -215,18 +218,47 @@ export default function createAtomicElementBaseView( type ) {
...
+			}
+
+			const link = this.getLink();
+
+			this.$el.removeAttr( 'href' );
+			this.$el.removeAttr( 'data-action-link' );
+
+			if ( link ) {
+				this.$el.attr( link.attr, link.value );
+			}
```
IshayMaya, line 283: split it?
```suggestion
		_applyResolvedAttributes() {
			if ( ! this._parent ) {
				return;
			}

			if ( this._shouldRecreateForTagChange() ) {
				return;
			}

			this._applyLinkAttributes();
		},

		_shouldRecreateForTagChange() {
			const resolvedTag = this.tagName();
			const currentTag = this.el.tagName.toLowerCase();

			if ( resolvedTag === currentTag ) {
				return false;
			}

			this._cacheResolvedTag( resolvedTag );
			this.rerenderEntireView();
			return true;
		},

		_applyLinkAttributes() {
			this.$el.removeAttr( 'href' );
			this.$el.removeAttr( 'data-action-link' );

			const link = this.getLink();

			if ( link ) {
				this.$el.attr( link.attr, link.value );
			}
		},
```

modules/atomic-widgets/assets/js/editor/create-atomic-element-base-view.js
IshayMaya: invalidate on destroy as well

modules/atomic-widgets/assets/js/editor/create-atomic-element-base-view.js
Omerisra6: I think it makes sense to handle that from the model

IshayMaya (approved):

PR #34241, opened 2026-01-14
Fix: Render relevant local styles when pasting/importing/duplicating component instances

ronros-elementor:
### ✨ PR Description
Purpose: Fix component instance rendering by loading local styles and assets when pasting, importing, or duplicating component instances in the editor.

Main changes:
- Added initLoadComponentDataAfterInstanceAdded to load component assets on paste/import operations via data hooks
- Refactored module organization by moving cleanup and regenerate functions from hooks/ to sync/ directory
- Implemented comprehensive test coverage for new asset loading functionality with single and array element handling


packages/packages/core/editor-components/src/sync/load-component-data-after-instance-added.ts
```diff
@@ -0,0 +1,24 @@
...
+
+	registerDataHook( 'after', 'document/elements/import', ( _args, result: V1Element | V1Element[] ) => {
+		load( result );
+	} );
+}
+
+function load( result: V1Element | V1Element[] ) {
+	const containers = Array.isArray( result ) ? result : [ result ];
+
+	loadComponentsAssets( containers.map( ( container ) => container.model.toJSON() as V1ElementData ) );
```
IshayMaya, line 19: doe it matter if it's still running in the background? can it break something?

packages/packages/core/editor-components/src/sync/load-component-data-after-instance-added.ts
```diff
@@ -0,0 +1,24 @@
+import { type V1Element, type V1ElementData } from '@elementor/editor-elements';
+import { registerDataHook } from '@elementor/editor-v1-adapters';
+
+import { loadComponentsAssets } from '../store/actions/load-components-assets';
+
+export function initLoadComponentDataAfterInstanceAdded() {
```
IshayMaya, line 6: just a thought, you think it worth unifying with the regenerate actions?

packages/packages/core/editor-components/src/sync/load-component-data-after-instance-added.ts
ronros-elementor: the only effect I noticed was the style being updated a bit after the initial rendering of the component instance, which is also the same defect as upon loading the editor
and all other side effects dispatch corresponding actions of the store, so in terms of react we're covered (and that is the only thing I can think of, i.e. the editing panel)

packages/packages/core/editor-components/src/sync/load-component-data-after-instance-added.ts
ronros-elementor: I could tap into your regeneration logic and rename it to something more general

IshayMaya: maybe keep it like this for now, if we will have one more action we can have a duplication events handler or something

IshayMaya (dismissed):

IshayMaya (dismissed):

IshayMaya (approved):

PR #34245, opened 2026-01-14
Internal: Add Components notification for core users [ED-22483]

Omerisra6:
### ✨ PR Description
Purpose: Add promotional notification to Components tab informing core users that Components feature will transition to Pro subscription while preserving existing content.

Main changes:
- Created ComponentsProNotification component displaying InfoAlert about upcoming Pro subscription requirement for Components feature
- Added isProUser utility function checking elementor.helpers.hasPro and elementorPro.config.isActive for user tier detection
- Integrated notification into Components tab rendering conditionally for non-Pro users with existing components


packages/packages/core/editor-components/src/components/components-tab/components-pro-notification.tsx
```diff
@@ -0,0 +1,28 @@
+import * as React from 'react';
+import { InfoAlert } from '@elementor/editor-ui';
+import { Box, Typography } from '@elementor/ui';
+import { __ } from '@wordpress/i18n';
+
+import { isProUser } from '../../utils/is-pro-user';
+
+export function ComponentsProNotification() {
+	if ( ! isProUser() ) {
```
IshayMaya: seems redundant to check it here, should be outside where we already have conditional rendering

packages/packages/core/editor-components/src/utils/is-pro-user.ts
```diff
@@ -0,0 +1,26 @@
...
+		};
+	};
+	elementorPro?: {
+		config?: {
+			isActive?: boolean;
+		};
+	};
+};
+
+export function isProUser(): boolean {
```
IshayMaya, line 14: maybe should be util is user package or something?  i see that we check it on other place as well and changing the types everywhere for it

Omerisra6: We check only in one more place I think it is fine for now

IshayMaya (dismissed):

IshayMaya (approved):

PR #34247, opened 2026-01-14
Internal: Add components overrides analytics events

MiryamOren:
### ✨ PR Description
Purpose: Add analytics tracking for component overrides functionality to monitor user interactions with property exposure, removal, and panel operations across the editor.

Main changes:
- Added four new analytics events (propertiesPanelOpened, propertiesGroupCreated, propertyExposed, propertyRemoved) to track component property management actions
- Updated all component tracking calls to include source parameter distinguishing user actions from MCP tool operations
- Refactored createUnpublishedComponent function to accept object parameters and standardized tracking across component lifecycle events


packages/packages/core/editor-components/src/components/component-panel-header/component-badge.tsx
```diff
@@ -4,38 +4,39 @@ import { ComponentPropListIcon } from '@elementor/icons';
 import { Badge, Box, keyframes, styled, ToggleButton } from '@elementor/ui';
 import { __ } from '@wordpress/i18n';
 
-export const ComponentsBadge = React.forwardRef< HTMLDivElement, { overridesCount: number; onClick: () => void } >(
```
MiryamOren, line 7: All changes in this file are naming & texts: changed all `overrides` to `overridableProps` / `exposedProperties`

packages/packages/core/editor-components/src/store/actions/set-overridable-prop.ts
```diff
@@ -94,5 +95,20 @@ export function setOverridableProp( {
...
+		const currentComponent = selectCurrentComponent( getState() );
+
+		trackComponentEvent( {
+			action: 'propertyExposed',
+			component_uid: currentComponent?.uid,
+			property_id: overridableProp.overrideKey,
+			property_path: propKey,
+			property_name: label,
+			element_type: widgetType ?? elType,
+		} );
```
Omerisra6, line 113: Do we have a way to add these logs as some kind of a side effect to the actions via the redux store?

I think it will be cleaner, to separate the actual logic from the events if we can.

packages/packages/core/editor-components/src/utils/tracking.ts
```diff
@@ -5,18 +5,43 @@ import { __getState as getState } from '@elementor/store';
 import { selectCreatedThisSession } from '../store/store';
 import { type ExtendedWindow } from '../types';
 
+type PerformedBy = 'user' | 'mcp_tool';
+
+let currentPerformedBy: PerformedBy = 'user';
+
+export const withMCPContext = async < T >( fn: () => Promise< T > ): Promise< T > => {
```
Omerisra6: What this function does?

MiryamOren: We should add analytics parameter of `performed_by: 'user' | 'mcp_tool'`
By wrapping the MCP tool handler with this function, it changes the `currentPerformedBy` variable to `mcp_tool` while the handler is running, and restores the last value once it's done, so the MCP handler doesn't have to pass this context to every function it calls that might, in the end, trigger event tracking

packages/packages/core/editor-components/src/store/actions/set-overridable-prop.ts
MiryamOren: We can't really do it, since the same state change can happen for different reasons.
For example, when deleting an element, we delete the related overridable from meta, but we shouldn't trigger an event in this case

packages/packages/core/editor-components/src/utils/tracking.ts
MiryamOren: * I changed the parameter name to `source`

Omerisra6: Can't we simply pass it as a parameter?

I think it is a bit confusing

Omerisra6 (dismissed):

Omerisra6 (approved):

PR #34254, opened 2026-01-14
Internal: Create control rerenders control

Omerisra6:
### ✨ PR Description
Purpose: Refactor control replacement logic to prevent unnecessary rerenders by moving createControl call from component body to dedicated hook-based component.

Main changes:
- Replaced inline createControl() with useControlReplacement hook to memoize control creation and prevent rerenders
- Extracted ControlWithReplacements component to handle control replacement logic with proper React rendering pattern
- Moved dynamic replacement definitions from global scope into test cases for better test isolation

### DIFF
--- a/packages/packages/core/editor-components/src/components/overridable-props/overridable-prop-control.tsx
+++ b/packages/packages/core/editor-components/src/components/overridable-props/overridable-prop-control.tsx
@@ -2,10 +2,10 @@ import * as React from 'react';
 import { type ComponentType } from 'react';
 import {
 	ControlReplacementsProvider,
-	createControl,
 	PropKeyProvider,
 	PropProvider,
 	useBoundProp,
+	useControlReplacement,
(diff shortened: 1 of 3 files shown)


packages/packages/core/editor-components/src/components/overridable-props/overridable-prop-control.tsx
```diff
@@ -97,10 +96,31 @@ export function OverridablePropControl< T extends object >( {
...
 					</ControlReplacementsProvider>
 				</PropKeyProvider>
 			</PropProvider>
 		</OverridablePropProvider>
 	);
 }
+
+type ControlComponentType = ComponentType< object & { OriginalControl: ComponentType } >;
+
+function ControlWithReplacements< T extends object >( {
```
IshayMaya, line 109: we need to find a better solution, for multi replacements

IshayMaya (approved):

PR #34265, opened 2026-01-15
Fix: Paste styles to include also global classes [ED-22387]

ronros-elementor:
### ✨ PR Description
Purpose: Extend paste-style command to copy global CSS classes along with local styles when pasting element styles in the editor canvas.

Main changes:
- Added pasteClasses function to merge global classes from clipboard element with target elements' existing classes
- Implemented getClassesWithoutLocalStyle to filter out local style IDs from copied classes, ensuring only global classes are transferred
- Enhanced test coverage with parameterized tests validating class merging, deduplication, and multi-element paste scenarios

[ED-22387]: https://elementor.atlassian.net/browse/ED-22387?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/packages/core/editor-canvas/src/style-commands/paste-style.ts
```diff
@@ -29,25 +31,60 @@ export function initPasteStyleCommand() {
...
 	}
 
 	const elementStyles = clipboardElement.styles;
 	const elementStyle = Object.values( elementStyles ?? {} )[ 0 ]; // we currently support only one local style
 
-	if ( ! elementStyle ) {
-		return;
+	const classesSetting = getClassesWithoutLocalStyle( clipboardContainer, elementStyle );
+	if ( classesSetting.length ) {
+		pasteClasses( atomicContainers, classesSetting );
```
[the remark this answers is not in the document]
ronros-elementor: Just to comment about that - it's intentional as we don't mind about any class used there, for now we only know of globals (and base, but these are fixed to the widget type), but we may have other style providers in the future, the system is built to support that. No need to explicitly list globals because it's both inaccurate, and contradicting to our architecture, where each package is only aware of its own artifacts, and imported ones, but globals are hooking up on to our infra, it's not treated as a core feature, unlike the paste-styles which is more generic in nature.

packages/packages/core/editor-canvas/src/style-commands/paste-style.ts
```diff
@@ -29,25 +31,60 @@ export function initPasteStyleCommand() {
...
 	}
 
-	const atomicContainers = containers.filter( isAtomicWidget ) as V1Element[];
+	if ( elementStyle ) {
```
IshayMaya, line 56: is that mean we running the callback even if we didn't paste? is it valid?

packages/packages/core/editor-canvas/src/style-commands/__tests__/paste-style.test.ts
```diff
@@ -447,138 +494,324 @@ describe( 'pasteStyles', () => {
...
+	describe( 'Pasting styles and classes', () => {
+		type SelectedElementState = {
+			id: string;
+			hasLocalStyle: boolean;
+			existingClasses: string[];
+		};
+
+		type ClipboardState = {
+			hasLocalStyle: boolean;
+			globalClasses: string[];
```
IshayMaya: not really critical but maybe we should also not be "aware" of global classes here just like you did on the actual code? (only local/not local)

packages/packages/core/editor-canvas/src/style-commands/paste-style.ts
ronros-elementor: What do you mean? this command runs after the user clicked on the "Paste Style" entry from the widget's context menu, and the callback is invoked only if the clipboard element is atomic, and only on the atomic elements selected (if none is selected it won't be executed either)

IshayMaya: oh so maybe rename the `pasteCallback` it implies it runs on the `pasteClasses` action

IshayMaya (approved):

PR #34286, opened 2026-01-16
Fix: Component draft handling [ED-22503]

MiryamOren:
### Background

**The Original Bug:**
1. User changes overridable props of a component
2. User saves as draft/autosave
3. User gets out of component edit mode and publishes the page the component is inside (triggers `update_statuses` endpoint)
4. **Result:** Only elements were copied from autosave to the published version - changes in overridable props were lost

While debugging, I found a related issue occurred with component rename - on rename of a draft component, all draft changes were lost (including elements data).

### Debugging
I checked `update_statuses`'s code and found out we only copy the elements' data from autosave to the published version, and any other changes were lost.

`archive` and `update-titles` endpoints were always updating the published version, ignoring any existing draft, which caused the published version to be mistakenly detected as more up-to-date than the draft, and therefore, in the next sessions, it was displayed instead of the draft, and the draft changes were lost.

### Solution

- **Fixed `update_statuses`** - Now properly copies all component data from autosave: elements, title, and component custom meta (overridable props, archived status)
- **Added `status` parameter** to `archive` and `update-titles` endpoints:
  - `publish` → Updates the autosave if it exists (then later, it will be published by `update_statuses`), or the published version, if there's no autosave
  - `draft` / `autosave` → Updates (or creates) the autosave 
- **Standardized autosave handling** in repository
- **Changed `beforeSave` execution to sequential** to prevent race conditions when multiple changes are applied to the same component *(TODO: consolidate to a single endpoint)*

### ✨ PR Description
Purpose: Fix component draft/autosave handling to properly preserve and publish component changes including metadata, titles, and archived status across document states.

Main changes:
- Refactored component repository to handle draft/autosave lifecycle with new get_component_for_edit and publish_component methods
- Added status parameter to archive and update-title endpoints to support draft/autosave operations before publishing
- Implemented copy_autosave_data_to_main_component_document_and_publish to transfer component custom metadata during publish workflow


includes/db.php
```diff
@@ -353,14 +353,18 @@ public function safe_copy_elementor_meta( $from_post_id, $to_post_id ) {
 	 * @param int $from_post_id Original post ID.
 	 * @param int $to_post_id   Target post ID.
 	 */
-	public function copy_elementor_meta( $from_post_id, $to_post_id ) {
+	public function copy_elementor_meta( $from_post_id, $to_post_id, $meta_keys = null ) {
```
ronros-elementor: Maybe name it something with "filter" to clarify the usage of this argument

```suggestion
	public function copy_elementor_meta( $from_post_id, $to_post_id, $filter_meta_keys = null ) {
```

modules/components/components-rest-api.php
```diff
@@ -310,7 +320,7 @@ private function get_overridable_props( \WP_REST_Request $request ) {
 				->build();
 		}
 
-		$document = $this->get_repository()->get( $component_id, true );
+		$document = $this->get_repository()->get( $component_id );
```
ronros-elementor, line 323: Wouldn't it mean we now always get the overridables of the publish version? If so how could we be aware of any new override that hasn't been published yet? (assuming I'm the author introducing draft changes, between refreshes I'd loose all data always, no?)

includes/db.php
```diff
@@ -304,7 +304,7 @@ public static function iterate_elementor_documents( $callback, $batch_size = 100
 
 				$callback( $document, $elements_data );
 
-				$processed_posts++;
+				++$processed_posts;
```
Omerisra6, line 307: why?

modules/components/components-repository.php
```diff
@@ -123,13 +121,68 @@ public function archive( $ids ) {
...
+			$main_id = $component->get_main_id();
+			$main_component = $this->get( $main_id, false );
+
+			$autosave = $main_component->get_newer_autosave();
+
+			if ( $autosave ) {
+				$autosave_id = $autosave->get_post()->ID;
+
+				// Copy component custom meta keys from the autosave to the main component.
+				Plugin::$instance->db->copy_elementor_meta( $autosave_id, $main_id, Component_Document::COMPONENT_CUSTOM_META_KEYS );
```
Omerisra6: Why do we need to filter the keys?

modules/components/documents/component.php
```diff
@@ -71,21 +77,19 @@ public function get_overridable_props(): Component_Overridable_Props {
 
 	public function archive() {
 		try {
-			$this->update_main_meta( self::ARCHIVED_META_KEY, json_encode( [
+			$this->update_json_meta( self::ARCHIVED_META_KEY, [
```
Omerisra6, line 80: Was it a bug or just a better way?

modules/components/components-repository.php
```diff
@@ -123,13 +121,68 @@ public function archive( $ids ) {
...
+		$should_create_autosave = in_array( $request_status, $autosave_statuses, true ) && ! $autosave_exists;
+
+		if ( ! $should_create_autosave ) {
+			return $component;
+		}
+
+		return $component->get_autosave( 0, true );
+	}
+
+	public function publish_component( Component_Document $component ): bool {
```
Omerisra6: shouldn't it come from the component document?

modules/components/components-repository.php
```diff
@@ -123,13 +121,68 @@ public function archive( $ids ) {
...
+		if ( ! $component ) {
 			return false;
 		}
-		$sanitized_title = sanitize_text_field( $title );
-		$doc->save( [ 'post_title' => $sanitized_title ] );
+
+		return $component->update_title( $title );
+	}
+
+	private function get_component_for_edit( int $component_id, string $request_status ): ?Component_Document {
```
Omerisra6: how other documents do that?

modules/components/documents/component.php
```diff
@@ -71,21 +77,19 @@ public function get_overridable_props(): Component_Overridable_Props {
...
-	public function get_is_archived() {
-		$archived_meta = $this->get_main_meta( self::ARCHIVED_META_KEY );
-		if ( ! $archived_meta ) {
-			return false;
-		}
-		return json_decode( $archived_meta, true );
+	public function get_is_archived(): bool {
+		$archived_meta = $this->get_json_meta( self::ARCHIVED_META_KEY );
+
+		return $archived_meta['is_archived'] ?? false;
```
Omerisra6, line 92: is that covers an empty array?

modules/components/documents/component.php
```diff
@@ -103,4 +107,37 @@ public function update_overridable_props( $data ): Parse_Result {
...
+
+	public function update_status( string $status ): bool {
+		if ( ! in_array( $status, [ Document::STATUS_PUBLISH, Document::STATUS_DRAFT, Document::STATUS_AUTOSAVE ], true ) ) {
+			return false;
+		}
+
+		return $this->update_post_field( 'post_status', $status );
+	}
+
+	private function update_post_field( string $field, $value ): bool {
```
Omerisra6, line 129: dont we have a similar method in the base document?

modules/components/transformers/component-instance-transformer.php
```diff
@@ -39,7 +39,8 @@ private function is_circular_reference( int $component_id ): bool {
 	}
 
 	private function get_rendered_content( int $component_id ): string {
-		$component = $this->get_repository()->get( $component_id );
+		$should_show_autosave = is_preview();
```
Omerisra6, line 42: So, before that PR, the user would never see autosave changes in the preview?

Is that the right way to check if it's preview mode? Check how we handle it in the global-classes.

modules/components/components-repository.php
```diff
@@ -28,32 +28,30 @@ public function all() {
 		$components = [];
 
 		foreach ( $posts as $post ) {
-			$doc = Plugin::$instance->documents->get( $post->ID );
+			$component = $this->get( $post->ID );
```
Omerisra6, line 31: why?

modules/components/components-repository.php
Omerisra6:
```suggestion
	private function get_component_for_edit( int $component_id, string $status ): ?Component_Document {
```

modules/components/components-repository.php
```diff
@@ -123,13 +121,68 @@ public function archive( $ids ) {
...
+
+		$autosave_statuses = [ Document::STATUS_AUTOSAVE, Document::STATUS_DRAFT ];
+		$autosave_exists = $component->is_autosave();
+		$should_create_autosave = in_array( $request_status, $autosave_statuses, true ) && ! $autosave_exists;
+
+		if ( ! $should_create_autosave ) {
+			return $component;
+		}
+
+		return $component->get_autosave( 0, true );
```
Omerisra6, line 164: What do these magic properties mean?

maybe extract to attributes?

modules/components/components-repository.php
```diff
@@ -123,13 +121,68 @@ public function archive( $ids ) {
...
+				$autosave_elements = $autosave->get_elements_data();
+				$autosave_title = $autosave->get_post()->post_title;
+
+				$success = $main_component->save( [
+					'elements' => $autosave_elements,
+					'settings' => [
+						'post_status' => Document::STATUS_PUBLISH,
+						'post_title' => $autosave_title,
+					],
+				] );
```
Omerisra6: Does that merge the changes?

If I have published elements they be merged with the autosave?

modules/components/components-repository.php
```diff
@@ -123,13 +121,68 @@ public function archive( $ids ) {
...
+		if ( ! $should_create_autosave ) {
+			return $component;
+		}
+
+		return $component->get_autosave( 0, true );
+	}
+
+	public function publish_component( Component_Document $component ): bool {
+		try {
+			$main_id = $component->get_main_id();
```
Omerisra6, line 169: Maybe we can extract this part to a method

modules/components/save-components-validator.php
```diff
@@ -42,7 +42,8 @@ public function validate( Collection $data ) {
 	}
 
 	private function validate_count( Collection $data ): array {
-		$count = $this->components->count() + $data->count();
+		$non_archived_components = $this->components->filter( fn ( $component ) => ! $component['is_archived'] );
```
Omerisra6, line 45: is that the right approach from the product point of view?

packages/packages/core/editor-components/src/sync/before-save.ts
```diff
@@ -24,10 +24,17 @@ export const beforeSave = ( { container, status }: Options ) => {
...
+		syncComponents( { elements, status } ),
 		setComponentOverridablePropsSettingsBeforeSave( { container } ),
-		updateComponentTitleBeforeSave(),
 	] );
 };
+
+// These operations run sequentially to prevent race conditions when multiple
+// edits occur on the same component simultaneously.
+// TODO: Consolidate these into a single PUT /components endpoint.
+const syncComponents = async ( { elements, status }: { elements: V1ElementData[]; status: DocumentSaveStatus } ) => {
```
Omerisra6, line 35: if we wont handle this in this pr: add in the ticket section about error handling

modules/components/documents/component.php
Omerisra6: Check that it works in 7.4

modules/components/components-repository.php
Omerisra6: add a cooment

includes/db.php
MiryamOren: Lint fix in a file i changed something else in

modules/components/components-repository.php
MiryamOren: Same logic, inside `Documents_Manager::ajax_save` :
```
		if ( Document::STATUS_AUTOSAVE === $status ) {
			// If the post is a draft - save the `autosave` to the original draft.
			// Allow a revision only if the original post is already published.
			if ( in_array( $document->get_post()->post_status, [ Document::STATUS_PUBLISH, Document::STATUS_PRIVATE ], true ) ) {
				$document = $document->get_autosave( 0, true );
			}
		}
```

modules/components/documents/component.php
MiryamOren: yes https://www.php.net/manual/en/migration70.new-features.php

modules/components/documents/component.php
MiryamOren: No ):

modules/components/save-components-validator.php
MiryamOren: Yes, I asked Aviv

Omerisra6 (approved):

PR #34305, opened 2026-01-18
Fix(editor-components): Revert overridables on copy, duplicate and component creation [ED-22423]

MiryamOren:
### Problem

We had three issues caused by overridable props not being properly cleaned on element re-creation:

1. **Copy component inner element with overridable setting/s outside component** - The overridable values have no meaning outside the component context, and we don't support editing overridable values outside component edit mode, which made the overridable values control not be shown at all, making these settings uneditable.

2. **Duplicate component inner element with overridable setting/s inside component** - The duplicated element kept the same `override_key`, causing two settings to be overwritten by the same override, which is not a behaviour we support, so it is confusing and unexpected

3. **Create component from inner elements** - Overridable setting remained in elements settings, but weren't synced with the new component's overridable props meta, making them only appear like they were exposed, while actually they aren't

### Solution

- Added sync hooks to revert overridables on `copy` and `duplicate` operations when editing a component
- Revert overridables when creating an unpublished component from existing elements

### ✨ PR Description
Purpose: Fix overridable property handling to revert component overrides to origin values during copy, duplicate, and component creation operations.

Main changes:
- Added sync hook to automatically revert overridable settings when copying or duplicating elements within component editing mode
- Refactored revert-element-overridable-setting.ts into revert-overridable-settings.ts with expanded utilities for container and element data processing
- Updated createUnpublishedComponent to revert overridables before storing component elements to prevent override propagation


packages/packages/core/editor-components/src/utils/revert-overridable-settings.ts
```diff
@@ -0,0 +1,207 @@
...
+
+	updateElementSettings( {
+		id: element.id,
+		props: { component_instance: updatedSetting },
+		withHistory: false,
+	} );
+}
+
+function revertElementSettings( element: V1Element ): void {
+	const settings = element.settings?.toJSON() ?? {};
```
ronros-elementor, line 195: Please use `isAtomicWidget` to eliminate this for any non-v4 widgets

ronros-elementor (changes requested):

packages/packages/core/editor-components/src/utils/revert-overridable-settings.ts
ronros-elementor: sorry was not aware that we actually block v3 widgets from being added into a component, consider it as resolved :)

ronros-elementor (approved):

packages/packages/core/editor-components/src/utils/revert-overridable-settings.ts
MiryamOren: We only allow v4 elements inside components:)

PR #34339, opened 2026-01-20
Internal: Show actual origin value on empty override [ED-22585]

IshayMaya:
### ✨ PR Description
Purpose: Fix component instance override panel to display actual origin values when overrides are cleared, resolving ED-22585 by implementing recursive value resolution through nested component hierarchies.

Main changes:
- Added useResolvedOriginValue hook that recursively resolves origin values through nested component overrides and originPropFields
- Updated override value resolution logic to prioritize resolved override, recursive origin, then fallback originValue
- Exported selectData selector and fixed React.Fragment key to include componentInstanceId preventing render issues

[ED-22585]: https://elementor.atlassian.net/browse/ED-22585?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/packages/core/editor-components/src/components/instance-editing-panel/instance-editing-panel.tsx
```diff
@@ -69,7 +69,7 @@ export function InstanceEditingPanel() {
 					) : (
 						<Stack direction="column" alignItems="stretch">
 							{ groups.map( ( group ) => (
-								<React.Fragment key={ group.id }>
+								<React.Fragment key={ group.id + componentInstanceId }>
```
ronros-elementor, line 72: Maybe better be on the safe, theoretical only, side (so that `aa` `bb` won't have the same key as `a` `abb`)

```suggestion
								<React.Fragment key={ `${group.id}-${componentInstanceId}` }>
```

packages/packages/core/editor-components/src/components/instance-editing-panel/use-resolved-origin-value.tsx
```diff
@@ -0,0 +1,121 @@
+import { type PropValue } from '@elementor/editor-props';
+import { __useSelector as useSelector } from '@elementor/store';
+
+import { componentInstanceOverridePropTypeUtil } from '../../prop-types/component-instance-override-prop-type';
+import { type ComponentInstanceOverride } from '../../prop-types/component-instance-overrides-prop-type';
+import { componentOverridablePropTypeUtil } from '../../prop-types/component-overridable-prop-type';
+import { selectData } from '../../store/store';
+import { type OverridableProp, type PublishedComponent } from '../../types';
+
+export function useResolvedOriginValue( override: ComponentInstanceOverride, overridableProp: OverridableProp ) {
```
ronros-elementor:
```suggestion
export function useResolvedOriginValue( override: ComponentInstanceOverride | null, overridableProp: OverridableProp ) {
```
and then you could remove the `as ComponentInstanceOverride` when calling this hook from `<OverridePropControl />`
(it's also wrong to assume it's non-nullable)

packages/packages/core/editor-components/src/components/instance-editing-panel/instance-editing-panel.tsx
MiryamOren, line 72: Is there a scenario where we have editing panels for two instances open at the same time? 
If the problem is that the content doesn't update when changing instances, maybe we should move the `componentInstanceId` key to the wrapper of the entire panel?

packages/packages/core/editor-components/src/components/instance-editing-panel/__tests__/override-prop-control.test.tsx
```diff
@@ -217,8 +217,182 @@ describe( '<OverridePropControl />', () => {
...
+			renderOverridePropControl( store, overridablePropWithNullOrigin, [ nestedOverridableWithNullValue ] );
+
+			// Assert
+			const input = screen.getByRole( 'textbox' );
+			expect( input ).toHaveValue( 'Innermost Origin Value' );
+		} );
+	} );
 } );
 
+function setupNestedComponents( config: {
```
ronros-elementor: I would maybe change a bit of these mocking functions, to allow passing a _set_ of props - to then make sure it matches the same props with their corresponding widget-level props (goes with my question regarding the ability to resolve the origin value without using the `propKey`)

ronros-elementor (changes requested):

packages/packages/core/editor-components/src/components/instance-editing-panel/instance-editing-panel.tsx
ronros-elementor: Miryam's comment is better 🙈 (resolving mine)

IshayMaya: yeah the issue is content update between components (the keys persist while they shouldn't)
i thought about putting the key above initially but it feels more like a hack (and also might be accidently removed)
i think that this is the right place for it since the purpose is to have the accurate unique key for each group

ronros-elementor (approved):

PR #34375, opened 2026-01-21
Internal: Flex & Div Block looses selection deleting an assigned class [ED-22353]

YaelAvitan:
### ✨ PR Description
Purpose: Fix selection loss when deleting assigned global classes from Class Manager by deferring document reload until panel closure instead of during class deletion.

Main changes:
- Moved document reload logic from delete-class.ts onDelete() to class-manager-panel.tsx onClose() handler to preserve element selection state
- Updated panel API to support async onClose callback that receives onOpen state, enabling post-close operations with state context
- Added Playwright e2e test verifying element overlay remains visible after deleting assigned classes and saving changes


packages/packages/libs/editor-ui/src/components/introduction-modal.tsx
```diff
@@ -48,6 +48,7 @@ export const IntroductionModal = ( { open, handleClose, title, children }: Intro
 					size={ 'medium' }
 					variant="contained"
 					sx={ { minWidth: '135px' } }
+					aria-label={ __( 'Got it introduction', 'elementor' ) }
```
YaelAvitan, line 51: this is for the test
because there are two 'Got it' buttons on the canvas

asafdl: should it be aria-label or aria-role? whats the better practice?

tests/playwright/sanity/modules/v4-tests/global-classes/global-classes-delete-class-overlay.test.ts
```diff
@@ -0,0 +1,42 @@
+import { expect } from '@playwright/test';
+import { parallelTest as test } from '../../../../parallelTest';
+import WpAdminPage from '../../../../pages/wp-admin-page';
+
+test.describe( 'Global Classes - delete assigned class keeps element overlay @v4-tests', () => {
```
asafdl: the describe is the same title as the test. which makes no sense
maybe `Global Classes - Delete Class` for describe?
and then change the filename aswell

packages/packages/core/editor-global-classes/src/components/class-manager/delete-class.ts
```diff
@@ -19,16 +19,14 @@ export const deleteClass = ( id: string ) => {
...
-
 	isDeleted = false;
 };
 
 export const hasDeletedItems = () => isDeleted;
 
 // When deleting a class, we remove it from all the documents that have it applied.
 // In order to reflect the changes in the active document, we need to reload it.
-const reloadDocument = () => {
+export const reloadDocument = () => {
```
asafdl: should move this function, its not used here and doesn't fit here

asafdl: also the comments don't make sense anymore

packages/packages/libs/editor-ui/src/components/introduction-modal.tsx
YaelAvitan: usually use aria-label

RobiFerentz (approved):

asafdl (approved):

PR #34420, opened 2026-01-23
Internal: Mixpanel.json being fetched on every page load [ED-22665]

louiswol94:
### ✨ PR Description
Purpose: Optimize Mixpanel configuration loading by conditionally fetching remote config only when event tracking is enabled, preventing unnecessary API calls on every page load.

Main changes:
- Wrapped get_remote_mixpanel_config() call in conditional check to execute only when $can_send_events is true
- Moved session replays and flags initialization inside conditional block to avoid unnecessary variable assignments
- Reduced remote API calls by skipping Mixpanel config fetch when tracking is disabled or not allowed


hein-obox (approved):

louiswol94:
Fixes https://github.com/elementor/elementor/issues/34206

PR #34500, opened 2026-01-27
Internal: Move Interaction Array To Prop Type Be [ED-22691]

elchugreeva:
## Summary
- Added adapter to wrap interactions data from v1 to v2 format before saving to DB
- Added adapter to unwrap interactions data from v2 to v1 format before sending to frontend
- Updated parser and validation to handle both formats

## Changes
- New `adapter.php` with `wrap_for_db()` and `unwrap_for_frontend()` methods
- Updated `module.php` with filters at appropriate priorities
- Updated `parser.php` and `validation.php` to decode v2 format

Related to ED-22691

### ✨ PR Description
Purpose: Refactor interactions data structure to wrap items array with PropType format for backend storage while maintaining backward compatibility with v1 format for frontend rendering.

Main changes:
- Added Interactions Adapter class to handle bidirectional transformation between v1 (direct array) and v2 (PropType-wrapped) formats
- Implemented data conversion filters in Module to wrap interactions on save and unwrap on load
- Refactored timing value extraction to support both 'number' and 'size' PropType formats using Adapter helper methods


modules/interactions/validation.php
```diff
@@ -318,13 +328,77 @@ private function is_valid_timing_config( $data ) {
...
+
+		return true;
+	}
+
+	/**
+	 * Validate timing value that can be either 'number' or 'size' type.
+	 * - number format: {$$type: 'number', value: 123}
+	 * - size format: {$$type: 'size', value: {size: 123, unit: 'ms'}}
+	 */
+	private function is_valid_timing_value( $data, $key, $min = null ) {
```
mike-elementor:
```suggestion
	private function is_valid_timing_value( $data, $key, $min = null, $max = null ) {
```

so that we use the api in full... (respecting the contract)

modules/atomic-widgets/elements/base/has-atomic-base.php
```diff
@@ -215,6 +218,11 @@ private function extract_prop_value( $data, $key, $default = '' ) {
 		$value = $data[ $key ];
 
 		if ( is_array( $value ) && isset( $value['$$type'] ) && isset( $value['value'] ) ) {
+			// Handle size format: {$$type: 'size', value: {size: X, unit: 'ms'}}
```
Batsirai-muchareva: Let's separate the concerns

by adding a new function that handles the size

```php
	private function extract_size_prop_value( $data, $key, $default = '' ): string {
		$value = $this->extract_prop_value_simple( $data, $key, $default );

		if ( ! array_key_exists( 'size', $value ) ) {
			return $default;
		}

		return $value['size'];
	}
```

modules/atomic-widgets/elements/base/has-atomic-base.php
```diff
@@ -215,6 +218,11 @@ private function extract_prop_value( $data, $key, $default = '' ) {
...
+			if ( 'size' === $value['$$type'] && is_array( $value['value'] ) && isset( $value['value']['size'] ) ) {
+				return $value['value']['size'];
```
Batsirai-muchareva: Are you not supposed to return the full string of size + unit here since we now have diff units so that the handles knows which unit to work with

modules/interactions/validation.php
```diff
@@ -318,13 +326,77 @@ private function is_valid_timing_config( $data ) {
...
+
+		$prop = $data[ $key ];
+
+		if ( ! isset( $prop['$$type'] ) ) {
+			return false;
+		}
+
+		// Accept 'number' format
+		if ( 'number' === $prop['$$type'] ) {
+			return $this->is_valid_number_prop_in_range( $data, $key, $min );
```
mike-elementor:
```suggestion
			return $this->is_valid_number_prop_in_range( $data, $key, $min, $max );
```

modules/interactions/validation.php
```diff
@@ -318,13 +326,77 @@ private function is_valid_timing_config( $data ) {
...
+		}
+
+		// Accept 'size' format
+		if ( 'size' === $prop['$$type'] ) {
+			return $this->is_valid_size_prop_in_range( $data, $key, $min );
```
mike-elementor:
```suggestion
			return $this->is_valid_size_prop_in_range( $data, $key, $min, $max );
```

Batsirai-muchareva:
@elchugreeva 
on files `modules/atomic-widgets/elements/base/has-atomic-base.php` and `modules/atomic-widgets/elements/base/has-template.php`
Maybe consider using a trait Has_Interactions and put the extraction logic to minimise duplicates

This is a suggestion but you are welcome to dismiss it if its a big hustle

modules/atomic-widgets/elements/base/has-atomic-base.php
elchugreeva: it is a future change, once front end knows how to work with units.

mike-elementor (approved):

Batsirai-muchareva (approved):

PR #34501, opened 2026-01-27
Internal: Class & Vars template import [ED-22165]

nicoladj77:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #

### ✨ PR Description
Purpose: Add global styles (classes and variables) import/export functionality to template library with user-selectable merge strategies for handling conflicts.

Main changes:
- Implemented three import modes (match_site, keep_create, keep_flatten) with modal dialog for user selection during template import
- Created utility classes for extracting, merging, and transforming global classes and variables in template snapshots with ID remapping
- Added global styles metadata embedding in template export JSON and processing during import via new AJAX endpoint


nicoladj77:
I've reviewed the PR and implemented some refactorings locally to improve clarity and maintainability. Here is a summary of the proposed changes:

1.  **`Template_Library_Import_Export_Utils` Refactoring**:
    *   Split the monolithic `apply_import_mode_to_content` method into distinct `process_variables_import` and `process_classes_import` methods.
    *   Simplified the `generate_unique_label` logic to be more concise.

2.  **`Manager` Class Cleanup**:
    *   Introduced a `get_validated_json_arg` helper method in `includes/template-library/manager.php`.
    *   Refactored `process_global_styles` to use this helper, removing repetitive JSON validation blocks.

3.  **Snapshot Builders De-duplication**:
    *   In both `Template_Library_Global_Classes_Snapshot_Builder` and `Template_Library_Variables_Snapshot_Builder`, extracted the core iteration logic into a private `process_snapshot` method.
    *   Refactored `merge_snapshot_and_get_id_map` and `create_all_as_new` to use this shared method with strategy callbacks for ID generation.

These changes significantly reduce code duplication and make the logic easier to follow and test.

nicoladj77:
To assist with the code review, here is a high-level architectural summary of the changes introduced in this PR (including the refactoring applied):

### 🎯 Goal
This PR enables the **Import/Export of Global Classes and Global Variables** within the Template Library. It ensures that when a user imports a template, the associated global styles are correctly handled based on a selected conflict resolution strategy.

### 🏗 Architecture & Components

The implementation is built around three main pillars:

#### 1. Central Utility (`Template_Library_Import_Export_Utils`)
This class acts as the orchestrator for the import process.
*   **Responsibility**: It routes the import logic based on the user's selected strategy (`match_site`, `keep_create`, `keep_flatten`).
*   **Refactoring Note**: The logic has been split into `process_variables_import` and `process_classes_import` to handle the two types of globals independently but sequentially (since classes can depend on variables).

#### 2. Snapshot Builders
*   **`Template_Library_Global_Classes_Snapshot_Builder`**
*   **`Template_Library_Variables_Snapshot_Builder`**
*   **Responsibility**:
    *   **Export**: They iterate over the template's elements to extract only the *used* global classes/variables and create a "snapshot" (a portable representation of the repository data).
    *   **Import**: They take an incoming snapshot and merge it into the current site's repository.
*   **Key Logic**: They handle the "ID Mapping". If a global exists or needs to be duplicated, they generate a new ID (e.g., `g-12345` -> `g-67890`) and return a map so the content can be updated.

#### 3. Template Manager Integration (`Manager`)
*   The `import_template` and `process_global_styles` methods in the Template Library Manager have been updated to accept `global_classes` and `global_variables` payloads.
*   It delegates the heavy lifting to the Utils class, ensuring the Manager remains a coordinator rather than containing business logic.

### 🔄 Data Flow (Import)

1.  **Payload**: The import request includes the Template Content + Global Classes Snapshot + Global Variables Snapshot + Import Mode.
2.  **Variable Processing**:
    *   Variables are processed first because Global Classes might reference them.
    *   Based on the mode, variables are either matched, created as new, or flattened.
    *   An `id_map` is generated for any renamed variables.
3.  **Class Processing**:
    *   Classes are processed next.
    *   Crucially, if variables were renamed, the Class Snapshot is first updated to point to the *new* variable IDs.
    *   Classes are then matched, created, or flattened.
4.  **Content Rewriting**:
    *   Finally, the actual Template Content (Elements/Widgets) is traversed.
    *   References to old Class IDs and Variable IDs are replaced with the new IDs from the `id_map`.

### 🧩 Import Modes

*   **`match_site`**: Tries to use existing globals if they match. If a conflict exists (same ID, different value), it generates a new ID to avoid overwriting site data.
*   **`keep_create`**: Ignores existing globals and forces the creation of new ones to ensure the template looks exactly as designed.
*   **`keep_flatten`**: Does not import globals. Instead, it converts the global references into raw values (e.g., `var(--global-color)` becomes `#FF0000`) on the elements themselves.

includes/template-library/sources/base.php
```diff
@@ -461,7 +462,7 @@ public function validate_quota( array $items ) {
 		return new \WP_Error( 'quota_error', 'This source does not support quota validation' );
 	}
 
-	public function prepare_import_template_data( $file_path ) {
+	public function prepare_import_template_data( $file_path, $import_mode = 'match_site' ) {
```
louiswol94, line 464: @nicoladj77 When I look at the current methods inside these sources, `prepare_import_template_data` etc, they don't know about specific features. So can we move the preparation logic into the global classes and global variables modules?

includes/template-library/sources/local.php
```diff
@@ -772,6 +773,28 @@ public function get_data( array $args ) {
 			$data['page_settings'] = $page->get_data( 'settings' );
 		}
 
+		if ( ! empty( $content ) ) {
+			if ( Template_Library_Import_Export_Utils::is_classes_feature_active() ) {
+				$class_ids = \Elementor\Modules\GlobalClasses\Utils\Template_Library_Global_Classes::extract_used_class_ids_from_elements( $content );
```
louiswol94: All over this PR, can you import the classes and just call `Template_Library_Global_Classes::method()` ?

includes/template-library/manager.php
```diff
@@ -682,6 +684,68 @@ public function import_from_json( array $args ) {
...
+	private function get_validated_json_arg( array $args, string $key, bool $is_optional = false ) {
+		if ( ! array_key_exists( $key, $args ) || null === $args[ $key ] || '' === $args[ $key ] ) {
+			return $is_optional ? null : new \WP_Error( "invalid_$key", "Invalid $key." );
+		}
+
+		$value = $args[ $key ];
+		if ( is_string( $value ) ) {
+			$decoded = json_decode( $value, true );
+			if ( JSON_ERROR_NONE !== json_last_error() ) {
+				return new \WP_Error( "invalid_$key", "Invalid JSON $key." );
```
louiswol94, line 759: How will these errors display on the Front end?

includes/template-library/manager.php
```diff
@@ -1064,13 +1129,13 @@ private function bulk_move_template_items( array $args ) {
...
 
-			if ( is_wp_error( $is_quota_valid ) ) {
-				return $is_quota_valid;
+			if ( is_wp_error( $quota_validation ) ) {
+				return $quota_validation;
 			}
 
-			if ( ! $is_quota_valid ) {
+			if ( false === $quota_validation ) {
 				return new \WP_Error( 'quota_error', 'The moving failed because it will pass the maximum templates you can save.' );
```
louiswol94, line 1161: Why do we need these changes?

modules/global-classes/utils/template-library-global-classes-element-transformer.php
```diff
@@ -0,0 +1,129 @@
...
+use Elementor\Core\Utils\Template_Library_Element_Iterator;
+use Elementor\Core\Utils\Template_Library_Import_Export_Utils;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+class Template_Library_Global_Classes_Element_Transformer {
+
+	public static function rewrite_elements_classes_ids( array $elements, array $id_map ): array {
```
louiswol94, line 14: Question, we could not use the existing process_element_export_import_content() which uses the on_import and on_export methods in the controls?

modules/global-classes/utils/template-library-global-classes-element-transformer.php
```diff
@@ -0,0 +1,129 @@
...
+	private static function flatten_class_values( array $class_values, array $element_data, array $items, ?array $ids_to_flatten ): array {
+		$updated_values = [];
+		$element_styles = $element_data['styles'] ?? [];
+
+		foreach ( $class_values as $class_id ) {
+			if ( ! is_string( $class_id ) || '' === $class_id ) {
+				continue;
+			}
+
+			$is_global_class = 0 === strpos( $class_id, 'g-' );
```
louiswol94: Is this a good check? Not sure what to replace it with, but technically the user could also save a class with this prefix?

assets/dev/js/editor/components/template-library/views/parts/import.js
```diff
@@ -89,6 +90,21 @@ TemplateLibraryImportView = Marionette.ItemView.extend( {
 			IntroductionClass: window.elementorModules.editor.utils.Introduction,
 		} );
 
+		if ( fileName.endsWith( '.json' ) ) {
+			try {
+				const jsonContent = JSON.parse( atob( fileData ) );
```
louiswol94, line 97: Is there a different way we can do this? Similar to `hasPageSettings` maybe. We can save meta data about the template if we need to (in the service).

assets/dev/js/editor/components/template-library/component.js
```diff
@@ -177,32 +178,74 @@ export default class Component extends ComponentModalBase {
...
-
-			return;
-		}
-
 		this.manager.layout.showLoadingView();
 
 		this.manager.requestTemplateContent( model.get( 'source' ), model.get( 'template_id' ), {
 			data: {
-				with_page_settings: withPageSettings,
+				with_page_settings: true,
```
louiswol94: How can this always be `true`?

assets/dev/js/editor/components/template-library/manager.js
```diff
@@ -841,6 +841,14 @@ const TemplateLibraryManager = function() {
 		return elementorCommon.ajax.addRequest( 'get_template_data', options );
 	};
 
+	this.hasGlobalStyles = function( templateData ) {
+		const hasClasses = templateData.global_classes?.items &&
+			Object.keys( templateData.global_classes.items ).length > 0;
+		const hasVariables = templateData.global_variables?.data &&
+			Object.keys( templateData.global_variables.data ).length > 0;
+		return hasClasses || hasVariables;
```
louiswol94, line 849:
```suggestion
			Object.keys( templateData.global_variables.data ).length > 0;

		return hasClasses || hasVariables;
```

assets/dev/scss/editor/_templates-modal.scss
```diff
@@ -2117,3 +2117,167 @@ label.elementor-template-library-order-label {
...
+	.dialog-message {
+		padding: 0;
+		overflow: visible;
+	}
+
+	.dialog-lightbox-widget-content {
+		width: 444px;
+		max-width: 444px;
+		border-radius: 4px;
+		box-shadow: 0 9px 46px 8px rgba(0, 0, 0, 0.12), 0 24px 38px 3px rgba(0, 0, 0, 0.14), 0 11px 15px -7px rgba(0, 0, 0, 0.2);
```
louiswol94: Are these styles necessary?

core/utils/template-library-import-export-utils.php
```diff
@@ -0,0 +1,397 @@
...
+	public static function is_classes_feature_active(): bool {
+		return Plugin::instance()->experiments->is_feature_active( Global_Classes_Module::NAME )
+			&& Plugin::instance()->experiments->is_feature_active( Atomic_Widgets_Module::EXPERIMENT_NAME );
+	}
+
+	public static function is_variables_feature_active(): bool {
+		return Plugin::instance()->experiments->is_feature_active( Variables_Module::EXPERIMENT_NAME )
+			&& Plugin::instance()->experiments->is_feature_active( Atomic_Widgets_Module::EXPERIMENT_NAME );
+	}
+}
```
louiswol94, line 281: In general I just feel like this is A LOT of code. I am not convinced we need so much.

includes/template-library/sources/cloud.php
```diff
@@ -197,6 +256,46 @@ private function prepare_template_export( $data ) {
 			'type' => $data['templateType'],
 		];
 
+		// Prefer snapshot stored in cloud content; fallback to current site snapshot.
+		$snapshot = $data['content']['global_classes'] ?? null;
```
louiswol94: What does this mean exactly?

louiswol94:
> To assist with the code review, here is a high-level architectural summary of the changes introduced in this PR (including the refactoring applied):
> 
> ### 🎯 Goal
> This PR enables the **Import/Export of Global Classes and Global Variables** within the Template Library. It ensures that when a user imports a template, the associated global styles are correctly handled based on a selected conflict resolution strategy.
> 
> ### 🏗 Architecture & Components
> The implementation is built around three main pillars:
> 
> #### 1. Central Utility (`Template_Library_Import_Export_Utils`)
> This class acts as the orchestrator for the import process.
> 
> * **Responsibility**: It routes the import logic based on the user's selected strategy (`match_site`, `keep_create`, `keep_flatten`).
> * **Refactoring Note**: The logic has been split into `process_variables_import` and `process_classes_import` to handle the two types of globals independently but sequentially (since classes can depend on variables).
> 
> #### 2. Snapshot Builders
> * **`Template_Library_Global_Classes_Snapshot_Builder`**
> * **`Template_Library_Variables_Snapshot_Builder`**
> * **Responsibility**:
>   
>   * **Export**: They iterate over the template's elements to extract only the _used_ global classes/variables and create a "snapshot" (a portable representation of the repository data).
>   * **Import**: They take an incoming snapshot and merge it into the current site's repository.
> * **Key Logic**: They handle the "ID Mapping". If a global exists or needs to be duplicated, they generate a new ID (e.g., `g-12345` -> `g-67890`) and return a map so the content can be updated.
> 
> #### 3. Template Manager Integration (`Manager`)
> * The `import_template` and `process_global_styles` methods in the Template Library Manager have been updated to accept `global_classes` and `global_variables` payloads.
> * It delegates the heavy lifting to the Utils class, ensuring the Manager remains a coordinator rather than containing business logic.
> 
> ### 🔄 Data Flow (Import)
> 1. **Payload**: The import request includes the Template Content + Global Classes Snapshot + Global Variables Snapshot + Import Mode.
> 2. **Variable Processing**:
>    
>    * Variables are processed first because Global Classes might reference them.
>    * Based on the mode, variables are either matched, created as new, or flattened.
>    * An `id_map` is generated for any renamed variables.
> 3. **Class Processing**:
>    
>    * Classes are processed next.
>    * Crucially, if variables were renamed, the Class Snapshot is first updated to point to the _new_ variable IDs.
>    * Classes are then matched, created, or flattened.
> 4. **Content Rewriting**:
>    
>    * Finally, the actual Template Content (Elements/Widgets) is traversed.
>    * References to old Class IDs and Variable IDs are replaced with the new IDs from the `id_map`.
> 
> ### 🧩 Import Modes
> * **`match_site`**: Tries to use existing globals if they match. If a conflict exists (same ID, different value), it generates a new ID to avoid overwriting site data.
> * **`keep_create`**: Ignores existing globals and forces the creation of new ones to ensure the template looks exactly as designed.
> * **`keep_flatten`**: Does not import globals. Instead, it converts the global references into raw values (e.g., `var(--global-color)` becomes `#FF0000`) on the elements themselves.

@nicoladj77 Can you add the JSON representations for each scenario perhaps?

assets/dev/js/editor/components/template-library/component.js
nicoladj77: @louiswol94 this is because I first ask the question about the classes and variable and then I ask the other question. So I set this to true and moved the logic to after asking the other question. I can roll back this, not sure if product cares about the order or not

assets/dev/scss/editor/_templates-modal.scss
nicoladj77: @louiswol94 they are needed because we can't use the design system in a backbone / marionette context. So I had to add styles, as you can see the template modals have lots of styling.

core/utils/template-library-import-export-utils.php
nicoladj77: @louiswol94 there is lots of logic involved in this as the class and variables are controlled by I think 4 different experiments. I actually copied the code from the I/E task maksim did.
The complexity here is that actually classes can use variables, you can have a class that refers to variables, so the logic is different between the two. And then you have the local class.

assets/dev/js/editor/components/template-library/views/parts/import.js
nicoladj77: @louiswol94 this is for when we upload a new template in the template library through a json file. 
Product has not exactly defined what should happen in this case, but I guess this is needed. But in this case all is client side, at this point so I just need to parse the json and see if it has global classes and variables. WDYT?

core/utils/template-library-import-export-utils.php
nicoladj77: @louiswol94 In any case I agree this is a lot of code, I am trying to slim it down. But the problem is that the task itself was very big, and it was also kind of hard to "divide into subtasks", as maybe approaching this with smaller subtask could have made the PR more manageable. But I also would want to have product start testing this sooner rather than later because the class / variables part is so complex that I am sure things will change.

includes/template-library/sources/base.php
nicoladj77: @louiswol94 agreed. Moving out as much as possible.

includes/template-library/manager.php
nicoladj77: @louiswol94 this would be an error in the ajax call and would be handled by the template modal, not by what we added.

assets/dev/js/editor/components/template-library/component.js
nicoladj77: @louiswol94 refactored

includes/template-library/sources/cloud.php
nicoladj77: @louiswol94 refactored

core/utils/template-library-import-export-utils.php
nicoladj77: @louiswol94 size greatly reduced

nicoladj77:
## High-Level Design: Global Styles Import/Export Architecture

### Overview

This implementation introduces a **filter-based, decoupled architecture** for importing and exporting global styles (classes and variables) within the Elementor template library. The core template library components remain agnostic of specific module implementations, relying on WordPress filters for extensibility.

---

### Architecture Diagram

```
┌─────────────────────────────────────────────────────────────────────┐
│                        Template Library                              │
│  (local.php, cloud.php, base.php, manager.php)                      │
│                                                                      │
│  Uses filters - knows nothing about global classes/variables:        │
│  • elementor/template_library/export/build_snapshots                │
│  • elementor/template_library/get_data/extract_snapshots            │
│  • elementor/template_library/import/process_content                │
└─────────────────────────────────────────────────────────────────────┘
                                   │
                                   ▼
┌─────────────────────────────────────────────────────────────────────┐
│                         Core Utils                                   │
│                                                                      │
│  Template_Library_Import_Export_Utils                                │
│  ├── ID generation, label handling, comparison utilities            │
│  ├── filter_items_by_ids(), build_filtered_order()                  │
│  └── Import mode constants (match_site, keep_create, keep_flatten)  │
│                                                                      │
│  Template_Library_Snapshot_Processor (abstract base)                 │
│  ├── merge_and_get_id_map() - merge strategy                        │
│  ├── create_all_as_new() - create strategy                          │
│  └── process_snapshot() - shared processing logic                   │
│                                                                      │
│  Template_Library_Element_Iterator                                   │
│  └── iterate() - recursive element traversal                        │
└─────────────────────────────────────────────────────────────────────┘
                                   │
                    ┌──────────────┴──────────────┐
                    ▼                              ▼
┌─────────────────────────────┐    ┌─────────────────────────────┐
│   Global Classes Module      │    │     Variables Module         │
│                              │    │                              │
│ Template_Library_Global_     │    │ Template_Library_Variables   │
│ Classes (facade)             │    │ (facade)                     │
│ ├── Filter callbacks         │    │ ├── Filter callbacks         │
│ └── process_global_classes   │    │ └── process_variables        │
│     _import()                │    │     _import()                │
│                              │    │                              │
│ Snapshot_Builder             │    │ Snapshot_Builder             │
│ (extends Snapshot_Processor) │    │ (extends Snapshot_Processor) │
│ ├── get_item_prefix(): 'g-'  │    │ ├── get_item_prefix(): 'e-gv-'│
│ ├── build_snapshot_for_*()   │    │ ├── build_snapshot_for_*()   │
│ └── Module-specific logic    │    │ └── Module-specific logic    │
│                              │    │                              │
│ Element_Transformer          │    │ Element_Transformer          │
│ └── Rewrites/flattens in     │    │ └── Rewrites/flattens in     │
│     element content          │    │     element content          │
└─────────────────────────────┘    └─────────────────────────────┘
```

---

### Import Flow

```
1. Template Import Request (with import_mode)
   │
   ▼
2. base.php::prepare_import_template_data()
   │
   ├── Applies filter: elementor/template_library/import/process_content
   │
   ▼
3. Variables Module (priority 10)
   ├── Processes variables snapshot
   ├── Builds id_map and ids_to_flatten
   └── Passes control to next filter
   │
   ▼
4. Global Classes Module (priority 20)
   ├── Applies: elementor/global_classes/import/transform_snapshot
   │   └── Variables module rewrites variable refs in classes
   ├── Processes classes snapshot
   ├── Transforms element content (rewrite IDs or flatten)
   └── Returns modified content
```

---

### Export Flow

```
1. Template Export Request
   │
   ▼
2. local.php/cloud.php::prepare_template_export()
   │
   ├── Applies filter: elementor/template_library/export/build_snapshots
   │
   ▼
3. Global Classes Module (priority 10)
   ├── Extracts used class IDs from elements
   └── Builds global_classes snapshot
   │
   ▼
4. Variables Module (priority 20)
   ├── Extracts used variable IDs from elements AND classes
   └── Builds global_variables snapshot
   │
   ▼
5. Snapshots merged into export data
```

---

### Import Modes

| Mode | Behavior |
|------|----------|
| `match_site` | Match by ID, create new ID if content differs |
| `keep_create` | Always create with new IDs |
| `keep_flatten` | Convert global styles to inline/local styles |

---

### Key Design Decisions

1. **Filter-based decoupling**: Template library uses `apply_filters()` instead of direct class calls
2. **Priority ordering**: Variables processed before classes (priority 10 vs 20) to handle cross-references
3. **Abstract base class**: `Template_Library_Snapshot_Processor` eliminates ~80 lines of duplicate code
4. **Facade pattern**: Thin facade classes (`Template_Library_Global_Classes`, `Template_Library_Variables`) provide clean filter callback entry points
5. **Cross-module filter**: `elementor/global_classes/import/transform_snapshot` allows variables module to rewrite refs in classes snapshot
6. **Single source of truth for constants**: Import mode constants defined once in `Template_Library_Import_Export_Utils`

nicoladj77:
@louiswol94 this is ready for re review, I refactored the approach a lot and also added testing with json snapshots to see how things are transformed

assets/dev/js/editor/components/template-library/component.js
```diff
@@ -177,32 +178,76 @@ export default class Component extends ComponentModalBase {
...
-			const insertTemplateHandler = this.getImportSettingsDialog();
-
-			insertTemplateHandler.showImportDialog( model );
-
-			return;
-		}
-
 		this.manager.layout.showLoadingView();
 
+		const shouldFetchPageSettings = null === withPageSettings ? model.get( 'hasPageSettings' ) : withPageSettings;
```
nicoladj77, line 229: @louiswol94 discussed with Nehama, we will remove this second pop up

assets/dev/scss/editor/_templates-modal.scss
nicoladj77: @louiswol94 ok now I get it removing them, I am refactoring the CSS anyway

louiswol94 (dismissed):

louiswol94 (dismissed):

assets/dev/js/editor/components/template-library/views/parts/import.js
```diff
@@ -91,6 +92,21 @@ TemplateLibraryImportView = Marionette.ItemView.extend( {
...
+				if ( elementor.templates.hasGlobalStyles( jsonContent ) ) {
```
nicoladj77, line 98: **Empty catch block** — This silently swallows JSON parse errors. If `atob(fileData)` or `JSON.parse()` fails for a non-trivial reason, there's no feedback anywhere. At minimum:
```js
} catch ( parseError ) {
    console.warn( 'Failed to parse template JSON for global styles check:', parseError );
}
```

core/utils/template-library-import-export-utils.php
```diff
@@ -0,0 +1,214 @@
...
+		}
+
+		$existing_labels[] = $label;
+
+		return $item;
+	}
+
+	public static function generate_unique_id( array $existing_ids, string $prefix = 'g-' ): string {
+		$existing = array_fill_keys( $existing_ids, true );
+
```
nicoladj77, line 108: **Potential infinite loop** — While astronomically unlikely with 7-char hex space (~268M combinations), this `do…while` has no safety exit. Consider adding a max-iterations guard:
```php
$attempts = 0;
do {
    if (++$attempts > 1000) {
        throw new \RuntimeException('Unable to generate unique ID');
    }
    // ...
} while (isset($existing[$id]));
```

core/utils/template-library-import-export-utils.php
```diff
@@ -0,0 +1,214 @@
...
+				++$counter;
+				$new_label = $prefix . $base_label . $counter;
+			}
+
+			if ( strlen( $new_label ) > $max_length ) {
+				$available_length = $max_length - strlen( $prefix . $counter );
+				$base_label = substr( $base_label, 0, $available_length );
+				$new_label = $prefix . $base_label . $counter;
+			}
+		} else {
```
nicoladj77, line 168: **`$available_length` used before assignment risk** — In this `else` branch, `$available_length` is only set inside the `if (strlen($new_label) > $max_length)` block, but is referenced in the `while` loop via null coalescing. While technically safe due to `??`, this is fragile. Initialize it explicitly:
```php
$available_length = strlen($original_label);
```
at the start of this `else` block for clarity.

modules/global-classes/utils/template-library-global-classes-snapshot-builder.php
```diff
@@ -0,0 +1,196 @@
...
+		$id = $item['id'] ?? '';
+
+		if ( '' === $id ) {
+			return $item;
+		}
+
+		$parsed = self::parse_snapshot_or_null( [
+			'items' => [ $id => $item ],
+			'order' => [ $id ],
+		] );
```
nicoladj77, line 115: **`is_matching_item` always returns `true`** — This means that when merging, any class with the same label is treated as identical regardless of its variants/props. If a user imports a template where class "Primary Button" has completely different styles than the site's "Primary Button", the incoming definition is silently discarded.

This is documented in the comment, but worth considering: should there at least be a log/warning when the definitions actually differ?

modules/variables/utils/template-library-variables-snapshot-builder.php
```diff
@@ -0,0 +1,236 @@
...
+use Elementor\Plugin;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+class Template_Library_Variables_Snapshot_Builder extends Template_Library_Snapshot_Processor {
+
+	private static ?self $instance = null;
+	private ?Variables_Repository $repository = null;
```
nicoladj77: **Mutable state in singleton** — `$repository` is set via `can_access_repository()` and later used in `load_current_data()` and `save_data()`. If `make()` is reused across different contexts in the same request, the cached `$repository` may be stale. Consider resolving the repository on each call to `load_current_data()` / `save_data()` instead of caching it.

packages/packages/core/editor-global-classes/src/components/global-styles-import-listener.tsx
```diff
@@ -0,0 +1,61 @@
...
+						frontend: {
+							items: importedClasses.items,
+							order: importedClasses.order,
+						},
+					} )
+				);
+			}
+
+			Promise.all( [ apiClient.all( 'preview' ), apiClient.all( 'frontend' ) ] ).then(
+				( [ previewRes, frontendRes ] ) => {
```
nicoladj77: **Missing `.catch()` on `Promise.all`** — If either `apiClient.all('preview')` or `apiClient.all('frontend')` rejects, this creates an unhandled promise rejection. Add error handling:
```ts
Promise.all([apiClient.all('preview'), apiClient.all('frontend')])
  .then(([previewRes, frontendRes]) => {
    // ...
  })
  .catch((err) => {
    console.error('Failed to refresh global classes after import:', err);
  });
```

includes/template-library/sources/local.php
```diff
@@ -744,6 +744,24 @@ public function get_data( array $args ) {
...
+				$template_id,
+				$data
+			);
+
+			if ( ! empty( $snapshots['global_classes'] ) ) {
+				$data['global_classes'] = $snapshots['global_classes'];
+			}
+
+			if ( ! empty( $snapshots['global_variables'] ) ) {
+				$data['global_variables'] = $snapshots['global_variables'];
```
nicoladj77: **DRY opportunity** — This exact `apply_filters` + conditional snapshot extraction block (lines 747-763 in the new code) is repeated verbatim in:
- `cloud.php::get_data()` 
- `cloud.php::prepare_template_export()`
- `cloud.php::format_resource_item_for_create()`
- `local.php::prepare_template_export()`

Consider extracting to a shared method on `base.php` or a trait:
```php
protected function attach_snapshots_to_data(array &$data, array $content, $template_id): void {
    $snapshots = apply_filters('elementor/template_library/export/build_snapshots', [], $content, $template_id, $data);
    if (!empty($snapshots['global_classes'])) $data['global_classes'] = $snapshots['global_classes'];
    if (!empty($snapshots['global_variables'])) $data['global_variables'] = $snapshots['global_variables'];
}
```

modules/global-classes/utils/template-library-global-classes.php
```diff
@@ -0,0 +1,103 @@
...
+			$data
+		);
+
+		$content = $result['content'];
+
+		switch ( $import_mode ) {
+			case Template_Library_Import_Export_Utils::IMPORT_MODE_KEEP_FLATTEN:
+				$content = Template_Library_Global_Classes_Element_Transformer::flatten_elements_classes( $content, $snapshot );
+				break;
+
```
nicoladj77: **Structural duplication with `Template_Library_Variables::process_variables_import`** — The `match_site` / `keep_create` / `keep_flatten` switch-case pattern is nearly identical between this method and its variables counterpart. The only differences are the builder/transformer classes and the key names. A shared strategy or template method could reduce ~60 lines of duplication.

nicoladj77 (commented):
## Code Review: Global Styles (Classes & Variables) Template Import/Export

Overall this is a well-structured implementation. The abstract `Template_Library_Snapshot_Processor` base class provides clean separation, and the three import modes (`match_site`, `keep_create`, `keep_flatten`) are handled consistently across both classes and variables. Below are findings grouped by severity.

---

### Critical / Must-Fix

1. **Singleton pattern with mutable state in `Template_Library_Variables_Snapshot_Builder`** — The `$instance` singleton holds a `$repository` field that is set in `can_access_repository()`. If the method is called in different contexts within the same request (e.g. import multiple templates), the stale `$repository` reference could lead to subtle bugs. Consider making `$repository` resolution stateless (compute it each call), or clearly scope the singleton lifetime.

2. **Empty `catch` block silently swallowing errors** in `import.js` — The outer `catch (parseError)` block at the JSON-parsing step is completely empty. If `JSON.parse(atob(fileData))` fails for a legitimate reason other than "not JSON", the error is silently swallowed. At minimum, log to `console.warn` so debugging isn't a black hole.

3. **Potential infinite loop in `generate_unique_id()`** — If all 7-char hex IDs with the given prefix are exhausted (extremely unlikely but theoretically possible), the `do…while` loop never terminates. Add a safety counter.

---

### Important Improvements

4. **Duplicated snapshot extraction pattern across sources** — The `apply_filters('elementor/template_library/export/build_snapshots', ...)` + conditional `$data['global_classes']` / `$data['global_variables']` block is copy-pasted identically in `local.php::get_data()`, `local.php::prepare_template_export()`, `cloud.php::get_data()`, `cloud.php::prepare_template_export()`, and `cloud.php::format_resource_item_for_create()`. Extract this into a shared method on the `base.php` source (or a trait) to follow DRY.

5. **`process_global_classes_import` and `process_variables_import` duplicate the switch/case structure** — Both `Template_Library_Global_Classes::process_global_classes_import()` and `Template_Library_Variables::process_variables_import()` repeat the same `match_site` / `keep_create` / `keep_flatten` switching pattern with near-identical logic (merge → remap → flatten fallback). Consider extracting a strategy pattern or shared method to reduce the duplication.

6. **`$available_length` may be used uninitialized** in `generate_unique_label()` — In the `else` branch (line ~192 of `template-library-import-export-utils.php`), `$available_length` is used in the `while` loop's fallback via `$available_length ?? strlen($original_label)`, but if `strlen($new_label) <= $max_length` was true initially, `$available_length` is never set, and the null coalescing covers it. However, on subsequent iterations inside the `while`, the `$available_length` from the `if` block inside the loop may shadow the outer scope. This is fragile — assign `$available_length = strlen($original_label)` at the start of the else-branch for clarity.

7. **Missing error handling for `Promise.all` in `GlobalStylesImportListener` (editor-global-classes)** — After the optimistic dispatch, `Promise.all([apiClient.all('preview'), apiClient.all('frontend')])` has `.then()` but no `.catch()`. If either API call fails, the unhandled rejection could surface in the console. Add `.catch()` to gracefully handle network failures.

---

### Suggestions / Nice-to-haves

8. **Test coverage gap for variables** — There are no tests for `Template_Library_Variables`, `Template_Library_Variables_Snapshot_Builder`, or `Template_Library_Variables_Element_Transformer`. The global-classes side appears to have tests elsewhere, but the variables side is entirely untested. Mirror the global-classes test approach.

9. **Magic strings** — Prefixes like `'g-'`, `'e-gv-'`, `'e-'`, length `7` for random strings, and type keys like `'class'`, `'local'` appear as inline literals. Some are already constants in `Template_Library_Import_Export_Utils`, but the rest would benefit from being constants too, especially `'g-'` which appears in both `Template_Library_Global_Classes_Snapshot_Builder::get_item_prefix()` and `Template_Library_Global_Classes_Element_Transformer::is_global_class_id()`.

10. **`is_matching_item` always returns `true` for global classes** — In `Template_Library_Global_Classes_Snapshot_Builder`, if two classes have the same label but completely different variants, the merge still reuses the existing class. This is a design decision (documented in the comment), but worth flagging — importing a class with the same name but different styles will silently drop the incoming definition. Consider at least logging this.

11. **Variable type resolution coupling** — `Variable_Type_Keys` hardcodes the mapping (`color → 'color'`, `font → 'string'`, `size → 'size'`). If a new variable type is added, this mapping must be manually updated. Consider deriving it from the prop type classes themselves if they expose a `get_resolved_type()` method.

12. **Dialog accessibility** — The `global-styles-dialog.js` dialog uses raw radio/checkbox inputs without ARIA labels on the radio group. The `<label>` wrapping helps, but adding `role="radiogroup"` and `aria-labelledby` on the options container would improve screen reader experience.

13. **`showFlatteningWarningIfNeeded` — i18n pluralization** — The flattening warning messages don't distinguish singular vs. plural ("Some styles"). For better UX, consider using `_n()` or similar to provide count-specific messages.

14. **Auto-save before template export** — In `save-template.js`, `$e.run('document/save/auto', { force: true })` is `try/catch`'d with a comment "Proceed even if auto-save fails". This is fine, but if it fails, the exported template may not include the latest edits. Worth a toast notification to the user.

---

### Architecture Positives

- Clean use of WordPress filter hooks (`elementor/template_library/export/build_snapshots`, `elementor/template_library/import/process_content`) to decouple the global-classes and variables modules from the template library core.
- The abstract `Template_Library_Snapshot_Processor` is well-designed — `merge_and_get_id_map()` vs `create_all_as_new()` with a shared `process_snapshot()` template method.
- Proper use of `CustomEvent` for cross-module communication in the editor (import listener pattern).
- The `sanitize_import_mode()` utility correctly defaults to `match_site` for invalid input.
- Good defensive coding throughout (null checks, `is_array()` guards, empty-array short-circuits).

core/utils/template-library-snapshot-processor.php
```diff
@@ -0,0 +1,208 @@
...
+namespace Elementor\Core\Utils;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+abstract class Template_Library_Snapshot_Processor {
+
+
+
```
louiswol94: Too many blank lines

modules/global-classes/utils/template-library-global-classes-element-transformer.php
```diff
@@ -0,0 +1,163 @@
...
+		$local_style_used = false;
+
+		foreach ( $class_values as $class_id ) {
+			if ( ! is_string( $class_id ) || '' === $class_id ) {
+				continue;
+			}
+
+			if ( ! self::should_flatten_class_id( $class_id, $items, $ids_to_flatten ) ) {
+				if ( self::is_global_class_id( $class_id ) ) {
+					if ( null !== $ids_to_flatten && isset( $items[ $class_id ] ) ) {
```
louiswol94: Three nested ifs 🙈

modules/variables/utils/template-library-variables-element-transformer.php
```diff
@@ -0,0 +1,127 @@
...
+
+		return $element_data;
+	}
+
+	private static function rewrite_variable_ids_recursive( $data, array $id_map, array $variable_types ) {
+		if ( ! is_array( $data ) ) {
+			return $data;
+		}
+
+		if ( isset( $data['$$type'] ) && in_array( $data['$$type'], $variable_types, true ) ) {
```
louiswol94: Could possibly do a hash map here instead of in_array.

core/utils/template-library-snapshot-processor.php
nicoladj77: I wonder Linting doesn't catch this

assets/dev/js/editor/components/template-library/component.js
```diff
@@ -2,6 +2,7 @@ import ComponentModalBase from 'elementor-api/modules/component-modal-base';
 import * as commands from './commands/';
 import * as commandsData from './commands-data/';
 import { SAVE_CONTEXTS } from './constants';
+import { showGlobalStylesDialog } from './views/parts/global-styles-dialog';
```
louiswol94, line 5: It seems like the content is transferred twice? Once to display to the user, and once to process on the server. For a large template (hundreds of elements, complex styles), this AJAX payload could be several megabytes of JSON.

But the alternative (a caching strategy) could add some complexity.

I am okay to keep it as is for simplicity's sake. But just worth knowing about it.

nicoladj77: @louiswol94 WDYM Exactly? So the complexity here is that "Import" might mean different things, but basically you could:
 - import a json file from another site, there is a button in the template library modal which allows you to imprt a json )
 - import from the site templates
 - import from the cloud library
So in case of the json file I show the dialog too. I am asking this to product because I am not sure it make sense to ask, but what shoul;d be the default behavior? but will create a follow up ticket for it so we can merge this into the beta

modules/variables/utils/template-library-variables-element-transformer.php
nicoladj77: @louiswol94 refactored https://github.com/elementor/elementor/pull/34501/changes/1204e66d99105b525eb90ec946df7c48cc340809

modules/global-classes/utils/template-library-global-classes-element-transformer.php
nicoladj77: @louiswol94 refactored https://github.com/elementor/elementor/pull/34501/changes/1204e66d99105b525eb90ec946df7c48cc340809

louiswol94 (approved):

PR #34523, opened 2026-01-28
Internal: Remove redundant overrides after deleting their corresponding overridable props [ED-22583]

ronros-elementor:
Continued @MiryamOren 's work on [this draft PR](https://github.com/elementor/elementor/pull/34490)

### ✨ PR Description
Purpose: Remove orphaned component override properties when their corresponding overridable props are deleted to prevent validation errors and data inconsistency.

Main changes:
- Added sanitization hook that filters invalid overridable props before rendering and deletes corresponding overrides
- Implemented caching mechanism to prevent redundant validation cycles using per-component sanitized attribute flags
- Refactored component instance editing to use context provider pattern for cleaner prop threading


packages/packages/core/editor-components/src/components/component-panel-header/use-overridable-props.ts
ronros-elementor, line 1: this is apparently a code duplication, we have this hook already in our store.ts code

packages/packages/core/editor-components/src/utils/tracking.ts
```diff
@@ -5,7 +5,7 @@ import { __getState as getState } from '@elementor/store';
 import { selectCreatedThisSession } from '../store/store';
 import { type ExtendedWindow } from '../types';
 
-export type Source = 'user' | 'mcp_tool';
+export type Source = 'user' | 'mcp_tool' | 'auto';
```
IshayMaya:
```suggestion
export type Source = 'user' | 'mcp_tool' | 'system';
```

packages/packages/core/editor-components/src/components/instance-editing-panel/override-prop-control.tsx
```diff
@@ -73,17 +75,18 @@ function OverrideControl( { overridableProp, overrides }: Props ) {
...
-	if ( ! propType ) {
-		return null;
-	}
-
 	const componentInstanceId = instanceValue.component_id?.value;
 
 	if ( ! componentInstanceId ) {
 		throw new Error( 'Component ID is required' );
 	}
+	const overridableProps = useOverridableProps( componentInstanceId );
```
IshayMaya: maybe we can move these logics up an keep the relevant data for the panel on the top level

packages/packages/core/editor-components/src/store/actions/delete-overridable-prop.ts
```diff
@@ -25,7 +28,9 @@ export function deleteOverridableProp( { componentId, propKey, source }: DeleteP
 		return;
 	}
 
-	revertElementOverridableSetting( prop.elementId, prop.propKey, prop.originValue, propKey );
+	if ( shouldRevertElementSetting ) {
```
IshayMaya: maybe this one should be splitted

packages/packages/core/editor-components/src/utils/filter-valid-overridable-props.ts
```diff
@@ -0,0 +1,89 @@
...
+				return overridableValue.override_key === outerKey;
+			}
+
+			return override.value.override_key === outerKey;
+		} ) ?? null
+	);
+}
+
+export function isExposedPropValid( prop: OverridableProp, visited: Set< string > = new Set() ): boolean {
+	if ( ! prop.originPropFields ) {
```
MiryamOren, line 35: Maybe add a comment to explain that this is the stop condition since this is the original / level zero  / most inner component

packages/packages/core/editor-components/src/utils/filter-valid-overridable-props.ts
```diff
@@ -0,0 +1,89 @@
...
+		return true;
+	}
+
+	const visitKey = `${ prop.elementId }:${ prop.overrideKey }`;
```
MiryamOren: Why do we need it? Is there a chance for a loop?

packages/packages/core/editor-components/src/utils/filter-valid-overridable-props.ts
```diff
@@ -0,0 +1,89 @@
...
+	const innerOverridableProp = getOverridableProp( { componentId, overrideKey: innerOverrideKey } );
+
+	if ( ! innerOverridableProp ) {
+		return false;
+	}
+
+	return isExposedPropValid( innerOverridableProp, visited );
+}
+
+export function filterValidOverridableProps( overridableProps: OverridableProps ): OverridableProps {
```
MiryamOren: Please move this function to the top of the file, as it is the exported function

packages/packages/core/editor-components/src/hooks/use-valid-overridable-props.ts
```diff
@@ -0,0 +1,23 @@
...
+		return undefined;
+	}
+
+	const filteredOverridableProps = filterValidOverridableProps( overridableProps );
+
+	const originalPropsArray = Object.entries( overridableProps.props ?? {} );
+	const propsToDelete = originalPropsArray.filter( ( [ key ] ) => ! filteredOverridableProps.props[ key ] );
+
+	propsToDelete.forEach( ( [ key ] ) => {
+		deleteOverridableProp( { componentId, propKey: key, source: 'system' }, false );
```
MiryamOren: Why do we need the `shouldRevertElementSetting` param? I think it will be better to revert element setting here too, no?

MiryamOren: Anyway if we need it please change it to `options?: { shouldRevertElementSetting = true }` instead of plain boolean for better readability

MiryamOren (commented):
.

packages/packages/core/editor-components/src/utils/filter-valid-overridable-props.ts
```diff
@@ -0,0 +1,89 @@
...
+			const overridableValue = componentOverridablePropTypeUtil.extract( override );
+
+			if ( overridableValue ) {
+				return overridableValue.override_key === outerKey;
+			}
+
+			return override.value.override_key === outerKey;
+		} ) ?? null
+	);
+}
```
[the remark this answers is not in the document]
ronros-elementor: not a duplication, here we look for the outer overrideKey, which could either be an override or an overridable prop value, while the latter looks for the override key of an override prop value only - so if the iterated prop value is an overridable - we would look for its origin value, and only if the origin value is an override - we would check its overrideKey

packages/packages/core/editor-components/src/utils/filter-valid-overridable-props.ts
ronros-elementor: I actually didn't think that part too deep (it was part of the original draft PR of yours) and it made sense to me when I ran through all the code changes, but now looking into it closer - really no reason for that 🙈

packages/packages/core/editor-components/src/store/store.ts
```diff
@@ -51,6 +52,7 @@ export const initialState: ComponentsState = {
 	path: [],
 	currentComponentId: null,
 	updatedComponentNames: {},
+	sanitizedComponents: {},
```
MiryamOren:
```suggestion
    // Components we already filtered/sanitized their overridable props (Rests on every document switch)
	sanitizedComponents: {},
```

packages/packages/core/editor-components/src/hooks/use-valid-overridable-props.ts
```diff
@@ -0,0 +1,31 @@
+import { addSanitizedComponent } from '../store/actions/add-sanitized-component';
+import { deleteOverridableProp } from '../store/actions/delete-overridable-prop';
+import { useIsSanitizedComponent, useOverridableProps } from '../store/store';
+import { type ComponentId, type OverridableProps } from '../types';
+import { filterValidOverridableProps } from '../utils/filter-valid-overridable-props';
+
+export function useValidOverridableProps( componentId: ComponentId | null ): OverridableProps | undefined {
```
MiryamOren: Maybe change to `useSanitizeOverridableProps`? sanitize to be consistent with the optimization mechanism, and sanitize instead of sanitizing to make it clear that we are changing the data in store to the sanitized value, not only returning it

MiryamOren (dismissed):

packages/packages/core/editor-components/src/store/store.ts
```diff
@@ -31,6 +33,10 @@ type ComponentsState = {
 	path: ComponentsPathItem[];
 	currentComponentId: V1Document[ 'id' ] | null;
 	updatedComponentNames: Record< number, string >;
+
+	// We use this map to flag any sanitized attribute of a given component
+	// This map currently resets in response to איק `editor/documents/open` command
```
MiryamOren: איק🥲
```suggestion
	// This map currently resets in response to the `editor/documents/open` command
```

packages/packages/core/editor-components/src/hooks/use-sanitized-overridable-props.ts
```diff
@@ -0,0 +1,31 @@
+import { deleteOverridableProp } from '../store/actions/delete-overridable-prop';
+import { updateComponentSanitizedAttribute } from '../store/actions/update-component-sanitized-attribute';
+import { useIsSanitizedComponent, useOverridableProps } from '../store/store';
+import { type ComponentId, type OverridableProps } from '../types';
+import { filterValidOverridableProps } from '../utils/filter-valid-overridable-props';
+
+export function useSanitizedOverridableProps( componentId: ComponentId | null ): OverridableProps | undefined {
```
MiryamOren: Not `sanitize`?
https://github.com/elementor/elementor/pull/34523#discussion_r2764481030

packages/packages/core/editor-components/src/store/store.ts
ronros-elementor: 🤦‍♂️

MiryamOren (approved):

PR #34537, opened 2026-01-29
Internal: [V4] Add inline editing support to the V4 button widget [ED-21765]

YaelAvitan:
### ✨ PR Description
Purpose: Add inline editing capability to V4 button widget by implementing HTML prop type and Inline_Editing_Control for rich text editing with XSS protection.

Main changes:
- Replace String_Prop_Type with Html_Prop_Type for button text property and add Inline_Editing_Control binding
- Update button template to sanitize HTML content using striptags filter allowing only safe formatting tags
- Modify inline editor toolbar to hide link button for clickable elements (buttons and linked elements)


packages/packages/libs/editor-controls/src/components/inline-editor-toolbar.tsx
```diff
@@ -44,6 +44,7 @@ export const InlineEditorToolbar = ( { editor, elementId, sx = {} }: InlineEdito
 	const toolbarRef = useRef< HTMLDivElement >( null );
 	const linkPopupState = usePopupState( { variant: 'popover' } );
 	const hasLinkOnElement = elementId ? checkIfElementHasLink( elementId ) : false;
```
asafdl: i guess should rework here to `isElementClickable` (button or has link)

packages/packages/libs/editor-controls/src/components/inline-editor-toolbar.tsx
```diff
@@ -183,6 +184,12 @@ export const InlineEditorToolbar = ( { editor, elementId, sx = {} }: InlineEdito
 const checkIfElementHasLink = ( elementId: ElementID ): boolean =>
 	!! getElementSetting< LinkPropValue >( elementId, 'link' )?.value?.destination;
 
+const checkIfElementIsButton = ( elementId: ElementID ): boolean => {
+	const container = getContainer( elementId );
+	const type = container?.model.get( 'widgetType' ) || container?.model.get( 'elType' );
```
asafdl: its never going to be elType

asafdl (dismissed):

modules/atomic-widgets/elements/atomic-button/atomic-button.php
```diff
@@ -47,12 +49,16 @@ protected static function define_props_schema(): array {
...
 				->default( [] ),
 
-			'text' => String_Prop_Type::make()
+			'text' => Html_Prop_Type::make()
 				->default( __( 'Click here', 'elementor' ) )
 				->description( 'The text displayed on the button.' ),
 
 			'link' => Link_Prop_Type::make(),
 
+			'tag' => String_Prop_Type::make()
```
RobiFerentz, line 58: Just a question: Why do we need this? Are we allowing the button be different tags? If so, do we not need to add the control for that in the UI?

RobiFerentz (dismissed):

modules/atomic-widgets/elements/atomic-button/atomic-button.php
YaelAvitan: For default, tiptap renders a p tag and we want it to still be a button tag
This is the way to do it

ronenboxer (approved):

PR #34548, opened 2026-01-30
Fix: Inactive experiments still show green dot [ED-22713]

louiswol94:
### ✨ PR Description
Purpose: Fix incorrect active state indicator in experiment feature settings by conditionally applying the active CSS class based on actual feature status.

Main changes:
- Changed indicator CSS classes from always-active to conditional based on is_feature_active() check
- Added logic to append 'e-experiment__title__indicator--active' class only when feature is actually active
- Removed hardcoded active state class that was causing inactive experiments to display green indicator

### DIFF
--- a/core/experiments/manager.php
+++ b/core/experiments/manager.php
@@ -640,7 +640,11 @@ class Manager extends Base_Object {
 
 		$is_feature_active = $this->is_feature_active( $feature['name'] );
 
-		$indicator_classes = 'e-experiment__title__indicator e-experiment__title__indicator--active';
+		$indicator_classes = 'e-experiment__title__indicator';
+
+		if ( $is_feature_active ) {
+			$indicator_classes .= ' e-experiment__title__indicator--active';
+		}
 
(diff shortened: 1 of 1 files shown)


core/experiments/manager.php
```diff
@@ -640,7 +640,11 @@ private function get_feature_settings_label_html( array $feature ) {
 
 		$is_feature_active = $this->is_feature_active( $feature['name'] );
 
-		$indicator_classes = 'e-experiment__title__indicator e-experiment__title__indicator--active';
+		$indicator_classes = 'e-experiment__title__indicator';
+
+		if ( $is_feature_active ) {
+			$indicator_classes .= ' e-experiment__title__indicator--active';
+		}
```
louiswol94, line 647: Slight regression was caused here: https://github.com/elementor/elementor/commit/1a055f3d779b7e9c3ff7cb86079ae236e9b021c1#diff-963dfdf51558facfee4943cf26dd74f61b3d98bb8cd76235a82f8075fc263013L645-L648

max-zu (approved):

PR #34567, opened 2026-02-02
Internal: Save synced variables to kit

asafdl:
### ✨ PR Description
Purpose: Add V3 compatibility layer to sync V4 design system variables (colors and typography) to V3 Global Styles format for backward compatibility.

Main changes:
- Created Kit_Stylesheet_Extended class that generates V3 CSS variable mappings using filter hook on variable rendering
- Added sync_to_v3 field to Variable entity allowing selective synchronization of individual variables to V3 format
- Implemented CSS renderer extension that appends V3-compatible CSS custom properties alongside V4 variable declarations


modules/design-system-sync/classes/kit-stylesheet-extended.php
```diff
@@ -0,0 +1,41 @@
...
+
+		$label = sanitize_text_field( $variable['label'] ?? '' );
+
+		if ( empty( $label ) ) {
+			return $css;
+		}
+
+		$v3_id = 'v4-' . $label;
+		$type = $variable['type'];
+		$global_type = $this->get_v3_global_type( $type );
```
murzin-nikita: $this->get_v3_global_type( $variable['type'] ?? '' );

murzin-nikita (approved):

modules/design-system-sync/classes/kit-stylesheet-extended.php
```diff
@@ -0,0 +1,40 @@
...
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+class Kit_Stylesheet_Extended {
+	public function register_hooks() {
+		add_filter( 'elementor/variables/css_entry_additional', [ $this, 'add_v3_mapping_for_variable' ], 10, 3 );
+	}
+
+	public function add_v3_mapping_for_variable( string $css, array $variable, string $id ): string {
```
mike-elementor, line 14: we need to keep in mind, that v4 vars, have more types (plus one virtual a.t.m.)
so I would add "white-list" of types, that are supported by V3 (fonts and colors)

modules/variables/classes/css-renderer.php
```diff
@@ -64,7 +64,11 @@ private function build_css_variable_entry( string $id, array $variable ): ?strin
 			return null;
```
mike-elementor, line 64: I would remove this. and simply print additional as a separate hook...
so that variables module is "pure" and has it's full logic, and desing-system-sync module does it's part

asafdl: i need the variable data and i want it to trigger on write var

asafdl: so discussed it with a 3rd party, and while i agree your approach has benifits, for now we are leaving as is.
We might refactor if your approach suites better when adding sync into global classes

RobiFerentz (approved):

PR #34583, opened 2026-02-03
Internal: Remove default frontend assets loader

hein-obox:
### ✨ PR Description
Purpose: Remove default frontend assets loader dependency from atomic widgets base classes to enable conditional script loading only when specific widgets are used.

Main changes:
- Removed automatic ATOMIC_WIDGETS_HANDLER script dependency from Atomic_Element_Base and Atomic_Widget_Base constructors
- Eliminated jQuery dependency from FRONTEND_HANDLERS_HANDLE script registration to reduce default bundle size
- Added explicit dependency declarations with clarifying comments in atomic-tabs and atomic-youtube widget registrations

### DIFF
--- a/modules/atomic-widgets/elements/base/atomic-element-base.php
+++ b/modules/atomic-widgets/elements/base/atomic-element-base.php
@@ -3,7 +3,6 @@
 namespace Elementor\Modules\AtomicWidgets\Elements\Base;
 
 use Elementor\Element_Base;
-use Elementor\Modules\AtomicWidgets\Elements\Loader\Frontend_Assets_Loader;
 use Elementor\Modules\AtomicWidgets\PropDependencies\Manager as Dependency_Manager;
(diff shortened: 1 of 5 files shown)


modules/atomic-widgets/elements/base/atomic-element-base.php
```diff
@@ -33,7 +32,6 @@ public function __construct( $data = [], $args = null ) {
 		$this->styles = $data['styles'] ?? [];
 		$this->interactions = $this->parse_atomic_interactions( $data['interactions'] ?? [] );
 		$this->editor_settings = $data['editor_settings'] ?? [];
-		$this->add_script_depends( Frontend_Assets_Loader::ATOMIC_WIDGETS_HANDLER );
```
hein-obox, line 36: I am not sure whether we need to load this dependency here e.g. inside the editor. But in the original way we are forcing the frontend handlers to load on every page.

PR #34598, opened 2026-02-03
POC: Test conditional loading of interaction handlers

hein-obox:
### ✨ PR Description
Purpose: Implement conditional loading of interaction handlers to prevent unnecessary script enqueueing when interactions are not used or experiment is inactive.

Main changes:
- Added conditional script dependency registration based on interaction IDs presence and experiment activation status
- Refactored enqueue_interactions() to only load scripts in preview mode or when actually enqueued
- Extracted FRONTEND_SCRIPT_HANDLE constant and created get_interactions_script_handles() method for centralized dependency management


hein-obox:
This is just some playing around from my side. This isn't fully tested. I have also not considered the Pro interactions.

Could we conditionally load the interactions handlers? Currently they are loaded on every page load.

modules/interactions/module.php
```diff
@@ -121,14 +122,18 @@ private function register_frontend_scripts() {
 	}
 
 	public function enqueue_interactions(): void {
-		wp_enqueue_script( 'motion-js' );
-		wp_enqueue_script( 'elementor-interactions' );
+		if ( Plugin::$instance->preview->is_preview_mode() ) {
```
Batsirai-muchareva, line 141: This is not correct.
This will stop interactions to work on frontend @hein-obox @heinvv

hein-obox: @Batsirai-muchareva 

Please test it :)

From my testing it worked.

I am loading it conditionally in the widget and element base files as a script dependency.

Please try to prove me wrong :). I am up for a challenge.

modules/interactions/module.php
```diff
@@ -121,14 +122,18 @@ private function register_frontend_scripts() {
...
+			wp_enqueue_script( 'motion-js' );
+			wp_enqueue_script( self::FRONTEND_SCRIPT_HANDLE );
+		}
 
-		wp_localize_script(
-			'elementor-interactions',
-			'ElementorInteractionsConfig',
-			$this->get_config()
-		);
+		if ( wp_script_is( self::FRONTEND_SCRIPT_HANDLE, 'enqueued' ) ) {
```
Batsirai-muchareva, line 146: I don't think this change is necessary since  wp_localize_script silently fails ( does not include the script to the page )

hein-obox: You might be right there. I am not so familiar with that context.

modules/interactions/module.php
Batsirai-muchareva: Challenge accepted 😎

PR #34618, opened 2026-02-04
Fix: Refactor frontend handlers cleanup to support multiple component instances [ED-22763]

MiryamOren:
### The Bug
When adding multiple instances of a component that contains a YouTube widget to a page, the first instance stopped working - the video wasn't displayed at all.
This happened because frontend handler cleanup was mapped by elementId, so when rendering a second element with the same internal ID, it would abort the handlers of the first element.
### Solution
Replaced the ID-based mapping with element reference-based mapping.
Now cleanup is done according to the actual DOM element, ensuring multiple component instances with the same internal element IDs work independently.

### ✨ PR Description
Purpose: Refactor frontend element lifecycle management to use WeakMap-based cleanup, fixing duplicate component instances sharing the same ID from incorrectly aborting each other's handlers.

Main changes:
- Replaced Map-based unmount tracking with WeakMap keyed by element instances to support multiple components with identical IDs
- Removed premature cleanup call from render event handler and added element parameter requirement to destroy event
- Added test coverage verifying handlers remain active when rendering duplicate elements with shared IDs


packages/packages/core/frontend-handlers/src/lifecycle-events.ts
```diff
@@ -30,11 +34,19 @@ export const onElementRender = ( {
 	elementType: string;
 	elementId: string;
 } ) => {
+	const existingEntry = unmountCallbacks.get( element );
+	if ( existingEntry ) {
+		existingEntry.controller.abort();
+		existingEntry.manualUnmount.forEach( ( callback ) => callback() );
```
IshayMaya: no need to delete as well?

packages/packages/core/frontend-handlers/src/lifecycle-events.ts
```diff
@@ -92,34 +104,26 @@ export const onElementRender = ( {
 			manualUnmount.push( unmount );
 		}
 	} );
-
-	if ( ! unmountElementTypeCallbacks.has( elementType ) ) {
```
IshayMaya, line 96: do we have extra/reduandant registrations now?

packages/packages/core/frontend-handlers/src/lifecycle-events.ts
```diff
@@ -92,34 +104,26 @@ export const onElementRender = ( {
...
-	}
-
-	unmountElementTypeCallbacks.get( elementType )?.set( elementId, () => {
-		controller.abort();
-
-		manualUnmount.forEach( ( callback ) => callback() );
-	} );
 };
 
 export const onElementSelectorRender = ( {
```
IshayMaya, line 105: is it still the correct name for this function?

MiryamOren: Yes, we record the selector unmount callbacks in the same map as the element-type unmount callbacks, but this function is still where we attach the handlers by selector & push the cleanup callbacks to the map

packages/packages/core/frontend-handlers/src/lifecycle-events.ts
MiryamOren: Not really, since in next lines (43-46) we set the initial values, which will override the existing entry
```
	const controller = new AbortController();
	const manualUnmount: ( () => void )[] = [];

	unmountCallbacks.set( element, { controller, manualUnmount } );
```

But if you find it confusing, we can delete it explicitly

IshayMaya: maybe we can have a cleanup function that runs in both
but if you think its too much, not a must

IshayMaya (dismissed):

IshayMaya (approved):

PR #34705, opened 2026-02-09
Internal: Align templated elements behavior [ED-22792]

IshayMaya:
### ✨ PR Description
Purpose: Refactor Twig-templated element rendering system to improve render caching, optimize child rerendering, and fix React component lifecycle management in empty views.

Main changes:
- Centralized render state management by extracting cache logic into createTwigRenderState and createRenderCacheState functions
- Optimized child element rendering to reuse existing views when DOM updates are skipped instead of recreating them
- Fixed React component unmounting in AtomicElementEmptyView by adding onBeforeRender cleanup and optional chaining in onDestroy


packages/packages/core/editor-canvas/src/legacy/create-templated-element-type.ts
```diff
@@ -67,17 +69,10 @@ export function createTemplatedElementView( {
...
-		element,
-	} );
+	const setup = createTwigRenderSetup( { renderer, element } );
 
 	return class extends BaseView {
-		#abortController: AbortController | null = null;
-		#childrenRenderPromises: Promise< void >[] = [];
-		#lastResolvedSettingsHash: string | null = null;
-		#domUpdateWasSkipped = false;
+		_abortController: AbortController | null = null;
```
MiryamOren: Why did you remove the #? (I'm not sure why we had it the first time, but is it a style change, or something else?)

modules/atomic-widgets/assets/js/editor/container/atomic-element-empty-view.js
```diff
@@ -6,18 +6,27 @@ export default class AtomicElementEmptyView extends Marionette.ItemView {
...
 
+	unmount = null;
+
 	renderReactDefaultElement( container ) {
 		const { unmount } = ReactUtils.render( <EmptyComponent container={ container } />, this.el );
 		this.unmount = unmount;
 	}
 
+	onBeforeRender() {
+		if ( this.unmount ) {
```
MiryamOren, line 18: Please add a comment to explain these lines

packages/packages/core/editor-canvas/src/legacy/__tests__/twig-rendering-utils.test.ts
```diff
@@ -0,0 +1,115 @@
...
+				renderNewChildren,
+			} );
+
+			// Assert
+			expect( renderNewChildren ).not.toHaveBeenCalled();
+			expect( childView1.render ).toHaveBeenCalled();
+			expect( childView2.render ).toHaveBeenCalled();
+		} );
+
+		it( 'should fall back to renderNewChildren when skipped but no children exist', () => {
```
MiryamOren: Why is this the wanted behaviour? why call `renderNewChildren` if `domUpdateWasSkipped === true`?

packages/packages/core/editor-canvas/src/legacy/create-templated-element-type.ts
```diff
@@ -67,17 +69,10 @@ export function createTemplatedElementView( {
 }: CreateTemplatedElementTypeOptions ): typeof ElementView {
 	const BaseView = createElementViewClassDeclaration();
 
-	const { templateKey, baseStylesDictionary, resolveProps } = setupTwigRenderer( {
-		type,
-		renderer,
-		element,
-	} );
+	const setup = createTwigRenderSetup( { renderer, element } );
```
MiryamOren: maybe
```suggestion
	const state = createTwigRenderState( { renderer, element } );
```

packages/packages/core/editor-canvas/src/legacy/create-templated-element-type.ts
```diff
@@ -119,92 +114,27 @@ export function createTemplatedElementView( {
...
+				signal: this._abortController?.signal,
+				setup,
+				buildContext: ( resolvedSettings ) => ( {
+					id: this.model.get( 'id' ),
+					type,
+					settings: resolvedSettings,
+					base_styles: setup.baseStylesDictionary,
+				} ),
+				attachContent: ( html: string ) => this.$el.html( html ),
+				transformSettings: ( settings ) => this.afterSettingsResolve( settings ),
```
MiryamOren:
```suggestion
				afterSettingsResolve: ( settings ) => this.afterSettingsResolve( settings ),
```

packages/packages/core/editor-canvas/src/legacy/twig-rendering-utils.ts
```diff
@@ -72,60 +58,87 @@ export function createAfterRender< TView extends TwigViewInterface >( view: TVie
...
-	renderer,
+	setup,
 	buildContext,
 	attachContent,
+	transformSettings,
 }: RenderTwigTemplateOptions< TView > ): Promise< void > {
 	view.triggerMethod( 'before:render:template' );
 
-	if ( signal.aborted ) {
+	if ( signal?.aborted ) {
```
MiryamOren, line 104: Why not wrap the whole rendering with signalized process as we did in `_renderTemplate` before this change? 
line 162 in main
`const process = signalizedProcess( this.#abortController?.signal as AbortSignal )`

packages/packages/core/editor-canvas/src/legacy/create-templated-element-type.ts
```diff
@@ -119,92 +114,27 @@ export function createTemplatedElementView( {
...
-
-		#collectChildrenRenderPromises() {
-			this.children?.each( ( childView: ElementView ) => {
-				if ( childView._currentRenderPromise ) {
-					this.#childrenRenderPromises.push( childView._currentRenderPromise );
-				}
+			await renderChildrenWithOptimization( {
+				children: this.children,
+				domUpdateWasSkipped: setup.cacheState.domUpdateWasSkipped,
+				renderNewChildren: () => super._renderChildren(),
```
MiryamOren: I think `renderNewChildren` is a bit confusing - it sounds like it's only for childern we now addded to the view, while it's actually for full render of existing childern too. Maybe it should be named `FullRenderChildren` / `recreateChildren`?

packages/packages/core/editor-canvas/src/legacy/twig-rendering-utils.ts
IshayMaya: the process there is still wrapped with it.
I tried to utilize it here as well but the process here is less linear and it came out less clear so i preferred to keep it like this

packages/packages/core/editor-canvas/src/legacy/__tests__/twig-rendering-utils.test.ts
IshayMaya: It was initially (even before these changes) meant to be a guard, components always have a children, elements as well (even if we remove  a children there should be an empty view as a single children) and simple widgets have no children anyway so the operation is pretty much the same

this can be removed tbh i was just afraid of edge cases with actual empty children for containers

IshayMaya: @MiryamOren added a comment to clarify that that this is a safe fallback to when children is empty.

packages/packages/core/editor-canvas/src/legacy/create-templated-element-type.ts
IshayMaya: you're right, changed it to be just `renderChildren`

MiryamOren (approved):

PR #34755, opened 2026-02-11
Internal: Custom effect on FE support [ED-22862]

Ntnelbaba:
Declare the FE part of Custom interaction
Only support it in register and render location

### ✨ PR Description
Purpose: Add custom effect support to the frontend interactions system, enabling users to define and apply custom animation effects alongside existing fade, slide, and scale effects.

Main changes:
- Added disabled "Custom" option to effect dropdown and customEffects control with conditional visibility based on effect selection
- Refactored EffectType and Direction controls to use dynamic component loading with conditional rendering when effect is not custom
- Updated AnimationPresetPropValue type to include optional custom_effect field and propagated through interaction creation utilities


packages/packages/core/editor-interactions/src/components/interaction-details.tsx
```diff
@@ -102,6 +105,7 @@ function useControlComponent( controlName: InteractionsControlType, isVisible: b
 export const InteractionDetails = ( { interaction, onChange, onPlayInteraction }: InteractionDetailsProps ) => {
 	const trigger = extractString( interaction.trigger, DEFAULT_VALUES.trigger );
 	const effect = extractString( interaction.animation.value.effect, DEFAULT_VALUES.effect );
+	const customEffects = interaction.animation.value[ 'custom-effects' ];
```
mike-elementor: a question, don't we want to align 'key' with `timing_config`
I'm fine either way:
`timing-config` or `custom_effects`

just don't think that mixing both is better...

mike-elementor: maybe, we should look it up in prop-types, if there are samples. so that we're more or less aligned e2e..

mike-elementor (changes requested):

mike-elementor (approved):

PR #34771, opened 2026-02-13
Internal: Style and refactor success/error atomic form msgs [ED-22414]

marcin-el:
### ✨ PR Description
Purpose: Refactor form status messages to use dedicated element types with built-in styling instead of generic Div_Block, enabling consistent UI presentation for success and error states.

Main changes:
- Created Form_Success_Message and Form_Error_Message classes extending Form_Message base class with predefined background/text colors and styling
- Replaced Div_Block element type with specific message types in build_status_message() method and updated default message text
- Registered new form message element types in module.js and module.php, added corresponding editor view factory function


modules/atomic-widgets/assets/js/editor/module.js
```diff
@@ -24,6 +25,8 @@ class Module extends elementorModules.editor.utils.Module {
 		elementor.elementsManager.registerElementType( createDivBlockType() );
 		elementor.elementsManager.registerElementType( createFlexboxType() );
 		elementor.elementsManager.registerElementType( createAtomicFormType() );
+		elementor.elementsManager.registerElementType( createFormMessageType( 'e-form-success-message' ) );
```
asafdl, line 28: why do these need to be mrionette views?

asafdl: we try to limit the use of marionette, is it possible to do this differently?

RobiFerentz: I'm pretty sure we have to do this for atomic container types. Otherwise we cannot have them contain V3 widgets on the canvas.
@asafdl

RobiFerentz (dismissed):

modules/atomic-widgets/assets/js/editor/module.js
```diff
@@ -24,6 +25,8 @@ class Module extends elementorModules.editor.utils.Module {
...
+		elementor.elementsManager.registerElementType( createFormMessageType( 'e-form-error-message' ) );
```
asafdl, line 29: @marcin-el can you try to sync with @Omerisra6 about `create-nested-templated-element-type.ts`
perhaps we can avoid creating a new div-block

marcin-el: I've already tried that but it's not possible without a broader refactor of the atomic form itself

asafdl (dismissed):

asafdl (approved):

PR #34774, opened 2026-02-14
Fix: Handle components inner elements id duplication [ED-22828]

MiryamOren:
### ✨ PR Description
Purpose: Fix component inner element ID duplication to prevent conflicts when multiple instances of the same component exist on a page by implementing hash-based ID generation.

Main changes:
- Implemented hash-based ID generation for component inner elements using nesting path to ensure unique IDs across multiple component instances
- Refactored render context system to support multiple context pairs instead of single context, enabling proper element ID tracking per instance
- Added getContainerByOriginId utility to resolve elements by their original ID within component instance scope for proper element lookup


assets/dev/js/editor/editor-base.js
```diff
@@ -422,6 +422,23 @@ export default class EditorBase extends Marionette.Application {
...
+		}
+
+		const view = $e.components.get( 'document' ).utils.findViewRecursive(
+			parent.children,
+			key,
+			value,
+			false,
+		);
+
+		return view ? view[ 0 ]?.getContainer() : null;
```
ronros-elementor: Maybe -
```suggestion
		return view?.[ 0 ]?.getContainer() ?? null;
```
?

modules/components/utils/format-component-elements-id.php
ronros-elementor, line 1: could it maybe be added to our general "parsing_utils" (`modules/components/utils/parsing-utils.php`) class? name-wise it seems quite relevant to this as well, maybe as a "format_element_ids" function in that class?

ronros-elementor (changes requested):

modules/components/utils/format-component-elements-id.php
MiryamOren: parsing-utils is used for validation & sanitization on data save, so IMO it doesn't really belong there.
If the suggestion is because `Format_Component_Elements_Id` is a bit too specific for a class, I can change it to a general `Component_Utils` class, WDYT?

assets/dev/js/editor/editor-base.js
MiryamOren: Done

ronros-elementor (approved):

PR #34785, opened 2026-02-15
Internal: Add caching to nested templated elements [ED-22993]

IshayMaya:
### ✨ PR Description
Purpose: Add caching mechanism to nested templated elements by tracking resolved settings hash to prevent unnecessary DOM updates and improve rendering performance.

Main changes:
- Implemented settings hash caching with `_lastResolvedSettingsHash` to skip redundant DOM updates when settings haven't changed, controlled by `_domUpdateWasSkipped` flag
- Extracted child rendering orchestration into reusable utility functions `rerenderExistingChildren()` and `waitForChildrenToComplete()` for consistent reuse across templated element types
- Refactored private fields from hash-private syntax (`#`) to protected underscore prefix (`_`) for improved accessibility and extendability in class hierarchies


packages/packages/core/editor-canvas/src/legacy/twig-rendering-utils.ts
ronros-elementor, line 1: I would consider changing the file name as it has very little twig-specific logic here and more of children-rendering logic

ronros-elementor (approved):

packages/packages/core/editor-canvas/src/legacy/twig-rendering-utils.ts
IshayMaya: I have another task to unify the process as well, added your comment to the ticket i'll rename it there as well

PR #34786, opened 2026-02-15
Internal: Admin projection v4/v3 colors [ED-22622]

murzin-nikita:
### ✨ PR Description
Purpose: Normalize V4 variable IDs to lowercase format in the design system sync to ensure consistent color variable mapping between Version 4 and Version 3 global colors in Elementor.

Main changes:
- Standardized V4 variable ID generation to lowercase format using new get_v4_variable_id() utility method
- Added V4 colors to global color selector with grouped UI sections and filter hook integration
- Implemented JavaScript editor script for automatic global data refresh on variable updates via window event listener


core/editor/data/globals/endpoints/base.php
```diff
@@ -11,7 +11,7 @@ protected function register() {
 		parent::register();
 
 		$args = [
-			'id_arg_type_regex' => '[\w]+',
+			'id_arg_type_regex' => '[\w-]+',
```
asafdl, line 14: ?

packages/packages/libs/editor-v1-adapters/src/dispatchers/dispatchers.ts
```diff
@@ -47,6 +47,17 @@ export function openRoute( route: string ) {
...
 }
 
+export function refreshGlobalsCache() {
+	const extendedWindow = window as unknown as ExtendedWindow;
+
+	try {
+		extendedWindow.$e?.components?.get?.( 'globals' )?.refreshGlobalData?.();
+	} catch ( error ) {
+		// eslint-disable-next-line no-console
+		console.warn( 'Failed to refresh globals cache:', error );
```
asafdl: printing this in console on client side probably isn't the way to log the error.

packages/packages/core/editor-variables/src/components/variables-manager/hooks/use-variables-manager-state.ts
```diff
@@ -72,6 +109,11 @@ export const useVariablesManagerState = () => {
 
 		if ( result.success ) {
 			await service.load();
+
+			if ( hasSyncRelatedChanges( originalVariables, variables, deletedVariables ) ) {
```
asafdl: i don't like the way its designed here.
1. i think we should be listening to state on backend not on frontend (since requests to change get sent to backend anyway
2. if we really need to listen to changes, we should add `onVariableChange` window event / store / php hook.
then in design sync module we catch it and refresh cache if needed
this logic shouldn't be injected into the variables frontend

packages/packages/libs/editor-v1-adapters/src/dispatchers/dispatchers.ts
```diff
@@ -47,6 +47,17 @@ export function openRoute( route: string ) {
 	}
 }
 
+export function refreshGlobalsCache() {
```
asafdl: again i think this needs to be done in backend if possible

murzin-nikita: @asafdl this is client-side cache(localStorage), so it should be here

core/editor/data/globals/endpoints/base.php
murzin-nikita: @asafdl this endpoint is used to fetch global values and originally it's not supported ids with dash. But v4 variables could consist dash in it - so I extended an endpoint.

murzin-nikita: sounds a bit risky :)

modules/design-system-sync/classes/global-colors-extension.php
```diff
@@ -97,6 +109,27 @@ public function add_v4_variables_section( Global_Colors $tab ) {
...
 
+	public function add_v4_variables_section_to_color_selector( array $items ): array {
+		$v4_colors = $this->get_v4_color_variables();
+
+		if ( empty( $v4_colors ) ) {
+			return $items;
+		}
+
+		foreach ( $v4_colors as $color ) {
+			$id = 'v4-' . $color['label'];
```
asafdl: do we need a tolower for labels? ask adar maybe?

core/editor/data/globals/endpoints/base.php
ronenboxer: Risky in what way?

modules/design-system-sync/assets/js/editor-variables-sync.js
```diff
@@ -0,0 +1,14 @@
+( function() {
+	'use strict';
+
+	let refreshTimeout;
+
+	window.addEventListener( 'variables:updated', () => {
+		clearTimeout( refreshTimeout );
+		refreshTimeout = setTimeout( () => {
```
ronenboxer, line 8: Can this work?
[queueMicrotask](https://developer.mozilla.org/en-US/docs/Web/API/Window/queueMicrotask)

modules/design-system-sync/classes/global-colors-extension.php
```diff
@@ -97,6 +109,27 @@ public function add_v4_variables_section( Global_Colors $tab ) {
...
+
+			$items[ $id ] = [
+				'id' => $id,
+				'title' => $color['label'],
+				'value' => strtoupper( $color['value'] ),
```
ronenboxer, line 134: Why does the value need to be formatted? Maybe we need to talk to the one in charge and unify it?

murzin-nikita: @ronenelementor for CSS it doesn't matter it's uppercase or lowercase. The issue here in variables manager it's stored in lowercase, but in v3 Global Colors we used uppercased value. So this was added only for UI view. 

Since it's only a virtual projection - we don't save in db these modified values, it's not an issue to change it to any case.

core/editor/data/globals/endpoints/base.php
murzin-nikita: @ronenelementor looks like it can allow to create v3 globals with dashes, not sure what's the purpose of such restriction. 

In general we can leave old regex for write/delete, and create a new one for read(with dash support). 
@asafdl or @davseve what do you think?

asafdl: i think adding dash to id seems fine to me

ronenboxer (approved):

PR #34798, opened 2026-02-16
Internal: Quick Start fatal error when remote is down [ED-22816]

hein-obox:
### ✨ PR Description
Purpose: Fix fatal error in Quick Start feature when remote API endpoints are unavailable by implementing graceful degradation and cached data fallback.

Main changes:
- Refactored EditorAssetsAPI to use stale cached data when fresh API requests fail, preventing fatal errors
- Added validation helpers (has_valid_nested_array, is_non_empty_array) to safely check data structure before accessing nested properties
- Consolidated duplicate remote data fetching logic across 6 modules (events-manager, apps, notifications, home, pro-free-trial-popup) into centralized EditorAssetsAPI


modules/home/assets/js/components/home-screen.js
```diff
@@ -6,8 +6,13 @@ import SideBarPromotion from './sidebar-promotion';
 import ExternalLinksSection from './external-links-section';
 import GetStarted from './get-started-section';
 import CreateWithAIBanner from './create-with-ai-banner';
+import LoadFallbackMessage from './load-fallback-message';
 
 const EditorScreen = ( props ) => {
+	if ( ! props.homeScreenData?.get_started ) {
+		return <LoadFallbackMessage />;
+	}
```
hein-obox, line 14: I am still waiting for approval for this from Nehama.

modules/notifications/api.php
```diff
@@ -30,35 +29,18 @@ public static function get_notifications_by_conditions( $force_request = false )
 	}
 
 	private static function get_notifications( $force_request = false ) {
-		$notifications = self::get_transient( '_elementor_notifications_data' );
```
hein-obox, line 33: The notifications have duplicate logic that is included in the Editor Assets class.

modules/pro-free-trial-popup/module.php
```diff
@@ -214,7 +225,7 @@ private function enqueue_scripts(): void {
 	 * @return array Popup data or empty array if not found
 	 */
 	private function extract_popup_data( array $external_data ): array {
-		if ( ! isset( $external_data['pro-free-trial-popup'] ) || ! is_array( $external_data['pro-free-trial-popup'] ) ) {
+		if ( ! EditorAssetsAPI::has_valid_nested_array( $external_data, [ 'pro-free-trial-popup', 0 ] ) ) {
```
louiswol94: There is double validation here?

Once in `extract_popup_data` and once in `get_external_data`

Also, I am just aware that we started enforcing a certain structure instead of just checking that it's not down. I guess this is okay since if the structure changes remotely the plugin might break.

includes/editor-assets-api.php
```diff
@@ -10,6 +10,10 @@ class EditorAssetsAPI {
 
 	const ASSETS_DATA_KEY = 'ASSETS_DATA_KEY';
 
+	const ASSETS_DATA_EXPIRATION = 'ASSETS_DATA_EXPIRATION';
+
+	const HTTP_STATUS_OK = 200;
```
louiswol94: Maybe use the consts from WordPress HTTP class?

modules/pro-free-trial-popup/module.php
```diff
@@ -13,6 +13,7 @@
 use Elementor\Core\Utils\Ab_Test;
 use Elementor\Core\Isolation\Elementor_Adapter;
```
louiswol94, line 14: Why is pro-free-trial-popup not migrated to EditorAssetsAPI for fetching?

Every other affected module now delegates the actual HTTP fetch+cache to EditorAssetsAPI. pro-free-trial-popup is the odd one out...

modules/notifications/api.php
```diff
@@ -1,12 +1,11 @@
 <?php
 namespace Elementor\Modules\Notifications;
 
+use Elementor\Includes\EditorAssetsAPI;
 use Elementor\User;
 
 class API {
 
-	const NOTIFICATIONS_URL = 'https://assets.elementor.com/notifications/v1/notifications.json';
```
louiswol94, line 8: Why remove the const?

core/common/modules/events-manager/module.php
```diff
@@ -88,18 +91,12 @@ private static function get_subscription() {
...
-		}
-
-		$data = json_decode( wp_remote_retrieve_body( $data ), true );
-
-		if ( empty( $data['mixpanel'] ) || ! is_array( $data['mixpanel'] ) ) {
-			return [];
-		}
-
-		return $data['mixpanel'];
+		return $editor_assets_api->get_assets_data( false, true );
```
louiswol94: Is it deliberate to always bypass cache here? If so,maybe add a comment explaining why.

modules/home/classes/transformations-manager.php
```diff
@@ -37,6 +38,10 @@ public function __construct( $home_screen_data ) {
 	}
 
 	public function run_transformations(): array {
+		if ( ! EditorAssetsAPI::has_valid_nested_array( $this->home_screen_data, [] ) ) {
```
louiswol94: The semantics are not obvious here with an empty array.

```
// In EditorAssetsAPI
public static function is_valid_data( $data ): bool {
    return static::is_non_empty_array( $data );
}
```

The call then reads as plain English:
```
if ( ! EditorAssetsAPI::is_valid_data( $this->home_screen_data ) ) {
    return [];
}
```

modules/pro-free-trial-popup/module.php
hein-obox: We had certain checks to enforce a certain structure. I now have added everywhere in a more consistent way.

Double validation: I have removed the duplicates and move the check to the constructor.

includes/editor-assets-api.php
hein-obox: Implemented

modules/pro-free-trial-popup/module.php
hein-obox: Good catch. Implemented.

modules/notifications/api.php
hein-obox: We can use the original version. I have reverted this change.

core/common/modules/events-manager/module.php
hein-obox: Comment added.
This how it was implemented.

I don't think it would hurt to cache the data for one hour to reduce the number of calls to the editor assets url. It could remove the cache bypass completely.

The disadvantage would be that it would take an hour for the ab testing experiment to be activated after updating editor assets, which might be annoying from a product perspective. But we rarely update these experiments.

What do you think?

Let's ask @max-zu as well.

modules/home/classes/transformations-manager.php
hein-obox: This is correct. Implemented.

includes/editor-assets-api.php
```diff
@@ -67,4 +79,43 @@ private function set_transient( $cache_key, $value, $expiration = '+12 hours' ):
 
 		return update_option( $cache_key, $data, false );
 	}
+
+	private function get_expiration_time(): string {
+		$expiration = $this->config( static::ASSETS_DATA_EXPIRATION );
+		return $expiration ? $expiration : '+1 hour';
```
max-zu: let's maybe move +1hour to const as the default value ?

includes/editor-assets-api.php
```diff
@@ -18,12 +20,22 @@ public function config( $key ): string {
 		return $this->config[ $key ] ?? '';
 	}
 
-	public function get_assets_data( $force_request = false ): array {
+	public function get_assets_data( $force_request = false, bool $skip_cache = false ): array {
```
max-zu: do we need an additional $skip_cache parameter ? $force_request should be enough. there is nothing bad to re-fetch data and put it in cache if force_request = true

core/common/modules/events-manager/module.php
max-zu: I would keep using the cache. One hour is not such a long period, especially considering that experiments don’t change that often.

includes/editor-assets-api.php
hein-obox: Done

includes/editor-assets-api.php
hein-obox: Completely removed. We don't use it anywhere else. We use a one our cache in most places and a 12 hour cache in one other place.

includes/editor-assets-api.php
```diff
@@ -59,12 +69,51 @@ private function get_transient( $cache_key ) {
 		return json_decode( $cache['value'], true );
 	}
 
-	private function set_transient( $cache_key, $value, $expiration = '+12 hours' ): bool {
+	private function set_transient( $cache_key, $value, $expiration ): bool {
```
louiswol94, line 72: Why remove the default? 🤔

hein-obox: It wasn't used anywhere. And we are overwriting it with the 1 hour value inside `get_expiration_time()`. I thought it cleaner to remove it.

louiswol94 (approved):

modules/pro-free-trial-popup/module.php
```diff
@@ -49,13 +50,17 @@ public function __construct() {
 			return;
 		}
 
+		if ( ! EditorAssetsAPI::has_valid_nested_array( $data, [ self::MODULE_NAME, 0 ] ) ) {
```
Ntnelbaba, line 53: @hein-obox $data is undefined 😄 

Image

PR #34810, opened 2026-02-17
Internal: Shape defines string prop type, but content is plain string [ED-23002]

marcin-el:
### ✨ PR Description
Purpose: Upgrade HTML property type from v2 to v3 to enforce type-safe string content wrapping and improve data structure consistency across atomic widgets.

Main changes:
- Replaced Html_V2_Prop_Type with Html_V3_Prop_Type across all atomic widgets (button, heading, paragraph, form, tabs) requiring content values to be wrapped as String_Prop_Type instead of plain strings
- Created new Html_V3_Prop_Type class with validation, sanitization, and nested children support; added corresponding Html_V3_Transformer to extract content during rendering
- Updated all prop shape definitions, dynamic tag mappings, and inline editing logic to use html-v3 type and properly extract string values from nested structure


migrations/operations/html-v2-to-html-v3.json
```diff
@@ -0,0 +1,22 @@
+{
+  "up": [
+    {
+      "op": { "fn": "move", "src": "value.content", "dest": "value.content.value", "clean": false }
+    },
+    {
+      "op": { "fn": "set", "path": "value.content.$$type", "value": "string" },
+      "condition": { "fn": "exists", "path": "value.content.value" }
```
asafdl, line 8: why do we need this condition?

modules/atomic-widgets/prop-types/base/untyped-prop-type.php
```diff
@@ -0,0 +1,70 @@
...
+
+namespace Elementor\Modules\AtomicWidgets\PropTypes\Base;
+
+use Elementor\Modules\AtomicWidgets\PropTypes\Contracts\Prop_Type;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+class Untyped_Prop_Type implements Prop_Type {
```
asafdl: isn't this more like "unkown object prop type"?

migrations/operations/html-v2-to-html-v3.json
```diff
@@ -0,0 +1,22 @@
+{
```
asafdl, line 1: migrations need separate pr in first

modules/atomic-widgets/props-resolver/transformers/settings/html-v3-transformer.php
```diff
@@ -0,0 +1,16 @@
...
+use Elementor\Modules\AtomicWidgets\PropsResolver\Props_Resolver_Context;
+use Elementor\Modules\AtomicWidgets\PropsResolver\Transformer_Base;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+class Html_V3_Transformer extends Transformer_Base {
+	public function transform( $value, Props_Resolver_Context $context ) {
+		return $value['content'] ?? '';
```
asafdl, line 14: ? do we not care about children? also isn't it value['content'].value now?

packages/packages/core/editor-canvas/src/transformers/settings/html-v3-transformer.ts
```diff
@@ -0,0 +1,10 @@
+import { createTransformer } from '../create-transformer';
+
+type HtmlV3Value = {
+	content: string | null;
+	children: Record< string, unknown >;
```
asafdl: here children is object

packages/packages/core/editor-canvas/src/legacy/tabs-model-extensions.ts
```diff
@@ -23,7 +23,10 @@ const tabModelExtensions: ModelExtensions = {
 			...paragraphElement,
 			settings: {
 				...paragraphElement.settings,
-				paragraph: htmlV2PropTypeUtil.create( { content: `Tab ${ position }`, children: [] } ),
+				paragraph: htmlV3PropTypeUtil.create( {
+					content: stringPropTypeUtil.create( `Tab ${ position }` ),
+					children: [],
```
asafdl, line 28: in php empty array is empty array, not object
need something like (object) [] to represent object

packages/packages/libs/editor-props/src/prop-types/html-v3.ts
```diff
@@ -0,0 +1,25 @@
+import { z } from '@elementor/schema';
+
+import { createPropUtils } from '../utils/create-prop-utils';
+import { type ChildElement } from './html-v2';
+import { stringPropTypeUtil } from './string';
+
+const childElementSchema: z.ZodType< ChildElement > = z.lazy( () =>
```
asafdl: lets keep this as unknown, aswell so it behaves like backend

marcin-el: ok done

packages/packages/core/editor-canvas/src/legacy/tabs-model-extensions.ts
marcin-el: but this is ts? children: [] is correct here, even if we serialize it to JSON it becomes [] and php reads it as empty array too?

packages/packages/core/editor-canvas/src/transformers/settings/html-v3-transformer.ts
marcin-el: good catch, fixed

modules/atomic-widgets/props-resolver/transformers/settings/html-v3-transformer.php
marcin-el: transformer receives the value after the resolver has resolved the shape, content is string-prop-type in the shape so resolver unwarps it to just the value before passing into transformer
we dont care about children as they are not needed on php side as twig just outputs the content?
children is also unknown type and since it doesnt contain $$type isTransformable returns false and passes children data untouched

modules/atomic-widgets/prop-types/base/untyped-prop-type.php
marcin-el: yeah, fixed

migrations/operations/html-v2-to-html-v3.json
marcin-el: because content can be null and then null content => move is a no-op => set creates `{"$$type": "string"}` with no value key, we have a test for this too

asafdl: so maybe set -> then move?

asafdl: or i don't know two moves? it seems like this will leave some states as wrong object, which will cause failures

modules/atomic-widgets/prop-types/html-v3-prop-type.php
```diff
@@ -0,0 +1,133 @@
...
+
+class Html_V3_Prop_Type extends Object_Prop_Type {
+	public static function get_key(): string {
+		return 'html-v3';
+	}
+
+	protected function define_shape(): array {
+		return [
+			'content' => String_Prop_Type::make(),
+			'children' => Unknown_Prop_Type::make(),
```
Ntnelbaba: Please add a description to this property so Angie can figure out what to put here if he will want to use it
Also, mark as optional

marcin-el: done

modules/atomic-widgets/prop-types/html-v3-prop-type.php
```diff
@@ -0,0 +1,135 @@
...
+	public static function get_key(): string {
+		return 'html-v3';
+	}
+
+	protected function define_shape(): array {
+		return [
+			'content' => String_Prop_Type::make(),
+			'children' => Unknown_Prop_Type::make()
+				->optional()
+				->description( 'Array of child element objects, each with id, type, optional content, and optional nested children.' ),
```
Ntnelbaba:
```suggestion
				->description( 'Array of child element objects, each with id, type, optional content, and optional nested children. The array is a simple array with no prop type schema, example: [{...}]' ),
```
@marcin-el Fill the missing example please

asafdl (approved):

modules/atomic-widgets/prop-types/base/unknown-prop-type.php
```diff
@@ -0,0 +1,63 @@
...
+use Elementor\Modules\AtomicWidgets\PropTypes\Concerns\Has_Meta;
+use Elementor\Modules\AtomicWidgets\PropTypes\Concerns\Has_Required_Setting;
+use Elementor\Modules\AtomicWidgets\PropTypes\Concerns\Has_Settings;
+use Elementor\Modules\AtomicWidgets\PropTypes\Contracts\Prop_Type;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+class Unknown_Prop_Type implements Prop_Type {
```
ronenboxer, line 14: This is excellent!

packages/packages/core/editor-canvas/src/legacy/tabs-model-extensions.ts
ronenboxer: That should be fine
A different question though, don't we want it to be null?

ronenboxer (approved):

PR #34816, opened 2026-02-17
Tweak: Login/signup flow [ED-22730]

louiswol94:
### ✨ PR Description
Purpose: Add Elementor Pro installation screen to onboarding flow, enabling connected users to install Pro directly during signup with popup-based authentication.
Main changes:
- Implemented two new REST endpoints (`install-pro` and `pro-install-screen`) with Pro eligibility checks and installation capability via Plugin_Installer
- Refactored Login component to use popup-based connect flow with event listeners instead of direct navigation, storing connection state in Redux
- Added Pro Install screen with installation UI, progress tracking, and dismiss functionality; displayed conditionally after successful authentication


packages/packages/apps/onboarding/src/components/app-content.tsx
```diff
@@ -54,11 +58,18 @@ export function AppContent( { onComplete, onClose }: AppContentProps ) {
...
+	const handleConnectSuccess = useCallback( async () => {
+		const result = await checkProInstallScreen();
+		actions.setShouldShowProInstallScreen( result.shouldShowProInstallScreen );
+		actions.setConnected( true );
+	}, [ actions, checkProInstallScreen ] );
+
+	const handleConnect = useElementorConnect( {
+		connectUrl: urls.connect,
+		onSuccess: handleConnectSuccess,
+	} );
```
max-zu, line 72: does it makes sense to move these hooks directly into the Login component? but to be honest, I’m not completely sure if we should do that. On one hand, the Login component is currently very “dumb” -  it just renders something and accepts an onConnect callback, so it can be reused anywhere, On the other hand, it’s unlikely that it will actually be reused anywhere else

max-zu (approved):

packages/packages/apps/onboarding/src/components/app-content.tsx
louiswol94: @max-zu I think it's a better call to keep the Login component "dumb".

It triggers `handleConnectSuccess` which updates multiple state fields and determines whether the Pro install screen shows next. That orchestration logic belongs in `app-content`, not in a screen component imo.

PR #34826, opened 2026-02-18
Fix: Prevent overridable props deletion on element move + fix overrides array structure [ED-22999]

MiryamOren:
## The original bug
When a user edits a component with exposed properties and moves an element with an exposed property, the system mistakenly deletes the element's overridable prop. This deletion then triggers another bug, which corrupts the overrides array structure in existing instances, causing them to disappear from the canvas.

### How to Reproduce (before the fix)
1. Create a component with 2 elements, where both have overridable props (e.g., 2 headings with overridable text)
2. Create an instance of this component
3. Override both props in order
4. Go back to the component editing mode
5. Move the first element under the second - do not move the second one.
6. Get out of edit mode
7. Publish the page and reload
8. **Result:** 
   - The instance disappears from the canvas
   - Attempting to open its editing panel from the structure panel shows a blank panel
   - Entering component edit mode shows the first overridable prop is missing
   - In the database: The overrides array has corrupted structure: `{"1": <override-2>}` instead of `[<override-2>]`
  
### Root Causes

**Bug 1: Overridable props deleted on element move**
- **Problem:** The `cleanup-overridable-props-on-delete` hook couldn't distinguish between actual delete operations and move operations (which are internally implemented as delete + create)
- **Impact:** Moving an element incorrectly triggered prop deletion

**Bug 2: Array structure corruption**  
- **Problem:** When filtering out deleted overrides, `array_filter()` preserved original keys, causing PHP to serialize the result as an associative array (e.g., `{"0": <override-0>, "2": <override-2>, "4": <override-4>}`) instead of an indexed array (e.g., `[<override-0>, <override-2>, <override-4>]`
- **Impact:** TypeScript couldn't parse the corrupted structure, causing instances to fail rendering

**Note:** Bug 2 can also occur independently when users legitimately delete overridable props.

### Bug 1
- Use commands trace in `cleanup-overridable-props-on-delete` hook to detect when deletion is part of a move operation and skip cleanup in those cases

### Bug 2
- Fix: Refactored `Overrides_Prop_Type::sanitize_value()` to wrap the results with array_value to ensure indexed array
- Data recovery will be done in a separate task: ED-23055

### ✨ PR Description
Purpose: Fix deletion of overridable component props during element move operations and ensure overrides array maintains indexed structure to prevent data corruption.

Main changes:
- Added move command detection in cleanup hook to prevent deletion of overridable props during element moves
- Wrapped array_filter with array_values in sanitize_value to convert filtered overrides to indexed array structure
- Enhanced deleteOverridableProp to support bulk deletion and added get_data method to format corrupted overrides data

[ED-22999]: https://elementor.atlassian.net/browse/ED-22999?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/phpunit/elementor/modules/components/prop-types/test-overrides-prop-type.php
ronros-elementor, line 1: could you maybe add a test of the serialization of the sanitized value? 
even as part of that test you added, just assert the value of `json_encode( $result[ 'value' ] )` to be equal to `'[ ... ]'`

because I'm not entirely sure the current tests would fail without the use of `array_values`

packages/packages/core/editor-components/src/store/actions/delete-overridable-prop.ts
```diff
@@ -1,33 +1,45 @@
...
-	propKey: string;
+	propKey: string | string[];
 	source: Source;
 };
 
 export function deleteOverridableProp( { componentId, propKey, source }: DeletePropParams ): void {
 	const overridableProps = selectOverridableProps( getState(), componentId );
 
-	if ( ! overridableProps ) {
+	if ( ! overridableProps || Object.keys( overridableProps.props ).length === 0 ) {
```
ronros-elementor, line 18: I think it can be nicer -

```suggestion
	const { props = {}, groups } = overridableProps ?? {};
	
	if ( Object.keys( props ).length === 0 ) {
```

packages/packages/core/editor-components/src/store/actions/delete-overridable-prop.ts
```diff
@@ -1,33 +1,45 @@
...
 		return;
 	}
 
-	const prop = overridableProps.props[ propKey ];
+	const propKeysToDelete = Array.isArray( propKey ) ? propKey : [ propKey ];
+	const deletedProps: OverridableProp[] = [];
```
ronros-elementor, line 23: I would port the `deletedProps` creation logic to another function

packages/packages/core/editor-components/src/sync/cleanup-overridable-props-on-delete.ts
```diff
@@ -1,17 +1,24 @@
...
 type DeleteCommandArgs = {
 	container?: V1Element;
 	containers?: V1Element[];
 };
 
 export function initCleanupOverridablePropsOnDelete() {
-	registerDataHook( 'dependency', 'document/elements/delete', ( args: DeleteCommandArgs ) => {
+	// This hook is not a real dependency - it doesn't block the execution of the command in any case, only perform side effect.
+	// We use `dependency` and not `after` hook because the `after` hook doesn't include the children of a deleted container
+	// in the callback parameters (as they already were deleted).
```
ronros-elementor, line 16: Thank you for that!

packages/packages/libs/editor-v1-adapters/src/data-hooks/register-data-hook.ts
```diff
@@ -65,8 +80,20 @@ export function registerDataHook< TArgs extends Args = Args, TResult = unknown >
 			return `${ command }--data--${ currentHookId }`;
 		}
 
-		apply( ...args: [ TArgs, TResult ] ) {
-			return callback( ...args );
+		apply( args: TArgs, result?: TResult ) {
+			const hookOptions: HookOptions = {};
+
+			const currentWindow = window as unknown as WindowWithDataHooks;
+			const commandsCurrentTrace = currentWindow.$e?.commands?.currentTrace;
```
ronros-elementor, line 87: 🤯

modules/components/widgets/component-instance.php
```diff
@@ -100,4 +100,18 @@ private function get_merged_overrides( $value ): array {
 
 		return $overrides;
 	}
+
+	public function get_data( $item = null ) {
```
ronros-elementor: still thinking about whether we should execute such logic on all component instances whenever in use, regardless of their state (i.e. may already be sanitized)

this is why we usually optimize for reads, and run such stuff upon write

though the alternative is to deal with it on-demand, upon save, and have our guardrails be implemented on the editor side (js), which means components with faulty overrides would remain broken on visitor view, until any change be done actively by the admin...

so not certain about any of our options regarding already-corrupted data 🤷‍♂️

ronros-elementor (changes requested):

ronros-elementor (approved):

PR #34859, opened 2026-02-23
Internal: Memoizing styles on both client and server sides [ED-23012]

ronros-elementor:
### ✨ PR Description
Purpose: Add memoization for global classes styles on both client and server to improve performance and prevent unnecessary re-rendering of CSS styles.

Main changes:
- Implemented server-side caching in Global_Classes_Repository with cache invalidation on context/data changes
- Added client-side differential updates in useStyleItems hook using Map-based cache to track style changes
- Fixed CSS ordering by reversing global classes array and filtering duplicate style definitions in StyleRenderer


packages/packages/core/editor-canvas/src/hooks/use-style-items.ts
```diff
@@ -11,26 +11,47 @@ import { useOnMount } from './use-on-mount';
...
 
 	const providerAndSubscribers = useMemo( () => {
 		return stylesRepository.getProviders().map( ( provider ): ProviderAndSubscriber => {
+			const providerKey = provider.getKey();
+
+			if ( ! styleItemsCacheRef.current.has( providerKey ) ) {
+				styleItemsCacheRef.current.set( providerKey, { orderedIds: [], itemsById: new Map() } );
+			}
+
+			const cache = styleItemsCacheRef.current.get( providerKey ) as StyleItemsCache;
```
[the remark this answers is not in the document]
ronros-elementor: This is a forbidden type annotation in our lint rules - [@typescript-eslint/no-non-null-assertion](https://typescript-eslint.io/rules/no-non-null-assertion)

modules/global-classes/atomic-global-styles.php
```diff
@@ -37,10 +37,10 @@ private function register_styles( Atomic_Styles_Manager $styles_manager ) {
 		$context = is_preview() ? Global_Classes_Repository::CONTEXT_PREVIEW : Global_Classes_Repository::CONTEXT_FRONTEND;
 
 		$get_styles = function () use ( $context ) {
-			return Global_Classes_Repository::make()->context( $context )->all()->get_items()->map( function( $item ) {
+			return Global_Classes_Repository::make()->context( $context )->all()->get_ordered_items()->map( function( $item ) {
 				$item['id'] = $item['label'];
 				return $item;
-			})->all();
+			})->reverse()->all();
```
MiryamOren: Please add a comment to explain why we apply the reverse

packages/packages/core/editor-canvas/src/hooks/use-style-items.ts
```diff
@@ -54,19 +75,17 @@ export function useStyleItems() {
...
 		() =>
 			Object.values( styleItems )
 				.sort( sortByProviderPriority )
 				.flatMap( ( { items } ) => items )
 				.sort( sortByStateType )
-				.sort( sortByBreakpoint( breakpointsOrder ) ),
-		// eslint-disable-next-line react-hooks/exhaustive-deps
-		[ styleItems, breakpointsOrder.join( '-' ) ]
+				.sort( sortByBreakpoint( breakpoints.map( ( breakpoint ) => breakpoint.id ) ) ),
+		[ styleItems, breakpoints ]
```
MiryamOren: how about 
```suggestion
	const breakpointsOrder = sortByBreakpoint( breakpoints.map( ( breakpoint ) => breakpoint.id ) );

	return useMemo(
		() =>
			Object.values( styleItems )
				.sort( sortByProviderPriority )
				.flatMap( ( { items } ) => items )
				.sort( sortByStateType )
				.sort( sortByBreakpoint( breakpointsOrder ) ),
		// eslint-disable-next-line react-hooks/exhaustive-deps
		[ styleItems, breakpointsOrder.join( '-' ) ]
```

packages/packages/core/editor-canvas/src/hooks/use-style-items.ts
```diff
@@ -157,3 +220,52 @@ function createProviderSubscriber( { provider, renderStyles, setStyleItems }: Cr
 		).flatMap( ( breakpointMap ) => Object.values( breakpointMap ) );
 	}
 }
+
+function getChangedStyleIds( previous: StylesCollection, current: StylesCollection ): string[] {
+	const changedIds: string[] = [];
```
MiryamOren, line 227: Not in this PR, but WDYT about making the providers pass the changed ids to subscribers?

packages/packages/core/editor-canvas/src/hooks/use-style-items.ts
```diff
@@ -157,3 +220,52 @@ function createProviderSubscriber( { provider, renderStyles, setStyleItems }: Cr
...
+}
+
+function getOrderedItems( cache: StyleItemsCache ): StyleItem[] {
+	return cache.orderedIds
+		.map( ( id ) => cache.itemsById.get( id ) )
+		.filter( ( items ): items is StyleItem[] => items !== undefined )
+		.flat();
+}
+
+function updateCacheItems( cache: StyleItemsCache, rendered: StyleItem[] ): void {
```
MiryamOren:
```suggestion
function updateCacheItems( cache: StyleItemsCache, changedItems: StyleItem[] ): void {
```

tests/playwright/sanity/modules/v4-tests/global-classes/global-classes-api-stress.test.ts
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	return { items, order };
+}
+
+test.describe.skip( 'Global Classes API Stress Test @stress', () => {
+	test.setTimeout( 600000 );
+
+	test( 'Create 100 global classes via REST API and verify CSS generation', async ( { page, apiRequests }, testInfo ) => {
+		logProgress( 'Starting API-based stress test...' );
```
MiryamOren, line 132: Nice!

tests/playwright/sanity/modules/v4-tests/global-classes/global-classes-api-stress.test.ts
```diff
@@ -0,0 +1,206 @@
...
+
+		const stylesheets = await page.evaluate( () => {
+			const links = document.querySelectorAll( 'link[rel="stylesheet"]' );
+			return Array.from( links ).map( ( link ) => ( link as HTMLLinkElement ).href );
+		} );
+
+		logProgress( `Found ${ stylesheets.length } stylesheets on frontend` );
+
+		const pageSize = await page.evaluate( () => document.documentElement.outerHTML.length );
+		logProgress( `Page HTML size: ${ ( pageSize / 1024 ).toFixed( 2 ) } KB` );
```
MiryamOren, line 200: Don't we want to assert pageSize is less than X?

tests/playwright/sanity/modules/v4-tests/global-classes/global-classes-stress.test.ts
```diff
@@ -0,0 +1,187 @@
...
+import Breakpoints from '../../../../assets/breakpoints';
+import type EditorPage from '../../../../pages/editor-page';
+import type StyleTab from '../../../../pages/atomic-elements-panel/style-tab';
+import { type Device } from '../../../../types/types';
+import WpAdminPage from '../../../../pages/wp-admin-page';
+import { deleteAllGlobalClasses } from './utils';
+import { timeouts } from '../../../../config/timeouts';
+
+/**
+ * This is an initial set of stress tests, currently meant to be executed manually+locally (against a valid .env file)
```
MiryamOren: Why only locally?

tests/playwright/sanity/modules/v4-tests/global-classes/global-classes-stress.test.ts
```diff
@@ -0,0 +1,187 @@
...
+	await styleTab.setBorderWidth( ( baseValue % 10 ) + 1, 'px' );
+	await styleTab.setBorderRadius( ( baseValue % 20 ) + 5, 'px' );
+	await styleTab.setBorderColor( colors[ ( baseValue + 3 ) % colors.length ] );
+
+	await styleTab.openSection( 'Background' );
+	await styleTab.setBackgroundColor( colors[ ( baseValue + 5 ) % colors.length ] );
+}
+
+function logProgress( message: string ): void {
+	const timestamp = new Date().toISOString();
```
MiryamOren, line 100: 🔥

tests/playwright/sanity/modules/v4-tests/global-classes/global-classes-stress.test.ts
```diff
@@ -0,0 +1,187 @@
...
+			logProgress( `Created and styled class ${ i + 1 }/${ CLASS_COUNT }: ${ className }` );
+
+			if ( 0 === ( i + 1 ) % 10 ) {
+				logProgress( `Progress: ${ i + 1 }/${ CLASS_COUNT } classes created and styled` );
+			}
+		}
+
+		logProgress( 'All classes created and styled. Publishing page...' );
+
+		await editor.publishPage();
```
MiryamOren, line 165: Don't we want to check that these styles are actually rendered? (like, look for the <style> tags)?

tests/playwright/sanity/modules/v4-tests/global-classes/global-classes-api-stress.test.ts
ronros-elementor: at this point I'm not running any assertions as this is just used to automate the population of 100 robust styles violently 😄

but the next step would be to have a stress tests suite running on a schedule, as it could potentially get quite resource-heavy, but in that scenario we would certainly want to run assertions based on some decided-upon thresholds

tests/playwright/sanity/modules/v4-tests/global-classes/global-classes-stress.test.ts
ronros-elementor: like the test above - I don't want anyone to think it's ready to be tested as part of our normal CI, since it's intended for stress testing only, will add a comment to emphasize it, and opened a [placeholder ticket](https://elementor.atlassian.net/browse/ED-23148) for now

also - there should not be any assertions either, and sure how the world would cope with the removal of `expect( true ).toBe( true );` (🫣), but I'll remove it.

regardless I have the global-classes-order test which tests on both canvas and frontend that we do render the styles, and that their ordering affects the final outcome

but as to these stress tests - they'll need to be polished anyway, but at a later time

tests/playwright/sanity/modules/v4-tests/global-classes/global-classes-stress.test.ts
ronros-elementor: could be invoked also against a wp-playground env indeed, changed the comment

packages/packages/core/editor-canvas/src/hooks/use-style-items.ts
ronros-elementor: we'll add that probably as part of [this ticket](https://elementor.atlassian.net/browse/ED-23150) I created, to have a proper selection-strategy for global classes

then this piece of information will be important, because the data we pass as previous/current is being sent from the provider level (as only the provider is in charge of tracking the before/after (for local styles it's driver from the widget.style, globals has their store...)

MiryamOren (dismissed):

MiryamOren (approved):

PR #34869, opened 2026-02-24
Internal: Final onboarding step [ED-23096]

louiswol94:
### ✨ PR Description
Purpose: Implement final onboarding step completion flow with loading screen and automatic redirect to page creation after user finishes setup wizard.

Main changes:
- Added completion screen component with animated progress bar displayed when user completes or skips final onboarding step
- Implemented automatic redirect to createNewPage URL after final step completion, with fallback to editor or dashboard URLs
- Removed onComplete callback prop and replaced with internal state management using isCompleting flag and redirectToNewPage function


tests/playwright/sanity/core/app/modules/e-onboarding/e-onboarding.test.ts
```diff
@@ -264,6 +275,30 @@ test.describe( 'E-Onboarding @e-onboarding', () => {
...
+		await page.route( '**/edit.php**', async ( route ) => {
+			redirectedUrl = route.request().url();
+			await route.fulfill( { status: 200, contentType: 'text/html', body: '<html></html>' } );
+		} );
+
+		await doAndWaitForProgress( page, () =>
+			page.getByRole( 'button', { name: 'Skip' } ).click(),
+		);
+
+		expect( progressRequests.at( -1 ) ).toMatchObject( {
```
hein-obox, line 288: Is there a way to make the step values more dynamic, so that we don't have update this object every time that we add or remove a step?

packages/packages/apps/onboarding/src/components/app-content.tsx
```diff
@@ -101,7 +103,29 @@ export function AppContent( { onComplete, onClose }: AppContentProps ) {
...
+			updateProgress.mutate(
+				{
+					skip_step: true,
+					complete: true,
+					step_index: stepIndex,
+					total_steps: totalSteps,
+				},
+				{
+					onSuccess: redirectToNewPage,
+					onError: redirectToNewPage,
```
nicoladj77, line 123: @louiswol94 what happens if there is an error and the mutation fails? Will the onboaridng be displayed again?

louiswol94: No, the onboarding won't show again even on error. `setIsCompleting(true)` is called synchronously before `updateProgress.mutate`, which immediately swaps the UI to `<CompletionScreen />`. Then both `onSuccess` and `onError` call `redirectToNewPage`, so the user is navigated away regardless of whether the API succeeded. The only edge case would be if the redirect itself fails (e.g. no valid URL), but `urls.dashboard` is always present as a final fallback.

hein-obox (approved):

PR #34878, opened 2026-02-24
Internal: Typography Settings in V3 [ED-22630]

asafdl:
### ✨ PR Description
Purpose: Integrate V4 Atomic typography classes into V3 global typography system by adding filter hook and format conversion for cross-version compatibility.

Main changes:
- Added filter hook and conversion logic to expose V4 Atomic classes in V3 typography selector with proper format mapping
- Updated editor UI labels from "Version 3/4" to "Global Fonts/Atomic Classes" with visual separators between groups
- Implemented comprehensive test coverage for V4-to-V3 typography conversion including edge cases and multi-class scenarios


modules/design-system-sync/classes/global-typography-extension.php
```diff
@@ -65,4 +68,72 @@ public function add_v4_classes_section( Global_Typography $tab ) {
...
+		$v4_items = [];
+
+		foreach ( $v4_classes as $class ) {
+			$label = sanitize_text_field( $class['label'] ?? '' );
+
+			if ( empty( $label ) ) {
+				continue;
+			}
+
+			$id = 'v4-' . $label;
```
RobiFerentz, line 91: is the label here unique? If not, could this not create id collision?

asafdl: labels are unique, can't have 2 classes with same name

RobiFerentz (approved):

PR #34879, opened 2026-02-24
Fix: Component instance editing panel crashes when switching components [ED-23098]

MiryamOren:
### ✨ PR Description
Purpose: Fix component instance editing panel crash caused by hook synchronization issues when switching between components by ensuring element data and settings are properly synced through context providers.

Main changes:
- Modified ElementProvider context to include settings parameter, replacing direct useElementSetting and useElementSettings hook calls to prevent stale data during component switches
- Added useSelectedElementSettings hook to synchronously fetch element and settings together, replacing separate useSelectedElement calls that could desynchronize
- Refactored TransformRepeaterControl to accept showChildrenPerspective prop instead of using useElementCanHaveChildren hook internally


packages/packages/core/editor-components/src/components/instance-editing-panel/instance-editing-panel.tsx
```diff
@@ -95,9 +91,14 @@ export function InstanceEditingPanel() {
...
+	const { element, elementType, settings } = useElement();
+	const extractedSettings = componentInstancePropTypeUtil.extract( settings.component_instance );
+
+	return {
+		settings: extractedSettings,
+		element: {
+			id: element?.id,
+			elementType,
+		},
+	};
```
IshayMaya:
```suggestion
	const { element, settings } = useElement();

	return {
		settings: componentInstancePropTypeUtil.extract( settings.component_instance ),
		element,
	};
```

packages/packages/core/editor-components/src/components/instance-editing-panel/override-prop-control.tsx
```diff
@@ -189,6 +189,8 @@ function OverrideControl( { overridableProp }: InternalProps ) {
 		return null;
 	}
 
+	const settings = getElementSettings< AnyTransformable >( elementId, Object.keys( elementType.propsSchema ) );
```
IshayMaya, line 192: we're not gonna have a problem with not listening to settings change here?

packages/packages/core/editor-components/src/extended/components/component-properties-panel/component-properties-panel.tsx
```diff
@@ -16,27 +15,20 @@ export const { panel, usePanelActions } = createPanel( {
...
-					<Panel>
-						<ComponentPropertiesPanelContent
-							onClose={ () => {
-								closePanel();
-								openEditingPanel();
-							} }
-						/>
-					</Panel>
-				</ElementProvider>
+				<Panel>
```
IshayMaya, line 24: 🔥

packages/packages/core/editor-editing-panel/src/components/editing-panel.tsx
```diff
@@ -24,7 +25,7 @@ export const { Slot: PanelHeaderTopSlot, inject: injectIntoPanelHeaderTop } = cr
 const { useMenuItems } = controlActionsMenu;
 
 export const EditingPanel = () => {
-	const { element, elementType } = useSelectedElement();
+	const { element, elementType, settings } = useSelectedElementSettings< AnyTransformable >();
```
IshayMaya: maybe assign `TValue = AnyTransformable ` in the hook to have it by default

packages/packages/core/editor-editing-panel/src/contexts/element-context.tsx
```diff
@@ -24,3 +26,13 @@ export function useElement() {
...
+
+export function usePanelElementSetting< TValue = AnyTransformable | null >( propKey: string ) {
+	const context = useContext( Context );
+
+	if ( ! context ) {
+		throw new Error( 'usePanelElementSetting must be used within a ElementProvider' );
+	}
+
+	return ( context.settings[ propKey ] ?? null ) as TValue | null;
+}
```
IshayMaya, line 38: maybe, just because it kinda util hook, also maybe we can name it without the "panel"?

```suggestion
export function usePanelElementSetting< TValue = AnyTransformable | null >( propKey: string ) {
	const { settings } = useElement();

	return ( settings[ propKey ] ?? null ) as TValue | null;
}
```

packages/packages/core/editor-editing-panel/src/hooks/use-active-style-def-id.ts
```diff
@@ -27,9 +27,7 @@ function useFirstAppliedClass( appliedClassesIds: string[] ) {
 }
 
 function useAppliedClassesIds( classProp: PropKey ) {
```
IshayMaya: seems like we can just remove it

packages/packages/libs/editor-controls/src/hooks/use-element-can-have-children.ts
```diff
@@ -1,10 +1,7 @@
 import { useMemo } from 'react';
-import { getContainer, useSelectedElement } from '@elementor/editor-elements';
-
-export const useElementCanHaveChildren = (): boolean => {
-	const { element } = useSelectedElement();
-	const elementId = element?.id || '';
+import { getContainer } from '@elementor/editor-elements';
 
+export const useElementCanHaveChildren = ( elementId: string ): boolean => {
```
IshayMaya: 🔥 i think it can just be a function now 
```suggestion
export const canElementHaveChildren = ( elementId: string ): boolean => {
	const container = getContainer( elementId );

	if ( ! container ) {
		return false;
	}

	return container.model.get( 'elType' ) !== 'widget';
};
```

packages/packages/core/editor-components/src/components/instance-editing-panel/override-prop-control.tsx
MiryamOren: No, because it's the settings of the component's inner element, and they can't be changed while the instance editing panel is open, only when we are in component edit mode

packages/packages/core/editor-components/src/components/instance-editing-panel/instance-editing-panel.tsx
MiryamOren: Done 👍

packages/packages/core/editor-editing-panel/src/contexts/element-context.tsx
MiryamOren: I don't like the `panel` too, but I'm not sure about `useElementSetting` - feel like it might be confusing?

IshayMaya (dismissed):

IshayMaya (dismissed):

IshayMaya (approved):

PR #34916, opened 2026-03-01
Internal: Form and Nested Children to Use TWIG [ED-23017]

asafdl:
### ✨ PR Description
Purpose: Migrate Atomic Form and Form Message elements from manual rendering to Twig templates, standardizing the template system and improving form state management through CSS class-based approach.

Main changes:
- Added Twig template files for atomic-form and form-message elements with template context methods
- Replaced data-attribute based form state with CSS class-based state management (form-state-{value})
- Implemented generates_class metadata pattern in editor for automatic CSS class generation from prop values


marcin-el (dismissed):

modules/atomic-widgets/elements/atomic-form/atomic-form.html.twig
```diff
@@ -0,0 +1,20 @@
+{%- set form_state = form_state | default(settings['form-state']) | default('default') -%}
```
murzin-nikita, line 1: this is always be default? since: 
```
$context['form_state'] = 'default';
```

modules/atomic-widgets/module.php
```diff
@@ -434,10 +434,10 @@ private function register_frontend_scripts() {
 	private function add_inline_styles() {
 		$inline_css = implode( '', [
 			'.e-heading-base a, .e-paragraph-base a { all: unset; cursor: pointer; }',
-			'.elementor-element[data-element_type="e-form"][data-form-state="default"] [data-e-state="success"],',
```
murzin-nikita, line 439: should we remove attribute generation in modules/atomic-widgets/elements/atomic-form/atomic-form.php

```
					Key_Value_Prop_Type::generate( [
						'key' => String_Prop_Type::generate( 'data-e-state' ),
						'value' => String_Prop_Type::generate( $state ),
					] ),
```

?

murzin-nikita: this attribute is used in e2e as well

modules/atomic-widgets/elements/atomic-form/atomic-form.html.twig
asafdl: build context works in two cases:
1. create a new form
2. frontend rendering

in both cases, we want to start with default state

modules/atomic-widgets/module.php
asafdl: yes, i don't think its used anymore

murzin-nikita (dismissed):

murzin-nikita (dismissed):

murzin-nikita (approved):

PR #34948, opened 2026-03-02
Internal: Design system sync - fix for breakpoints [ED-23013]

murzin-nikita:
### ✨ PR Description
Purpose: Fix breakpoint handling in design system sync to properly generate responsive CSS for typography classes across desktop, tablet, and mobile breakpoints.

Main changes:
- Refactored breakpoint logic to process all normal state variants per breakpoint instead of only desktop defaults
- Added responsive breakpoint suffix handling for typography properties in V3 format conversion
- Updated CSS generation to output breakpoint-specific styles with device media queries


tests/playwright/sanity/modules/v4-tests/design-sync/design-system-sync-typography.test.ts
```diff
@@ -0,0 +1,91 @@
...
+import WpAdminPage from '../../../../pages/wp-admin-page';
+import { deleteAllGlobalClasses, openClassManager, saveAndCloseClassManagerViaDialog, startSyncToV3 } from '../global-classes/utils';
+
+const CLASS_NAME = 'SyncTypoTest';
+const BREAKPOINT_FONT_SIZES: Record<string, number> = {
+	desktop: 74,
+	tablet: 48,
+};
+
+test.describe( 'Design System Sync - Typography V4→V3 @v4-tests', () => {
```
RobiFerentz: Missing AAA comments

modules/design-system-sync/classes/global-typography-extension.php
```diff
@@ -136,4 +173,18 @@ private function convert_v4_props_to_v3_format( array $v4_props ): array {
 
 		return $v3_format;
 	}
+
+	private function parse_css_size_value( string $css_value ): array {
+		if ( preg_match( '/^(-?[\d.]+)\s*([a-z%]+)$/i', $css_value, $matches ) ) {
```
murzin-nikita: v3 font size control expects array, v4 provide to us just a string value

RobiFerentz (dismissed):

modules/design-system-sync/classes/classes-provider.php
```diff
@@ -87,6 +88,35 @@ public static function get_default_breakpoint_props( array $variants ): array {
 		return [];
 	}
 
+	public static function get_all_normal_state_variant_props( array $variants ): array {
```
asafdl, line 55: this seems to be very similar to the other function `get_default_breakpoint_props`
do we need both of these?

asafdl: i feel like we can group and maybe filter all data to one place, don't need to send the entire props

modules/design-system-sync/classes/global-typography-extension.php
```diff
@@ -98,6 +99,27 @@ public function add_v4_classes_to_typography_selector( array $items ): array {
...
+				if ( Breakpoints_Manager::BREAKPOINT_KEY_DESKTOP === $breakpoint ) {
+					continue;
+				}
+
+				if ( empty( $bp_props ) ) {
+					continue;
+				}
+
+				$resolved_bp_props = $props_resolver->resolve( $schema, $bp_props );
+				$bp_values = $this->convert_v4_props_to_v3_format( $resolved_bp_props );
```
asafdl: this is strange, you resolve the prop into pure css, then turn it into array with [ size, unit ]?

modules/design-system-sync/classes/global-typography-extension.php
```diff
@@ -98,6 +99,27 @@ public function add_v4_classes_to_typography_selector( array $items ): array {
 			$resolved_props = $props_resolver->resolve( $schema, $props );
 			$value = $this->convert_v4_props_to_v3_format( $resolved_props );
 
+			$variants_props = $class['variants_props'] ?? [];
```
asafdl, line 96: i feel like a lot of this logic can be simplified, if the classes provider sends only necessary data + we don't resolve data we need to transform after we resolve it

modules/design-system-sync/classes/kit-stylesheet-extended.php
```diff
@@ -29,14 +32,26 @@ public function add_v3_mapping_css( $post_css ) {
...
-		if ( ! empty( $synced_classes ) ) {
-			$css_entries = array_merge( $css_entries, $this->get_classes_css_entries( $synced_classes ) );
+		if ( ! empty( $grouped_class_entries[ Breakpoints_Manager::BREAKPOINT_KEY_DESKTOP ] ) ) {
+			$css_entries = array_merge( $css_entries, $grouped_class_entries[ Breakpoints_Manager::BREAKPOINT_KEY_DESKTOP ] );
 		}
 
 		if ( ! empty( $css_entries ) ) {
 			$post_css->get_stylesheet()->add_raw_css(
-				':root { ' . implode( ' ', $css_entries ) . ' }'
+				$kit_selector . ' { ' . implode( ' ', $css_entries ) . ' }'
```
asafdl: its global typography, why does it matter if it sits per post class?

modules/design-system-sync/classes/kit-stylesheet-extended.php
```diff
@@ -29,14 +32,26 @@ public function add_v3_mapping_css( $post_css ) {
 		}
 
 		$synced_classes = Classes_Provider::get_synced_classes();
+		$grouped_class_entries = ! empty( $synced_classes ) ? $this->get_classes_css_entries( $synced_classes ) : [];
 
-		if ( ! empty( $synced_classes ) ) {
-			$css_entries = array_merge( $css_entries, $this->get_classes_css_entries( $synced_classes ) );
+		if ( ! empty( $grouped_class_entries[ Breakpoints_Manager::BREAKPOINT_KEY_DESKTOP ] ) ) {
+			$css_entries = array_merge( $css_entries, $grouped_class_entries[ Breakpoints_Manager::BREAKPOINT_KEY_DESKTOP ] );
```
asafdl: the logic here seems very strange, we merge to `$css_entries` only desktop, then we iterate over everything again, and call add_raw_css which probably write to file and not a function that should be iterated.
why can't we prepare the data before and just iterate once?

modules/design-system-sync/module.php
```diff
@@ -73,4 +76,12 @@ public function enqueue_editor_styles() {
 	public function clear_classes_cache() {
 		Classes\Classes_Provider::clear_cache();
 	}
+
+	public function clear_css_cache() {
```
asafdl: whats this related to? how does deleting the post meta help?

tests/phpunit/elementor/modules/design-system-sync/classes/test-global-typography-extension.php
```diff
@@ -314,6 +314,107 @@ public function test_add_v4_classes_to_typography_selector__skips_empty_label()
 		$this->assertEmpty( $result );
 	}
 
+	public function test_add_v4_classes_to_typography_selector__includes_responsive_breakpoint_values() {
+		// Arrange
+		$kit = Plugin::$instance->kits_manager->get_active_kit();
+		$kit->update_json_meta( '_elementor_global_classes', [
+			'items' => [
```
asafdl, line 369: this is just wrong, data in global classes looks like:
```
"props":{"width":{"$$type":"size","value":{"size":400,"unit":"px"}},"font-family":{"$$type":"string","value":"Verdana"}},```
it doesn't look like `'font-family' => 'Roboto',`

modules/design-system-sync/classes/global-typography-extension.php
murzin-nikita: @asafdl I need to render size value into v3 control, it accepts array value

asafdl: so can we do this without transforming from prop? the issue i have is we take 
{ $$type: size, value: .... } resolve it into css string, then use regex to "convert to v3 format" to { size:, unit:}

asafdl (approved):

PR #34984, opened 2026-03-04
Internal: Add Mixpanel Events to the onboarding 2.0 [ED-22732]

max-zu:
## PR Checklist

image

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #

### ✨ PR Description
Purpose: Add Mixpanel analytics tracking to the onboarding 2.0 flow to monitor user journey, feature adoption, and completion metrics.

Main changes:
- Implemented comprehensive Mixpanel event tracking across onboarding steps (login, persona selection, theme installation, pro features) with queue management for offline scenarios
- Added TrackingProvider context with automatic initialization based on user connection status and tracking consent from connect flow
- Integrated events package registration and expanded editor-starter to use Portal-based rendering with preview iframe header control


core/common/modules/connect/apps/base-app.php
```diff
@@ -666,14 +666,15 @@ protected function print_popup_close_script( $url ) {
...
-					'elementor/connect/success/<?php echo esc_attr( Utils::get_super_global_value( $_REQUEST, 'callback_id' ) ); //phpcs:ignore WordPress.Security.NonceVerification.Recommended -- Nonce verification is not required here. ?>',
-					<?php echo wp_json_encode( $data ); ?>
+					'elementor/connect/success/' + cbId,
+					eventData
 				);
 
-				opener.dispatchEvent( new CustomEvent( 'elementor/connect/success' ),
-					<?php echo wp_json_encode( $data ); ?>
-				);
+				opener.dispatchEvent( new CustomEvent( 'elementor/connect/success/' + cbId, { detail: eventData } ) );
```
louiswol94: What about the original `elementor/connect/success` event without the `cbid`? There are multiple apps listening to this already.

core/common/modules/connect/apps/base-app.php
```diff
@@ -666,14 +666,15 @@ protected function print_popup_close_script( $url ) {
...
 				opener.jQuery( 'body' ).trigger(
-					'elementor/connect/success/<?php echo esc_attr( Utils::get_super_global_value( $_REQUEST, 'callback_id' ) ); //phpcs:ignore WordPress.Security.NonceVerification.Recommended -- Nonce verification is not required here. ?>',
-					<?php echo wp_json_encode( $data ); ?>
+					'elementor/connect/success/' + cbId,
+					eventData
 				);
 
-				opener.dispatchEvent( new CustomEvent( 'elementor/connect/success' ),
-					<?php echo wp_json_encode( $data ); ?>
-				);
```
louiswol94, line 676: Good catch!!!!

packages/apps/onboarding/src/analytics/init-tracking.ts
```diff
@@ -0,0 +1,20 @@
+import { canSendEvents, initializeAndEnableTracking, setCanSendEvents } from '@elementor/events';
+
+import type { ConnectSuccessData } from './events';
+
+export function updateLibraryConnectConfig( data: ConnectSuccessData ): void {
+	const config = window.elementorCommon?.config;
```
louiswol94:
```suggestion
    const extendedWindow = window as unknown as ExtendedWindow;
	const config = extendedWindow.elementorCommon?.config;
```
And you need to `import { type ExtendedWindow } from '../../../types';`

packages/.eslintignore
```diff
@@ -1,5 +1,5 @@
 node_modules/
-packages/**/dist/
+**/dist/
```
louiswol94, line 2: ?

packages/apps/onboarding/src/components/app.tsx
```diff
@@ -57,7 +58,9 @@ export function App( props: AppProps ) {
 				<DirectionProvider rtl={ window.document.dir === 'rtl' }>
 					<ThemeProvider colorScheme={ colorScheme } palette="argon-beta">
 						<ToastProvider>
-							<AppContent { ...props } />
+							<TrackingProvider>
```
hein-obox, line 59: Nice

max-zu:
Re **packages/.eslintignore** (line 2, [discussion](https://github.com/elementor/elementor/pull/34984#discussion_r2888663844)): the `**/dist/` entry is intentional. The old pattern `packages/**/dist` did not match `apps/*/dist` (e.g. `apps/onboarding/dist`), so ESLint was still linting built output under `apps/`. Using `**/dist/` ignores all nested `dist` folders in the repo (both under `packages/` and under `apps/`).

louiswol94 (dismissed):

louiswol94 (dismissed):

packages/apps/onboarding/src/hooks/use-onboarding-event.ts
```diff
@@ -0,0 +1,63 @@
...
+import { useTrackingState } from '../analytics/tracking-context';
+
+export function useOnboardingEvent() {
+	const { isActive, activate } = useTrackingState();
+
+	return {
+		trackOnboardingInitialized: () => trackOnboardingInitialized( isActive ),
+		trackLoginType: ( loginType: 'social_login' | 'elementor_login' | 'guest' ) =>
+			trackLoginType( isActive, loginType ),
+		trackConnect: ( success: boolean, error?: string ) => trackConnect( isActive, success, error ),
```
louiswol94: `useOnboardingEvent` recreates all functions every render?

louiswol94 (dismissed):

packages/apps/onboarding/src/hooks/use-onboarding-event.ts
max-zu: good catch!

elementor.php
```diff
@@ -45,7 +45,7 @@
 define( 'ELEMENTOR_ASSETS_URL', ELEMENTOR_URL . 'assets/' );
 
 if ( ! defined( 'ELEMENTOR_EDITOR_EVENTS_MIXPANEL_TOKEN' ) ) {
-	define( 'ELEMENTOR_EDITOR_EVENTS_MIXPANEL_TOKEN', '' );
+	define( 'ELEMENTOR_EDITOR_EVENTS_MIXPANEL_TOKEN', '150605b3b9f979922f2ac5a52e2dcfe9' );
```
louiswol94: Please remove

packages/apps/onboarding/src/steps/components/site-features/pro-plan-notice.tsx
```diff
@@ -27,12 +29,20 @@ interface LicenseNoticeProps {
...
 export function ProPlanNotice( { planName }: LicenseNoticeProps ) {
-	const { urls } = useOnboarding();
+	const { urls, choices } = useOnboarding();
+	const { trackProFeaturesSelected } = useOnboardingEvent();
 	const comparePlansUrl = urls.comparePlans;
 	const theme = useTheme();
 
 	const isOne = 'One' === planName;
 
+	const handleComparePlansClick = useCallback( () => {
```
nicoladj77, line 39: @max-zu I will leave the comment here but it applies basically to the whole PR. 
I think that this approach ends up coupling too much UI and logic. 
So imagine that at some point I would want to do something else when this is clicked, and I would need to add it here too. It doesn't look right. 
So I was more for emitting custom events and listening to them using an event listener, but Mr Smith also rightfully suggested that the best way in react would be to use redux and create a custom redux middleware and in the redux  middlewar just listen to the changes.

nicoladj77 (commented):
Just to expand on the previous comment, moving tracking out of the React components and into the data layer will make this much more maintainable. Since we are using Redux (`@elementor/store`), the standard best practice is to use a **Redux Middleware**.

This keeps our UI components "dumb" (they don't need to know about Mixpanel) and centralizes all analytics logic in one place.

Here is a quick example of how we could implement this:

### 1. Create the Tracking Middleware
Create a new file `analytics/middleware.ts` that listens to Redux actions:

```typescript
import { type Middleware } from '@elementor/store';
import { trackPersonaSelected, trackStepViewed } from './onboarding-tracking';

export const analyticsMiddleware: Middleware = ( store ) => ( next ) => ( action ) => {
    // Let the action run first so the state updates
    const result = next( action );
    const state = store.getState();

    // Intercept specific actions and fire Mixpanel events
    switch ( action.type ) {
        case 'onboarding/setUserChoice': {
            const { key, value } = action.payload;
            if ( key === 'building_for' ) {
                trackPersonaSelected( state.onboarding.isTrackingActive, value );
            }
            break;
        }
        case 'onboarding/nextStep':
        case 'onboarding/goToStep': {
            trackStepViewed( state.onboarding.isTrackingActive, state.onboarding.currentStepId );
            break;
        }
    }

    return result;
};
```

### 2. Register the Middleware
In `app.tsx`, register it before creating the store:

```tsx
import { __createStore, __addMiddleware } from '@elementor/store';
import { analyticsMiddleware } from '../analytics/middleware';

// ...
if ( ! existingStore ) {
    __addMiddleware( analyticsMiddleware );
    existingStore = __createStore();
}
```

### 3. Clean up the Components!
Now we can completely remove `useOnboardingEvent` from files like `building-for.tsx`:

```tsx
// Before:
function handleSelect( value: BuildingForValue ) {
    trackPersonaSelected( value ); // ❌ Leakage! UI shouldn't know about tracking
    actions.setUserChoice( 'building_for', value );
    onComplete( { building_for: value } );
}

// After:
function handleSelect( value: BuildingForValue ) {
    // ✅ Clean. The middleware handles tracking automatically
    actions.setUserChoice( 'building_for', value );
    onComplete( { building_for: value } );
}
```

### What about clicks that *don't* update the state?
For UI interactions that only need tracking (like clicking an external "Compare Plans" link), we can dispatch a "No-Op" action. 

Add an empty reducer to `slice.ts`:
```typescript
trackUiInteraction: (state, action: PayloadAction<{ targetName: string }>) => {
    // Do nothing to the state! This is just a message for the middleware to catch.
},
```

Then the component just fires an intent:
```tsx
const handleComparePlansClick = useCallback(() => {
    actions.trackUiInteraction({ targetName: 'compare plans' });
}, [actions]);
```

This approach makes our components much easier to test, reduces re-renders (since we aren't passing tracking context everywhere), and keeps the Mixpanel logic completely decoupled. WDYT?

packages/apps/onboarding/src/steps/components/site-features/pro-plan-notice.tsx
max-zu: refactoring is not the part of my task.

max-zu:
> Just to expand on the previous comment, moving tracking out of the React components and into the data layer will make this much more maintainable. Since we are using Redux (`@elementor/store`), the standard best practice is to use a **Redux Middleware**.
> 
> This keeps our UI components "dumb" (they don't need to know about Mixpanel) and centralizes all analytics logic in one place.
> 
> Here is a quick example of how we could implement this:
> 
> ### 1. Create the Tracking Middleware
> Create a new file `analytics/middleware.ts` that listens to Redux actions:
> 
> ```ts
> import { type Middleware } from '@elementor/store';
> import { trackPersonaSelected, trackStepViewed } from './onboarding-tracking';
> 
> export const analyticsMiddleware: Middleware = ( store ) => ( next ) => ( action ) => {
>     // Let the action run first so the state updates
>     const result = next( action );
>     const state = store.getState();
> 
>     // Intercept specific actions and fire Mixpanel events
>     switch ( action.type ) {
>         case 'onboarding/setUserChoice': {
>             const { key, value } = action.payload;
>             if ( key === 'building_for' ) {
>                 trackPersonaSelected( state.onboarding.isTrackingActive, value );
>             }
>             break;
>         }
>         case 'onboarding/nextStep':
>         case 'onboarding/goToStep': {
>             trackStepViewed( state.onboarding.isTrackingActive, state.onboarding.currentStepId );
>             break;
>         }
>     }
> 
>     return result;
> };
> ```
> 
> ### 2. Register the Middleware
> In `app.tsx`, register it before creating the store:
> 
> ```tsx
> import { __createStore, __addMiddleware } from '@elementor/store';
> import { analyticsMiddleware } from '../analytics/middleware';
> 
> // ...
> if ( ! existingStore ) {
>     __addMiddleware( analyticsMiddleware );
>     existingStore = __createStore();
> }
> ```
> 
> ### 3. Clean up the Components!
> Now we can completely remove `useOnboardingEvent` from files like `building-for.tsx`:
> 
> ```tsx
> // Before:
> function handleSelect( value: BuildingForValue ) {
>     trackPersonaSelected( value ); // ❌ Leakage! UI shouldn't know about tracking
>     actions.setUserChoice( 'building_for', value );
>     onComplete( { building_for: value } );
> }
> 
> // After:
> function handleSelect( value: BuildingForValue ) {
>     // ✅ Clean. The middleware handles tracking automatically
>     actions.setUserChoice( 'building_for', value );
>     onComplete( { building_for: value } );
> }
> ```
> 
> ### What about clicks that _don't_ update the state?
> For UI interactions that only need tracking (like clicking an external "Compare Plans" link), we can dispatch a "No-Op" action.
> 
> Add an empty reducer to `slice.ts`:
> 
> ```ts
> trackUiInteraction: (state, action: PayloadAction<{ targetName: string }>) => {
>     // Do nothing to the state! This is just a message for the middleware to catch.
> },
> ```
> 
> Then the component just fires an intent:
> 
> ```tsx
> const handleComparePlansClick = useCallback(() => {
>     actions.trackUiInteraction({ targetName: 'compare plans' });
> }, [actions]);
> ```
> 
> This approach makes our components much easier to test, reduces re-renders (since we aren't passing tracking context everywhere), and keeps the Mixpanel logic completely decoupled. WDYT?

I agree that the solution you suggested looks more interesting, but in my opinion it’s super over-engineered, and you actually answered your own arguments in your own comment (or Mr. Smith did) about “coupling too much UI and logic.”

First, moving all these triggers into the data layer won’t really eliminate the coupling between UI and logic (for example, handling events like clicks, hovers, etc.), which Mr. Smith himself mentioned below and suggested workarounds for in the form of some no-op actions.

Also, the proposed approach assumes very strong coupling to a specific technology (Redux), while the more modern React approach is to use hooks and contexts. So essentially the current approach could be moved much more easily and cleanly into context/custom hooks.

Additionally, event tracking is a fairly simple action. But if we implement it via Redux middleware, we’ll need to understand and maintain the entire chain of business logic just to figure out how a particular action works, what it should trigger, etc. Adding new events or changing existing ones could easily turn into a big headache.

So the current solution should be considered a compromise between simplicity, speed of implementation, and architecture. That said, the option you suggested is definitely interesting.

nicoladj77:
> I agree that the solution you suggested looks more interesting, but in my opinion it’s super over-engineered, and you actually answered your own arguments in your own comment (or Mr. Smith did) about “coupling too much UI and logic.”
> 
> First, moving all these triggers into the data layer won’t really eliminate the coupling between UI and logic (for example, handling events like clicks, hovers, etc.), which Mr. Smith himself mentioned below and suggested workarounds for in the form of some no-op actions.
> 
> Also, the proposed approach assumes very strong coupling to a specific technology (Redux), while the more modern React approach is to use hooks and contexts. So essentially the current approach could be moved much more easily and cleanly into context/custom hooks.
> 
> Additionally, event tracking is a fairly simple action. But if we implement it via Redux middleware, we’ll need to understand and maintain the entire chain of business logic just to figure out how a particular action works, what it should trigger, etc. Adding new events or changing existing ones could easily turn into a big headache.
> 
> So the current solution should be considered a compromise between simplicity, speed of implementation, and architecture. That said, the option you suggested is definitely interesting.

I actually think that this would make code more mantainable, and you could probably only have one action fired, the UI interaction, and have different middlewares, or one middleware, respond to it. and do different things. In this way you would really decouple and something similar to what you did with event listeners, with the difference that here the event listiener here is a redux middleware. 
I wanted more to start a discussion on this as this approach ( hardcoding stuff on cliks ) is something we ( me included in some tracking I did ) have already used, but it feels very "un javascript". Hooks serve a different purpose IMHO, and they are born from the need to reuse in a totally functional context as react has become.
In any case I don't think we should reafctor everything at this point, but wanted to hear your thoughts on this/

louiswol94 (approved):

PR #35002, opened 2026-03-05
Internal: Make menus lib reactive [ED-23216]

ronros-elementor:
### ✨ PR Description
Purpose: Refactor menus library to use React's useSyncExternalStore for reactive state updates when menu items are dynamically registered after component mount.

Main changes:
- Implemented pub/sub subscription pattern with notify mechanism to track menu item registration changes
- Replaced useMemo with useSyncExternalStore and snapshot-based caching for reactive menu state management
- Added test coverage for dynamic menu item registration and component re-rendering behavior


packages/packages/libs/menus/src/create-use-menu-items.ts
```diff
@@ -13,22 +13,41 @@ type GroupedMenuItems< TGroups extends string > = Record<
...
-					MenuItem: injection.component,
-				} ) );
-
-				return {
-					...carry,
-					[ groupName ]: items,
-				};
-			}, {} as GroupedMenuItems< TGroups > );
-		}, [] );
+	let cachedMenuItems: GroupedMenuItems< TGroups > | null = null;
```
Omerisra6: Do we really need cache here?

It just few items for each menu

Omerisra6 (dismissed):

packages/packages/libs/menus/src/create-use-menu-items.ts
ronros-elementor: afaik it's a requirement of the useSyncExternalStore - the returned value should somehow be cached, I also get an error when trying to return a new object without any sort of memoization

but will rewrite it a bit to make the code slightly cleaner

Omerisra6 (approved):

PR #35016, opened 2026-03-08
Fix: Labels and input ids in atomic form should match. [ED-23203]

RobiFerentz:


davseve (approved):

modules/atomic-widgets/elements/atomic-form/atomic-form.php
davseve, line 1: Can we add a small test that check that the label 'input-id' and the input 'css-id' have the same value? 
WDYT?

PR #35019, opened 2026-03-08
Internal: Add the 'Edit in Variables manager' tooltip and button [ED-23139]

YaelAvitan:
### ✨ PR Description
Purpose: Add interactive edit button with tooltip to V4 color variable list control to enable quick access to Variables manager panel.

Main changes:
- Added edit button with pencil icon and "Edit in Variables manager" tooltip to color variable list rows
- Implemented OpenPanelFromEvent component to handle custom event-driven Variables manager panel opening via window events
- Created shared SCSS styles for edit buttons across V4 list controls with hover states and tooltip positioning


modules/design-system-sync/controls/v4-color-variable-list.php
```diff
@@ -28,6 +28,12 @@ public function content_template() {
 				<div class="elementor-repeater-fields" role="listitem">
 					<div class="elementor-repeater-row-controls e-v4-color-variable-list__row">
 						<span class="elementor-control-title e-v4-color-variable-list__title">{{{ title }}}</span>
+						<span class="e-v4-color-variable-list__edit-btn-wrapper">
```
asafdl, line 31: https://github.com/elementor/elementor/blob/f6fb9aa6ead68848577ad9c5a5f0a9b34c387d50/packages/packages/core/editor-variables/src/components/open-panel-from-url.tsx
why do we need to use an event, and not url?

YaelAvitan: This button lives in a PHP template, outside the React tree, so it can't call the usePanelActions().open() hook directly like other React components can. The custom event bridges this gap — the button dispatches a window event, and a React component (OpenPanelFromEvent) listens for it and calls open() internally. The URL approach only runs once on page load, so using it from a button click would require a full page reload.

mike-elementor (approved):

asafdl (approved):

PR #35045, opened 2026-03-10
Internal: Show promotion notification on component publishing insufficient permissions [ED-23258]

Omerisra6:
## Summary
- When a user's Pro subscription expires and they attempt to publish a page containing draft components, the `updateStatuses` API call fails with a 403 `insufficient_permissions` error. Previously this error propagated silently, blocking the save with no user feedback.
- This PR catches the `insufficient_permissions` error code from the API response and shows a **promotion notification** (with an "Upgrade Now" CTA) instead of failing silently. Non-permission errors are still re-thrown to preserve existing behavior.

image

## Test plan
- [x] Verify that when Pro is deactivated/expired and a page with draft components is saved, a promotion notification appears with "Upgrade Now" button
- [x] Verify that the page save itself still completes (the permission error is swallowed, not re-thrown)
- [x] Verify that other API errors (e.g. network failures) still propagate normally
- [x] Unit tests pass: `npx jest --testPathPattern="update-components-before-save"`

### ✨ PR Description
Purpose: Add promotion notification when component publishing fails due to insufficient permissions to guide free users toward upgrading to pro version.
Main changes:
- Wrapped apiClient.updateStatuses call in try-catch to handle insufficient_permissions error without breaking save flow
- Created notifyPublishUpgrade function displaying promotion alert with upgrade CTA when permissions check fails
- Enhanced PromotionSnackbar styling to prevent text wrapping and ensure proper layout alignment with panel


packages/packages/core/editor-components/src/sync/publish-draft-components-in-page-before-save.ts
```diff
@@ -1,10 +1,17 @@
...
+import { notify } from '@elementor/editor-notifications';
+import { AxiosError } from '@elementor/http-client';
+import { __ } from '@wordpress/i18n';
 
 import { apiClient } from '../api';
 import { type DocumentSaveStatus } from '../types';
 import { getComponentDocuments } from '../utils/get-component-documents';
 
+const INSUFFICIENT_PERMISSIONS_ERROR_CODE = 'insufficient_permissions';
+const PUBLISH_UPGRADE_URL = 'https://go.elementor.com/go-pro-components-edit/';
```
IshayMaya, line 12: maybe we should have a shared file for these links? its in use in 2 more places

IshayMaya (approved):

packages/packages/core/editor-components/src/sync/publish-draft-components-in-page-before-save.ts
Omerisra6: I think we should move all assets, like texts and links to a json file from editor-assets

PR #35055, opened 2026-03-10
Internal: Update components empty states and promotions [ED-23257]

IshayMaya:
### ✨ PR Description
Purpose: Update Components feature UI to replace text-based PRO badges with crown icon promotions and enhance empty states for non-Pro users across editor interfaces.

Main changes:
- Replaced text "PRO" badges with crown icon badges in context menus and promotion links for create/edit component actions
- Added new ProUpgradeEmptyState component with upgrade CTA for users without Pro when components list is empty
- Created reusable ComponentsUpgradeAlert component to display promotion notices in components tab and instance editing panel


packages/packages/core/editor-components/src/components/__tests__/components-tab.test.tsx
```diff
@@ -231,58 +250,73 @@ describe( 'ComponentsTab', () => {
...
+			setHasPro( false );
+			act( () => {
+				dispatch( slice.actions.load( [] ) );
+			} );
+
+			// Act
+			renderWithStore( <Components />, store );
+
+			// Assert
+			expect( screen.queryByText( 'Create New Components' ) ).not.toBeInTheDocument();
```
MiryamOren: Using only the text 'Create New Components' as the upgrade notification selector is a bit confusing IMO.
Can we use the second phrase  ("Your Pro subscription has expired. Reactivate to enable components again.") which is explicitly the upgrade notification text, so it will be more clear & reliable (in future we might have the text 'Create New Components' in the components tab)

packages/packages/core/editor-components/src/components/components-tab/components-list.tsx
```diff
@@ -28,7 +30,11 @@ export function ComponentsList() {
 	const isEmpty = ! components?.length;
 
 	if ( isEmpty ) {
-		return searchValue.length ? <EmptySearchResult /> : <EmptyState />;
+		if ( searchValue.length ) {
+			return <EmptySearchResult />;
```
MiryamOren, line 34: It's a problem we had before, but if you can fix it now - I think we should not render the search component when there are no components at all - it looks pretty weird

packages/packages/core/editor-components/src/components/instance-editing-panel/instance-editing-panel.tsx
```diff
@@ -46,6 +50,40 @@ export function InstanceEditingPanel() {
 					componentInstanceId={ componentInstanceId }
 				/>
 			</ComponentInstanceProvider>
+			{ ! hasPro && <InstanceEditUpgradeAlert /> }
+		</Box>
+	);
+}
+
+function InstanceEditUpgradeAlert() {
```
MiryamOren: This component is almost identical to `ComponentsProNotification`, (if I'm not missing something, it's the same except for the texts & url) 
Can we create one shared component for both?

MiryamOren (approved):

PR #35115, opened 2026-03-15
Feature: Add svg-src prop type with Twig template for Atomic SVG [ED-22671]

Omerisra6:
## Problem

The `e-svg` atomic widget rendered entirely in its PHP `render()` method — it had no Twig template. This meant the SVG content was never exposed as a template variable (`settings.svg`), making it impossible to use the SVG widget as a component in the overridable-props/component architecture (which relies on Twig template context for parent→child settings overrides).

Moving to a Twig template required the SVG's processed HTML to be available as `settings.svg.html`. The existing `image-src` transformer just passes through `{ id, url }` and is shared by all image widgets — adding SVG-specific processing (fetch → DOMParser → `fill="currentColor"` → inline styles) to it would break the contract for other consumers.

Additionally, the dynamic tag category mapping relied on a fragile `is_svg` meta flag on `Image_Src_Prop_Type` instead of proper type discrimination.

### 1. New `svg-src` prop type instead of reusing `image-src`
Client-side transformer routing is keyed by `$$type` — different transformers for `image-src` vs `svg-src`. `image-src` validates "exactly one of id/url", while `svg-src` allows both. Dynamic tag category mapping now uses `instanceof Svg_Src_Prop_Type` instead of a fragile `is_svg` meta flag.

### 2. JSON migration instead of union/runtime compatibility
Legacy SVG widgets stored data as `image-src`. Instead of maintaining a `Union_Prop_Type` to accept both formats at runtime, we use Elementor's existing JSON-based prop type migration system (`migrations/manifest.json` + `operations/image-src-to-svg-src.json`). The migration is schema-driven and scoped per-widget and per-prop — it only triggers when the stored `$$type` doesn't match the widget schema's expected type. This means the image widget (which still expects `image-src`) is unaffected.

### 3. SVG processing in transformer layer, not the widget
Moved SVG fetching, processing (`fill="currentColor"`, inline styles), and sanitization from the widget's `render()` into `Svg_Src_Transformer`. This ensures all settings handling happens in the transformer layer (symmetric with the client-side transformer) and keeps the widget class clean — just schema + controls + styles + template path.

### 4. ID-to-URL resolution (following video-src pattern)
Both PHP and TS transformers resolve attachment `id` → `url` via `wp_get_attachment_url` / `getMediaAttachment`, handling the import scenario where only `id` exists after re-upload (import transformers intentionally set `url: null`).

### 5. Inline SVG only, no `` fallback
The main branch always rendered inline `<svg>` or nothing. The `` fallback was a behavior regression — it doesn't support `fill="currentColor"`, inline styles, or SVG manipulation.

### 6. `data-interaction-id` on wrapper only
Dropped `data-interaction-id` from the inner `<svg>` element (client-side never added it). The wrapper `<div>`/`<a>` carries the interaction ID, consistent with all other atomic widgets.

### PHP
- **`Svg_Src_Prop_Type`** — same `{ id, url }` shape as `Image_Src_Prop_Type`, relaxed validation (allows both id + url), own `$$type` key (`svg-src`).
- **`Svg_Src_Transformer`** — resolves `id` → `url`, fetches SVG content (local file for attachments, filesystem for Elementor assets, HTTP for remote), processes (`fill`, styles), sanitizes, returns `{ html, url }`.
- **`Atomic_Svg`** — simplified to schema + controls + styles + template path. No `render()` override needed.
- **Import/export transformers** mirror the `image-src` equivalents.
- **Dynamic tag mapping** — `instanceof Svg_Src_Prop_Type` instead of `is_svg` meta flag.
- **Migration** — `image-src-to-svg-src` JSON migration updates `$$type` from `image-src` to `svg-src` on document load, only for props whose widget schema expects `svg-src`.

### TypeScript
- **`svgSrcPropTypeUtil`** — prop type util for `svg-src` validation.
- **`svgSrcTransformer`** — client-side transformer with `getMediaAttachment` for ID→URL resolution, SVG fetch, DOMParser processing.
- **`SvgMediaControl`** — uses `svgSrcPropTypeUtil` directly for prop binding.
- **No changes to `image-src.ts`** — the migration handles backward compatibility at the data layer.

## Test Plan

- [x] Client-side transformer tests — ID resolution, URL fallback, fetch, SVG processing
- [x] Controls registry tests
- [x] `tsc --noEmit` passes
- [x] ESLint passes
- [x] PHP snapshot tests
- [ ] Manual: open editor with existing SVG widget (legacy `image-src` data) → verify migration runs and inline SVG renders in canvas
- [ ] Manual: pick new SVG → verify data saves as `svg-src`, inline SVG renders
- [ ] Manual: import template with SVG widget → verify SVG renders (ID-only after import)

### ✨ PR Description
Purpose: Add dedicated svg-src prop type with Twig template rendering to replace image-src-based SVG handling in Atomic Widget system

Main changes:
- Introduced Svg_Src_Prop_Type class with dedicated transformers for settings, import, and export operations
- Refactored Atomic_Svg widget to use Has_Template trait with Twig-based rendering instead of PHP render method
- Added TypeScript svgSrcTransformer with DOMPurify sanitization and currentColor fill attribute handling


tests/phpunit/elementor/modules/atomic-widgets/elements/__snapshots__/Test_Has_Template__test_render with data set Atomic SVG default__1.txt
```diff
@@ -1 +0,0 @@
...998 29.0057H18.934C19.9485 29.0057 20.9215 29.4088 21.6388 30.1261C22.3562 30.8435 22.7592 31.8165 22.7592 32.831V42.775C22.7592 43.7895 22.3562 44.7625 21.6388 45.4799C20.9215 46.1973 19.9485 46.6003 18.934 46.6003H8.98998C7.97545 46.6003 7.00248 46.1973 6.2851 45.4799C5.56772 44.7625 5.1647 43.7895 5.1647 42.775V32.831C5.1647 31.8165 5.56772 30.8435 6.2851 30.1261ZM30.0959 31.5824C31.7457 29.9326 33.9834 29.0058 36.3165 29.0058C38.6497 29.0058 40.8873 29.9326 42.5371 31.5824C44.187 33.2322 45.1138 35.4698 45.1138 37.803C45.1138 40.1362 44.187 42.3738 42.5371 44.0236C40.8873 45.6734 38.6497 46.6003 36.3165 46.6003C33.9834 46.6003 31.7457 45.6734 30.0959 44.0236C28.4461 42.3738 27.5193 40.1362 27.5193 37.803C27.5193 35.4698 28.4461 33.2322 30.0959 31.5824ZM36.3165 31.6843C34.6938 31.6843 33.1374 32.329 31.99 33.4764C30.8425 34.6239 30.1978 36.1802 30.1978 37.803C30.1978 39.4258 30.8425 40.9821 31.99 42.1296C33.1374 43.2771 34.6938 43.9217 36.3165 43.9217C37.9393 43.9217 39.4956 43.2771 40.6431 42.1296C41.7906 40.9821 42.4352 39.4258 42.4352 37.803C42.4352 36.1802 41.7906 34.6239 40.6431 33.4764C39.4956 32.329 37.9393 31.6843 36.3165 31.6843Z" fill="currentColor"></path></svg></div>
```
IshayMaya, line 1: why these 2 were removed?

tests/phpunit/elementor/modules/atomic-widgets/__snapshots__/Test_Atomic_Svg__test__should_render_svg_wrapped_in_link with data set ID link__1.txt
```diff
@@ -1 +1 @@
-			<a href="#element-id" target="_blank" class="e-svg-base" data-interaction-id="abcd123"><svg xmlns="http://www.w3.org/2000/svg" data-interaction-id="abcd123" fill="currentColor" style="width: 100%; height: 100%; overflow: unset;" width="100" height="100"><path d="M0 0h100v100H0z"></path></svg></a>		
\ No newline at end of file
+			<a href="#element-id" target="_blank" class="e-svg-base" data-interaction-id="abcd123"><svg xmlns="http://www.w3.org/2000/svg" fill="currentColor" style="width: 100%; height: 100%; overflow: unset;" width="100" height="100"><path d="M0 0h100v100H0z"></path></svg></a>
```
IshayMaya, line 1: all those `data-interaction-id` changes are valid?
did we have a duplication on the editor as well?

tests/phpunit/elementor/modules/atomic-widgets/__snapshots__/Test_Atomic_Svg__test__render_svg_with_action_link__1.txt
```diff
@@ -1 +1 @@
-			<button data-action-link="https://very.dynamic.content.elementor" target="_blank" class="e-svg-base" data-interaction-id="linked456"><svg xmlns="http://www.w3.org/2000/svg" data-interaction-id="linked456" fill="currentColor" style="width: 100%; height: 100%; overflow: unset;" width="100" height="100"><path d="M0 0h100v100H0z"></path></svg></button>		
\ No newline at end of file
+			<button data-action-link="https://very.dynamic.content.elementor" target="_blank" class="e-svg-base" data-interaction-id="linked456"><svg xmlns="http://www.w3.org/2000/svg" fill="currentColor" style="width: 100%; height: 100%; overflow: unset;" width="100" height="100"><path d="M0 0h100v100H0z"></path></svg></button>
```
IshayMaya, line 1: isn't that an issue?

packages/packages/libs/editor-controls/src/controls/svg-media-control.tsx
```diff
@@ -61,7 +62,10 @@ export const SvgMediaControl = createControl( () => {
 					$$type: 'image-attachment-id',
 					value: selectedAttachment.id,
 				},
-				url: null,
+				url: {
+					$$type: 'url',
+					value: selectedAttachment.url,
+				},
```
IshayMaya:
```suggestion
				url: urlPropTypeUtil.create( selectedAttachment.url ),
```

packages/packages/libs/editor-controls/src/controls/svg-media-control.tsx
```diff
@@ -61,7 +62,10 @@ export const SvgMediaControl = createControl( () => {
 					$$type: 'image-attachment-id',
 					value: selectedAttachment.id,
 				},
-				url: null,
+				url: {
```
IshayMaya: it was loaded from the BE before?

packages/packages/core/editor-canvas/src/transformers/shared/svg-src-transformer.ts
```diff
@@ -0,0 +1,82 @@
+import { getMediaAttachment } from '@elementor/wp-media';
```
IshayMaya, line 2: check if we have a security issue

tests/phpunit/elementor/modules/atomic-widgets/elements/__snapshots__/Test_Has_Template__test_render with data set Atomic SVG default__1.txt
Omerisra6: they used no settings, we need the settings to fetch the svg

tests/phpunit/elementor/modules/atomic-widgets/__snapshots__/Test_Atomic_Svg__test__should_render_svg_wrapped_in_link with data set ID link__1.txt
Omerisra6: same

tests/phpunit/elementor/modules/atomic-widgets/__snapshots__/Test_Atomic_Svg__test__render_svg_with_action_link__1.txt
Omerisra6: It is its a blocker for now, so I pause the work on this

packages/packages/libs/editor-controls/src/controls/svg-media-control.tsx
Omerisra6: removed

packages/packages/core/editor-canvas/src/transformers/shared/svg-src-transformer.ts
Omerisra6: added a sanitize step, we might need to use a package to be more secure

modules/atomic-widgets/elements/atomic-svg/atomic-svg.html.twig
ronros-elementor, line 1: probably not for this PR, but we should consider adding our own atomic extension to twig, which would handle wrapper attributes, safe-tag selection, click-attributes, etc
as most of the content in our templates currently is duplicated logic pieces

ronros-elementor (approved):

modules/atomic-widgets/elements/atomic-svg/atomic-svg.html.twig
Omerisra6: https://elementor.atlassian.net/browse/ED-23592

PR #35125, opened 2026-03-15
Fix: Component inner elements interactions [ED-23347]

MiryamOren:
### ✨ PR Description
Purpose: Fix interaction tracking for component inner elements by preserving original element IDs when components are nested or instantiated multiple times.

Main changes:
- Added origin_id property to atomic elements and widgets to track original element IDs before hashing
- Updated Format_Component_Elements_Id to preserve origin_id while generating unique hashed IDs for nested components
- Modified all templates and views to use get_interaction_id() method that returns origin_id or falls back to current ID


Omerisra6:
What about the marionette view? Does it need to change?

Is it a breaking change for people with older versions? 
Do we need to open a pr for forms or other pro elements?

Omerisra6 (dismissed):

Omerisra6 (approved):

PR #35134, opened 2026-03-15
Fix: Resolve custom effects misconfiguration for Angie [ED-23280]

Ntnelbaba:
## Summary
- Rename schema field `customEffect` → `customEffects` in `proSchema` to match the field name used by the MCP tool handler
- Default `relativeTo` to `'viewport'` in `createConfig` so custom effect interactions use the correct default
- Add `custom-effect-to-prop-value.ts` converter that transforms plain `{ keyframes: [...] }` JS objects (as provided by Angie) into fully-typed `PropValue` structures
- Accept `PlainCustomEffect` type in `createAnimationPreset` and `createInteractionItem`, auto-converting via `toCustomEffectPropValue`

## Test plan
- [ ] Ask Angie to create a custom effect interaction and verify no errors are thrown
- [ ] Verify the interaction is saved and applied correctly on the element
- [ ] Run unit tests: `cd packages && npx jest packages/core/editor-interactions --no-coverage`

## Dependencies
None

### ✨ PR Description
Purpose: Fix custom effects schema property name from customEffect to customEffects to align with animation configuration system and prevent runtime misconfiguration errors.

Main changes:
- Renamed schema property from customEffect to customEffects in proSchema for consistent plural naming convention
- Added toCustomEffectPropValue utility function with PlainCustomEffect type guards to handle plain and PropValue formats
- Updated createAnimationPreset to conditionally include custom_effect only when customEffects is defined, preventing undefined prop values


packages/packages/core/editor-interactions/src/utils/prop-value-utils.ts
```diff
@@ -46,7 +47,7 @@ export const createBoolean = ( value: boolean ): BooleanPropValue => ( {
 export const createConfig = ( {
 	replay,
 	easing = 'easeIn',
-	relativeTo = '',
+	relativeTo = 'viewport',
```
mike-elementor, line 50: @elchugreeva can you check? what effect changes in default `relativeTo` might affect? (if anything)

Ntnelbaba: Seems like this is a required field, this was one of the issues in the element not getting the custom effect**s**

dima-belkin (approved):

packages/packages/core/editor-interactions/src/utils/custom-effect-to-prop-value.ts
```diff
@@ -0,0 +1,145 @@
...
+const TRANSFORM_MOVE_TYPE = 'transform-move';
+const TRANSFORM_SKEW_TYPE = 'transform-skew';
+const UNIT_PERCENT = '%';
+const UNIT_DEG = 'deg';
+const UNIT_PX = 'px';
+
+export type PlainKeyframeValue = {
+	opacity?: number;
+	scale?: { x: number; y: number };
+	rotate?: { x: number; y: number; z: number };
```
Batsirai-muchareva, line 20: @Ntnelbaba aren't these value supposed to be strings since we deal with different units

packages/packages/core/editor-interactions/src/utils/custom-effect-to-prop-value.ts
```diff
@@ -0,0 +1,145 @@
...
+} );
+
+const toKeyframeStopSettingsPropValue = ( plain: PlainKeyframeValue ): PropValue => {
+	const value: Record< string, PropValue > = {};
+	if ( plain.opacity !== undefined ) {
+		const percent = plain.opacity <= 1 ? Math.round( plain.opacity * 100 ) : plain.opacity;
+		value.opacity = toSizePropValue( percent );
+	}
+	if ( plain.scale !== undefined ) {
+		value.scale = toDimensionalNumberPropValue( TRANSFORM_SCALE_TYPE, plain.scale, { x: 1, y: 1, z: 1 } );
```
Batsirai-muchareva, line 88: @mike-elementor didn't we remove the scale z confirm here

packages/packages/core/editor-interactions/src/utils/custom-effect-to-prop-value.ts
```diff
@@ -0,0 +1,145 @@
...
+	}
+	if ( plain.rotate !== undefined ) {
+		value.rotate = toDimensionalSizePropValue(
```
Batsirai-muchareva, line 91: I would prefer to use existing rotateTransformPropTypeUtil but for now its fine we can handle it later

PR #35173, opened 2026-03-17
Fix: Dynamic `post-custom-field` in components [ED-22335]

MiryamOren:
Summary
Fixes two issues with post-custom-field dynamic tags when used inside components:

1. **Validation failure on save** 
  Dynamic tag select controls (like the field key selector in post-custom-field) were validated against an enum of available options (e.g., the post's actual custom fields). When used inside a component, the field doesn't exist on the component itself — it exists on the post containing the component. This caused save/publish to fail with validation errors.
  **Fix**:
Remove enum validation for dynamic tag select controls entirely. The validation was removed globally (not just for components) because deleting an existing custom field from a post/page that references it via a post-custom-field dynamic tag shouldn't break validation either.

2. **Stale cache on document switch** 
The dynamic tags cache is global and doesn't account for document context. When switching from component edit mode back to a post, the cached value (empty, from the component context) was used instead of fetching the correct value from the post context.
**Fix**:
Clear the dynamic tags cache before opening a document, ensuring values are fetched with the correct context.

### ✨ PR Description
Purpose: Fix dynamic tag caching issue preventing post-custom-field values from updating correctly in component inner elements.

Main changes:
- Removed enum validation from select control type conversion to allow dynamic custom field values
- Added ClearDynamicTagsCache hook to refresh dynamic tags when opening documents in editor
- Updated test expectations to remove enum settings from key property validation


modules/atomic-widgets/dynamic-tags/dynamic-tags-converter.php
```diff
@@ -29,17 +29,9 @@ public static function convert_control_to_prop_type( array $control ) {
...
 			case 'textarea':
-				$prop_type = String_Prop_Type::make()
-					->default( $control['default'] ?? null );
-				break;
-
 			case 'select':
 				$prop_type = String_Prop_Type::make()
 					->default( $control['default'] ?? null );
-
-				if ( ! isset( $control['collection_id'] ) || empty( $control['collection_id'] ) ) {
```
IshayMaya, line 40: any idea why we had this?

modules/atomic-widgets/dynamic-tags/dynamic-tags-converter.php
```diff
@@ -29,17 +29,9 @@ public static function convert_control_to_prop_type( array $control ) {
 		switch ( $control_type ) {
 			case 'text':
 			case 'textarea':
-				$prop_type = String_Prop_Type::make()
```
IshayMaya, line 32: is it related to components?
why don't we want a default?

assets/dev/js/editor/components/documents/hooks/ui/open/clear-dynamic-tags-cache.js
```diff
@@ -0,0 +1,15 @@
+export class ClearDynamicTagsCache extends $e.modules.hookUI.Before {
+	getCommand() {
+		return 'editor/documents/open';
```
IshayMaya, line 3: not sure how it can be done but i think we dont want if to run if the dynamic module doesn't load

IshayMaya (approved):

PR #35223, opened 2026-03-22
Fix: Resolve stale element references in undoable operations for atomic nested elements [ED-22825]

Omerisra6:
## Summary

Undo/redo operations on atomic nested elements (e.g. Tabs widget children) fail because the undoable closures hold stale `V1Element` references. Atomic elements render children asynchronously via client-side Twig templates, so the Marionette view tree may not contain the recreated child views when an undo/redo closure executes.

This PR adds a **fallback resolution strategy** to all undoable operations and legacy commands:

- **`resolveContainer()`** — tries `container.lookup()` first, then falls back to `elementor.getContainer(id)` with DOM-connection validation
- **Model-tree fallback** — when both view-based lookups fail, operations manipulate the Backbone model tree directly using `addModelToParent` / `removeModelFromParent` with `{ silent: true }`. When the parent is mid-render, its `_renderChildren` naturally picks up the silently-added models; when the parent has a live view, the two-step resolution in the legacy commands prefers the view-based path
- **`findModelById`** — new utility in `document/component.js` to traverse the model tree independently of the view tree

### Why `rerenderAtomicAncestor` was removed

The initial implementation included a `rerenderAtomicAncestor` helper that would force-render the nearest atomic widget after a silent model-tree change. Testing showed this is unnecessary:

- **Legacy commands** use a two-step resolution: first try `findContainerById` to get a fresh container, prefer the view-based `addElement` path when a live view exists, and only fall through to the silent model-tree path when the parent is genuinely mid-render — in which case `_renderChildren` picks up the model automatically
- **Undoable actions** in the turbo-repo follow the same pattern via `resolveContainer` + `addModelToParent`/`removeModelFromParent`

## Changes

| File | Change |
|------|--------|
| `document/component.js` | Added `findModelById` utility for model-tree traversal |
| `elements/commands/create.js` | Two-step container resolution with `findContainerById` fallback, `addViaModelTree` for viewless containers |
| `elements/commands/delete.js` | Same two-step resolution pattern, `deleteViaModelTree` for viewless containers |
| `resolve-element.ts` | New shared utilities: `resolveContainer`, `findModelInDocTree`, `addModelToParent`, `removeModelFromParent` |
| `create-elements.ts` | Uses `resolveContainer` + model-tree fallback in undo/redo |
| `remove-elements.ts` | Uses `resolveContainer` + model-tree fallback in undo/redo |
| `duplicate-elements.ts` | Uses `resolveContainer` + model-tree fallback in undo/redo |
| `move-elements.ts` | Uses `resolveContainer` in undo/redo |
| `types.ts` | Added `BackboneModel`, `BackboneCollection` types and extended `ExtendedWindow` |
| `create-elements.test.ts` | Tests for model-tree fallback on undo/redo |
| `remove-elements.test.ts` | Updated test expectations for `resolveContainer` behavior |

## Test plan

- [x] Unit tests pass for `resolve-element.ts`, `create-elements.test.ts`, `remove-elements.test.ts`
- [x] Manual QA: Add Tabs widget → add tab → add element to tab → undo all → redo all (all cycles work, fourth tab renders correctly)
- [x] Manual QA: Delete tab child → undo → redo (all cycles work)
- [x] Manual QA: Multi-step undo all → redo all (widget removal + re-creation works)
- [x] No new console errors from the implemented changes

### ✨ PR Description
Purpose: Fix stale element reference errors in undo/redo operations for nested atomic elements by implementing fallback model-tree manipulation when views are unavailable.

Main changes:
- Added resolveContainer utility with DOM connectivity check and fallback model lookup to handle async-rendered nested elements
- Implemented direct model-tree manipulation methods (findModelById, addModelToParent, removeModelFromParent) as fallback for detached views
- Updated all undoable operations (create, delete, duplicate, move, remove) to use container resolution with model-tree fallback


Omerisra6:
## Why the fallback was added to all undoable operations

The core problem is that **all undoable operations** (`createElements`, `removeElements`, `duplicateElements`, `moveElements`) store `V1Element` references in closures and rely on `container.lookup()` during undo/redo to obtain fresh references. For regular elements this works because the Marionette view tree is synchronously updated — `lookup()` always finds the recreated view.

**Atomic nested elements** (Tabs, Accordion, etc.) break this contract because their children are rendered **asynchronously** via client-side Twig templates. When an undo/redo fires:

1. The old `V1Element` in the closure points to a destroyed view
2. `container.lookup()` returns `null` — the new view hasn't been created yet
3. `elementor.getContainer(id)` also fails — it relies on `findViewRecursive` which searches the view tree
4. The operation silently skips, leaving the UI out of sync with the model

### The fix: a two-tier resolution strategy

```
resolveContainer(container, id)
├── 1. container.lookup()          → works for regular elements (zero overhead)
├── 2. elementor.getContainer(id)  → works for elements that moved in the view tree
└── 3. (both fail) → model-tree fallback
         ├── addModelToParent / removeModelFromParent  (silent: true)
         └── rerenderAncestor → invalidateRenderCache + render
```

### Why not just fix the specific operation where we saw the bug?

- **Any operation can target atomic children** — users can add, delete, duplicate, or reorder tabs. All four operations produce the exact same stale-reference problem.
- **The fallback is zero-cost for regular elements** — `resolveContainer()` first tries `lookup()` and `getContainer()`. The model-tree path is only reached when **both** return null, which only happens for atomic children whose views haven't rendered yet. Regular elements never hit this path.
- **Consistency prevents subtle regressions** — if only `removeElements` had the fallback but `createElements` didn't, a simple "add tab → undo → redo" cycle would break while "delete tab → undo → redo" would work. This kind of inconsistency is extremely hard to debug.
- **The `{ silent: true }` + `rerenderAncestor()` pattern is safe** — silent model manipulation avoids triggering Marionette's incremental `_addChild`/`_removeChild` (which would conflict with the atomic Twig rendering), and `rerenderAncestor()` calls `invalidateRenderCache()` to bypass the settings-hash cache before triggering a full re-render through the correct Twig pipeline.

assets/dev/js/editor/document/component.js
```diff
@@ -77,6 +77,27 @@ export default class Component extends ComponentBase {
...
+					const childElements = model.get( 'elements' );
+
+					if ( childElements ) {
+						const found = this.utils.findModelById( id, childElements );
+
+						if ( found ) {
+							return found;
+						}
+					}
+				}
```
Omerisra6: simplify

packages/packages/libs/editor-elements/src/sync/__tests__/resolve-element.test.ts
Omerisra6, line 1: do an integration test in one of the undoable actions using this and delete this test

packages/packages/libs/editor-elements/src/sync/create-elements.ts
```diff
@@ -52,49 +62,90 @@ export const createElements = ( {
...
 
 				return { createdElements };
 			},
 			undo: ( _: { elements: CreateElementParams[] }, { createdElements }: CreatedElementsResult ) => {
-				[ ...createdElements ].reverse().forEach( ( { container } ) => {
-					const freshContainer = container.lookup?.();
+				// Atomic nested elements (e.g. Tabs) render children asynchronously via Twig, so
+				// the view tree may not contain the child views when undo/redo runs. resolveContainer
+				// tries lookup() then getContainer(); if both fail, we fall back to direct model-tree
+				// manipulation with silent:true and re-render the atomic ancestor. See ED-22825.
```
Omerisra6: the comment should be much shorter (one line)

packages/packages/libs/editor-elements/src/sync/create-elements.ts
```diff
@@ -52,49 +62,90 @@ export const createElements = ( {
...
+					} );
 
-					if ( freshContainer ) {
-						deleteElement( {
-							container: freshContainer,
-							options: { useHistory: false },
-						} );
-					}
-				} );
+				rerenderAncestor( atomicToRerender );
```
Omerisra6: Do that only if we have elements to render and remove the atomic keyword...

packages/packages/libs/editor-elements/src/sync/create-elements.ts
```diff
@@ -52,49 +62,90 @@ export const createElements = ( {
...
+								options,
+								containerId: element.id,
+								parentContainerId: freshParent.id,
+								atomicAncestorId,
+							} );
+
+							return;
+						}
+
+						if ( addModelToParent( parentContainerId, model as Record< string, unknown > ) ) {
```
Omerisra6: use proper typing

packages/packages/libs/editor-elements/src/sync/resolve-element.ts
```diff
@@ -0,0 +1,93 @@
...
+	const byId = getContainer( id );
+
+	if ( isConnected( byId ) ) {
+		return byId;
+	}
+
+	return null;
+}
+
+export function findModelInDocTree( id: string ): BackboneModel | null {
```
ronros-elementor: maybe -
```suggestion
export function findModelInDocument( id: string ): BackboneModel | null {
```
?

don't think the tree structure has any relevancy to this function in particular

assets/dev/js/editor/document/elements/commands/create.js
```diff
@@ -74,6 +84,22 @@ export class Create extends $e.modules.editor.document.CommandHistoryBase {
 
 		return result;
 	}
+
+	addViaModelTree( container, model, options ) {
```
ronros-elementor: if this will be also exported as a util, it could be consumed on the `ts` side similar to how you dealt with `findModelById`

and then you could remove that duplicated logic from the `resolve-element` module (also - not sure it should be in that file to begin with, because it doesn't resolve an element, but updates a model)

assets/dev/js/editor/document/elements/commands/delete.js
```diff
@@ -62,6 +71,32 @@ export class Delete extends $e.modules.editor.document.CommandHistoryBase {
 		return containers;
 	}
 
+	deleteViaModelTree( container ) {
```
ronros-elementor: same suggestion as with `addViaModelTree`

ronros-elementor (changes requested):

ronros-elementor (dismissed):

ronros-elementor (approved):

PR #35254, opened 2026-03-23
Internal: Clear pregenerated CSS files upon style changes [ED-23383]

ronros-elementor:
# Editor Canvas CSS Override Problem – Analysis & Solution

## Status (current branch)

The approach below is **implemented** in the editor canvas and styles repository: editable providers declare `pregeneratedLinkPattern`, `useStyleItems` removes matching `<link>` tags after dynamic styles are cached, and `resetRemovedProviders()` runs on `attach-preview`.

One subtlety worth keeping in the design record: **`StylesProvider.subscribe` callbacks receive `previous` and `current` collections only from the global classes provider** (`subscribeWithStates` in `global-classes-styles-provider.ts`). The document elements provider fires `styleRerenderEvents` with `() => cb()` — **no prev/current**. Incremental updates (`getChangedStyleIds` → partial re-render) therefore run only for globals; local/element styles always take the full `createItems` path on each notification. Link removal cannot rely on “inside `updateItems` after `changedIds`” or it would **never** run for local edits. The production code removes links **before** branching on `hasDiffInfo`, gated only by `hasCache && pregeneratedLinkPattern`.

The narrative below was brought in line with the branch: the main correction is **where** link removal runs relative to `hasDiffInfo` / `changedIds`. If you see a mismatch, treat `packages/packages/core/editor-canvas/src/hooks/use-style-items.ts` and each provider’s `subscribe` as source of truth.

---

## Problem Statement

In the editor canvas iframe, v4 atomic widget styles exist in **two places simultaneously**:

1. **Pregenerated CSS files** (`<link>` tags) – Created by PHP (`Atomic_Styles_Manager`), served from `/elementor/css/`, with selectors like `.elementor .class-name`.
2. **Dynamic `<style>` tags** – Created by TypeScript (`StyleRenderer`) via React Portal into the iframe's `<head>`, with **identical selectors**.

When a user **clears** a CSS property (e.g., removes text color), the property is deleted from the in-memory style definition (`mergeProps` sets the key to `null`, which causes `delete props[key]`). The dynamic `<style>` tag is re-rendered **without** that property. However, the pregenerated CSS file still contains the old value. Since both use the same CSS specificity (`.elementor .class-name`), the pregenerated CSS value remains in effect — the user sees no change.

**Key constraint**: We do NOT set `unset` or any placeholder value when clearing — we simply omit the property entirely from the rendered CSS. This is by design to keep the style state clean.

---

### CSS Style Rendering Hierarchy (Priority Order)

| Layer  | PHP Class                   | TS Provider                      | Priority     | CSS Handle Pattern                      | Editable (`updateProps`) |
| ------ | --------------------------- | -------------------------------- | ------------ | --------------------------------------- | ------------------------ |
| Base   | `Atomic_Widget_Base_Styles` | `elementBaseStylesProvider`      | 10 (default) | `base-{breakpoint}`                     | No                       |
| Global | `Atomic_Global_Styles`      | `globalClassesStylesProvider`    | 30           | `global-{context}-{breakpoint}`         | Yes                      |
| Local  | `Atomic_Widget_Styles`      | `documentElementsStylesProvider` | 50           | `local-{postId}-{context}-{breakpoint}` | Yes                      |

### How Styles Appear in the Canvas iframe `<head>`

```html
<!-- V3 post CSS (handled by useDocumentsCssLinks - removed from DOM and re-rendered in Portal) -->
<!-- <link id="elementor-post-123-css" ... /> — moved to Portal -->

<!-- V4 atomic pregenerated CSS (NOT handled - remains in DOM as-is) -->
<link id="base-desktop-css" href=".../elementor/css/base-desktop.css" rel="stylesheet" media="all" />
<link id="global-preview-desktop-css" href=".../elementor/css/global-preview-desktop.css" rel="stylesheet" />
<link id="local-123-preview-desktop-css" href=".../elementor/css/local-123-preview-desktop.css" rel="stylesheet" />

<!-- V4 dynamic style tags (rendered by StyleRenderer via Portal) -->
<style>.elementor .my-class{color:red;font-size:16px;}</style>
<!-- After user clears color: -->
<style>.elementor .my-class{font-size:16px;}</style>
<!-- But base-desktop-css still has: .elementor .my-class{color:blue;} → color:blue wins by source order -->
```

### Key Finding

`useDocumentsCssLinks` only handles `elementor-post-*-css` link tags (v3). The v4 atomic CSS link tags (`base-*-css`, `global-*-css`, `local-*-css`) are **not** managed — they remain in the DOM untouched during the editor session.

### Selector Matching

Both PHP (`Styles_Renderer`) and TS (`createStylesRenderer`) produce identical CSS selectors:

- Base selector prefix: `.elementor`
- Class pattern: `.{cssName}`
- Combined: `.elementor .{cssName}`
- With state: `.elementor .{cssName}:hover`

### Why CSS Layers / Specificity Approaches Don't Work

CSS `@layer` and other specificity-based approaches only resolve conflicts between **competing declarations** for the same property. Our problem is the **absence** of a declaration — when the user clears a value, we omit the property entirely. There is nothing for the cascade to resolve; the only declaration is in the pregenerated CSS, and it applies by default. For any specificity-based approach to work, we'd need to emit `color: unset` when clearing — which contradicts the design principle of not polluting the real state.

---

### Concept

Remove pregenerated CSS `<link>` tags **on-demand** from the canvas iframe once dynamic styles for that provider are already represented in the StyleRenderer cache. Each editable provider declares a `pregeneratedLinkPattern` for its `<link id="…">` tags. The canvas subscriber removes matching links on a **post-cache** notification (usually the first real style change), not by instrumenting `updateProps` directly.

This approach combines:
- **From surgical approach**: Provider-specific patterns; only providers that subscribe and declare a pattern participate
- **From wholesale approach**: Remove entire `<link>` tags (not individual CSSOM rules) for simplicity

### Why This Works

1. **Base styles are NOT editable** — `elementBaseStylesProvider` has no `pregeneratedLinkPattern` and a no-op `subscribe`, so `base-*-css` links are never targeted. This is correct because base styles are read-only defaults.

2. **Editable providers trigger removal after the first successful dynamic render** — The canvas subscriber runs on each provider notification. Once the in-memory cache has been populated (`orderedIds.length > 0`), the next run removes matching `<link>` tags. That typically coincides with the first user-driven style change, but the trigger is **subscriber + warm cache**, not a direct hook on `updateProps`.

3. **Dynamic styles already exist** — Removal runs only when `hasCache` is true, so dynamic `<style>` tags from an earlier `createItems` pass are already in the iframe. Removing the `<link>` tags is safe (no FOUC).

4. **One-time removal per provider key** — A `Set` of provider keys avoids redundant DOM work across repeated notifications.

### How It Works

```
┌─────────────────────────────────────────────────────────────────────────┐
│ 1. Page Load                                                            │
│    └─ PHP renders iframe with pregenerated <link> tags                  │
│       └─ base-desktop-css, global-preview-desktop-css, local-*-css      │
│                                                                         │
│ 2. attach-preview fires                                                 │
│    └─ StyleRenderer renders dynamic <style> tags via Portal             │
│    └─ All styles now exist in BOTH <link> and <style> (identical)       │
│                                                                         │
│ 3. User edits a global class (first edit after cache warm)              │
│    └─ Global subscribe fires with (previous, current)                   │
│    └─ Subscriber: remove links if hasCache && pattern (before diff path)│
│    └─ Then incremental updateItems(previous, current) if hasDiffInfo    │
│                                                                         │
│ 4. User edits an element's local style (first edit after cache warm)    │
│    └─ styleRerenderEvents → subscribe calls cb() with no arguments      │
│    └─ Subscriber: same removal gate (hasCache && pattern)               │
│    └─ hasDiffInfo is false → full createItems (no prev/current)         │
│                                                                         │
│ 5. Subsequent edits                                                     │
│    └─ Links already removed; removal is a no-op                         │
└─────────────────────────────────────────────────────────────────────────┘
```

#### 1. Extend `StylesProvider` Type

**`editor-styles-repository/src/types.ts`**:

```typescript
export type StylesProvider = {
  // ...
  subscribe: ( callback: ( previous?: StylesCollection, current?: StylesCollection ) => void ) => () => void;
  pregeneratedLinkPattern?: RegExp;
};
```

(`previous` / `current` are optional in the type; in practice only the global classes implementation passes both — see **Status** above.)

#### 2. Extend `createStylesProvider`

**`editor-styles-repository/src/utils/create-styles-provider.ts`**:

```typescript
export type CreateStylesProviderOptions = {
  // ... existing fields ...
  pregeneratedLinkPattern?: RegExp;
};

export function createStylesProvider({
  // ... existing params ...
  pregeneratedLinkPattern,
}: CreateStylesProviderOptions): StylesProvider {
  return {
    // ... existing fields ...
    pregeneratedLinkPattern,
  };
}
```

#### 3. Add Patterns to Editable Providers

**`global-classes-styles-provider.ts`**:

```typescript
export const globalClassesStylesProvider = createStylesProvider({
  key: GLOBAL_CLASSES_PROVIDER_KEY,
  priority: 30,
  pregeneratedLinkPattern: /^global-.*-css$/,
  // ... rest unchanged
});
```

**`document-elements-styles-provider.ts`**:

```typescript
export const documentElementsStylesProvider = createStylesProvider({
  key: () => `${ELEMENTS_STYLES_PROVIDER_KEY_PREFIX}${getCurrentDocumentId()}`,
  priority: 50,
  pregeneratedLinkPattern: /^local-\d+-.*-css$/,
  // ... rest unchanged
});
```

Note: `elementBaseStylesProvider` does NOT get a pattern (it's read-only).

#### 4. Create Removal Utility

**`editor-canvas/src/utils/pregenerated-links-removal.ts`**:

```typescript
import { getCanvasIframeDocument } from '@elementor/editor-v1-adapters';

const removedProviderKeys = new Set<string>();

export function removeProviderPregeneratedLinks(
  providerKey: string,
  pattern: RegExp
): void {
  if (removedProviderKeys.has(providerKey)) {
    return;
  }

  const iframeDocument = getCanvasIframeDocument();
  if (!iframeDocument) {
    return;
  }

  const links = iframeDocument.head.querySelectorAll<HTMLLinkElement>(
    'link[rel="stylesheet"]'
  );

  links.forEach((link) => {
    const id = link.getAttribute('id') ?? '';
    if (pattern.test(id)) {
      link.remove();
    }
  });

  removedProviderKeys.add(providerKey);
}

export function resetRemovedProvider
<!--start_gitstream_placeholder-->
### ✨ PR Description
Purpose: Fix CSS caching issue by implementing automatic removal of pregenerated stylesheet links when style changes occur in Elementor editor preview mode.

Main changes:
- Added is_editor_or_preview() method to detect WordPress preview or Elementor iframe preview states
- Implemented pregenerated CSS link removal mechanism that clears cached stylesheets upon first render
- Replaced is_preview() calls with is_editor_or_preview() across atomic widgets and global classes modules

<!--end_gitstream_placeholder-->

### CODE REVIEW

File: includes/preview.php [hunk 1]
```diff
@@ -194,6 +194,18 @@ public function is_preview_mode( $post_id = 0 ) {
 		return true;
 	}
 
+	/**
+	 * Whether WordPress post preview or Elementor preview iframe is active.
+	 *
+	 * @since 4.2.0
```
IshayMaya: 4.1.0 ?

includes/preview.php
```diff
@@ -194,6 +194,18 @@ public function is_preview_mode( $post_id = 0 ) {
...
+	 * @access public
+	 *
+	 * @return bool
+	 */
+	public function is_editor_or_preview() {
+		return is_preview() || $this->is_preview_mode();
```
IshayMaya, line 206: preview mode is the editor?

packages/packages/core/editor-styles-repository/src/providers/document-elements-styles-provider.ts
```diff
@@ -28,6 +29,7 @@ export const documentElementsStylesProvider = createStylesProvider( {
 		return `${ ELEMENTS_STYLES_PROVIDER_KEY_PREFIX }${ documentId }`;
 	},
 	priority: 50,
+	pregeneratedLinkPattern: PREGENERATED_LINK_PATTERN,
```
IshayMaya: maybe it should be a callback that tests the pattern?

tests/playwright/pages/atomic-elements-panel/style-tab.ts
```diff
@@ -14,7 +14,7 @@ export const STYLE_SECTIONS = {
 	EFFECTS: 'Effects',
 } as const;
 
-const SIZE_SECTION_LABELS = {
+type SIZE_SECTION_LABELS = {
```
IshayMaya: would enum make sense?

includes/preview.php
ronros-elementor: yes exactly

IshayMaya (approved):

PR #35281, opened 2026-03-25
Internal: Prevent ViewDestroyedError on link field editing in atomic elements [ED-23356]

IshayMaya:
### **Summary:**

- Fix **ViewDestroyedError** thrown when rapidly editing the link field on a div-block with children (e.g., typing and deleting quickly)

- **Root cause 1**: isTagChanged() compared this.tagName() (lowercase "a") against this.el.tagName (uppercase "A") without normalizing case, causing it to always return true and trigger unnecessary rerenderEntireView() calls on every keystroke

- **Root cause 2**: rerenderEntireView() was called after BaseElementView.prototype.renderOnChange, which schedules an unguarded async .then() callback via renderDataBindings. When rerenderEntireView() destroyed the view, that pending callback fired on the destroyed view, triggering Marionette's ViewDestroyedError

- **Fix 1**: Add .toLowerCase() to the DOM tag comparison in isTagChanged() — prevents spurious view recreation when the tag hasn't actually changed

- **Fix 2**: Move the isTagChanged check before the base class renderOnChange call and return early — when the tag genuinely changed, the view will be destroyed and fully re-rendered, so there's no need to run incremental updates on a doomed view

### ✨ PR Description
Purpose: Fix ViewDestroyedError in atomic elements by preventing unnecessary view rerendering when link field changes don't affect the actual DOM tag structure.

Main changes:
- Added early tag-change check before base renderOnChange to prevent view destruction when tag remains unchanged
- Implemented case-insensitive tag comparison and parent validation in isTagChanged method to avoid false positives
- Added unit tests verifying rerender logic for actual tag changes versus case-only differences


modules/atomic-widgets/assets/js/editor/create-atomic-element-base-view.js
```diff
@@ -205,14 +210,10 @@ export default function createAtomicElementBaseView( type ) {
...
 			this.$el.addClass( this.getClasses() );
-
-			if ( this.isTagChanged( changed ) ) {
-				this.rerenderEntireView();
-			}
 		},
 
 		isTagChanged( changed ) {
-			return ( changed?.tag !== undefined || changed?.link !== undefined ) && this._parent && this.tagName() !== this.el.tagName;
+			return ( changed?.tag !== undefined || changed?.link !== undefined ) && this._parent && this.tagName() !== this.el.tagName.toLowerCase();
```
ronros-elementor: would it be best also to lower case the tagName() returned value?

IshayMaya: shannon-sharpe-suit-meme-why-not

ronros-elementor (approved):

PR #35285, opened 2026-03-25
Fix: Inline Editing Styling issues [ED-23337]

asafdl:
Changed inline editing component to insert itself into existing DOM element without completely replacing it, this fixes the style hierarchy

### ✨ PR Description
Purpose: Fix inline editing styling issues by refactoring React root management and editor mounting to preserve element styles and prevent CSS inheritance conflicts.

Main changes:
- Refactored inline editor to mount directly into content element instead of replacing parent, preserving original element structure and inherited styles
- Centralized React root management in view lifecycle with shared root instance across replacements, replacing per-instance root creation
- Removed ClickAwayListener wrapper and implemented custom click-outside detection to fix toolbar positioning and pointer event issues


packages/packages/libs/editor-controls/src/components/inline-editor.tsx
```diff
@@ -65,9 +64,12 @@ export const InlineEditor = React.forwardRef( ( props: InlineEditorProps, ref )
 		onEditorCreate,
 		wrapperClassName,
 		onSelectionEnd,
+		mountElement = null,
 	} = props;
 
 	const containerRef = useRef< HTMLDivElement >( null );
+	const onBlurRef = useRef( onBlur );
```
RobiFerentz, line 57: Why the need to keep onBlur as a ref here?

RobiFerentz (commented):
Just the one comment.

packages/packages/libs/editor-controls/src/components/inline-editor.tsx
asafdl: The onBlurRef ensures the ReactiveAnimationFrame callback inside useEditor.onBlur always calls the latest onBlur function — even if the parent re-renders with a new callback reference.

packages/packages/libs/editor-controls/src/components/inline-editor.tsx
```diff
@@ -170,47 +186,24 @@ export const InlineEditor = React.forwardRef( ( props: InlineEditorProps, ref )
...
-			<Wrapper
-				containerRef={ containerRef }
-				editor={ editor }
-				sx={ sx }
-				onBlur={ onBlur }
-				className={ wrapperClassName }
-			>
+			<Wrapper containerRef={ containerRef } sx={ sx } className={ wrapperClassName }>
 				<EditorContent ref={ ref } editor={ editor } />
 			</Wrapper>
```
ronenboxer: Is the wrapper still needed?
```suggestion
	return ( <Box ref={ containerRef } { ...sx } className={ className }>
		<EditorContent ref={ ref } editor={ editor } />
	</Box> );
```

ronenboxer (dismissed):

ronenboxer (approved):

PR #35334, opened 2026-03-29
Fix: Site Settings back button at root level does not close the panel [ED-23598]

Ntnelbaba:
## Summary
- Clicking the back button on the **main Site Settings page** now correctly closes the panel (same as the X button), clearing `active-document` from the URL
- **Root cause**: `back.js` used `panelHistory.length === 1` to detect the root, but `getHistory('panel')` returns the full panel history including routes from before Site Settings was opened — so history length was often > 1, bypassing the guard. `$e.routes.back('panel')` then navigated to a prior panel route without switching the document back, leaving the panel stuck
- **Fix**: replaced the length check with a route check (`panel/global/menu`). When at root, delegates to `panel/global/close` which properly switches the document and clears the query param

## Test plan
- [ ] Open Site Settings → navigate into a sub-section (e.g. Global Colors) → click back → should return to the main Site Settings page
- [ ] From the main Site Settings page, click the back button → panel should close and return to the normal editor, with `active-document` removed from the URL
- [ ] Confirm the close (X) button still works the same way
- [ ] Confirm ESC key (bound to `panel/global/back`) also closes correctly from the root page

## Jira
[https://elementor.atlassian.net/browse/ED-23598](https://elementor.atlassian.net/browse/ED-23598)

### ✨ PR Description
Purpose: Fix Site Settings back button behavior to properly close panel when navigating from root menu level instead of getting stuck.

Main changes:
- Replace generic history length check with explicit route detection for Site Settings menu page
- Update back button logic to trigger close dialog when at root route instead of checking history depth
- Improve code formatting consistency in isGlobalRoute method and dialog creation for better readability

[ED-23598]: https://elementor.atlassian.net/browse/ED-23598?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/kits/assets/js/commands/back.js
+++ b/core/kits/assets/js/commands/back.js
@@ -1,174 +1,174 @@
 import { EditorOneEventManager } from 'elementor-editor-utils/editor-one-events';
 
 export class Back extends $e.modules.CommandBase {
-  document = null;
-  confirmDialog = null;
-  unsavedChangesDialog = [];
+	document = null;
+	confirmDialog = null;
+	unsavedChangesDialog = [];
 
-  apply() {
-    const panelHistory = $e.routes.getHistory('panel');
-    const currentRoute = panelHistory[panelHistory.length - 1]?.route;
+	apply() {
+		const panelHistory = $e.routes.getHistory( 'panel' );
+		const currentRoute = panelHistory[ panelHistory.length - 1 ]?.route;
 
-    // When at the root Site Settings page, back should behave like close —
-    // switching back to the initial document and clearing the active-document query param.
-    if (currentRoute === this.component.getNamespace() + '/menu') {
-      $e.run('panel/global/close');
-      return;
-    }
+		// When at the root Site Settings page, back should behave like close —
+		// switching back to the initial document and clearing the active-document query param.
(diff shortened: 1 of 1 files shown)


core/kits/assets/js/commands/back.js
```diff
@@ -7,11 +7,12 @@ export class Back extends $e.modules.CommandBase {
 
 	apply() {
 		const panelHistory = $e.routes.getHistory( 'panel' );
+		const currentRoute = panelHistory[ panelHistory.length - 1 ]?.route;
 
-		// When there's no more previous pages to navigate back to,
-		// prompt the user with a confirmation dialog asking if they would like to exit.
-		if ( 1 === panelHistory.length ) {
-			this.getCloseConfirmDialog( event ).show();
```
Batsirai-muchareva, line 14: Since you remove getCloseConfirmDialog what will happen when if inside like layout and ai want to exit do I get a confimation before save

Batsirai-muchareva (approved):

PR #35368, opened 2026-03-31
Internal: Add prop base value support [ED-23616]

IshayMaya:
### ✨ PR Description
Purpose: Add base value support to prop control system to enable proper inheritance of component instance property values from parent components.

Main changes:
- Replaced placeholder-based inheritance with baseValue property throughout PropProvider, PropKeyProvider, and useBoundProp hook
- Modified ObjectPropKeyProvider and ArrayPropKeyProvider to merge sibling properties from baseValue when value is null
- Updated OverrideControl component to resolve and pass baseValue instead of placeholder for component override properties

### DIFF
--- a/packages/packages/core/editor-components/src/components/instance-editing-panel/override-prop-control.tsx
+++ b/packages/packages/core/editor-components/src/components/instance-editing-panel/override-prop-control.tsx
@@ -119,7 +119,7 @@ function OverrideControl( { overridableProp }: InternalProps ) {
 		propValue,
 		baseValue: resolvedBaseValue,
 		placeholderValue,
-	} = resolveValueAndBaseValue( matchingOverride, overrideValue, resolvedOriginValues, propKey );
+	} = resolveOverrideValues( matchingOverride, overrideValue, resolvedOriginValues, propKey );
 
 	const value = {
 		[ overridableProp.overrideKey ]: propValue,
@@ -224,7 +224,7 @@ function OverrideControl( { overridableProp }: InternalProps ) {
 	);
 }
 
-function resolveValueAndBaseValue(
+function resolveOverrideValues(
 	matchingOverride: ComponentInstanceOverride | null,
 	overrideValue: AnyTransformable | null,
 	resolvedOriginValues: ElementSettings,
--- a/packages/packages/core/editor-components/src/components/instance-editing-panel/override-prop-control.tsx
+++ b/packages/packages/core/editor-components/src/components/instance-editing-panel/override-prop-control.tsx
@@ -115,11 +115,12 @@ function OverrideControl( { overridableProp }: InternalProps ) {
 		return null;
 	}
 
-	const {
-		propValue,
-		baseValue: resolvedBaseValue,
-		placeholderValue,
-	} = resolveOverrideValues( matchingOverride, overrideValue, resolvedOriginValues, propKey );
+	const { propValue, baseValue: resolvedBaseValue } = resolveOverrideValues(
+		matchingOverride,
+		overrideValue,
+		resolvedOriginValues,
+		propKey
+	);
 
 	const value = {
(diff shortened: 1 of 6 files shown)


packages/packages/core/editor-components/src/components/instance-editing-panel/override-prop-control.tsx
```diff
@@ -281,21 +286,22 @@ function OverrideControl( { overridableProp }: InternalProps ) {
 
 type ElementSettings = Record< string, AnyTransformable | null >;
 
-function resolveValueAndPlaceholder(
+function resolveValueAndBaseValue(
```
Omerisra6: now we return placeholder + base lets come up with a better name

packages/packages/libs/editor-controls/src/bound-prop-context/prop-key-context.tsx
```diff
@@ -81,7 +92,7 @@ const ArrayPropKeyProvider = ( { children, bind }: PropKeyProviderProps ) => {
 	const { path } = useContext( PropKeyContext ) ?? {};
 
 	const setValue = ( value: PropValue, options?: CreateOptions ) => {
-		const newValue = [ ...( context.value ?? [] ) ];
+		const newValue = [ ...( context.value ?? context.baseValue ?? [] ) ];
```
Omerisra6, line 95: What is the case for this?

Is that supposed to be here? shouldnt we control it from the components infra?

packages/packages/core/editor-components/src/components/instance-editing-panel/override-prop-control.tsx
```diff
@@ -281,21 +286,22 @@ function OverrideControl( { overridableProp }: InternalProps ) {
...
 	const overrideValue = matchingOverride ? resolveOverridePropValue( matchingOverride ) : null;
 
-	const placeholderSettings = unwrapOverridableSettings( settingsWithInnerOverrides );
-	const inheritedValue = placeholderSettings[ propKey ] ?? null;
+	const unwrappedSettings = unwrapOverridableSettings( settingsWithInnerOverrides );
+	const inheritedValue = unwrappedSettings[ propKey ] ?? null;
 	const isInheritedDynamic = isDynamicPropValue( inheritedValue );
 
 	const propValue = isInheritedDynamic && ! matchingOverride ? inheritedValue : overrideValue;
+	const baseValue = isInheritedDynamic ? null : inheritedValue;
```
Omerisra6: why?

packages/packages/core/editor-components/src/components/instance-editing-panel/override-prop-control.tsx
Omerisra6, line 1: can you explain the difference between base and placholder?

packages/packages/libs/editor-controls/src/bound-prop-context/prop-key-context.tsx
IshayMaya: This is needed for compound props ( for example image with url, resolution ). 
When a user overrides only one nested key the value from the provider is null because no explicit override exists yet. Without baseValue as a fallback, the spread produces { resolution: newVal } and the sibling url is lost.

It has to live in PropKeyProvider rather than in the components infra because this is the layer that constructs the merged object before calling the parent setValue.

packages/packages/core/editor-components/src/components/instance-editing-panel/override-prop-control.tsx
IshayMaya: we currently have a custom solution for dynamics (there's a ticket to change it when we'll have the requirements)

as for the placeholder and baseValue - was redundant, removed the placeholder

Omerisra6 (approved):

PR #35376, opened 2026-03-31
Fix: Overridden components migration [ED-23607]

asafdl:
### ✨ PR Description
Purpose: Fix migration handling for overridden component props by adding support for Override_Prop_Type traversal in the migrations orchestrator.

Main changes:
- Added migrate_override_value() method to handle Override_Prop_Type migration by resolving prop types from component schemas
- Implemented resolve_override_value_prop_type() in Component_Override_Parser to extract correct prop types for override values
- Extended migration logic to recursively process override values nested within overridable props and component instances

### DIFF
--- a/modules/atomic-widgets/prop-type-migrations/migrations-orchestrator.php
+++ b/modules/atomic-widgets/prop-type-migrations/migrations-orchestrator.php
@@ -8,7 +8,9 @@ use Elementor\Modules\AtomicWidgets\PropTypes\Base\Array_Prop_Type;
 use Elementor\Modules\AtomicWidgets\PropTypes\Base\Object_Prop_Type;
 use Elementor\Modules\AtomicWidgets\PropTypes\Contracts\Prop_Type;
 use Elementor\Modules\AtomicWidgets\PropTypes\Union_Prop_Type;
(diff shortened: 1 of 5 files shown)


modules/atomic-widgets/prop-type-migrations/migrations-orchestrator.php
```diff
@@ -234,6 +236,10 @@ private function migrate_prop( &$value, Prop_Type $prop_type ): bool {
 					$has_changes = true;
 				}
 			}
+		} elseif ( $actual_prop_type instanceof Override_Prop_Type && is_array( $value['value'] ) ) {
```
Omerisra6, line 239: Maybe we could do it later because I know that PR is urgent, but maybe we should make this service extendable, then the components module will extend it

ronenboxer (approved):

Omerisra6 (approved):

modules/atomic-widgets/prop-type-migrations/migrations-orchestrator.php
asafdl: not sure exactly what the suggestion here is, but im sure that we can find a better way to model this part of the code, it's becoming very specific

PR #35294, opened 2026-03-25
Internal: `OverridePropControl`:Resolve all inner element settings [ED-23420]

MiryamOren:
## Summary
Currently, the override prop control resolves only the specific prop being edited (via `useResolvedOriginValue`). This PR changes the approach to resolve **all** settings of the innermost element — preparing for the next PR where we'll use these full settings to compute prop dependencies.

### How it works
The core addition is `walkDownOverridesChain` — a recursive function that traverses nested component instances from the outermost instance down to the innermost element. At each level it:
1. Finds the intermediate component instance element
2. Reads its overrides and collects them into an `OverridesMapping`, translating override keys across nesting levels (outer key → inner key)
3. Finds the next level's overridable prop and recurses.

Once the innermost element is found, its full settings are read via `getElementSettings`, the collected overrides are applied, and the current instance's overrides are applied on top. 
The prop value displayed in the control is now derived from these resolved settings:
```
// packages/packages/core/editor-components/src/components/instance-editing-panel/override-prop-control.tsx:161

const propValue = resolvedElementSettings[ propKey ];
```

### Other changes
- Removed `useResolvedOriginValue` hook (replaced by the new approach)
- Simplified `isExposedPropValid` by reusing `walkDownOverridesChain`

### ✨ PR Description
Purpose: Refactor override prop control to resolve complete chain of nested component instance settings and apply overrides recursively [ED-23420]

Main changes:
- Implemented resolveOverridesChain utility to recursively walk nested component instances and collect override mappings
- Added applyOverridesToSettings and unwrapOverridableSettings functions to merge overrides with element settings at all nesting levels
- Removed useResolvedOriginValue hook and replaced recursive origin value resolution with direct settings resolution from override chain

_Generated by LinearB AI and added by gitStream._
<sub>AI-generated content may contain inaccuracies. Please verify before using.
💡 **Tip:** You can customize your AI Description using **Guidelines** [Learn how](https://docs.gitstream.cm/automation-actions/#describe-changes)</sub>

[ED-23420]: https://elementor.atlassian.net/browse/ED-23420?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #35429, opened 2026-04-09
Fix: Component instance editing panel crashes on first load for nested components [ED-23660]

MiryamOren:
### The Bug 
When adding to the page a component that has a nested inner component with exposed-further props, the instance editing panel crashes on first load. 
On subsequent loads, or when the component was already in page,  it works correctly.

### Root cause 
We sent a separate request to get each component's overridable props. The outer component's data loaded first, and the panel tried to resolve the overrides chain before the inner component's data was loaded to the store, causing `resolveOverridesChain` to throw.
This crash was introduced by [a recent change](https://github.com/elementor/elementor/pull/35294) in `filterValidOverridableProps`, which now throws error when a nested component's overridable props are not found in the store. Before that change, the same situation caused a silent bug: the exposed-further props were filtered out and not shown in the panel.

### Fix
Changed the `GET /overridable-props` endpoint to accept an array of component IDs and return a map of results, instead of supporting a single component ID only.
All components' overridable props are now fetched in a single request - (whether on initial page load or when a new component is added to the page), guaranteeing that a component and its nested components' data are loaded to the store together before the instance editing panel renders.

### ✨ PR Description
Purpose: Fix crash in component instance editing panel by refactoring overridable-props REST API endpoint to support batch fetching of multiple component IDs instead of single component requests.

Main changes:
- Refactored REST API endpoint from single componentId to componentIds array parameter with batch processing and error handling in response metadata
- Updated TypeScript API client to fetch multiple component overridable props in single request with typed error responses
- Consolidated component overridable props loading logic to eliminate sequential API calls and filter already-loaded components

### DIFF
--- a/packages/packages/core/editor-components/src/store/actions/load-components-overridable-props.ts
+++ b/packages/packages/core/editor-components/src/store/actions/load-components-overridable-props.ts
@@ -10,22 +10,18 @@ export async function loadComponentsOverridableProps( componentIds: number[] ) {
 		return;
 	}
 
-	const { data, meta } = await apiClient.getOverridableProps( unloadedIds );
+	const { data } = await apiClient.getOverridableProps( unloadedIds );
 
-	for ( const [ componentId, overridableProps ] of Object.entries( data ) ) {
-		if ( ! overridableProps ) {
-			continue;
+	componentIds.forEach( ( componentId ) => {
+		if ( ! data[ componentId ] ) {
+			return;
 		}
 
 		dispatch(
 			slice.actions.setOverridableProps( {
-				componentId: Number( componentId ),
-				overridableProps,
+				componentId,
+				overridableProps: data[ componentId ],
 			} )
 		);
-	}
-
-	if ( meta?.errors && Object.keys( meta.errors ).length ) {
-		throw new Error( `Failed to load overridable props for some components: ${ JSON.stringify( meta.errors ) }` );
-	}
+	} );
 }
--- a/packages/packages/core/editor-components/src/store/actions/load-components-overridable-props.ts
+++ b/packages/packages/core/editor-components/src/store/actions/load-components-overridable-props.ts
@@ -3,31 +3,29 @@ import { __dispatch as dispatch, __getState as getState } from '@elementor/store
 import { apiClient } from '../../api';
 import { selectIsOverridablePropsLoaded, slice } from '../store';
 
-export function loadComponentsOverridableProps( componentIds: number[] ) {
-	if ( ! componentIds.length ) {
+export async function loadComponentsOverridableProps( componentIds: number[] ) {
+	const unloadedIds = componentIds.filter( ( id ) => ! selectIsOverridablePropsLoaded( getState(), id ) );
+
+	if ( ! unloadedIds.length ) {
 		return;
 	}
 
-	return Promise.all( componentIds.map( loadComponentOverrides ) );
-}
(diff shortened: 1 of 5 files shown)


packages/packages/core/editor-components/src/store/actions/load-components-overridable-props.ts
```diff
@@ -3,31 +3,29 @@ import { __dispatch as dispatch, __getState as getState } from '@elementor/store
...
+			} )
+		);
 	}
 
-	const overridableProps = await apiClient.getOverridableProps( componentId );
-
-	if ( ! overridableProps ) {
-		return;
+	if ( meta?.errors && Object.keys( meta.errors ).length ) {
+		throw new Error( `Failed to load overridable props for some components: ${ JSON.stringify( meta.errors ) }` );
```
IshayMaya: should we throw an error here? do we handle it somewhere?  i think its a bit weird since we handle the response only only then throw the error

packages/packages/core/editor-components/src/store/actions/load-components-overridable-props.ts
```diff
@@ -3,31 +3,29 @@ import { __dispatch as dispatch, __getState as getState } from '@elementor/store
...
 
-	if ( isOverridablePropsLoaded ) {
-		return;
+		dispatch(
+			slice.actions.setOverridableProps( {
+				componentId: Number( componentId ),
+				overridableProps,
+			} )
+		);
 	}
```
IshayMaya: maybe?
```suggestion
	componentIds.forEach( ( id ) => {
		if ( ! data[ id ] ) {
			return;
		}

		dispatch( slice.actions.setOverridableProps( { componentId: id, overridableProps: data[ id ] } ) );
	} );
```

packages/packages/core/editor-components/src/store/actions/load-components-overridable-props.ts
```diff
@@ -3,31 +3,29 @@ import { __dispatch as dispatch, __getState as getState } from '@elementor/store
...
-async function loadComponentOverrides( componentId: number ) {
-	const isOverridablePropsLoaded = selectIsOverridablePropsLoaded( getState(), componentId );
+	for ( const [ componentId, overridableProps ] of Object.entries( data ) ) {
+		if ( ! overridableProps ) {
+			continue;
+		}
 
-	if ( isOverridablePropsLoaded ) {
-		return;
+		dispatch(
```
IshayMaya: wdyt about creating a action for this? (update all the components at once in the store)

MiryamOren: It's not *all* components - on page load we fetch the components that exist in the page, and when adding a new component from the components panel, we load its & nested components data.

I think there's no need to add an action for multiple components, since react is [automatically batching the dispatches](https://dev.to/edriso/react-automatic-batching-how-many-re-renders-1273), so no unnecessary re-renders are caused as a result ץ

packages/packages/core/editor-components/src/store/actions/load-components-overridable-props.ts
MiryamOren: Deleted

packages/packages/core/editor-components/src/store/actions/load-components-overridable-props.ts
MiryamOren: Done

IshayMaya (approved):

PR #35459, opened 2026-04-13
Fix: Migrations run on frontend and preview [ED-23565] 

asafdl:
## 1. Problem & Context

**Ticket:** ED-23565

Migrations weren't running on frontend/preview renders, only on save. This caused documents with unmigrated data to display incorrectly until manually saved in the editor. Now applying migrations on document load via a filter hook, ensuring frontend always shows migrated structure without requiring editor intervention.

## 2. What Changed (Where)

- **includes/frontend.php**: Added `elementor/document/load/data` filter hook after fetching elements data to enable runtime migrations
- **modules/atomic-widgets/prop-type-migrations/migrations-orchestrator.php**: Added cache invalidation when migrating documents to prevent stale data

## 3. How It Works

Frontend loads document → `get_elements_data()` returns raw data → new `elementor/document/load/data` filter intercepts → migrations apply transformations → migrated data flows to rendering pipeline. The cache invalidation in the orchestrator ensures that after migration, the document cache is cleared so subsequent reads get fresh data.

## 4. Risks

Cache invalidation happens during migration write—if migrations run frequently on load without proper guards, could cause cache thrashing. Verify migrations only execute once per document version and that the filter hook doesn't double-migrate already-migrated content.


ronenelementor (dismissed):

modules/atomic-widgets/prop-type-migrations/migrations-orchestrator.php
```diff
@@ -337,6 +337,8 @@ private function migrate_doc( array $data, Document $document ): array {
 			$document->get_post()->ID,
 			Document::ELEMENTOR_DATA_META_KEY,
 			function( $migrated_data ) use ( $document ) {
+				$document->delete_meta( Document::CACHE_META_KEY );
```
IshayMaya, line 340: why is that needed?

includes/frontend.php
```diff
@@ -1147,6 +1147,19 @@ public function get_builder_content( $post_id, $with_css = false ) {
 
 		$data = $document->get_elements_data();
 
+		/**
+		 * Filters document elements data after loading.
+		 *
+		 * Allows modification of elements data when loading (not saving).
+		 * Useful for migrations, transformations, or data enrichment.
+		 *
+		 * @since 3.35.0
```
davseve: Will we merge this filter to 3.35?
3.35 is dead 🪦

davseve (approved):

PR #35462, opened 2026-04-13
Fix: Harden global classes endpoint against migration fetch timeouts [ED-23544]

Omerisra6:
## PR Checklist

## PR Type

- [x] Bugfix

## Summary

* Fix: Prevent global classes from disappearing when the remote migration manifest fetch times out or fails

## Description

**Problem:** The `GET /global-classes` endpoint could hang for up to 10 seconds (or until PHP `max_execution_time`) when `migrations.elementor.com` was slow or unreachable. This caused the client-side store to remain empty, making all global classes appear to vanish. The client-side `PopulateStore` component also lacked error handling, silently swallowing fetch failures.

**Solution — 5 layers of resilience:**

1. **Gate migration with `is_active()` check** — Skip the migration orchestrator entirely when the `e_bc_migrations` experiment is inactive, avoiding the remote fetch altogether in most cases.

2. **Cache remote manifest via WordPress transients** — Cache the manifest with a 12-hour TTL. On transient miss, attempt a remote fetch and persist a stale copy via `wp_option`. On fetch failure, fall back to the stale copy. This ensures that even if the remote server is down, previously-fetched manifests are reused.

3. **Reduce `wp_remote_get` timeout** — Lowered from 10s to 3s to fail fast on unreachable hosts instead of blocking the entire request.

4. **Catch `\Throwable` in `route_wrapper`** — Changed from `\Exception` to `\Throwable` to also catch PHP `\Error` (e.g., `Maximum execution time exceeded`), returning a proper REST API error response instead of a raw 500.

5. **Add `.catch()` to client-side `PopulateStore`** — Prevents silent promise rejections so errors are logged and the Redux store state is predictable.

### Key decisions

- **Show classes without migration over showing nothing** — If the manifest can't be fetched and no cache exists, classes are returned unmigrated rather than failing the entire endpoint. This trades off potential stale prop-type formats for guaranteed class visibility.
- **Stale-while-revalidate pattern** — The `wp_option`-based stale fallback ensures that even after the transient expires and a fresh fetch fails, the last known good manifest is used.

## Test instructions

1. Enable the `e_bc_migrations` experiment
2. Block network access to `migrations.elementor.com` (e.g. via hosts file or firewall rule)
3. Open the editor and verify global classes load correctly (using cached/stale manifest)
4. Disable the `e_bc_migrations` experiment
5. Verify global classes still load (migration is skipped entirely)

## Quality assurance

Fixes ED-23544

### DIFF
--- a/modules/atomic-widgets/prop-type-migrations/migrations-loader.php
+++ b/modules/atomic-widgets/prop-type-migrations/migrations-loader.php
@@ -9,6 +9,9 @@ if ( ! defined( 'ABSPATH' ) ) {
 }
 
 class Migrations_Loader {
+	const MANIFEST_TRANSIENT_KEY = 'elementor_migrations_manifest';
+	const MANIFEST_TRANSIENT_TTL = 12 * HOUR_IN_SECONDS;
+
 	private static ?self $instance = null;
 
 	private ?array $manifest = null;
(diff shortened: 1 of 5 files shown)


Omerisra6:
Closing in favor of a more targeted fix — reducing only the wp_remote_get timeout from 10s to 3s.

PR #35464, opened 2026-04-13
Internal: Expose history item ID from undo/redo commands and data hooks [ED-23506]

MiryamOren:
### Some small expansions of the history infrastructure
- Return the original history item ID from `Undo` and `Redo` commands
- Extend `registerDataHook` to pass `currentHistoryItemId` on `HookOptions

These changes^ enable data hooks to correlate side effects with the history item that triggered them.

- Forward `options` and `meta` through `OverridePropControl.setValue` to support `{ withHistory: false }` downstream.

All these changes are infrastructure to support undo/redo of overridable prop operations in elementor-pro.

## 1. Problem & Context

Overridable prop changes needed to correlate undo/redo operations with their origin history items for proper state restoration. Without exposing history item IDs, the system couldn't track which history entry triggered a prop override change, breaking undo/redo functionality for component instance overrides. Jira ticket: ED-23506.

## 2. What Changed (Where)

- `redo.js` / `undo.js`: Commands now return `originHistoryItemId` from navigate operation
- `history/manager.js`: `navigate()` returns the history item being undone/redone
- `override-prop-control.tsx`: `setValue` accepts optional `meta` parameter to pass history context through
- `register-data-hook.ts`: Data hooks receive `currentHistoryItemId` in options via `getCurrentId()` adapter

## 3. How It Works

Undo/redo commands capture the history item from `navigate()` and return its ID. This ID flows through the prop system via `SetValueMeta` to `setInstanceValue()`, where it becomes available to data hooks through `getCurrentId()`. The window adapter exposes `elementor.documents.getCurrent().history.getCurrentId()` so React code can read the current history item during prop updates, enabling proper correlation of override changes with their triggering history operations.

## 4. Risks

None significant. Optional chaining (`?.`) guards against missing history manager. Backward compatible — existing callers ignoring return values/new params continue working unchanged.

[ED-23506]: https://elementor.atlassian.net/browse/ED-23506?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/packages/core/editor-components/src/components/instance-editing-panel/override-prop-control.tsx
```diff
@@ -164,10 +165,14 @@ function OverrideControl( { overridableProp }: InternalProps ) {
...
-		setInstanceValue( {
-			...instanceValue,
-			overrides: componentInstanceOverridesPropTypeUtil.create( newOverrides ),
-		} );
+		setInstanceValue(
+			{
+				...instanceValue,
+				overrides: componentInstanceOverridesPropTypeUtil.create( newOverrides ),
+			},
+			options,
```
Omerisra6, line 173: is it a new api?

MiryamOren: The `setValue` we get from `useBoundProp` already has these parameters, so it's actually a bug we havn't passed them before

Omerisra6 (approved):

PR #35466, opened 2026-04-13
Fix: Sanitize faulty styles order data to allow saving modifications upon validation [ED-23644]

ronros-elementor:
## 1. Problem & Context

Faulty `order` array data in global classes (orphaned IDs, non-strings, duplicates) was blocking save operations after validation. This sanitizes order arrays at read time to ensure consistency between `items` and `order`, allowing modifications to persist. Ticket: ED-23644.

## 2. What Changed (Where)

- `global-classes-parser.php`: Added `sanitize_order()` to reconcile order array with actual item IDs
- `global-classes-repository.php`: Sanitizes order on every `all()` call using new parser method
- `import-export-customization/runners/import.php`: Sanitizes order before parsing imported data
- `import-export/import-runner.php`: Same sanitization for legacy import path
- `template-library-global-classes-snapshot-builder.php`: Sanitizes order in snapshot parsing
- `test-atomic-global-styles.php`: Added test verifying orphaned order entries are filtered out

## 3. How It Works

`sanitize_order()` filters out non-strings and duplicates, then splits the order into two groups: existing IDs (preserving user ordering) and missing IDs (sorted alphabetically). The sanitization runs at every read boundary—repository fetch, import, and snapshot parsing—ensuring corrupt data never reaches validation. Trade-off: Missing items always append alphabetically rather than preserving any legacy order hints.

## 4. Risks

None significant. Sanitization is defensive and idempotent. Worst case: alphabetic fallback for missing items differs from user's original intent if order was previously corrupted, but that data was already unusable.


tests/phpunit/elementor/modules/global-classes/test-global-classes-sanitization.php
```diff
@@ -0,0 +1,32 @@
+<?php
```
Omerisra6: I think its redundant if we have the integration test

modules/global-classes/global-classes-repository.php
```diff
@@ -54,6 +55,8 @@ public function all( bool $force = false ): Global_Classes {
 			$all = $kit->get_json_meta( static::META_KEY_FRONTEND );
 		}
 
+		$all['order'] = Global_Classes_Sanitization::order( $all['items'] ?? [], $all['order'] ?? [] );
```
Omerisra6: don't we need to add it to all other actions? (when saving and updating)

modules/global-classes/utils/global-classes-sanitization.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace Elementor\Modules\GlobalClasses\Utils;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+class Global_Classes_Sanitization {
+	public static function order( array $items, array $order ): array {
```
Omerisra6: maybe it should be apart of the `global-classes-parser`

Omerisra6 (approved):

PR #35467, opened 2026-04-13
Fix: [V4] Registering handlers after DOMContentLoaded is done [ED-23643]

ronenelementor:
## 1. Problem & Context

Handlers (link actions, form submissions) weren't attaching to dynamically-rendered atomic widgets on frontend because registration fired before DOM was ready. ED-23643 fixes this by deferring handler initialization until DOMContentLoaded and observing subsequent mutations.

## 2. What Changed (Where)

| File | Change |
|------|--------|
| `frontend-handlers/src/init.ts` | Boot handlers on DOMContentLoaded; add MutationObserver for dynamic elements |
| `handlers.js` | Refactor `handleLinkActions()` to accept registration element, extract actual link target; refactor form handler setup |
| `atomic-element-base.php` | Move link attributes generation from template to PHP; store as pre-escaped string in settings |
| `has-atomic-base.php` | Compute link attributes server-side via `get_link_attributes_string()` |
| `*.twig` templates | Consume pre-computed `link.attributes` string instead of building href/target inline |
| JS editor view | Update tag resolution to prefer link tag; refactor `getLink()` → `getLinkAttributes()` returning object |
| TypeScript view, tests | Mirror link attributes handling changes |

## 3. How It Works

**Problem path:** Editor renders element → frontend loads → handlers register before DOM ready → dynamic atoms never get listeners.

**New flow:** Init waits for DOMContentLoaded → scans existing `[data-e-type]` elements → MutationObserver catches additions → each triggers `onElementRender()` which fires registered callbacks (link action, form handler).

**Link attributes:** Server generates `href="..." target="_blank"` string during `get_atomic_settings()`, stored in `link.attributes`. Templates output raw. Frontend applies via `getLinkAttributes()` returning keyed object. Selector changed from `:has(> selector)` to `:has(selector)` for broader parent matching.

## 4. Risks

**Handler double-attachment on editor + frontend contexts:** Both `elementor/element/render` event and MutationObserver could fire. Mitigated by `onElementRender()` being idempotent per element ID.

**MutationObserver performance:** Unbounded observation could thrash on large DOMs. Mitigated by batching mutations via RAF and deduplication via Set.


modules/atomic-widgets/elements/base/has-atomic-base.php
```diff
@@ -5,6 +5,7 @@
 use Elementor\Element_Base;
 use Elementor\Modules\AtomicWidgets\Controls\Base\Atomic_Control_Base;
 use Elementor\Modules\AtomicWidgets\Controls\Section;
+use Elementor\Modules\AtomicWidgets\Elements\Atomic_Button\Atomic_Button;
```
marcin-el: unused import?

modules/atomic-widgets/assets/js/frontend/handlers.js
```diff
@@ -20,47 +21,85 @@ registerBySelector( {
...
+
+	if ( ! alpineId ) {
 		return;
 	}
 
-	const handler = ( event ) => {
-		if ( actionLinkElement && actionLinkElement !== element && ! actionLinkElement.contains( event.target ) ) {
-			return;
-		}
+	const url = actionLinkElement.dataset.actionLink;
```
marcin-el: shouldn't this also be
```
actionLinkElement.dataset.actionLink ||
		actionLinkElement.getAttribute( 'href' ) ||
		''
```
like on line 78?

ronenboxer: I saw that our twig had too many ifs and dynamic attributes, so I consolidated it into a function that prints the attributes into a new setting, so it can't have an href attribute when it's an action link, it will have this dataset and some alpine attributes
https://github.com/elementor/elementor/blob/3b5d79afcfed164f82c34d475e4e7ed8dda9a0c4/modules/atomic-widgets/elements/base/has-atomic-base.php#L373

we can still have it as a safe fallback but i think its redundant, WDYT?

ronenelementor:
Pro test passing
https://github.com/elementor/elementor-pro/actions/runs/24400363359/job/71274102088#step:4:17

marcin-el (dismissed):

packages/packages/core/editor-canvas/src/legacy/create-templated-element-type.ts
```diff
@@ -214,5 +214,40 @@ export function createTemplatedElementView( {
 
 			return originId ?? id;
 		}
+
+		_getLinkAttributes( settings: { [ key: string ]: unknown } ) {
+			const linkAttributes = Object.entries( this._handleActionLinkAttributes( settings ) )
+				.map( ( [ key, value ] ) => `${ key }="${ value }"` )
```
mykytamurzin, line 220: should be escaped?

modules/atomic-widgets/assets/js/frontend/handlers.js
```diff
@@ -20,47 +21,85 @@ registerBySelector( {
 	callback: ( { element } ) => handleAtomicFormSubmit( element ),
 } );
 
-function handleLinkActions( element ) {
-	const actionLinkElement = element.matches( ACTION_LINK_SELECTOR )
-		? element
-		: element.querySelector( ACTION_LINK_SELECTOR );
-	const url = actionLinkElement?.dataset.actionLink;
+function registerAtomicLinkAlpineData( actionLinkElement, registrationElement ) {
```
mykytamurzin: As for me this file is structured in a bad way - plain functions with mixing logic, which is hard to read and understand what's going on. I suggest to split the things by their responsibility: 
1. `alpine/` — all Alpine data factories and lifecycle (registerAtomicLinkAlpineData, registerAtomicFormAlpineData, refreshDom, the getAlpineId helpers). One place that owns the Alpine dependency.
2. `link-action.js` — pure link logic (shouldFireLinkActionHandler, allowlist) 
3. `form.js` — pure form logic (buildAtomicFormPayload, submitAtomicForm, field helpers, setFormState)
4. `handlers.js` — entry points that wire Alpine registrations to the business logic via registerBySelector

modules/atomic-widgets/elements/base/has-atomic-base.php
```diff
@@ -241,14 +241,21 @@ final public function get_stack( $with_common_controls = true ) {
...
-		) );
+		);
+		$props['attributes'] = Attributes_Prop_Type::generate( $merged_attribute_values );
+
+		$parsed = Render_Props_Resolver::for_settings()->resolve( $schema, $props );
+		$link_attributes = isset( $parsed['link'] ) ? $this->get_link_attributes_string( $parsed['link'] ) : '';
 
-		return Render_Props_Resolver::for_settings()->resolve( $schema, $props );
+		if ( ! empty( $link_attributes ) ) {
+			$parsed['link_attributes'] = $link_attributes;
```
mykytamurzin: can we put it inside link setting? just to have everything link related into one place - `settings.link`(`settings.link.href`, `settings.link.tag`, `settings.link.attributes`). Probably this will help to simplify conditions in twig templates

modules/atomic-widgets/assets/js/frontend/handlers.js
ronenboxer: I agree
I have a task with a PR just for that

packages/packages/core/editor-canvas/src/legacy/create-templated-element-type.ts
ronenboxer: I don't think so
It didn't get escape so far and these are controlled attributes that we are in charge of (`x-data` and `x-on:click`). data-action-link is either way not escaped (twig has it ` | raw`)

modules/atomic-widgets/elements/base/has-atomic-base.php
ronenboxer: Done

mykytamurzin (dismissed):

mykytamurzin (approved):

ronenelementor:
Pro test passing
https://github.com/elementor/elementor-pro/actions/runs/24460013620/job/71473919537#step:4:17

PR #35475, opened 2026-04-14
Internal: Twig-based rendering for div-block and flexbox containers [ED-23173]

Omerisra6:
## Summary

Adds Twig template rendering for **e-div-block** and **e-flexbox** elements, aligning them with the existing **atomic-tabs** approach for unified editor/frontend rendering.

Gated behind a new hidden experiment: **e_twig_containers** (inactive by default, dev release status).

### Jira
https://elementor.atlassian.net/browse/ED-23173

## Problem

Currently, `e-div-block` and `e-flexbox` render their HTML through PHP's `before_render()` / `after_render()` methods. This means the editor (JavaScript) and frontend (PHP) use completely different rendering paths, making it difficult to achieve rendering parity. The `atomic-tabs` element already solved this by using Twig templates — the same template renders on both sides.

### 1. Extend, don't replace (backward compatibility)
- Created **`Div_Block_Twig`** and **`Flexbox_Twig`** PHP classes that extend the original `Div_Block` and `Flexbox` classes
- These subclasses add the `Has_Element_Template` trait (same pattern as `Atomic_Tabs`)
- When the experiment is **off**, the original classes are registered — zero behavioral change
- When the experiment is **on**, the Twig variants are registered instead

### 2. Shared Twig macros for common rendering logic
- Created `_macros.html.twig` with reusable macros: `render_base_classes`, `render_data_attributes`, `render_custom_attributes`, `render_link_attributes`, `render_interactions`
- Both `div-block.html.twig` and `flexbox.html.twig` import and use these macros
- Macros are registered as shared templates via `Has_Element_Template::get_shared_templates()`

### 3. Twig templates handle dynamic HTML tags
- Both elements support configurable HTML tags (`div`, `a`, `button`, `header`, etc.)
- The Twig template dynamically resolves the tag based on `settings.tag` and `settings.link`
- Link elements switch to `<a>` tags; `button` tags use `data-action-link` instead of `href`

### 4. Editor-side fixes for templated views
- **attributes() override**: Templated views now return only `data-model-cid` from `attributes()`, preventing Marionette from setting stale attributes that conflict with Twig-rendered output
- **Tag swap on setting change**: `_attachTwigContent` now detects when the Twig-rendered HTML has a different tag than the current DOM element and performs a full element replacement (create new element → copy attributes + innerHTML → replaceWith → update Marionette's `$el`)

### 5. Bug fix in editor element type override
- **Pre-existing bug found**: `init-legacy-views.ts` referenced `elementsManager._elementTypes` but the actual property is `elementTypes` (no underscore prefix)
- **Fix**: `_elementTypes` → `elementTypes` in both `init-legacy-views.ts` and `types.ts`

### 6. Test isolation
- `e_twig_containers` excluded from auto-activation in `tests/bootstrap.php` so existing div-block/flexbox snapshot tests continue to pass unchanged
- Twig-specific tests directly instantiate the Twig variant classes

## Files Changed

| File | Change |
|------|--------|
| `modules/atomic-widgets/module.php` | Register experiment, conditional element registration |
| `modules/atomic-widgets/elements/base/has-element-template.php` | Shared macros registration, PHP 8.0 compat fix |
| `modules/atomic-widgets/elements/base/_macros.html.twig` | New: Shared Twig macros for common rendering |
| `modules/atomic-widgets/elements/div-block/div-block-twig.php` | New: Twig variant of Div_Block |
| `modules/atomic-widgets/elements/div-block/div-block.html.twig` | New: Twig template |
| `modules/atomic-widgets/elements/flexbox/flexbox-twig.php` | New: Twig variant of Flexbox |
| `modules/atomic-widgets/elements/flexbox/flexbox.html.twig` | New: Twig template |
| `packages/.../create-nested-templated-element-type.ts` | Fix: attributes() override + tag swap logic |
| `packages/.../init-legacy-views.ts` | Fix: `_elementTypes` → `elementTypes` |
| `packages/.../types.ts` | Fix: type definition to match runtime |
| `tests/.../test-twig-containers.php` | PHPUnit: snapshot + parity tests |
| `tests/bootstrap.php` | Exclude experiment from auto-activation |
| `tests/playwright/.../twig-containers-editor.test.ts` | New: Playwright E2E tests for editor rendering |

## How to Test

1. Activate the experiment:
   ```
   wp elementor experiments activate e_twig_containers
   ```
2. Open the Elementor editor, add a div-block or flexbox with child widgets
3. **Editor**: Elements should render correctly in the canvas (Twig rendering in editor)
4. **Tag changes**: Change the HTML tag via settings → element tag updates in the canvas
5. **Link**: Add a link → element switches to `<a>` tag with `href`
6. **CSS ID**: Set a CSS ID → `id` attribute appears on the element
7. **Frontend**: View the page — inspect HTML to confirm structure matches editor
8. **Experiment off**: Verify everything still works as before (original PHP rendering)

## Test Plan
- [x] PHPUnit snapshot tests for Twig rendering output
- [x] PHPUnit parity tests (Twig output matches non-templated PHP output)
- [x] Playwright E2E: div-block/flexbox render with correct attributes
- [x] Playwright E2E: tag changes via settings update DOM element
- [x] Playwright E2E: link changes tag to anchor
- [x] Playwright E2E: nested containers render correctly
- [x] Playwright E2E: CSS ID applies id attribute
- [ ] Manual: Experiment off = no behavioral change

## 1. Problem & Context

Standardizes rendering for div-block and flexbox containers using Twig templates instead of PHP-driven HTML. This unifies editor/frontend rendering paths under a feature flag (`e_twig_containers`), reducing divergence between preview and published output. The change is gated behind an experiment to validate behavior parity before rollout. [ED-23173]

## 2. What Changed (Where)

| File | Change |
|------|--------|
| `modules/atomic-widgets/elements/div-block/div-block-twig.php` | New Twig variant of Div_Block |
| `modules/atomic-widgets/elements/flexbox/flexbox-twig.php` | New Twig variant of Flexbox |
| `modules/atomic-widgets/elements/base/_macros.html.twig` | Shared Twig macros for link attributes, classes, data attributes |
| `modules/atomic-widgets/elements/div-block/div-block.html.twig` | Twig template for div-block rendering |
| `modules/atomic-widgets/elements/flexbox/flexbox.html.twig` | Twig template for flexbox rendering |
| `modules/atomic-widgets/elements/base/has-atomic-base.php` | Extracts `transform_link_for_render()` for PHP link transformation |
| `modules/atomic-widgets/elements/base/has-template.php` | Adds shared templates mechanism, macro registration |
| `modules/atomic-widgets/elements/base/has-element-template.php` | Same shared templates pattern for nested elements |
| `modules/atomic-widgets/module.php` | Feature flag registration, conditional element registration |
| `packages/core/editor-canvas/src/legacy/create-nested-templated-element-type.ts` | Tag-swap logic for dynamic element updates, attribute injection |
| `packages/core/editor-canvas/src/legacy/create-templated-element-type.ts` | Removes client-side link transformation (now handled server-side via templates) |
| Twig templates (button, heading, image, paragraph, svg) | Migrated to macro-based link rendering |
| `tests/phpunit/elementor/modules/atomic-widgets/test-twig-containers.php` | Comprehensive PHPUnit tests for output parity |
| `tests/playwright/sanity/modules/v4-tests/twig-containers-editor.test.ts` | Editor-specific Playwright tests |

## 3. How It Works

**Registration flow:**
1. `Module.php` checks `e_twig_containers` flag → registers `Div_Block_Twig`/`Flexbox_Twig` or original classes
2. Twig variants inherit base functionality but override `get_templates()` to return `.html.twig` paths
3. `has-template.php` merges shared macros (`_macros.html.twig`) with element-specific templates via `get_shared_templates()`

**Rendering:**
- **Server-side (PHP):** `transform_link_for_render()` is now a no-op for Twig elements; link transformation moved into templates
- **Templates:** Macros handle link attributes (`render_link_attributes`) with action-link detection (button tag → `data-action-link`, anchor → `href`)
- **Editor (TS):** `create-nested-templated-element-type.ts` now handles tag swapping when `oldEl.tagName !== newEl.tagName` (e.g., div→a when link added), preserves overlay HTML, and injects `data-model-cid`

**Key decisions:**
- Link transformation moved from PHP (`_getLinkAttributes`) to Twig macros for consistency
- Action links (button tags) use `data-action-link` instead of `href` to prevent navigation
- Tag-swap logic clones attributes to new element before replacing to avoid losing state

## 4. Risks

**Tag swap timing:** The `needsTagSwap` branch creates a new DOM element and calls `setElement()`, which could trigger unintended re-initialization if child views aren't properly detached. Tests validate children persist, but monitor for edge cases with deeply nested structures or Alpine.js directives.

**Feature flag leakage:** Tests explicitly exclude `e_twig_containers` from auto-activation (line 68 in `bootstrap.php`). If flag logic fails, prod could mix Twig/non-Twig rendering for the same element types, breaking assumptions in editor interactions or styles. Ensure flag checks are consistent across frontend/editor contexts.

**Template registration collision:** `get_shared_templates()` returns a fixed key (`'elementor/macros'`). If multiple element types register concurrently, last-write-wins. Current code checks `is_registered()` before re-registering, but race conditions during parallel tests could cause flakiness.

[ED-23173]: https://elementor.atlassian.net/browse/ED-23173?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/packages/core/editor-canvas/src/legacy/init-legacy-views.ts
```diff
@@ -101,7 +101,7 @@ function tryRegisterElement(
 		const canOverrideExisting = canBeNestedTemplated( element ) && isAlreadyRegistered;
 
 		if ( canOverrideExisting ) {
-			elementsManager._elementTypes[ type ] = new ResolvedElementType();
+			elementsManager.elementTypes[ type ] = new ResolvedElementType();
```
IshayMaya, line 104: why?

tests/bootstrap.php
```diff
@@ -64,7 +64,9 @@
 
 // The following action activates all registered experiments in order for them to be able to be tested.
 add_action( 'elementor/experiments/feature-registered', function ( Experiments_Manager $experiments_manager, array $experimental_data ) {
-	$exclude = [];
+	$exclude = [
```
IshayMaya, line 67: why exclude?

modules/atomic-widgets/elements/flexbox/flexbox.html.twig
```diff
@@ -0,0 +1,29 @@
+{% set tag = settings.tag | default('div') %}
```
IshayMaya: can we use the same template for all? (going to have grid as well soon)

modules/atomic-widgets/module.php
```diff
@@ -279,8 +291,10 @@ private function register_widgets( Widgets_Manager $widgets_manager ) {
 	}
 
 	private function register_elements( Elements_Manager $elements_manager ) {
-		$elements_manager->register_element_type( new Div_Block() );
-		$elements_manager->register_element_type( new Flexbox() );
+		$use_twig_containers = Plugin::$instance->experiments->is_feature_active( self::EXPERIMENT_TWIG_CONTAINERS );
+
+		$elements_manager->register_element_type( $use_twig_containers ? new Div_Block_Twig() : new Div_Block() );
```
IshayMaya, line 296: i think that nested templated should use the lightest possible version of the base view, we might be hiding existing issues, and future fixes might be applied on the wrong files.
check if we can have 2 different bases (one that will be deprecated) or an extended version of it under the ff

tests/bootstrap.php
Omerisra6: because I still want all tests to run with the old version, you think we shouldnt?

modules/atomic-widgets/elements/flexbox/flexbox.html.twig
Omerisra6: added macros to make it composable

modules/atomic-widgets/module.php
Omerisra6: We are overriding everything we need; all attributes and render methods come from Twig.

I don't see an easy way to disconnect it...

and we will keep using this view anyway

packages/packages/core/editor-canvas/src/legacy/init-legacy-views.ts
Omerisra6: The `_elementTypes` doesn't exist, it was dead code, the right name is `elementTypes`

tests/bootstrap.php
IshayMaya: i think we want it to run with the new one no? so we will catch any issues, its going to be the default eventually

Omerisra6: I think we should run the one on prod; we can run the other version as well. but not instead, WDYT?

packages/packages/core/editor-canvas/src/legacy/create-nested-templated-element-type.ts
```diff
@@ -240,14 +246,21 @@ export function createNestedTemplatedElementView( {
...
+				targetEl.setAttribute( attr.name, attr.value );
 			} );
 
-			oldEl.setAttribute( 'draggable', 'true' );
+			targetEl.setAttribute( 'draggable', 'true' );
+			targetEl.innerHTML = overlayHTML + newEl.innerHTML;
 
-			const overlayHTML = this.getHandlesOverlay()?.get( 0 )?.outerHTML ?? '';
-			oldEl.innerHTML = overlayHTML + newEl.innerHTML;
+			if ( needsTagSwap ) {
```
IshayMaya, line 260: why isnt it handled in the twig?

packages/packages/core/editor-canvas/src/legacy/create-nested-templated-element-type.ts
```diff
@@ -240,14 +246,21 @@ export function createNestedTemplatedElementView( {
 
 			this._destroyAlpine();
```
IshayMaya, line 248: not sure what happened but it saw an issue with link in link detection that stopped working in components

Omerisra6: its related to ronen's changes

packages/packages/core/editor-canvas/src/legacy/create-nested-templated-element-type.ts
Omerisra6: Because we need to update the Marionette view and set the new element.

For widgets, we don't need to do it because the wrapper div is a Marionette element.

modules/atomic-widgets/elements/base/_macros.html.twig
```diff
@@ -0,0 +1,24 @@
...
+{%- endmacro -%}
+
+{%- macro render_custom_attributes(settings, editor_attributes) -%}
+{% if settings._cssid is defined and settings._cssid is not empty %} id="{{ settings._cssid | e }}"{% endif %}
+{%- if settings.attributes is defined and settings.attributes is not empty %} {{ settings.attributes | raw }}{% endif %}
+{%- if editor_attributes is defined and editor_attributes is not empty %} {{ editor_attributes | raw }}{% endif %}
+{%- endmacro -%}
+
+{%- macro render_link_attributes(settings, tag) -%}
+{%- if settings.link is defined and settings.link.href is defined and settings.link.href is not empty -%}
```
ronenboxer:
```suggestion
{%- if settings.link is defined and settings.attributes is not empty -%}
```

modules/atomic-widgets/elements/base/_macros.html.twig
```diff
@@ -0,0 +1,24 @@
...
+{%- if tag == 'button' %} data-action-link="{{ settings.link.href | e }}"{% else %} href="{{ settings.link.href | e }}"{% endif -%}
+{%- if settings.link.target is defined and settings.link.target is not empty %} target="{{ settings.link.target | e }}"{% endif -%}
```
ronenboxer:
```suggestion
{{ settings.link.attributes | raw }}
```

modules/atomic-widgets/elements/atomic-button/atomic-button.html.twig
```diff
@@ -1,16 +1,17 @@
+{% import 'elementor/macros' as m %}
 {% if settings.text is not empty %}
     {%- set allowed_tags = '<b><strong><sup><sub><s><em><i><u><del><span><br>' %}
     {% set classes = settings.classes | merge( [ base_styles.base ] ) | join(' ') %}
 	{% set id_attribute = settings._cssid is not empty ? 'id=' ~ settings._cssid | e('html_attr') : '' %}
-    {% if settings.link and settings.link.attributes is not empty %}
-        <{{ settings.link.tag | e('html_tag') }}
-            {{ settings.link.attributes | raw }}
+    {% if settings.link is defined and settings.link.href is defined and settings.link.href is not empty %}
+        <{{ settings.link.tag | default('a') | e('html_tag') }}
```
ronenboxer: This is unnecessary since the default value for link.tag is already set as 'a'
```suggestion
        <{{ settings.link.tag | e('html_tag') }}
```

packages/packages/core/editor-canvas/src/legacy/create-nested-templated-element-type.ts
ronenboxer: hi

modules/atomic-widgets/elements/atomic-heading/atomic-heading.html.twig
```diff
@@ -1,3 +1,4 @@
+{% import 'elementor/macros' as m %}
```
asafdl, line 1: if you are going over all the twig templates we have a task to use whitespace trimmers 
https://symfony.com/blog/better-white-space-control-in-twig-templates
so might aswell
https://elementor.atlassian.net/browse/ED-22789

modules/atomic-widgets/elements/flexbox/flexbox.html.twig
asafdl: but why do they have different twig templates?

IshayMaya (approved):

PR #35531, opened 2026-04-19
Internal: Context mcp data divided into resources [ED-23679]

eavichay-elementor:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #

## 1. Problem & Context

Split monolithic context resource into granular, independently-updatable resources (selected-element, editor-state, general-context, available-widgets) to enable fine-grained MCP subscriptions and reduce unnecessary data transmission. Ticket: ED-23679.

## 2. What Changed (Where)

| File | Change |
|------|--------|
| `canvas-mcp.ts` | Import and init 4 new resources; reorder initialization |
| `document-structure-resource.ts` | Improved type safety (proper V1 types); added v3/v4 version detection; listen to document-switch events |
| `available-widgets-resource.ts` | New resource exposing v3/v4 widgets with metadata; dual URIs for all vs. v4-only |
| `selected-element-resource.ts` | New resource tracking selection, properties, version; filters defaults; supports deselect |
| `editor-state-resource.ts` | New resource: page title, preview text snapshot, viewed-screen label |
| `general-context-resource.ts` | New resource: timezone, post-id, page-url, plugin list |
| `widgets-schema-resource.ts` | V3 control metadata extraction; filter available widgets; improved schema fallback for v3 |
| `build-composition/tool.ts` | Assertion: v4 widgets only; fail fast on v3 usage |
| `configure-element/tool.ts` | Validation: reject v3 elements with clear message |
| `get-element-config/tool.ts` | Type extraction refactored; v3 guard added |
| `element-data-util.ts` | New utility: widget availability checks, version detection |
| `variables-resource.ts` | Added v4 version tag; event-driven updates; cleaner build logic |
| `elementor-v3-mcp/context.ts` | Removed `getPageOverView()` (moved to canvas MCP as document-structure) |
| `elementor-v3-mcp/resources.ts` | Removed PAGE_OVERVIEW resource exports |

## 3. How It Works

Each resource independently manages its data lifecycle: available-widgets scans cache on v1-ready and broadcasts updates; selected-element listens to select/deselect commands; editor-state watches doc-switch events; general-context tracks timezone/post-id/url. Tools now validate element type against widget cache upfront—if v3 or unknown, reject with helpful routing message. Document structure moved from v3-mcp to canvas-mcp and enhanced with atomic_props_schema detection for version tagging.

## 4. Risks

Type definitions slightly fragile across files (SelectionContainer, UnknownVersionElementInstanceData)—inconsistency if V1 APIs shift. V3 control metadata extraction could be incomplete if control.options has non-standard shape. Tools fail hard on v3 elements; ensure docs/messaging guide users to v3-mcp tools. Removed PAGE_OVERVIEW export breaks v3-mcp consumers if they depended on it.

[ED-23679]: https://elementor.atlassian.net/browse/ED-23679?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Omerisra6 (commented):
## Overall

The PR splits context into separate resources which is a good direction, but the implementation has a few recurring patterns that I think we should address before merging:

1. **Every resource independently re-implements the same "serialize → compare → notify" boilerplate.** This should be a shared utility or built into `MCPRegistryEntry`.
2. **`JSON.stringify` for equality comparison** is O(n) on every event, doesn't guarantee key ordering, and in some cases the same data gets serialized twice.
3. **`resolveElementVersion`** is duplicated across files instead of shared.
4. **Resource files mix too many concerns** — data fetching, DOM walking, change detection, debouncing, type definitions — making them hard to reason about and test.

See inline comments for specifics.

packages/packages/core/editor-canvas/src/mcp/resources/editor-state-resource.ts
```diff
@@ -0,0 +1,146 @@
...
+	const notifyIfChanged = () => {
+		const serialized = JSON.stringify( buildState() );
+		if ( serialized === lastSerializedState ) {
+			return;
+		}
+		lastSerializedState = serialized;
+		sendResourceUpdated( { uri: EDITOR_STATE_URI } );
+	};
+
+	const scheduleDebouncedContentRefresh = () => {
```
Omerisra6: This serialize-compare-notify pattern is duplicated across every resource file in this PR (`available-widgets-resource`, `selected-element-resource`, `general-context-resource`, and the existing `document-structure-resource`).

This is a cross-cutting concern that should be extracted into a shared utility, e.g.:
```ts
const notifyIfChanged = createChangeNotifier(sendResourceUpdated, EDITOR_STATE_URI);
```
That way if the notification strategy ever changes (batching, throttling, error handling), there is one place to update.

packages/packages/core/editor-canvas/src/mcp/resources/editor-state-resource.ts
```diff
@@ -0,0 +1,146 @@
...
+			}
+			cachedPageContent = nextContent;
+			notifyIfChanged();
+		}, PAGE_CONTENT_DEBOUNCE_MS );
+	};
+
+	const onEditorStateEvent = () => {
+		notifyIfChanged();
+		scheduleDebouncedContentRefresh();
+	};
```
Omerisra6: The debounce here causes a problematic double-notification:
1. `notifyIfChanged()` fires immediately with **stale** `cachedPageContent` (from the previous document)
2. 1 second later, `scheduleDebouncedContentRefresh()` scrapes the new DOM and fires again

Consumers get wrong data first, then a correction.

Also — the events being listened to (`documents/switch`, `documents/attach-preview`) are **infrequent** (user switching documents), so debouncing doesn't solve a real performance problem here.

If the concern is that the preview DOM isn't ready when `attach-preview` fires, debouncing with a hardcoded 1s delay is the wrong fix — it's just hoping the DOM is ready by then. Better to listen to an event that guarantees readiness.

I'd suggest removing the debounce and just reading the DOM synchronously on each (rare) event.

packages/packages/core/editor-canvas/src/mcp/resources/editor-state-resource.ts
```diff
@@ -0,0 +1,146 @@
...
+		documents?: {
+			getCurrent?: () => {
+				config?: {
+					settings?: {
+						post_title?: string;
+					};
+				};
+			};
+		};
+	};
```
Omerisra6, line 22: `ElementorWindow` is redefined inline here instead of reusing types from `@elementor/editor-documents` — which is exactly what `document-structure-resource.ts` was refactored to do **in this same PR**. This is inconsistent.

packages/packages/core/editor-canvas/src/mcp/resources/editor-state-resource.ts
```diff
@@ -0,0 +1,146 @@
...
+		walk( clone );
+		const text = content.join( ' ' );
+		if ( text.length > PAGE_CONTENT_CHARACTER_LIMIT ) {
+			return text.slice( 0, PAGE_CONTENT_CHARACTER_LIMIT ) + '...';
+		}
+		return text;
+	} catch {
+		return null;
+	}
+}
```
Omerisra6, line 104: `getPageContentFromPreview` is ~30 lines of complex DOM logic (cloning, stripping editor UI, recursive tree walk with context tracking). This is a separate concern from resource registration and should be extracted to a utility — both for readability and testability.

packages/packages/core/editor-canvas/src/mcp/resources/available-widgets-resource.ts
```diff
@@ -0,0 +1,72 @@
...
+	const { resource, sendResourceUpdated } = reg;
+
+	let currentPayload: string | null = null;
+	let currentPayloadV4Only: string | null = null;
+
+	const buildContents = ( filterFunction: ( x: AvailableWidget ) => boolean = () => true ) => {
+		const widgets = getAvailableWidgets().filter( filterFunction );
+		return {
+			contents: [
+				{
```
Omerisra6, line 16: Bug: `buildContents` hardcodes `uri: AVAILABLE_WIDGETS_URI` in the returned contents regardless of the filter passed in. When `notifyIfChanged` sends the v4 update via `...buildContents(filter)`, the response still contains the "all" URI, not `AVAILABLE_WIDGETS_URI_V4`.

packages/packages/core/editor-canvas/src/mcp/resources/available-widgets-resource.ts
```diff
@@ -0,0 +1,72 @@
...
+					text: JSON.stringify( widgets, null, 2 ),
+				},
+			],
+		};
+	};
+
+	const notifyIfChanged = () => {
+		const allWidgets = getAvailableWidgets();
+		const allWidgetsV4 = allWidgets.filter( ( w ) => w.version === 'v4' );
+		const next = JSON.stringify( allWidgets );
```
Omerisra6: Double serialization: `notifyIfChanged` calls `JSON.stringify(allWidgets)` for comparison, then immediately calls `buildContents()` which calls `JSON.stringify(widgets, null, 2)` on the same data again.

packages/packages/core/editor-canvas/src/mcp/resources/selected-element-resource.ts
```diff
@@ -0,0 +1,217 @@
...
+
+function resolveElementVersion( container: SelectionContainer, widgetType: string | null ): WidgetVersion {
+	if ( container.model?.config?.atomic ) {
+		return 'v4';
+	}
+	if ( widgetType && getWidgetsCache()?.[ widgetType ]?.atomic_props_schema ) {
+		return 'v4';
+	}
+	return 'v3';
+}
```
Omerisra6, line 159: `resolveElementVersion` is duplicated here and in `document-structure-resource.ts` (and a third variant exists as `getWidgetVersion` in `element-data-util.ts`).

Please unify into a single shared function — three places that know how to determine v3 vs v4 is a maintenance risk.

packages/packages/core/editor-canvas/src/mcp/resources/general-context-resource.ts
```diff
@@ -0,0 +1,108 @@
...
+						text: JSON.stringify( buildPayload(), null, 2 ),
+					},
+				],
+			};
+		}
+	);
+
+	pushUpdateIfChanged();
+	setInterval( pushUpdateIfChanged, TIME_UPDATE_INTERVAL_MS );
+};
```
Omerisra6, line 99: `setInterval` fires every 60 seconds even when nothing meaningful changed — just the clock ticking. This pushes unnecessary notifications to consumers.

Since `today` changes every minute regardless, the serialize-compare check on line 81 will almost always detect a "change" and trigger a notification. The time should be computed on-demand when the resource is read (in the `resource()` handler), not pushed continuously.

packages/packages/core/editor-canvas/src/mcp/resources/general-context-resource.ts
```diff
@@ -0,0 +1,108 @@
...
+					};
+				};
+			};
+		};
+	};
+};
+
+export const GENERAL_CONTEXT_URI = 'elementor://context/general';
+
+const TIME_UPDATE_INTERVAL_MS = 60_000;
```
Omerisra6: Question: does the LLM actually need `today` here?

- The LLM client (Angie) likely already has the current time injected via its system prompt, making this redundant.
- `today` is the **sole reason** the `setInterval` on line 108 exists. Everything else in this payload (`postId`, `currentPage`, `plugins`, `timezone`) is static for the editor session lifetime.
- It poisons the change detection: because `today` ticks every minute, `JSON.stringify` comparison will fire a notification every 60 seconds even though nothing the consumer cares about (page, post, plugins) actually changed.
- If the LLM needs timezone context, the `timezone` field (which is static) is sufficient.

If `today` is removed, the `setInterval` goes away, the resource becomes static, and the change detection becomes meaningful.

packages/packages/core/editor-canvas/src/mcp/resources/widgets-schema-resource.ts
```diff
@@ -11,6 +11,50 @@ import {
...
+
+const V3_LAYOUT_CONTROL_TYPES = new Set( [ 'section', 'tab', 'tabs' ] );
+
+type V3ControlMetadataEntry = {
+	default?: unknown;
+	type?: string;
+	options?: unknown;
+};
+
+function extractV3ControlsMetadata( controls: unknown ): Record< string, V3ControlMetadataEntry > {
```
Omerisra6, line 24: This is a stripped-down reimplementation of `mapControlsToSchema` that already exists in `elementor-v3-mcp/src/context.ts` (line 119).

Both do the same thing:
- Filter out layout control types (`section`, `tab`, `tabs`)
- Extract `type`, `default`, `options` (as keys)

But `mapControlsToSchema` is more complete — it also handles recursive `fields`, `size_units`, `onValue`/`return_value`, and deep-merges with global control config from `elementor.config.controls`.

This version drops all of that. If the goal is to give the LLM "non-authoritative hints" about v3 controls, why provide less complete hints when a more complete implementation already exists?

Please either reuse `mapControlsToSchema` from `elementor-v3-mcp`, or extract the shared logic into `element-data-util.ts` (which this PR already created as a shared utility location).

packages/packages/core/editor-canvas/src/mcp/resources/editor-state-resource.ts
eavichay-elementor: This implementation was taken as-is from the Angie repository. The purpose of this PR is to split the resources, optimizing can be delegated to tech debts.

packages/packages/core/editor-canvas/src/mcp/resources/editor-state-resource.ts
eavichay-elementor: un-debounced

packages/packages/core/editor-canvas/src/mcp/resources/available-widgets-resource.ts
eavichay-elementor: fixedd

packages/packages/core/editor-canvas/src/mcp/resources/editor-state-resource.ts
eavichay-elementor: This type was intentionally defined locally after attempting to reuse existing types. The `ExtendedWindow` from `@elementor/editor-documents` accesses different parts of the `elementor` object (`documents` manager vs `$previewContents`, different config paths).

Locally extending `Window` is a common pattern across the codebase to handle the 4 coexisting project versions, each exposing different shapes on the global object. Unifying these types is a broader effort outside the scope of this PR.

packages/packages/core/editor-canvas/src/mcp/resources/widgets-schema-resource.ts
eavichay-elementor: Intentionally minimal. This provides basic hints for v3 fallback without over-promising schema accuracy for legacy controls.

The `elementor-v3-mcp` package is scheduled for deletion as v4 coverage expands, so reusing from it would create a dependency on code that's going away.

packages/packages/core/editor-canvas/src/mcp/resources/general-context-resource.ts
eavichay-elementor: Removed `today` field and `setInterval`. Now listens to document/settings events instead.

Omerisra6 (approved):

PR #35536, opened 2026-04-20
Fix: Undo for global class style properties [ED-23496]

Omerisra6:
## Summary

Fixes undo/redo for global class style properties [ED-23496].

**Two issues fixed:**

1. **Undo didn't remove newly-added properties** (`use-styles-fields.ts`, `store.ts`, `types.ts`): The `updateProps` reducer always *merged* incoming props with existing ones. When undoing a property addition, the previous state (which lacked that property) was merged with the current state, leaving the added property intact. Added a `mode: 'replace'` option to `updateProps` that directly replaces variant props instead of merging. The undo callback now uses `mode: 'replace'` with `prevProps`, making the intent explicit.

2. **Redo re-created classes with a new ID** (`use-apply-and-unapply-class.ts`, `global-classes-styles-provider.ts`, `types.ts`): Without a custom `redo` callback, `undoable` defaulted to calling `do` again, which generated a *new* class ID via `generateId()`. Subsequent redo'd prop-update history items still referenced the *original* ID, causing `"Global class not found"` errors. Added a custom `redo` callback that reuses the stored `createdId`, and updated the `create` action to accept an optional `id` parameter.

## Test plan

- [ ] Select a div-block, create a new global class
- [ ] Set properties (font-family, text color, font-size)
- [ ] Press Cmd+Z repeatedly — each step should visibly revert one change
- [ ] Undo past the class creation (class disappears from panel)
- [ ] Press Cmd+Shift+Z repeatedly — each step should re-apply one change
- [ ] Verify redo re-creates the class with correct properties
- [ ] Verify no console errors during the full undo/redo cycle

[ED-23496]: https://elementor.atlassian.net/browse/ED-23496?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## 1. Problem & Context

Global class undo/redo was broken for style property changes. The redo path wasn't implemented, and style updates weren't properly reverting to previous state during undo. ED-23496.

## 2. What Changed (Where)

| File | Change |
|------|--------|
| `use-apply-and-unapply-class.ts` | Added `redo` handler to recreate class and reapply it with preserved ID |
| `use-styles-fields.ts` | Fixed undo to call `updateProps` with `mode: 'replace'` instead of merge |
| `global-classes-styles-provider.ts` | Extended `create` action to accept optional `id` parameter for redo reconstruction |
| `store.ts` | Added `mode` field ('merge' \| 'replace') to conditionally replace or merge variant props |
| `editor-styles-repository/types.ts` | Updated `UpdatePropsActionPayload` and `create` action signature to support mode and custom IDs |
| `css-class-selector.test.tsx` | Fixed test assertions to expect correct class ID (`new-class-id-1` instead of `new-class-id-2`) |
| `global-classes-undo-redo.test.ts` | New e2e test validating undo/redo through multiple property changes |

## 3. How It Works

On class creation, the action now stores `{ prevActiveId, createdId }` undo data. During undo, it removes the class; during redo, it reapplies using the same `createdId` and sets `mode: 'replace'` to atomically restore all style properties without merging stale state. The `create` action now accepts an optional `id` to enable deterministic recreation.

## 4. Risks

Style updates with `mode: 'replace'` now bypass merge logic—if non-class styles exist on the same element variant, they're lost on redo. Mitigated by scoping replace mode to class-driven updates in the editing panel. Test coverage added for multi-property redo path.


packages/packages/core/editor-editing-panel/src/components/css-classes/use-apply-and-unapply-class.ts
```diff
@@ -119,6 +119,16 @@ export function useCreateAndApplyClass() {
 
 					setActiveId( prevActiveId );
 				},
+				redo: (
```
Omerisra6, line 122: how was it "Redoing" a class before that change?

packages/packages/core/editor-editing-panel/src/hooks/use-styles-fields.ts
```diff
@@ -191,6 +195,18 @@ function getCurrentProps( style: StyleDefinition | null, meta: StyleDefinitionVa
...
 }
 
+function buildUndoProps( prevProps: Props, currentProps: Props ): Props {
+	const result: Props = { ...prevProps };
+
+	for ( const key of Object.keys( currentProps ) ) {
+		if ( ! ( key in prevProps ) ) {
+			result[ key ] = null;
+		}
+	}
```
Omerisra6: Why do we need that change? how is it different then just passing the classes?

packages/packages/core/editor-global-classes/src/global-classes-styles-provider.ts
```diff
@@ -41,7 +41,7 @@ export const globalClassesStylesProvider = createStylesProvider( {
 		resolveCssName: ( id: string ) => {
 			return selectClass( getState(), id )?.label ?? id;
 		},
-		create: ( label, variants: StyleDefinitionVariant[] = [] ) => {
+		create: ( label, variants: StyleDefinitionVariant[] = [], existingId ) => {
```
Omerisra6: just name it `id` its not existing

packages/packages/libs/editor-v1-adapters/src/undoable/undoable.ts
```diff
@@ -41,7 +41,16 @@ export function undoable< TPayload extends Payload, TDoReturn, TUndoReturn >(
 ): ( payload?: Payload ) => TDoReturn {
 	actions.redo ??= actions.do;
 
-	const _addHistoryItem = options.debounce ? debounce( addHistoryItem, options.debounce.wait ) : addHistoryItem;
+	let batchFirstDoReturn: Awaited< TDoReturn > | undefined;
+
+	const onHistoryItemAdded = ( item: HistoryItem ) => {
```
Omerisra6: Instead of changing it in the history system level, can we make that fix (debounce/batch) on the classes level? and why in the first place do we need it?

packages/packages/core/editor-editing-panel/src/components/css-classes/use-apply-and-unapply-class.ts
Omerisra6: Before this change, there was no custom `redo` callback. When `undoable.ts` doesn't receive a `redo` action, it falls back to re-running `do`. For class creation, `do` calls `createAction(classLabel)` which generates a **new** ID every time. So on redo, the class was re-created with a different ID (e.g., `g-2` instead of the original `g-1`).

The problem: subsequent V1 history items for property updates still reference the **original** ID. So when those property-update history items try to redo, `updateProps` looks up the original ID in the store, doesn't find it (because the class now has a new ID), and throws `GlobalClassNotFoundError`.

The custom `redo` callback fixes this by passing the original `createdId` back into `createAction`, ensuring the re-created class has the same ID that all downstream history items reference.

packages/packages/core/editor-editing-panel/src/hooks/use-styles-fields.ts
Omerisra6: When undoing, `prevProps` only contains the keys that existed before the change. But if the user *added* a new property (e.g., added `font-family` when none existed before), `prevProps` simply won't have a `font-family` key at all.

If we just pass `prevProps` as-is to `updateProps`, the reducer calls `mergeProps` which merges `prevProps` on top of the current state. Since `font-family` is absent from `prevProps` (not `null`, just missing), `mergeProps` leaves the current `font-family` value untouched — the property silently survives the undo.

`buildUndoProps` fixes this by comparing `prevProps` with `currentProps`: any key that exists in `currentProps` but not in `prevProps` is explicitly set to `null` in the result. This tells `mergeProps` to actually delete it.

In short: passing `prevProps` alone means "restore these values", while `buildUndoProps` means "restore these values AND remove anything that was added since".

packages/packages/core/editor-global-classes/src/global-classes-styles-provider.ts
Omerisra6: Good point, will rename to `id`.

packages/packages/libs/editor-v1-adapters/src/undoable/undoable.ts
Omerisra6: The debounce/batch issue isn't specific to global classes — it's a general problem in `undoable.ts` that affects any debounced undoable action.

Here's what happens without the fix: when you type "775" into a font-size field, the debounce batches 3 rapid calls (7 → 77 → 775) into a single history item. Each call updates `doReturn` (which holds `prevProps`). By the time the history item is created, `doReturn` points to the state from the *last* call (prevProps = {font-size: 77}), not the *first* (prevProps = {}).

So when you undo, it restores to font-size=77 (an intermediate state) instead of font-size=empty (the actual state before you started typing). That's fundamentally wrong for any debounced operation.

Moving this to the classes level would mean every consumer of `undoable` with debounce would need to independently track their own "first doReturn in batch" — duplicating the same logic. Since the problem is in how `undoable` manages debounce batches, fixing it at that level is the right place. It's a one-line addition (`batchFirstDoReturn`) that correctly captures the baseline for all debounced undoable actions.

packages/packages/core/editor-global-classes/src/global-classes-styles-provider.ts
```diff
@@ -50,8 +50,7 @@ export const globalClassesStylesProvider = createStylesProvider( {
 				throw new GlobalClassLabelAlreadyExistsError( { context: { label } } );
 			}
 
-			const existingIds = Object.keys( classes );
-			const id = generateId( 'g-', existingIds );
+			id ??= generateId( 'g-', Object.keys( classes ) );
```
annab1: Replace `??=` please :)

packages/packages/core/editor-global-classes/src/store.ts
```diff
@@ -132,6 +132,7 @@ export const slice = createSlice( {
 				meta: StyleDefinitionVariant[ 'meta' ];
 				props: Props;
 				custom_css?: CustomCss | null;
+				mode?: 'merge' | 'replace';
```
annab1, line 135: Please set `merge` as the default and remove the question mark

packages/packages/core/editor-global-classes/src/global-classes-styles-provider.ts
```diff
@@ -41,7 +41,7 @@ export const globalClassesStylesProvider = createStylesProvider( {
 		resolveCssName: ( id: string ) => {
 			return selectClass( getState(), id )?.label ?? id;
 		},
-		create: ( label, variants: StyleDefinitionVariant[] = [] ) => {
+		create: ( label, variants: StyleDefinitionVariant[] = [], id ) => {
```
annab1: Can we add a type?

annab1 (dismissed):

MiryamOren (approved):

PR #35566, opened 2026-04-23
Internal: Export function `doAfterRender` from editor-canvas [ED-23760]

MiryamOren:
## 1. Problem & Context

Exporting `doAfterRender` utility to support undo/redo cleanup for overridable props on element deletion (ED-23760). Introduces a public API for deferring callbacks until templated element rendering completes.

## 2. What Changed (Where)

- **editor-canvas/index.ts**: Export new `doAfterRender` function
- **legacy/types.ts**: Add `TemplatedElementView` class extending `ElementView` with `_doAfterRender` method
- **legacy/create-templated-element-type.ts**: Update return type from `ElementView` to `TemplatedElementView`
- **utils/after-render.ts**: New utility handling async render callbacks with type guard
- **editor-components/create-component-type.ts**: Replace `ElementView` with `TemplatedElementView` in imports/signatures

## 3. How It Works

`doAfterRender` collects element views, filters those supporting `_doAfterRender`, chains their promises, executes callback once all render cycles complete (or immediately if none pending). Type-safe via `hasDoAfterRender` discriminator.

## 4. Risks

None apparent. Breaking change limited to internal type refinement (`ElementView` → `TemplatedElementView`); only affects declared signatures, not runtime behavior.


packages/packages/core/editor-canvas/src/utils/after-render.ts
```diff
@@ -0,0 +1,26 @@
+import { getContainer } from '@elementor/editor-elements';
```
Omerisra6, line 1: Don't you want to reuse it in the views?

Omerisra6 (approved):

PR #35581, opened 2026-04-26
Fix: Atomic widget drop selects wrapper instead of widget [ED-23790]

Omerisra6:
## Problem & solution

When an atomic v4 widget was dropped on the canvas, the surrounding `e-flexbox` wrapper got selected instead of the widget itself. Root cause: the wrapper's Twig template renders **asynchronously**, but Marionette's `addElement` contract is **synchronous** — it returns a `Container` to the caller immediately, who then walks `.view`, registers history, and selects. With async rendering, the child view doesn't exist yet at that moment, so the chain falls back to the wrapper.

Fix: at the v4 element view level (`createNestedTemplatedElementType`), when the wrapper isn't rendered yet, `addElement` returns a `V1Element`-shaped pending stub whose `lookup()` upgrades to the real `Container` as soon as the registry has it. The legacy `create.js` `restore()` path falls back to a model-level removal when undo happens before the wrapper finishes rendering.

## PR Type
- [x] Bugfix

## Summary

* Fix atomic widget drop selecting the wrapper instead of the widget when the v4 wrapper renders asynchronously [ED-23790]

## Description

* Added `createPendingElement` (`packages/packages/core/editor-canvas/src/legacy/create-pending-element.ts`) — returns a `V1Element`-shaped stub whose `lookup()` returns the real `Container` once the wrapper view registers, and defers selection until the wrapper's `render` event.
* `createNestedTemplatedElementType` overrides `addElement` only when `!isRendered`, so the fix is **isolated to v4 atomic elements** and never affects v3 paths.
* Reuses existing helpers from `@elementor/editor-elements` (`addModelToParent`, `findModelInDocument`, `generateElementId`, `getContainer`) instead of going through `window.$e` directly. Exposed `addModelToParent`, `findModelInDocument`, `removeModelFromParent`, `resolveContainer` in the package's public API.
* `assets/dev/js/editor/document/elements/commands/create.js` — `restore()` flattened (early-return on real Container) and falls back to `removeModelFromParent` when `lookup()` still resolves to a stub (undo-before-render race).

### Why this design

* **Stub-with-lookup** instead of making `addElement` async: blocking the synchronous command bus would ripple across ~50 callers across `elementor` + `elementor-pro`.
* **Model-level removal in `restore()`** instead of synthesizing a fake `Container`: the parent's Backbone collection is the source of truth, and Backbone events make the wrapper render correctly without the child afterwards.
* **Injection point at `createNestedTemplatedElementType`**: localizes the async↔sync bridge to v4 and matches the existing pluggable pattern.

## Test instructions

1. Activate `e_atomic_elements` and `e_twig_containers` experiments.
2. Open a new page in the editor.
3. Drag a v4 atomic widget (e.g. Heading) onto the canvas.
4. Expected: the widget itself is selected (not the wrapping flexbox); the editing panel opens for the widget.
5. Press Ctrl/Cmd+Z — both wrapper and widget are removed.
6. Press Ctrl/Cmd+Shift+Z — both are restored.
7. Press Ctrl/Cmd+Z again — both are removed cleanly (no orphan wrapper).

Automated coverage: `tests/playwright/sanity/modules/v4-tests/atomic-widget-drop-wrapping.test.ts`.

## Quality assurance

Fixes ED-23790

[ED-23790]: https://elementor.atlassian.net/browse/ED-23790?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## 1. Problem & Context

Dropping atomic widgets into the canvas was selecting the auto-generated flexbox wrapper instead of the widget itself, breaking user expectations. ED-23790 fixes selection logic to ensure the widget gets selected, not its container.

## 2. What Changed (Where)

- **create.js**: Refactored undo logic to handle atomic widgets specially—skip deletion for atomic widgets and remove them from parent instead
- **create-nested-templated-element-type.ts**: Override `addElement()` to create pending elements before wrapper renders, ensuring correct selection
- **create-pending-element.ts** (new): Creates elements in model before view renders; defers selection until wrapper renders to guarantee child exists
- **types.ts**: Added `addElement()` and `getContainer()` method signatures; expanded imports
- **editor-elements exports**: Expose `addModelToParent`, `findModelInDocument`, `removeModelFromParent`, `resolveContainer`
- **atomic-widget-drop-wrapping.test.ts** (new): E2E tests verify widget selection and undo/redo correctness

## 3. How It Works

When dropping an atomic widget, the nested template view intercepts `addElement()`. If the wrapper hasn't rendered yet, it creates a "pending" container by synchronously adding the model to the parent, then defers selection via `selectChildWhenWrapperRenders()`. On undo, the create command now detects atomic widgets and calls `removeModelFromParent()` instead of delete, preserving data integrity.

## 4. Risks

Pending elements with mock settings/view objects could cause issues if code assumes real view instances exist—mitigate by strict `isRendered` checks before using view methods. Undo logic bifurcation adds complexity; ensure all widget-add paths flow through this code.


assets/dev/js/editor/document/elements/commands/create.js
```diff
@@ -16,7 +16,20 @@ export class Create extends $e.modules.editor.document.CommandHistoryBase {
 				options,
 			} );
 		} else {
-			$e.run( 'document/elements/delete', { container: data.containerToRestore } );
+			const containerToRestore = data.containerToRestore?.lookup?.() ?? data.containerToRestore;
```
IshayMaya: to be on the safe side, try to only apply these change on v4 element (maybe override from there, or with data hook)

packages/packages/core/editor-canvas/src/legacy/create-nested-templated-element-type.ts
```diff
@@ -101,6 +102,11 @@ export function createNestedTemplatedElementView( {
 	const AtomicElementBaseView = legacyWindow.elementor.modules.elements.views.createAtomicElementBase( type );
 	const parentRenderChildren = AtomicElementBaseView.prototype._renderChildren;
 	const parentOpenEditingPanel = AtomicElementBaseView.prototype._openEditingPanel;
+	const parentAddElement = (
+		AtomicElementBaseView.prototype as unknown as {
+			addElement: ( data: Partial< V1ElementModelProps >, options?: object ) => unknown;
```
IshayMaya: add it to the actual type

packages/packages/core/editor-canvas/src/legacy/create-pending-element.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export function createPendingElement(
+	wrapperView: ElementView,
+	data: Partial< V1ElementModelProps >,
+	options: { edit?: boolean; at?: number } = {}
+): { getContainer: () => V1Element } | undefined {
+	const parentContainer = ( wrapperView as unknown as { getContainer: () => V1Element } ).getContainer();
+	const model: Partial< V1ElementModelProps > = { ...data };
+
+	if ( ! model.id ) {
```
IshayMaya, line 20: please check if this is needed.

packages/packages/core/editor-canvas/src/legacy/create-pending-element.ts
```diff
@@ -0,0 +1,71 @@
...
+		model.elements = [];
+	}
+
+	const added = addModelToParent( parentContainer.id, model as V1ElementModelProps, options );
+
+	if ( ! added ) {
+		return undefined;
+	}
+
+	const childModel = findModelInDocument( model.id );
```
IshayMaya: why do we need to find it?

packages/packages/core/editor-canvas/src/legacy/create-pending-element.ts
```diff
@@ -0,0 +1,71 @@
...
+		lookup() {
+			return getContainer( childId ) ?? pendingContainer;
+		},
+	};
+
+	if ( options.edit !== false ) {
+		selectChildWhenWrapperRenders( wrapperView, childId );
+	}
+
+	return {
```
IshayMaya: need to return what the original function returns

packages/packages/core/editor-canvas/src/legacy/create-nested-templated-element-type.ts
```diff
@@ -381,6 +387,14 @@ export function createNestedTemplatedElementView( {
 			this._doAfterRender( () => parentOpenEditingPanel.call( this, options ) );
 		},
 
+		addElement( data: Partial< V1ElementModelProps >, options?: { edit?: boolean; at?: number } ) {
```
IshayMaya, line 386: check if we can apply the same solution here and on tabs

Omerisra6: It's not a quick fix, we can create a ticket and check how we can fix it from the root (making history model based)

packages/packages/core/editor-canvas/src/legacy/create-pending-element.ts
Omerisra6: Still needed

IshayMaya (approved):

PR #35609, opened 2026-04-28
Internal: Disconnect v4 atomic containers from Marionette rendering [ED-23927]

Omerisra6:
## PR Type
- [x] Refactoring (no functional changes, no api changes)
- [x] Bugfix

## Summary

* Removed JS rendering logic in `createAtomicElementBaseView` that's been superseded by Twig templates and `createNestedTemplatedElementView`.
* Fixed empty top-level div-block / flexbox crashing on drop because `draggable="true"` was being forced by `_attachTwigContent`.

## Description

After v4 atomic containers (div-block, flexbox, etc.) moved to Twig templates rendered via `createNestedTemplatedElementView`, large parts of `createAtomicElementBaseView` are duplicating what Twig and the nested templated view already do.

Removed (all unreachable / overridden once Twig owns the render):

* Tag resolution chain — `tagName`, `_resolveTag`, the resolved-tag cache, `rerenderEntireView`, `_shouldRecreateForTagChange`, `_applyResolvedAttributes`, `_resolvePropValue`. Twig sets the tag from `settings.tag` / `settings.link.tag`.
* Class generation chain — `className`, `getClasses`, `getClassString`, `getBaseClass`, `getGeneratedClasses`. The `render_base_classes` macro emits `elementor-element`, `elementor-element-{id}`, `e-con`, `e-atomic-element` and base styles. The `generates_class` meta path (only used by `atomic-form.php` for `form-state-{value}`) is now applied directly inside `atomic-form.html.twig`.
* `attributes()` and `renderOnChange()` — both fully overridden in `createNestedTemplatedElementView` (`attributes()` returns only `data-model-cid`; `renderOnChange()` calls `this.render()`). Everything they used to do — `data-id`, `data-element_type`, `data-e-type`, `data-interaction-id`, `_cssid`, custom attributes, link attributes — is produced by `_macros.html.twig` and copied onto `$el` by `_attachTwigContent`.
* Link/dynamic helpers — `_hasLink`, `getLinkAttributes`, `_applyLinkAttributes`, `getDynamicLinkValue`, `handleDynamicLink`. Replaced by `render_link_attributes` macro server- and editor-side.
* Test suites for the removed code.

Bugfix:

`_attachTwigContent` was unconditionally calling `targetEl.setAttribute( 'draggable', 'true' )` after copying attributes from the Twig output. For top-level e-con elements, `initDraggable` in `views/base.js:1185` returns early (long-standing `hasClass( '.e-con' )` selector typo, dot included), so `html5Draggable` never wires up. The element was native-draggable but had no `dragstart` handler — dragging the empty body fired a native drag with no `element:dragged` set, and `base-container.js:onDrop` then crashed on `Object.entries( undefined )`. Removing the line restores the previous behavior: top-level empty containers move only via the handle. Nested elements are unaffected because their `initDraggable` runs (via `parents( '.e-con' )`) and `html5Draggable` sets `draggable="true"` itself on init (`jquery-html5-dnd.js:39`).

## Test instructions

1. Open the editor, add a top-level Div block.
2. With it empty, try dragging it from its body — should not be draggable (was crashing on drop before this fix).
3. Drag the same div-block by its handle — moves correctly.
4. Add a child div-block; drag the nested empty div-block by its body — moves correctly with the icon helper as the drag image.
5. Set Tag (Div / Header / Section / a / button) — element re-renders with the correct tag.
6. Add a Link → URL/Action → tag swaps to `a` / `button` and `href` / `data-action-link` are set.
7. Set CSS ID, custom Attributes, custom classes — all reflected on the element.
8. Add a Form, toggle States control (Normal / Success / Error) — the form's class flips between `form-state-default` / `form-state-success` / `form-state-error` and the appropriate Form Message child becomes visible.
9. Re-save and reload — frontend renders identically.

## Quality assurance

Fixes ED-23927

## 1. Problem & Context

Removes the experimental Twig-based rendering for atomic containers (div-block, flexbox). The experiment is being discontinued, making Twig template support a permanent feature of the base classes rather than a conditional experiment.

## 2. What Changed (Where)

| Component | Change |
|-----------|--------|
| `div-block.php`, `flexbox.php` | Added `Has_Element_Template` trait + `get_templates()` method directly to base classes |
| `module.php` | Removed `EXPERIMENT_TWIG_CONTAINERS` constant and conditional registration logic; now unconditionally registers `Div_Block` and `Flexbox` |
| Test files | Deleted `Div_Block_Twig`/`Flexbox_Twig` classes; renamed tests from `*_twig_*` to `*_*` |
| Editor JS | Removed tag resolution, link attribute handling, and class generation from Marionette view |
| Playwright tests | Removed `e_twig_containers` experiment activation |

## 3. How It Works

Base container classes now permanently include template rendering via `Has_Element_Template` trait. No experiment flag checks—registration uses only `Div_Block()` and `Flexbox()`. Removed large swaths of editor-side logic that dynamically resolved tags/links; presumably moved to template-side or permanently handled differently post-Marionette disconnect.

## 4. Risks

**Test coverage loss**: Deleted 1100+ lines of unit tests comparing Twig vs non-Twig output. Verify snapshot tests still validate rendering parity. **Editor regression**: Removed tag/link resolution logic without confirming replacement mechanism exists (ties into "disconnect from Marionette" but no migration code visible here). **Incomplete template migration**: JS deletions suggest template now owns rendering, but no new template logic added—verify `div-block.html.twig` and `flexbox.html.twig` handle all former JS responsibilities.


modules/atomic-widgets/assets/js/editor/create-atomic-element-base-view.js
```diff
@@ -869,83 +549,6 @@ export default function createAtomicElementBaseView( type ) {
...
-
-			if ( ! transformer ) {
-				return prop.value?.origin_value;
-			}
-
-			const transformed = transformer( prop.value, { key: 'overridable', renderContext } );
-
-			return this._resolvePropValue( transformed, renderContext );
-		},
-
```
IshayMaya, line 948: something's broke in the structure

IshayMaya (dismissed):

IshayMaya (approved):

PR #35620, opened 2026-04-29
Internal: [V4] form actions email 2 [ED-23842]

ronenboxer:
## 1. Problem & Context

Refactoring form email actions to support multiple email recipients per action (ED-23842). Migrates from single `email` prop type to `emails` with array-based recipient fields (`to`, `cc`, `bcc`), enabling richer email workflows while maintaining backward compatibility via migration ops.

## 2. What Changed (Where)

| Component | Change |
|-----------|--------|
| **PHP Backend** | `Emails_Prop_Type` created; `Email_Form_Action_Control` extends `Chips_Control`; `Atomic_Form` refactored to support multiple email actions via `get_emails_prop_settings()` and helper methods |
| **TypeScript Controls** | `EmailFormActionControl` extracted into modular field components; `ChipsControl` enhanced with `freeChips` prop for custom email entry |
| **Prop Types** | `emailsPropTypeUtil` added; registry updated to use new prop type |
| **Migrations** | `email-to-emails.json` migration wraps `to/cc/bcc` scalars into string arrays |
| **Tests** | Email control tests refactored for array-based `to` field; migration roundtrip tests added |

## 3. How It Works

Backend generates multiple email action keys (`email`, `email_2`, etc.) via `get_email_action_count()` filter. Each action registers as chips option + dedicated `Email_Form_Action_Control` with `freeChips=true`. Frontend's new `EmailChipsField` component validates emails on blur/Enter/comma and maintains string-array structure. Migration handles schema conversion: up wraps scalars in arrays, down extracts first value (lossy but safe for old clients).

## 4. Risks

**Data loss on downgrade**: Migration's down path extracts only first email from arrays—additional recipients silently dropped if reverting. **Filter dependency**: Email action count hardcoded to 1 unless explicitly filtered; Pro version might need coordination. **Test coverage**: New email validation logic (`isValidEmail`) untested at component level; edge cases (special chars, internationalized domains) should be verified.

[ED-23842]: https://elementor.atlassian.net/browse/ED-23842?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/atomic-widgets/elements/atomic-form/atomic-form.php
```diff
@@ -125,91 +117,89 @@ protected static function define_props_schema(): array {
...
-			'email' => Email_Prop_Type::make()
-				->set_dependencies( $email_dependencies )
-				->meta( Overridable_Prop_Type::ignore() )
-				->default( [
-					'to' => String_Prop_Type::generate( self::get_default_recipient_email() ),
-					'from' => String_Prop_Type::generate( self::get_default_sender_email() ),
-				] ),
+		];
+
+		$props = array_merge( $props, self::get_emails_prop_settings() );
```
mykytamurzin: you can add here as many arrays as you want

mykytamurzin: I mean it's not necessary to do few array_merge's

packages/packages/libs/editor-controls/src/controls/email-form-action-control.tsx
```diff
@@ -28,8 +58,111 @@ const EmailField = ( { bind, label, placeholder }: { bind: string; label: string
 	</PropKeyProvider>
 );
 
+const EmailChipsField = ( { fieldLabel, placeholder }: { fieldLabel: string; placeholder?: string } ) => {
```
mykytamurzin: it's already 3rd component in one file, let's split them a bit

modules/atomic-widgets/elements/atomic-form/atomic-form.php
marcin-el: done

packages/packages/libs/editor-controls/src/controls/email-form-action-control.tsx
marcin-el: done

migrations/operations/email-to-emails.json
```diff
@@ -0,0 +1,62 @@
+{
+  "up": [
+    { "op": { "fn": "set", "path": "$$type", "value": "emails" } },
+
+    { "op": { "fn": "move", "src": "value.to", "dest": "__temp_to" } },
```
mykytamurzin, line 5: should we add exists condition here?

asafdl: uhhhhhh, need to test this

ronenboxer: `tests/phpunit/elementor/modules/atomic-widgets/prop-type-migrations/test-email-to-emails-migration.php`
like this? @asafdl

modules/atomic-widgets/controls/types/chips-control.php
```diff
@@ -23,6 +24,19 @@ public function set_options( array $options ): self {
 	public function get_props(): array {
 		return [
 			'options' => $this->options,
+			'freeChips' => $this->free_chips,
 		];
 	}
+
+	public function allow_free_chips(): self {
```
asafdl: allow and block params?

modules/atomic-widgets/elements/atomic-form/atomic-form.php
```diff
@@ -118,98 +110,96 @@ protected static function define_props_schema(): array {
 				->default( 'default' )
 				->meta( 'generates_class', 'form-state-{value}' ),
 			'actions-after-submit' => String_Array_Prop_Type::make()
+				->initial_value( [ String_Prop_Type::generate( self::ACTION_EMAIL ) ] )
```
asafdl, line 112: maybe string_array_prop_type::generate?

modules/atomic-widgets/elements/atomic-form/atomic-form.php
```diff
@@ -420,4 +410,70 @@ public static function is_instance_form( $instance ): bool {
 	public function render_markdown(): string {
 		return '';
 	}
+
+	private static function get_emails_prop_settings() {
```
asafdl: there has got to be a better way to write this

modules/atomic-widgets/elements/atomic-form/atomic-form.php
```diff
@@ -420,4 +410,70 @@ public static function is_instance_form( $instance ): bool {
...
+		return array_merge( ...array_map( [ self::class, 'get_indexed_email_prop_settings' ], range( 0, self::EMAIL_ACTION_COUNT - 1 ) ) );
+	}
+
+	private static function get_indexed_email_prop_settings( $index = 0 ) {
```
asafdl: why do we keep refering to index? email one and email two?
perhaps we can abstract it to primary and secondary or something?

ronenboxer:
Relevant pro tests passing
https://github.com/elementor/elementor-pro/actions/runs/25667278936/job/75345312452#step:6:21

modules/atomic-widgets/controls/types/chips-control.php
ronenboxer: Allow method is for true, while block is for false, would you rather i create one method with a parameter?

modules/atomic-widgets/elements/atomic-form/atomic-form.php
ronenboxer: No can do
generate wraps it in {type, value}. the resolver crashes because it expects only the value

modules/atomic-widgets/controls/types/chips-control.php
asafdl: yes

modules/atomic-widgets/elements/atomic-form/atomic-form.php
ronenboxer: Changed

modules/atomic-widgets/elements/atomic-form/atomic-form.php
ronenboxer: Because write now it's 2 emails, tomorrow there might be more

ronenboxer: I also wanted to prevent code duplication

packages/packages/libs/editor-controls/src/controls/chips-control.tsx
```diff
@@ -15,42 +15,56 @@ export type ChipsOption = {
...
 	const selectedValues: string[] = ( value || [] )
 		.map( ( item ) => stringPropTypeUtil.extract( item ) )
 		.filter( ( val ): val is string => val !== null );
 
-	const selectedOptions = selectedValues
-		.map( ( val ) => options.find( ( opt ) => opt.value === val ) )
-		.filter( ( opt ): opt is ChipsOption => opt !== undefined );
+	const selectedOptions = selectedValues.map( ( val ) => toChipsOption( val, options ) );
+
+	const handleChange = ( _: SyntheticEvent, newValue: ( ChipsOption | string )[] ) => {
```
asafdl, line 35: why do we need to keep deduping every change? can't we block adding values that already exist?

modules/atomic-widgets/elements/atomic-form/atomic-form.php
asafdl: the code is pretty complex written this way, but it's already written so your choice to keep it

packages/packages/libs/editor-controls/src/controls/chips-control.tsx
ronenboxer: You're right

modules/atomic-widgets/elements/atomic-form/atomic-form.php
ronenboxer: It's the original code, placed in a loop

ronenboxer:
Tests pass
https://github.com/elementor/elementor-pro/actions/runs/25733742243/job/75568116327

modules/atomic-widgets/controls/types/chips-control.php
ronenboxer: Done

asafdl (approved):

PR #35628, opened 2026-04-30
Internal: Add support for date picker settings [ED-23655]

RobiFerentz:
## 1. Problem & Context

Adding date range picker support to the atomic widgets system (ED-23655). Enables form builders to accept min/max date constraints as a single control input.

## 2. What Changed (Where)

| Layer | Changes |
|-------|---------|
| **Backend (PHP)** | New `Date_Range_Control`, `Date_Range_Prop_Type`, `Date_String_Prop_Type` classes; transformer to handle "Invalid Date" strings |
| **Frontend (TS/TSX)** | New `DateRangeControl`, `DateStringControl` components; transformer registration; prop type utilities |
| **Integration** | Module registration across control registry, settings transformers, prop-type exports |

## 3. How It Works

Control accepts `{min, max}` object with ISO date strings (YYYY-MM-DD). Frontend renders two date pickers side-by-side with client-side validation: max ≥ min triggers error state. Backend transformer normalizes "Invalid Date" placeholders to empty strings. Both layers validate via `date_create_from_format()` / dayjs, ensuring consistency.

## 4. Risks

**Date format brittleness**: Hardcoded YYYY-MM-DD format assumes no localization. If date localization needed later, transformer logic and format constant require updates. Mitigate: Document format assumption; consider i18n for `DATE_FORMAT` constant if regional support planned.


modules/atomic-widgets/prop-types/date-range-prop-type.php
```diff
@@ -0,0 +1,50 @@
...
+		];
+	}
+
+	protected function validate_value( $value ): bool {
+		if ( ! parent::validate_value( $value ) ) {
+			return false;
+		}
+
+		$min = $this->extract_iso_date( $value['min'] ?? null );
+		$max = $this->extract_iso_date( $value['max'] ?? null );
```
marcin-el, line 30: `validate_value` only parses `Y-m-d` so for the time variant extract_iso_date returns null and the min/max check is skipped?

RobiFerentz: yes, there is no point validating min max for time, because 13:00 can be both before and after 14:00 if it's passing day boundries.

marcin-el (dismissed):

modules/atomic-widgets/controls/types/min-max-date-time-control.php
```diff
@@ -0,0 +1,29 @@
...
+class Min_Max_Date_Time_Control extends Atomic_Control_Base {
+	private string $variant = 'date';
+
+	public function get_type(): string {
+		return 'min-max-date-time';
+	}
+
+	/** @var 'date'|'time' $variant Specifies the control type: either 'date' or 'time'. */
+	public function set_variant( string $variant ): self {
+		$this->variant = $variant;
```
asafdl: add check for date / time

RobiFerentz: Done.

packages/packages/libs/editor-controls/src/controls/date-string-control.tsx
```diff
@@ -0,0 +1,103 @@
...
+	variant?: DateStringControlVariant;
+	inputDisabled?: boolean;
+	ariaLabel?: string;
+	error?: boolean;
+};
+
+const DATE_FORMAT = 'YYYY-MM-DD';
+const TIME_FORMAT = 'HH:mm';
+
+export const DateStringControl = createControl(
```
asafdl: i think it should be two controls, with two prop types.
date control and time control
two prop-types, date-prop-type and time-prop-type

packages/packages/libs/editor-controls/src/controls/min-max-date-time-control.tsx
```diff
@@ -0,0 +1,108 @@
...
+	if ( ! minIso || ! maxIso || [ minIso, maxIso ].some( ( v ) => v === 'Invalid Date' ) ) {
+		return false;
+	}
+
+	return maxIso < minIso;
+};
+
+const RANGE_ERROR_MESSAGE = __( 'Max date must be on or after Min date', 'elementor' );
+
+export const MinMaxDateTimeControl = createControl( ( { variant = 'date' }: MinMaxDateTimeControlProps ) => {
```
asafdl: i think we should abstract this, rangeParentControl (childControl, validateCB, ErrorMsg)

asafdl (approved):

PR #35629, opened 2026-04-30
Tweak: Create file upload form atom [ED-23657]

mykytamurzin:
## 1. Problem & Context

Adds file upload support to atomic form widgets (ED-23657). Conditionally enables for Elementor Pro ≥4.1.0 to maintain backward compatibility while introducing attachment handling infrastructure.

## 2. What Changed (Where)

- **PHP controls**: New `Attachment_Type_Control` class for attachment metadata management
- **Atomic form**: Refactored `define_default_children()` to conditionally inject file upload field; extracted form field submission into `appendFormField()` helper
- **Frontend handlers**: Enhanced `getAtomicFormFieldValue()` and form submission to handle `FileList` objects with array expansion
- **TypeScript registry**: Added `'e-form-file-upload'` and `'e-form-date-picker'` to form field types; registered `AttachmentTypeControl` in controls registry
- **React control**: New `AttachmentTypeControl` component wrapping `SelectControl` with inline help text about upload behavior

## 3. How It Works

File upload field conditionally renders when Pro ≥4.1.0. On form submission, files are extracted via `getAtomicFormFieldValue()` which returns `FileList` directly (not stringified). The refactored `appendFieldValue()` handles polymorphic input—expands arrays/FileLists into indexed FormData keys (`form_fields[0][value][0]`, etc.) while preserving scalars. This matches server-side array unpacking for multi-file handling.

## 4. Risks

FileList handling assumes server correctly processes indexed FormData arrays. If backend expects flat structure, submissions break. Conditional Pro version check could mask missing 4.1.0 artifacts. No validation on file type/size at submission layer—rely on server-side enforcement.


RobiFerentz (commented):
I will just share the transcript of cursor's review 🙉 
https://cursor.com/dashboard/shared-chats?shareId=branch-comparison-review-Y-Gab6-gVVQj

RobiFerentz (dismissed):

modules/atomic-widgets/assets/js/frontend/form-handlers.js
```diff
@@ -201,6 +201,10 @@ function getAtomicFormFieldLabel( field, form ) {
 }
 
 function getAtomicFormFieldValue( inputs, type ) {
+	if ( 'file' === type ) {
+		return inputs.files.length > 0 ? inputs.files : '';
```
asafdl: who uses this value, inputs.files is what? string of filenames?
if it is, then empty list, will be emptry string anyway

modules/atomic-widgets/assets/js/frontend/form-handlers.js
```diff
@@ -243,7 +247,11 @@ async function submitAtomicForm( payload ) {
 		formData.append( `form_fields[${ index }][name]`, field.name );
 		formData.append( `form_fields[${ index }][options]`, JSON.stringify( field.options ) );
 
-		if ( Array.isArray( field.value ) ) {
+		if ( field.value instanceof FileList ) {
```
asafdl: the original funciton is kind of messy, can we clean up here a bit formData = handleFileValue(formData)

modules/atomic-widgets/elements/atomic-form/atomic-form.php
```diff
@@ -282,28 +282,36 @@ protected function define_default_children() {
 
 			$this->build_label( __( 'Message', 'elementor' ), $prefix . 'message' ),
 			$this->build_input( __( 'Your message', 'elementor' ), 'textarea', $prefix . 'message' ),
+		];
 
-			$this->build_checkbox_row( __( 'Checkbox', 'elementor' ), $prefix . 'checkbox' ),
+		// File-upload lives in elementor-pro; only seed it when Pro 4.1.0+ is installed.
```
asafdl: this comment is uselss

packages/packages/libs/editor-controls/src/controls/attachment-type-control.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { ControlFormLabel } from '../components/control-form-label';
+import { createControl } from '../create-control';
+import { SelectControl, type SelectOption } from './select-control';
+
+type Props = {
+	label?: string;
+	options: SelectOption[];
+};
+
+export const AttachmentTypeControl = createControl( ( { label, options }: Props ) => {
```
asafdl, line 15: this seems like a simple abstraction around select control, wdyt, can we move this logic into a variant of select control or would it make no sense?

mykytamurzin: yes, there is only InfoAlert is added. 

I would like to keep Select primitive as simple as possible, extending it with InfoAlert only for now looks too  coupled, so it should be something more generic.

asafdl (dismissed):

RobiFerentz (approved):

PR #35671, opened 2026-05-04
Internal: Add support for time picker form widget and time range prop type [ED-23656]

RobiFerentz:
## 1. Problem & Context

Adding time picker widget and time-range prop type support to atomic widgets system, mirroring existing date-range infrastructure.

## 2. What Changed (Where)

| Category | Changes |
|----------|---------|
| **Prop Types** | New `Time_String_Prop_Type` (ISO time validation) and `Time_Range_Prop_Type` (min/max object) in PHP |
| **Controls** | New `TimeStringControl` and `TimeRangeControl` in TypeScript; added to form element registry |
| **Transformers** | New `Time_Range_Transformer` (PHP/TS); refactored `Date_Range_Transformer` to use null coercion instead of 'Invalid Date' strings |
| **Form Elements** | Added `e-form-time-picker` to form field types |
| **Utils** | Extracted date/time parsing logic to shared `date-time.ts` utilities; added `parseTimeString()` for HH:mm format |

## 3. How It Works

Time inputs flow through `TimeStringControl` → `parseTimeString()` (parse HH:mm to Dayjs) → `TimePicker` UI → format back to HH:mm on change. `TimeRangeControl` wraps two `TimeStringControl` instances (min/max) within shared prop context. Backend transformers normalize null/empty values consistently. Unlike dates, time range doesn't validate ordering (accepts end < start).

## 4. Risks

**Coercion logic change**: `Date_Range_Transformer` now coerces empty values to `null` instead of empty strings—verify consumers handle null correctly. **Format assumption**: Time format hardcoded to HH:mm; localization/seconds handling not exposed. **No validation**: Time range accepts invalid orderings silently (by design per test).

### DIFF
--- a/modules/atomic-widgets/controls/types/time-range-control.php
+++ b/modules/atomic-widgets/controls/types/time-range-control.php
@@ -8,7 +8,6 @@ if ( ! defined( 'ABSPATH' ) ) {
 }
 
 class Time_Range_Control extends Atomic_Control_Base {
-
 	public function get_type(): string {
 		return 'time-range';
 	}
--- a/modules/atomic-widgets/controls/types/time-range-control.php
+++ b/modules/atomic-widgets/controls/types/time-range-control.php
@@ -1,0 +1,19 @@
+<?php
+namespace Elementor\Modules\AtomicWidgets\Controls\Types;
+
+use Elementor\Modules\AtomicWidgets\Controls\Base\Atomic_Control_Base;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+class Time_Range_Control extends Atomic_Control_Base {
+
+	public function get_type(): string {
+		return 'time-range';
+	}
+
+	public function get_props(): array {
+		return [];
+	}
+}
--- a/modules/atomic-widgets/props-resolver/transformers/settings/time-range-transformer.php
+++ b/modules/atomic-widgets/props-resolver/transformers/settings/time-range-transformer.php
@@ -10,23 +10,14 @@ if ( ! defined( 'ABSPATH' ) ) {
 }
 
 class Time_Range_Transformer extends Transformer_Base {
-	const INVALID_DATE = 'Invalid Date';
-
 	public function transform( $value, Props_Resolver_Context $context ) {
 		if ( empty( $value ) ) {
 			return null;
 		}
 
-		$result = [];
-
-		if ( isset( $value['min'] ) ) {
-			$result['min'] = self::INVALID_DATE !== $value['min'] ? $value['min'] : '';
-		}
-
-		if ( isset( $value['max'] ) ) {
-			$result['max'] = self::INVALID_DATE !== $value['max'] ? $value['max'] : '';
-		}
-
-		return $result;
+		return [
+			'min' => empty( $value['min'] ) ? null : $value['min'],
+			'max' => empty( $value['max'] ) ? null : $value['max'],
+		];
 	}
 }
--- a/modules/atomic-widgets/props-resolver/transformers/settings/time-range-transformer.php
+++ b/modules/atomic-widgets/props-resolver/transformers/settings/time-range-transformer.php
@@ -1,0 +1,33 @@
+<?php
+
+namespace Elementor\Modules\AtomicWidgets\PropsResolver\Transformers\Settings;
+
+use Elementor\Modules\AtomicWidgets\PropsResolver\Props_Resolver_Context;
+use Elementor\Modules\AtomicWidgets\PropsResolver\Transformer_Base;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+class Time_Range_Transformer extends Transformer_Base {
+	const INVALID_DATE = 'Invalid Date';
+
+	public function transform( $value, Props_Resolver_Context $context ) {
+
+		if ( empty( $value ) ) {
+			return null;
+		}
+
+		$result = [];
+
+		if ( isset( $value['min'] ) ) {
+			$result['min'] = self::INVALID_DATE !== $value['min'] ? $value['min'] : '';
+		}
+
+		if ( isset( $value['max'] ) ) {
+			$result['max'] = self::INVALID_DATE !== $value['max'] ? $value['max'] : '';
+		}
+
+		return $result;
+	}
+}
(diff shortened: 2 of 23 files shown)


packages/packages/libs/editor-controls/src/utils/date-time.ts
```diff
@@ -0,0 +1,42 @@
...
+	if ( Number.isNaN( h ) || Number.isNaN( m ) ) {
+		return null;
+	}
+
+	const base = ( dayjs as unknown as { default: () => Dayjs } ).default();
+	return base
+		.hour( h )
+		.minute( m )
+		.second( Number.isNaN( s ) ? 0 : s )
+		.millisecond( 0 );
```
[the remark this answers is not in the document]
RobiFerentz: The TimePicker from the UI library already handles out of bound numbers for the time values. No need to do double checks on the client. Only the output it returns interests me.

packages/packages/libs/editor-props/src/prop-types/time-string.ts
```diff
@@ -0,0 +1,7 @@
+import { z } from '@elementor/schema';
+
+import { createPropUtils } from '../utils/create-prop-utils';
+
+export const timeStringPropTypeUtil = createPropUtils( 'time-string', z.string() );
```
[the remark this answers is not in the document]
RobiFerentz: The time picker input will take of this.

modules/atomic-widgets/controls/types/time-range-control.php
```diff
@@ -0,0 +1,19 @@
...
+namespace Elementor\Modules\AtomicWidgets\Controls\Types;
+
+use Elementor\Modules\AtomicWidgets\Controls\Base\Atomic_Control_Base;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+class Time_Range_Control extends Atomic_Control_Base {
+
```
davseve: Remove empty line

modules/atomic-widgets/props-resolver/transformers/settings/time-range-transformer.php
```diff
@@ -0,0 +1,33 @@
...
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+class Time_Range_Transformer extends Transformer_Base {
+	const INVALID_DATE = 'Invalid Date';
+
+	public function transform( $value, Props_Resolver_Context $context ) {
+
```
davseve: Remove redundant empty line

packages/packages/libs/editor-controls/src/hooks/use-form-field-suggestions.ts
```diff
@@ -13,6 +13,8 @@ const FORM_FIELD_WIDGET_TYPES = [
 	'e-form-checkbox',
 	'e-form-radio-button',
 	'e-form-select',
+	'e-form-date-picker',
```
RobiFerentz, line 16: This was added because I forgot to do this for the date picker PR

davseve (dismissed):

modules/atomic-widgets/props-resolver/transformers/settings/time-range-transformer.php
```diff
@@ -0,0 +1,32 @@
...
+use Elementor\Modules\AtomicWidgets\PropsResolver\Props_Resolver_Context;
+use Elementor\Modules\AtomicWidgets\PropsResolver\Transformer_Base;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+class Time_Range_Transformer extends Transformer_Base {
+	const INVALID_DATE = 'Invalid Date';
```
mykytamurzin: as I understand this magic string is produced by dayjs when it can't parse the value. 

Can we use just null instead of connecting to magic strings of 3rd party lib?

packages/packages/libs/editor-controls/src/utils/date-time.ts
```diff
@@ -0,0 +1,42 @@
+import type { Dayjs } from 'dayjs';
+import * as dayjs from 'dayjs';
+
+export const DATE_FORMAT = 'YYYY-MM-DD';
+export const TIME_FORMAT = 'HH:mm';
```
mykytamurzin, line 5: are we supporting seconds?

Time_String_Prop_Type::ISO_TIME_REGEX - supports seconds, but this format is not

packages/packages/libs/editor-controls/src/utils/date-time.ts
```diff
@@ -0,0 +1,42 @@
...
+export const INVALID_DATE = 'Invalid Date';
```
mykytamurzin: Again magic string

packages/packages/libs/editor-controls/src/utils/date-time.ts
```diff
@@ -0,0 +1,42 @@
...
+	if ( ! raw ) {
+		return null;
+	}
+
+	const parsed = ( dayjs as unknown as { default: ( s?: string | number | Date ) => Dayjs } ).default( raw );
+
+	return isValidDayjs( parsed ) ? parsed : null;
+}
+
+export function parseTimeString( raw: string ): Dayjs | null {
```
mykytamurzin, line 21: Equivalent to dayjs(raw, 'HH:mm:ss', true)

packages/packages/libs/editor-controls/src/utils/date-time.ts
RobiFerentz: Yeah, I wanted to support seconds on the backend (other wanted to go further and do timezones etc, but we don't have the time to put everything in now), but for now the controls and widget will not support them.

packages/packages/libs/editor-controls/src/utils/date-time.ts
RobiFerentz: This is the string returned by dayjs. I need to check against it so I know what to do (in this case we I could just parse an invalid date and get the the string programmatically, but feels silly.

modules/atomic-widgets/props-resolver/transformers/settings/time-range-transformer.php
RobiFerentz: I removed it from the php transformer, as it's not needed, we empty the values if the they're invalid (they're not usable by the native time input if they're not valid anyway and are silently ignored)

packages/packages/libs/editor-controls/src/utils/date-time.ts
RobiFerentz: It's not really, daysjs doesn't parse time only string without some custom parser.

mykytamurzin (approved):

packages/packages/libs/editor-controls/src/utils/date-time.ts
RobiFerentz: Removed

PR #35677, opened 2026-05-04
Internal: Import design system: server side [ED-23914]

MiryamOren:
## 1. Problem & Context

Refactoring import system to support design-system as a top-level import type alongside settings, moving from hardcoded "settings" inclusion checks to flexible type routing. Extracts complex import logic into reusable utilities for maintainability.

## 2. What Changed (Where)

- **import.php route**: Added `include` array parameter and renamed `settings` to `customization` for API contracts
- **import.php process**: Extracted data-building logic into `build_runner_data()` method; `get_runners_name()` now filters by `should_import()`
- **Global_Classes_Repository**: Added `get_order()`, `update_order_and_labels()`, `delete_all()` for structured state management
- **global-classes import runner**: Simplified from 100+ LOC of merge logic to delegating `Import_Utils::import_classes()` with conflict resolution strategy
- **global-classes/import-export-utils/import-utils.php** (new): 336 LOC handling validation, conflict resolution (skip/replace/rename/override-all), and transaction-wrapped imports
- **variables import runner**: Refactored `merge_collections()` to `resolve_collections()` with result tracking; added `build_label_type_map()` for conflict detection

## 3. How It Works

Import flow: route accepts `include` type list → runners query `should_import()` with shared `$data` → runner resolves conflict strategy from `customization[type]` → utilities handle per-item validation, label/ID conflict resolution, and ACID-wrapped persistence. Design-system and settings now follow identical import patterns.

## 4. Risks

Transaction rollback assumes `$wpdb->query()` works consistently across DB drivers—consider wrapping in try-catch if supporting non-MySQL. ID collision retry loop in `generate_unique_id()` (infinite loop risk if namespace exhausted, though 2^28 space mitigates). No validation that imported IDs/labels don't conflict with in-flight WordPress operations; race conditions possible if concurrent imports run.


app/modules/import-export-customization/data/routes/import.php
```diff
@@ -69,9 +69,15 @@ protected function get_args(): array {
 				'description' => 'Session ID for import operations',
 				'required' => true,
 			],
-			'settings' => [
```
Omerisra6, line 72: Couldnt it be a BC problem?

modules/global-classes/import-export-customization/runners/import.php
```diff
@@ -37,125 +35,64 @@ private function is_classes_enabled( array $data ): bool {
...
-		$override_all = ! empty( $data['customization']['settings']['classesOverrideAll'] );
-
-		if ( $override_all ) {
-			$repository->put(
-				$imported_classes['items'],
-				$imported_classes['order']
-			);
+	private function get_include_type( array $data ): ?string {
+		$is_settings_import = in_array( 'settings', $data['include'] ?? [], true );
+		$is_design_system_import = in_array( 'design-system', $data['include'] ?? [], true );
```
Omerisra6, line 63: move to a const and reuse

modules/global-classes/import-export-customization/runners/import.php
```diff
@@ -37,125 +35,64 @@ private function is_classes_enabled( array $data ): bool {
...
 
-			return $imported_classes;
+		$include = null;
+		if ( $is_settings_import ) {
+			$include = 'settings';
+		} elseif ( $is_design_system_import ) {
+			$include = 'design-system';
 		}
```
Omerisra6, line 70: why?

modules/global-classes/import-export-customization/runners/import.php
```diff
@@ -37,125 +35,64 @@ private function is_classes_enabled( array $data ): bool {
...
 	public function import( array $data, array $imported_data ): array {
-		$kit = Plugin::$instance->kits_manager->get_active_kit();
-
-		$file_name = Import_Export_Customization::FILE_NAME;
-		$global_classes = ImportExportUtils::read_json_file( "{$data['extracted_directory_path']}/{$file_name}.json" );
-
-		if ( ! $kit || ! $global_classes ) {
-			return [];
-		}
+		$global_classes_dir = $data['extracted_directory_path'] . '/global-classes';
```
Omerisra6, line 38: maybe we should extract that to a method as well (get classes dir)?

modules/global-classes/import-export-customization/runners/import.php
```diff
@@ -37,125 +35,64 @@ private function is_classes_enabled( array $data ): bool {
...
+			case 'settings':
+				return ! empty( $data['customization']['settings']['classesOverrideAll'] )
+					? 'override-all'
+					: 'merge';
+			case 'design-system':
+				return $data['customization']['design-system']['conflict_resolution'] ?? 'skip';
+			default:
+				throw new \Exception( 'Classes should not be imported.' );
 		}
+	}
```
Omerisra6, line 59: Maybe we could create a Class for the include type or even for the import, and then we can handle retrieval there:

`Include_Type::make(data)->is_settings() ....`

that would also make us type safe

modules/global-classes/import-export-customization/runners/import.php
```diff
@@ -37,125 +35,64 @@ private function is_classes_enabled( array $data ): bool {
...
-			return $existing;
+			return $active_kit;
 		}
 
 		$previous_kit = Plugin::$instance->kits_manager->get_kit( $previous_kit_id );
 
-		return Global_Classes_Repository::make( $previous_kit )->all()->get();
-	}
-
-	private function merge_classes( array $existing, array $imported ): array {
```
Omerisra6, line 109: why dont we need it now?

modules/global-classes/import-export-utils/import-utils.php
```diff
@@ -0,0 +1,336 @@
...
+				continue;
+			}
+
+			if ( 'rename' === $action ) {
+				$existing_labels = array_keys( $label_to_id_map );
+
+				$new_label = ImportExportUtils::resolve_label_conflict( $import_entry['label'], $existing_labels );
+				$sanitized_item['label'] = $new_label;
+
+				$label_to_id_map[ strtolower( $new_label ) ] = $import_entry['id'];
```
Omerisra6, line 170: Bug: `$label_to_id_map[ strtolower( $new_label ) ] = $import_entry['id']` is set before the id-collision check below, which can replace `$sanitized_item['id']` via `generate_unique_id()`. The map then points at a stale id, and any later import entries that resolve via this map will be wrong.

Move the assignment after `$sanitized_item['id'] = $new_id` and use the final id:
```php
$label_to_id_map[ strtolower( $sanitized_item['label'] ) ] = $new_id;
```

modules/global-classes/import-export-utils/import-utils.php
```diff
@@ -0,0 +1,336 @@
...
+				'label' => $sanitized_item['label'],
+			];
+
+			if ( 'rename' === $action ) {
+				$result['renamed'][] = [
+					'import_entry' => $import_entry,
+					'result_entry' => $result_entry,
+				];
+			} else {
+				$result['created'][] = [ 'import_entry' => $import_entry ];
```
Omerisra6, line 196: `created` doesn't include `result_entry`, unlike `replaced` and `renamed`. Since the id can be regenerated by `generate_unique_id()` above, callers have no way to know the final id.

Suggest mirroring the shape:
```php
$result['created'][] = [
    'import_entry' => $import_entry,
    'result_entry' => $result_entry,
];
```

modules/global-classes/import-export-utils/import-utils.php
```diff
@@ -0,0 +1,336 @@
...
+			}
+		}
+
+		$has_changes = ! empty( $added_classes_order ) || ! empty( $modified_classes ) || ! empty( $deleted_classes );
+
+		if ( $has_changes ) {
+			$new_order = array_merge( $added_classes_order, $previous_order );
+			$repository->update_order_and_labels( $new_order, $added_classes_labels );
+
+			do_action( 'elementor/global_classes/update', Global_Classes_Repository::CONTEXT_FRONTEND, [
```
Omerisra6, line 206: This action fires while we're still inside the `START TRANSACTION` opened in `import_classes`. A few side-effects to consider:
- Listeners that write join the transaction and get rolled back together (might be intentional, just not obvious).
- Listeners that read may see uncommitted state depending on isolation level.
- External side-effects (cache invalidation, webhooks, etc.) fire even if we later roll back.

Consider returning the change-set from `do_import` and dispatching the action from `import_classes` after `COMMIT`. Same applies to `clean_post_cache()` calls inside `create_new_class` / `replace_existing_class`.

modules/variables/import-export-customization/runners/import.php
```diff
@@ -44,27 +53,75 @@ public function import( array $data, array $imported_data ): array {
...
+					'id' => $variable->id(),
+					'label' => $variable->label(),
+				],
+			];
+		}
+
+		return $result;
+	}
+
+	private function resolve_conflict_resolution( array $data ): string {
```
Omerisra6, line 97: is it the same logic as in the global classes runner?

modules/variables/import-export-customization/runners/import.php
```diff
@@ -44,27 +53,75 @@ public function import( array $data, array $imported_data ): array {
...
+		switch ( $include_type ) {
+			case 'settings':
+				return ! empty( $data['customization']['settings']['variablesOverrideAll'] )
+					? 'override-all'
+					: 'merge';
+			case 'design-system':
+				return $data['customization']['design-system']['conflict_resolution'] ?? 'skip';
+			default:
+				throw new \Exception( 'Variables should not be imported.' );
+		}
```
Omerisra6, line 109: same as above

modules/variables/import-export-customization/runners/import.php
```diff
@@ -92,18 +149,53 @@ private function get_existing_collection( Variables_Repository $repository, arra
...
 		$existing_ids = array_keys( $existing->all() );
 
+		$result = self::EMPTY_RESULT;
+
 		foreach ( $imported->all() as $variable ) {
 			if ( $variable->is_deleted() ) {
 				continue;
 			}
 
+			$import_entry = [
```
Omerisra6, line 168: can we extract into a method to simplify it?

modules/global-classes/import-export-customization/runners/import.php
MiryamOren: I moved this logic to the import-utils. We just read the classes differently now, but we still support the `merge` option

modules/variables/import-export-customization/runners/import.php
MiryamOren: Yes

modules/global-classes/import-export-utils/import-utils.php
MiryamOren: Unfortunately, we can't return the change set and apply it at once, as we can't hold that size of data without PHP running out of memory.
* If there are write listeners, their changes should be rolled back after the data rollback
* Yor'e right about the `elementor/global_classes/update` hook, I'll move it after the commit

modules/global-classes/import-export-utils/import-utils.php
MiryamOren: Right, I'l fix

modules/global-classes/import-export-utils/import-utils.php
MiryamOren: Good catch! fixed

app/modules/import-export-customization/data/routes/import.php
MiryamOren: No, I'm not changing the route's API (except for adding a new `include` option)
The args were outdated - you can see that in the callback, we're not trying to get the `settings` parameter, but do try to get `include` and `customization` parameters
image

modules/global-classes/import-export-customization/runners/import.php
MiryamOren: Moved back into a const

modules/variables/import-export-customization/runners/import.php
MiryamOren: Yes

PR #35691, opened 2026-05-05
Fix: Angie access to form input id [ED-22527]

asafdl:
## 1. Problem & Context

Form input fields were inaccessible because the schema validation wasn't properly filtering non-configurable properties. The fix refactors `isPropKeyConfigurable` to accept both prop key and type, enabling context-aware filtering of internal properties like `_cssid`, `classes`, and `attributes`.

## 2. What Changed (Where)

- **widgets-schema-resource.ts**: Moved property filtering into `Object.entries().filter()` chain using enhanced `isPropKeyConfigurable` check; removed post-processing deletion loop.
- **validate-input.ts**: Updated `isPropKeyConfigurable` call to pass `propSchema` parameter for type-aware validation.
- **props-to-llm-schema.ts**: Rewrote `isPropKeyConfigurable(key, propType?)` to allow LLM-configurable overrides via `propType.meta.llm_configurable`; removed unused `propsSchemaToJsonSchema` function.

## 3. How It Works

Non-configurable keys are now checked in two stages: (1) if key is in blocklist, check if propType explicitly allows it via `llm_configurable` meta flag; (2) otherwise, property is configurable. This lets special properties opt-in to exposure when needed, rather than blanket exclusion.

## 4. Risks

Minor: The signature change to `isPropKeyConfigurable` adds an optional parameter. Callers passing only the key still work but won't benefit from type-aware filtering—confirm all call sites get updated. No schema version bump needed since this is internal API.


packages/packages/libs/editor-props/src/utils/props-to-llm-schema.ts
```diff
@@ -186,38 +186,17 @@ function convertPropTypeToJsonSchema( propType: PropType ): JsonSchema7 {
...
-
 export const nonConfigurablePropKeys = [ '_cssid', 'classes', 'attributes' ] as readonly string[];
 
-export function isPropKeyConfigurable( propKey: string ): boolean {
-	return ! nonConfigurablePropKeys.includes( propKey );
+export function isPropKeyConfigurable( propKey: string, propType?: PropType ): boolean {
+	if ( ! nonConfigurablePropKeys.includes( propKey ) ) {
+		return true;
+	}
+	return !! ( propType?.meta && ! Array.isArray( propType.meta ) && propType.meta.llm_configurable );
```
mykytamurzin:
```suggestion
	return !! ( ! Array.isArray( propType?.meta ) && propType?.meta?.llm_configurable );
```

mykytamurzin (dismissed):

Ntnelbaba (approved):

PR #35714, opened 2026-05-06
Fix: Deletion of applied classes throws an error [ED-23966] 

ronros-elementor:


packages/packages/core/editor-design-system/src/design-system-panel.tsx
```diff
@@ -15,6 +16,7 @@ export const { panel, usePanelStatus, usePanelActions } = createPanel( {
 	},
 	onClose: async () => {
 		changeEditMode( 'edit' );
+		await reloadCurrentDocument();
```
IshayMaya, line 19: Can you open an optimization ticket to run it only if there was a class deletion operation? this is something we already had in the global classes panel and i'm not sure when but seems like it was removed from there as well.
Here i think it's even more relevant since we have more tabs to the panel and this operation is very specific and only relevant to class deletion.

IshayMaya (approved):

PR #35734, opened 2026-05-07
Internal: Add query-filter repeater control + prop type [ED-23990]

IshayMaya:
## 1. Problem & Context

Adds infrastructure for filtering query results through a repeatable UI control. Enables dynamic query filtering with type-safe prop handling and configurable key-value pairs. Supports ED-23990 implementation by providing reusable components for query filter management across atomic widgets.

## 2. What Changed (Where)

**Backend (PHP)**
- `modules/atomic-widgets/controls/types/query-filter-repeater-control.php` — New control class with fluent setters for allowed keys, key config, label, and chips placeholder
- `modules/atomic-widgets/prop-types/query-filter-prop-type.php` — Object prop type defining `key` (string) and `values` (query array) structure
- `modules/atomic-widgets/prop-types/query-filter-array-prop-type.php` — Array wrapper for query filter prop type

**Frontend (TypeScript)**
- `packages/libs/editor-controls/src/controls/query-filter-repeater-control.tsx` — 243-line React control with key selection, query chips integration, and per-key value caching
- `packages/libs/editor-props/src/prop-types/query-filter.ts` — Zod schemas and TypeScript types for query filter props
- `packages/core/editor-editing-panel/src/controls-registry/controls-registry.tsx` — Registry entries mapping control type to component and prop type util
- Export updates in both libs' index files

## 3. How It Works

Control enforces unique keys across filter items using `usedKeys` set tracking. Add button disables when `allowedKeys` are exhausted. Key selection dropdown marks already-used keys as disabled except for the current item's key.

Value persistence happens via `valuesByKeyRef` — when switching keys, current values cache by old key, and switching back restores them. Prevents data loss during key exploration.

`QueryFilterKeyConfig` drives per-key behavior: `queryEndpoint` presence controls whether values field renders. Query chips control passes through endpoint URL and params for async data fetching.

Repeater isn't sortable (`isSortable={false}`) — filters likely need stable ordering or order doesn't matter for query semantics.

## 4. Risks

**Type safety gap**: `keyConfig` record lookup returns `QueryFilterKeyConfig | undefined`, but `currentKeyConfig?.queryEndpoint` usage assumes it exists when key is valid. If `allowedKeys` includes keys not in `keyConfig`, control breaks. Consider validation or fallback defaults.

**Memory leak potential**: `valuesByKeyRef` persists indefinitely in component lifetime. If keys change frequently or control rerenders often, unbounded cache growth possible. Probably negligible in practice but worth monitoring.

[ED-23990]: https://elementor.atlassian.net/browse/ED-23990?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/atomic-widgets/controls/types/query-filter-repeater-control.php
```diff
@@ -0,0 +1,60 @@
...
+namespace Elementor\Modules\AtomicWidgets\Controls\Types;
+
+use Elementor\Modules\AtomicWidgets\Controls\Base\Atomic_Control_Base;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+class Query_Filter_Repeater_Control extends Atomic_Control_Base {
+	private array $allowed_types = [];
```
Omerisra6: rethink the naming for this

modules/atomic-widgets/controls/types/query-filter-repeater-control.php
```diff
@@ -0,0 +1,60 @@
...
+		return $this;
+	}
+
+	public function get_props(): array {
+		return [
+			'allowedTypes'  => $this->allowed_types,
+			'typeOptions'   => (object) $this->type_options,
+			'repeaterLabel' => $this->repeater_label,
+			'placeholder'   => $this->placeholder,
+			'addLabel'      => $this->add_label,
```
Omerisra6: bad naming!

modules/atomic-widgets/prop-types/query-filter-array-prop-type.php
```diff
@@ -0,0 +1,19 @@
...
+namespace Elementor\Modules\AtomicWidgets\PropTypes;
+
+use Elementor\Modules\AtomicWidgets\PropTypes\Base\Array_Prop_Type;
+use Elementor\Modules\AtomicWidgets\PropTypes\Contracts\Prop_Type;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+class Query_Filter_Array_Prop_Type extends Array_Prop_Type {
```
Omerisra6, line 11: Maybe we can find a more abstract name for that (if we want to use the same prop type in other places)

packages/packages/libs/editor-controls/src/controls/query-filter-repeater-control.tsx
```diff
@@ -0,0 +1,253 @@
...
+import { PopoverContent } from '../components/popover-content';
+import { PopoverGridContainer } from '../components/popover-grid-container';
+import { RepeaterHeader } from '../components/repeater/repeater-header';
+import { createControl } from '../create-control';
+import { QueryChipsControl } from './query-chips-control';
+import { SelectControl } from './select-control';
+
+export type QueryFilterTypeOption = {
+	label: string;
+	queryEndpoint?: {
```
Omerisra6: where does it comes from?

modules/atomic-widgets/prop-types/query-filter-array-prop-type.php
IshayMaya: Considered renaming to something more generic, but the class is literally "array of Query_Filter_Prop_Type" — its registry key (query-filter-array) and its TS counterpart (queryFilterArrayPropTypeUtil) are tied.
We can revisit if a second consumer of the shape shows up, for now the paired Object/Array class pattern matches what we do everywhere else (Query_Array_Prop_Type, Classes_Prop_Type).

packages/packages/libs/editor-controls/src/controls/query-filter-repeater-control.tsx
IshayMaya: from the schema, based on the source selected.

modules/atomic-widgets/prop-types/query-filter-prop-type.php
```diff
@@ -0,0 +1,27 @@
...
+	protected function define_shape(): array {
+		return [
+			'key'    => String_Prop_Type::make()->required(),
+			'values' => Query_Array_Prop_Type::make()->default( [] ),
+		];
+	}
+
+	public function get_default() {
+		return null;
+	}
```
Omerisra6: why?

packages/packages/libs/editor-controls/src/controls/query-filter-repeater-control.tsx
```diff
@@ -0,0 +1,243 @@
...
+
+	const keySelectOptions = useMemo(
+		() =>
+			allowedKeys.map( ( itemKey ) => ( {
+				value: itemKey,
+				label: keyConfig[ itemKey ]?.label ?? itemKey,
+				disabled: usedKeys.has( itemKey ),
+			} ) ),
+		[ allowedKeys, keyConfig, usedKeys ]
+	);
```
Omerisra6: check if it we can move this logic to the parent

modules/atomic-widgets/prop-types/query-filter-prop-type.php
IshayMaya: removed

Omerisra6 (approved):

PR #35740, opened 2026-05-07
Internal: Optimize querying global classes' corresponding posts [ED-23991]

ronros-elementor:
## 1. Problem & Context

Global class lookups were O(n) post queries; now cached via a class_id → post_id map stored on the kit. Eliminates redundant meta queries for every style resolution and handles migration/deduplication transparently.

## 2. What Changed (Where)

| File | Change |
|------|--------|
| `global-classes-post-ids.php` | New class: map-based lookup with backfill, dedup, cache invalidation hooks |
| `global-class-post.php` | Replaced `find_by_class_id()` query with `Global_Classes_Post_IDs` lookup; removed `menu_order` param |
| `global-classes-repository.php` | Use post_ids map instead of iterating all posts; inline preview label/meta cleanup |
| `atomic-global-styles.php` | Extract label resolution into `is_preview()` helper; simplify `transform_classes_names()` |
| `module.php` | Register post_ids hooks; add meta key to kit import preserve list |
| Tests + JS | Parallel API fetch; test coverage for preview/frontend transitions |

## 3. How It Works

On class creation, `Global_Classes_Post_IDs::set()` stores `{class_id → post_id}` in kit meta. Lookups hit cache first; stale entries trigger `backfill()` which queries by meta_key, deduplicates (keeps oldest), and rebuilds map. Deletion hooks keep maps clean across kits. Preview label overrides now stored separately in `Global_Classes_Labels`, not post titles—prevents frontend CSS selector drift during drafts.

## 4. Risks

**Stale map entries after external post deletion**: Mitigated by `deleted_post` hook sweeping all kit maps. **Concurrent creates of same class_id**: Dedup logic in backfill handles this. **Migration lag**: First lookup on dirty install triggers backfill—transparent but adds latency once.


modules/global-classes/global-class-post.php
```diff
@@ -79,10 +81,6 @@ public function get_label(): string {
 		return $this->post->post_title;
 	}
 
-	public function get_order(): int {
-		return (int) $this->post->menu_order;
-	}
-
```
ronros-elementor, line 87: we're not really using it

modules/global-classes/global-classes-post-ids.php
```diff
@@ -0,0 +1,221 @@
...
+			array_merge(
+				[ Global_Class_Post::META_KEY_ID ],
+				$class_ids,
+				[ Global_Class_Post_Type::CPT, 'publish' ]
+			)
+		);
+		// phpcs:enable
+
+		// phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared -- $sql is already prepared via $wpdb->prepare() above.
+		$rows = $wpdb->get_results( $sql );
```
ronros-elementor, line 210: using the wpdb->prepare should prevent such vulnerability

eavichay-elementor (changes requested):
Don't forget to add the new meta key to sanitize in main wp-api hook.

eavichay-elementor (approved):

PR #35754, opened 2026-05-07
Internal: Skip migrations on global classes export and bump request timeout [ED-23093]

Omerisra6:
## PR Type
- [x] Bugfix

## Summary

* Fix server hang / 502 on settings export by bypassing prop-type migrations during export and raising the design-system import/export request timeout from 20s to 60s.

## Description

Exporting a kit with many global classes (~1000+) caused the server to hang and the request to fail with a 502, even after the previous fix that memoized `Style_Parser` (#35730). Runtime tracing showed that `Global_Class_Post::to_array()` invokes `Migrations_Orchestrator::migrate()` for every unmigrated class, costing roughly 500ms per class. With 1000 classes that adds up to ~8 minutes of pure migration work — well beyond any sensible request budget.

Migrations are an import/read concern, not an export concern. The exporter should write whatever is currently stored, not lazily migrate it. This PR adds a parallel "raw" read path used only by the export runners:

- `Global_Class_Post::to_array_raw()` reads stored data without triggering migrations (factored a private `build_array()` helper to avoid duplicating the array shape).
- `Global_Classes_Repository::each_item_raw()` mirrors `each_item()` but yields raw (un-migrated) class data.
- The customization global-classes export runner switches to `each_item_raw()` and drops its now-redundant per-item parser/sanitize step.
- The legacy global-classes export runner drops its post-fetch `Global_Classes_Parser::parse()` step (no behavioral need on export — the data was already stored through the parsed write path).

After this change the global-classes runner completes a 1001-class export in ~452ms (verified via runtime instrumentation, instrumentation removed before commit).

Additionally, raised `IMPORT_REQUEST_TIMEOUT_MS` and `EXPORT_REQUEST_TIMEOUT_MS` in the design-system hooks from 20s → 60s to give the server more room before the client aborts.

## Test instructions

1. Have a kit with a large number of global classes (~1000+).
2. Trigger a settings-only export (`include: ['settings']`, `customization.settings.classes: true`).
3. Expect the export to complete in < 1s instead of hanging or 502'ing.
4. Trigger the corresponding import flow on a fresh site and confirm classes are imported correctly (migrations still run on import / on read in the editor).

## Quality assurance

Fixes ED-23093

## 1. Problem & Context

Global classes export was running migrations on every class during serialization, causing timeouts on large datasets. Migrations are unnecessary during export since we're just persisting existing data. The fix skips migrations entirely during export and extends request timeout from 20s to 2 minutes as a safety net for large exports (ticket: ED-23093).

## 2. What Changed (Where)

- `global-class-post.php`: Added `$should_migrate` parameter to `get_data()` and `to_array()`, defaulting to `true` for backward compatibility
- `global-classes-repository.php`: Threaded `$should_migrate` through `each_item()` to control migration behavior at the repository level
- `import-export-customization/runners/export.php`: Passes `false` for migration flag, removed `sanitize_item()` method and parser dependency
- `global-classes-rest-api.php`: Reduced `MAX_ITEMS` constant from 5000 to 1000 (likely separate optimization)
- `use-export-request.ts` & `use-import-request.ts`: Bumped request timeout from 20s to 120s
- `design-system-header-menu.tsx`: Added `CircularProgress` spinner during import/export instead of disabling button

## 3. How It Works

The migration flag flows through the call chain: `Export::export()` → `Repository::each_item($cb, false)` → `Global_Class_Post::to_array(false)` → `Global_Class_Post::get_data(false)`. When `false`, the migration check at line 95-97 is skipped entirely. The export runner no longer validates/sanitizes items through the parser—it trusts the stored data is already valid since it passed validation on import/create. Frontend now shows a progress spinner instead of a disabled state during long-running operations.

## 4. Risks

**Data integrity**: Export now skips parser validation that previously caught malformed data (test case removed shows `<script>` tags were being sanitized). If corrupted data exists in the DB, it will be exported as-is. Mitigation relies on import-time validation catching issues, but consider re-adding a lightweight validation pass if corruption reports surface.

**MAX_ITEMS reduction** (5000→1000): This is an unrelated change that could break existing workflows expecting larger batch sizes. If this is intentional rate-limiting, document why; otherwise, revert or extract to separate PR.

[ED-23093]: https://elementor.atlassian.net/browse/ED-23093?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/packages/core/editor-design-system/src/import/hooks/use-import-request.ts
```diff
@@ -7,7 +7,7 @@ import { type ConflictStrategy } from '../types';
 
 const IMPORT_BASE_PATH = 'elementor/v1/import-export-customization';
 
-const IMPORT_REQUEST_TIMEOUT_MS = 20_000;
+const IMPORT_REQUEST_TIMEOUT_MS = 120_000;
```
MiryamOren, line 10:
```suggestion
// 2 minutes
const IMPORT_REQUEST_TIMEOUT_MS = 120_000;
```

modules/global-classes/global-class-post.php
```diff
@@ -134,6 +142,16 @@ public function to_array(): array {
 		return $result;
 	}
 
+	public function get_raw_data(): array {
```
MiryamOren: let's add an arg of `should_migrate` / `run_migration` with default `true`, instead of the new functions

PR #35768, opened 2026-05-10
Internal: Change span controls to use text instead of numbers [ED-23856]

RobiFerentz:


modules/atomic-widgets/props-resolver/transformers/styles/span-transformer.php
```diff
@@ -11,6 +11,10 @@
 
 class Span_Transformer extends Transformer_Base {
 	public function transform( $value, Props_Resolver_Context $context ) {
-		return $value || 0 === $value ? 'span ' . $value : null;
+		if ( ! is_string( $value ) ) {
```
asafdl: this is probably useless

RobiFerentz: Removed

davseve (approved):

PR #35789, opened 2026-05-10
Fix: Import kit: Fix global classes legacy import format [ED-24027]

MiryamOren:
## 1. Problem & Context

Legacy global classes import was overwriting kit metadata (sync maps, labels, order) on kit creation during import. This broke the format expected by the new design system sync. Ticket: ED-24027. Fix preserves critical metadata across imports by filtering meta keys before kit creation.

## 2. What Changed (Where)

- `core/kits/manager.php`: Added `get_meta_to_preserve()` to extract metadata from current kit before creating new one
- `modules/global-classes/module.php`: Hooked into `elementor/kit/meta_to_preserve_on_kit_import` to register 4 meta keys (order, frontend labels, preview labels, sync map)
- `modules/variables/module.php`: Same pattern for variables metadata preservation (single key)
- `modules/global-classes/import-export-utils/legacy-import-utils.php`: Replaced `delete_all()` + `put()` with atomic `apply_changes()` call to maintain metadata integrity

## 3. How It Works

Modules register meta keys via `elementor/kit/meta_to_preserve_on_kit_import` filter. When `create_new_kit()` runs, it pulls metadata from the active kit using `get_meta_to_preserve()`, then passes that to `create()` alongside settings. The legacy import logic now uses `apply_changes()` with explicit added/deleted/order operations instead of destructive `delete_all()` + `put()`, ensuring metadata survives the override-all strategy. Early return added when `items` is empty to skip unnecessary writes.

## 4. Risks

None — this is defensive code. The filter hook returns empty array by default, so metadata preservation is opt-in per module. The `apply_changes()` refactor maintains existing import behavior while respecting preserved metadata.

[ED-24027]: https://elementor.atlassian.net/browse/ED-24027?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/global-classes/module.php
```diff
@@ -74,9 +75,23 @@ public function __construct() {
...
+				[ $this, 'add_meta_to_preserve_on_kit_import' ]
+			);
 		}
 	}
 
+	public function add_meta_to_preserve_on_kit_import( array $meta_keys ): array {
+		return array_merge( $meta_keys, [
+			Global_Classes_Order::META_KEY,
+			Global_Classes_Labels::META_KEY_FRONTEND,
+			Global_Classes_Labels::META_KEY_PREVIEW,
```
ronros-elementor, line 90: just wondering - do we want the preview as well?

ronros-elementor (dismissed):

modules/global-classes/module.php
MiryamOren: Yes, if we preserve the existing classes, there's no reason to delete the draft state

max-zu (dismissed):

ronros-elementor (approved):

PR #35798, opened 2026-05-11
Fix: Build compositions tool now deletes failed attempt [ED-24035]

eavichay-elementor:
## 1. Problem & Context

The composition builder was leaving orphaned elements in the DOM when `createElement` failed during composition build. This caused state inconsistency and potential memory leaks, especially problematic in the MCP tool context where failed attempts would accumulate. The fix adds cleanup to delete partially-created containers on error, with component support temporarily disabled until ED-24035 is resolved.

## 2. What Changed (Where)

- **composition-builder.ts**: Added `deleteElement` to API, wrapped `createElement` in try-catch to cleanup failed attempts, added type safety with `V1ElementModelProps`, temporarily disabled component support
- **tool.ts**: Passed `deleteElement` to CompositionBuilder initialization
- **element-data-util.ts**: Filtered out Component widgets from LLM support as temporary workaround

## 3. How It Works

Entry point is `build()` where `createElement` now runs in a try-catch per child node. On failure, it attempts to retrieve the partially-created container by `modelTree.id` and explicitly deletes it before re-throwing. The `buildModelTree` method now returns properly-typed `V1ElementModelProps` with explicit `elType: 'widget'` to ensure type consistency. Component-related code is commented out with TODO markers rather than removed, preserving context for future restoration.

## 4. Risks

**Type casting on line 296**: `modelTree as CreateElementParams['model']` suppresses type mismatch between `V1ElementModelProps` and expected createElement parameter. If the types diverge further, this cast could hide runtime errors. Mitigation: The explicit `elType` assignment should align types, but verify createElement actually accepts this shape.

**Cleanup race condition**: `getContainer(modelTree.id)` might return null if the failed createElement never registered the container, making the delete a no-op. This is acceptable (error still propagates), but doesn't clean up other potential side effects from partial initialization.


packages/packages/core/editor-canvas/src/composition-builder/composition-builder.ts
```diff
@@ -282,15 +290,21 @@ export class CompositionBuilder {
...
+				} );
+				this.rootContainers.push( newElement );
+				await this.awaitViewRender( newElement );
+			} catch ( e: unknown ) {
+				const attempToRestoreInvalidContainer = this.api.getContainer( modelTree.id as string );
+				if ( attempToRestoreInvalidContainer ) {
+					this.api.deleteElement( { container: attempToRestoreInvalidContainer } );
+				}
+				throw e;
+			}
```
Omerisra6, line 307: What is the flow you are trying to catch?

Does it always throws?

packages/packages/core/editor-canvas/src/composition-builder/composition-builder.ts
```diff
@@ -94,15 +98,26 @@ export class CompositionBuilder {
...
 			skipDefaultChildren: true,
-			elements: children,
+			elements: children as V1ElementModelProps[ 'elements' ],
 			editor_settings: {
 				title: node.getAttribute( 'configuration-id' ) ?? undefined,
 			},
+			elType: 'widget',
 		};
 
+		// TODO: Restore this code once components are working in compositions
```
Omerisra6, line 111: why does components dont work?

Omerisra6 (approved):

PR #35826, opened 2026-05-12
Fix: Global classes search uses contains matching and limit no longer blocks autocomplete [ED-23973]

Omerisra6:
## PR Checklist

## PR Type
- [x] Bugfix

## Summary

* Fixed: typing in the CSS class selector while at the class limit now shows existing classes in the autocomplete dropdown instead of replacing the dropdown with an error message.
* Fixed: clicking "Class Manager" in the limit-reached empty state now correctly opens the panel when the Design System experiment is off.
* Improved: global class search in both the Design System panel and the CSS class selector autocomplete now normalizes the query (strips non-alphanumeric characters like leading dots) before matching, so searching \`.btn\` or \`btn\` both find \`btn-primary\`.

## Description

Three issues addressed from [ED-23973](https://elementor.atlassian.net/browse/ED-23973):

**Bug fix — autocomplete blocked at class limit**

When a user reached the class limit and typed in the CSS class selector input, \`validate\` was called on every \`inputChange\` event. Since \`hasReachedLimit\` returned \`true\`, an error was immediately set and the \`ListboxComponent\` was replaced entirely with the error text — hiding all existing classes. The fix: when the limit is reached, \`useCreateAction\` returns no \`onCreate\`, so the "Create" option simply does not appear and the dropdown shows existing classes normally. Format validation errors still surface normally.

**Bug fix — "Class Manager" button did nothing when Design System experiment is off**

When the \`e_editor_design_system_panel\` experiment is off, dispatching \`elementor/open-global-classes-manager\` had no listener — that event is only consumed by \`DesignSystemEntrypoints\` which mounts only when the experiment is on. Added \`OpenPanelFromEvent\` in \`editor-global-classes\`: a component that listens for the event and calls \`usePanelActions().open()\` (the panels API, with all its side effects). Registered only when the experiment is off, mirroring the conditional pattern used by \`ClassManagerButton\`.

**Search improvement — contains matching with normalization**

- \`use-filter-options.ts\`: passes explicit \`matchFrom: 'any'\` to \`createFilterOptions\` and normalizes the search input (strips \`[^a-zA-Z0-9_-]\`) before the filter runs, so typing \`.btn-\` finds \`btn-primary\`.
- \`global-classes-list.tsx\`: applies the same normalization to the debounced search value in \`useFilteredCssClasses\`.

## Test instructions

**Bug fix — autocomplete at limit:**
1. Create 1000 global classes (or temporarily lower \`MAX_CLASSES\` in \`global-classes-styles-provider.ts\`).
2. Select an element and open the Style tab.
3. Type any character in the CSS class selector input.
4. **Expected:** the dropdown shows existing matching classes and a "Limit reached" empty state. **Before fix:** only an error message was shown.

**Bug fix — Class Manager button (experiment off):**
1. Disable the \`e_editor_design_system_panel\` experiment.
2. Reach the class limit so the "Limit reached" empty state appears.
3. Click the "Class Manager" button in the empty state.
4. **Expected:** the Class Manager panel opens. **Before fix:** nothing happened.

**Search improvement:**
1. Create global classes named \`btn-primary\`, \`text-center\`.
2. In the Design System → Classes search, type \`.btn\` or \`btn\`.
3. **Expected:** \`btn-primary\` is shown. Works the same in the CSS class selector autocomplete.

## Quality assurance

Fixes ED-23973

[ED-23973]: https://elementor.atlassian.net/browse/ED-23973?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## 1. Problem & Context

Two issues with global classes autocomplete: search was exact-match only (blocking partial matches), and hitting the class limit prevented creation attempts entirely instead of gracefully handling the constraint. Ticket: ED-23973.

## 2. What Changed (Where)

| File | Change |
|------|--------|
| `use-filter-options.ts` | Normalize search input by stripping non-alphanumeric chars, enable `matchFrom: 'any'` for contains matching |
| `css-class-selector.tsx` | Extract limit check into early return; add `LimitReachedEmptyState` component with Class Manager link; return `isAtLimit` flag |
| `global-classes-list.tsx` | Apply same normalization to global classes manager search |
| `global-classes-styles-provider.ts` | Reduce `MAX_CLASSES` from 5000 to 1000 |
| `init.ts` | Register `OpenPanelFromEvent` listener (non-design-system path) |
| `open-panel-from-event.tsx` | New component listening for custom event to open global classes panel |
| Test files | Update test labels and assertions for normalized search behavior |

## 3. How It Works

Search normalization strips special chars (`/[^a-zA-Z0-9_-]/g`) before filtering, enabling "my-class" to match "my--class". At creation limit, `useCreateAction()` returns early with `isAtLimit: true`, signaling UI to render `LimitReachedEmptyState` instead of suppressing validation errors. Users see "Class Manager" button to reduce class count. Cross-module event dispatch (`EVENT_OPEN_GLOBAL_CLASSES_MANAGER`) lets CSS selector panel trigger global classes manager when user navigates from limit state.

## 4. Risks

Normalization could over-match unintended classes if separator rules prove too loose. Early limit return removes error messaging from validation—UX clarity depends entirely on `LimitReachedEmptyState` rendering. Event-based communication between modules is implicit; breaking the listener registration in `init.ts` silently fails.


packages/packages/core/editor-editing-panel/src/components/creatable-autocomplete/use-filter-options.ts
```diff
@@ -10,7 +16,10 @@ export function useFilterOptions< TOption extends Option >( parameters: {
 } ) {
 	const { options, selected, onCreate, entityName } = parameters;
 
-	const filter = createFilterOptions< InternalOption< TOption > >();
+	const filter = createFilterOptions< InternalOption< TOption > >( {
+		matchFrom: 'any',
+		stringify: ( option ) => normalizeClassSearch( option.label ),
```
ronros-elementor: is this required? if all our classes anyway go through validation, they shouldn't have any non class characters in their name to begin with, no?

if anything - I would leverage this function to strip out all non alphanumeric characters (dashes and underscores) from all options, so that only alphanumeric characters would have any weight on the search

so searching font-size-md would also find font-size--md, fontsize-md...

(but this one is up to you to decide 🤭)
I would maybe remove it though, if it doesn't really change anything in the options labels, as they should already be validated

ronros-elementor (changes requested):

packages/packages/core/editor-editing-panel/src/components/creatable-autocomplete/use-filter-options.ts
Omerisra6: removed

ronros-elementor (dismissed):

ronros-elementor (approved):

PR #35829, opened 2026-05-12
Fix: Optimize dynamic class caching by flagging on save [ED-24058]

Omerisra6:
## PR Checklist

## PR Type
- [x] Refactoring (no functional changes, no api changes)

## Summary

* Optimize dynamic value detection for atomic styles caching — flag is now computed once at save time instead of scanning all style data on every page render.

## Description

**Problem (ED-24058):** `Atomic_Styles_Manager` was calling the recursive `contains_dynamic_value()` scan inside the inner breakpoints loop on every page render — an O(N×M) scan over all style props for every style group × every breakpoint.

**Solution:** Compute and store a boolean `has_dynamic` flag at the point of save:

1. **`Dynamic_Value_Detector`** (`modules/atomic-widgets/utils/dynamic-value-detector.php`) — new shared utility that extracts the recursive scan logic from `Atomic_Styles_Manager`.

2. **Global Classes** — `Global_Class_Post` now stores `META_KEY_HAS_DYNAMIC_VALUES` (`_elementor_global_class_has_dynamic`) on every `create()` / frontend `update_data()`. `Global_Classes_Repository::has_any_dynamic_in_ids()` provides an O(1) DB lookup across a set of class IDs.

3. **Local Widget Styles** — `Atomic_Widget_Styles` now hooks into `elementor/document/after_save` (priority 21) to traverse element styles and store `META_KEY_HAS_DYNAMIC_STYLES` (`_elementor_has_dynamic_widget_styles`) on the post.

4. **`Atomic_Styles_Manager::register()`** — extended with an optional `?callable $is_dynamic = null` third parameter. When provided it replaces the runtime `Dynamic_Value_Detector` scan; the decision is still memoized so it is evaluated at most once per style group per render cycle.

5. **Callers wired up:**
   - `Atomic_Global_Styles` passes `is_dynamic` → reads the stored global class flag via `has_any_dynamic_in_ids()`.
   - `Atomic_Widget_Styles` passes `is_dynamic` → reads `META_KEY_HAS_DYNAMIC_STYLES` from post meta.
   - `Atomic_Widget_Base_Styles` passes `fn() => false` — base styles never contain dynamic values.

**Backward compatibility:** Posts/classes that have never been re-saved won't have the new meta key → `is_dynamic` returns `false`, meaning they'll use the file cache. Any truly dynamic values in old unsaved posts would need a re-save to be treated as dynamic again. This is an acceptable tradeoff since the dynamic tag meta is populated on every editor save.

## Test instructions

1. Open a page in the Elementor editor that uses global classes.
2. Add a dynamic tag to a global class style prop (e.g. color → dynamic → post custom field).
3. Save the page/global class.
4. View the page on the frontend — the style should render as inline (not cached to file).
5. Open a page with a widget that uses a static style — verify it is served from a CSS file (check Network tab for `elementor/css/*.css`).
6. Remove the dynamic tag, save again — style should switch back to file-based caching.

## Quality assurance

Fixes #ED-24058

### DIFF
--- a/modules/atomic-widgets/dynamic-tags/dynamic-tags-module.php
+++ b/modules/atomic-widgets/dynamic-tags/dynamic-tags-module.php
@@ -2,9 +2,12 @@
 
 namespace Elementor\Modules\AtomicWidgets\DynamicTags;
 
+use Elementor\Core\Base\Document;
 use Elementor\Modules\AtomicWidgets\DynamicTags\ImportExport\Dynamic_Transformer as Import_Export_Dynamic_Transformer;
 use Elementor\Modules\AtomicWidgets\PropsResolver\Render_Props_Resolver;
 use Elementor\Modules\AtomicWidgets\PropsResolver\Transformers_Registry;
+use Elementor\Modules\AtomicWidgets\Utils\Utils as Atomic_Utils;
+use Elementor\Modules\GlobalClasses\Global_Classes_Dynamic_Index;
 use Elementor\Plugin;
 
 if ( ! defined( 'ABSPATH' ) ) {
@@ -12,6 +15,8 @@ if ( ! defined( 'ABSPATH' ) ) {
 }
 
 class Dynamic_Tags_Module {
+
+	const META_KEY_POST_DYNAMIC = '_elementor_post_styles_dynamic';
 
 	private static ?self $instance = null;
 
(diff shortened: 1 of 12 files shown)


modules/atomic-widgets/dynamic-tags/dynamic-tags-module.php
```diff
@@ -13,6 +16,8 @@
 
 class Dynamic_Tags_Module {
 
+	const META_KEY_POST_DYNAMIC = '_elementor_post_styles_dynamic';
```
Omerisra6, line 19: Why do we need a meta specially for that? Why can't we just add the dynamic flag to the cache dependency array when saving a class?

modules/atomic-widgets/dynamic-tags/dynamic-tags-module.php
```diff
@@ -77,6 +82,88 @@ public function register_hooks() {
...
+				return;
+			}
+
+			if ( Dynamic_Value_Detector::contains_dynamic_value( $element_data['styles'] ?? [] ) ) {
+				$is_dynamic = true;
+			}
+		} );
+
+		update_post_meta( $post_id, self::META_KEY_POST_DYNAMIC, $is_dynamic ? '1' : '0' );
+	}
```
Omerisra6, line 146: same question it seems redundant

PR #35850, opened 2026-05-13
Fix: Components: exposed link in link controls does not update in time [ED-23746]

Omerisra6:
## Problem & solution

Link-in-link controls in Components did not behave correctly:

- A sibling/exposed `LinkControl` only re-evaluated its restriction on its own value change or on a same-`elementId` `elementor:inline-link-changed` event, so panel `setValue` writes produced no cross-control signal.
- When the restriction auto-closed an active control, the partial value remained, leaving stale state.
- The infotip **"Take me there"** CTA could target elements that are not navigable from the current editing context (e.g. inner Twig elements or targets in another document scope), which broke or confused the instance editing flow when selection jumped away from the panel.
- Inner elements of component instances (rendered from Twig, without a registered V1 Backbone view) had `getContainer(id)?.view?.el === null`, so ancestor/descendant restriction was missed on first render and only appeared after a later re-check triggered by typing.

**Fix:** React to global settings writes via the V1 `document/elements/set-settings` command end, clear partial values on auto-close, add a preview-DOM fallback in `getElementDOM` for elements without a V1 view, and **show "Take me there" only when the restriction target’s DOM resolves to the same document as the active editor document** (nearest `[data-elementor-id]` on the target node vs `getCurrentDocumentId()`). No separate React context or override-panel wrapper is required for CTA suppression.

Fixes [ED-23746](https://elementor.atlassian.net/browse/ED-23746).

## PR Type

- [x] Bugfix

### 1. Cross-control reactivity on settings change

**Bug:** Assigning a link in one exposed control did not disable the related sibling/ancestor control's "+" toggle until the user typed.

**Fix:** Subscribe `LinkControl` to `commandEndEvent('document/elements/set-settings')` via `__privateUseListenTo`, and drop the overly narrow `elementId` filter on the `elementor:inline-link-changed` listener.

**Files:**
- `packages/packages/libs/editor-controls/src/controls/link-control.tsx`
- `packages/packages/libs/editor-controls/src/controls/__tests__/link-control.test.tsx` — `should re-evaluate restriction when any element settings change via V1 command`

---

### 2. Partial value not cleared when restriction auto-closes the active control

**Bug:** When `debouncedCheckRestriction` forced the control closed because a new restriction appeared, the previously typed partial URL was kept in the prop.

**Fix:** When auto-closing, call `setValue(null)` (mirroring manual toggle-off) and use a stable debounced callback (`useDebouncedCallback` from `@elementor/utils`) so the debounced closure stays correct.

**Files:**
- `packages/packages/libs/editor-controls/src/controls/link-control.tsx`
- `packages/packages/libs/utils/src/use-debounced-callback.ts` (new)
- `packages/packages/libs/utils/src/index.ts`
- `packages/packages/libs/utils/src/__tests__/use-debounced-callback.test.tsx` (new)
- `packages/packages/libs/editor-controls/src/controls/__tests__/link-control.test.tsx` — `should clear partial value when restriction forcibly closes the active control`

---

### 3. "Take me there" must not navigate to inaccessible / cross-document targets

**Bug:** The CTA could appear for targets that are not in the active document (or not represented by a V1 view in the current preview), leading to broken navigation or leaving the instance editing panel in a bad state.

**Fix:** In `RestrictedLinkInfotip`, derive whether the target is in the **current** document by comparing the nearest `[data-elementor-id]` ancestor on `getContainer(elementId)?.view?.el` with `getCurrentDocumentId()`. Only then render the CTA; otherwise the infotip still explains the restriction but without the button. CTA click always uses `selectElement`.

**Files:**
- `packages/packages/libs/editor-controls/src/components/restricted-link-infotip.tsx`
- `packages/packages/libs/editor-controls/src/controls/__tests__/link-control.test.tsx` — `should hide "Take me there" button when target element lives in a different document` (and default mocks so existing CTA tests stay stable)

**Note:** An earlier iteration used a small `LinkNavigationProvider` context; that was **removed** in favor of this document-scoped check only.

---

### 4. Ancestor/descendant restriction missing for inner component elements

**Bug:** Inner elements of component instances have no registered V1 Backbone view, so `getContainer(innerElementId)?.view?.el` returns `null` and DOM-based restriction checks failed until a later re-check.

**Fix:** In `getElementDOM`, when the container view element is missing, fall back to querying the preview iframe document for `[data-id="…"]`.

**Files:**
- `packages/packages/libs/editor-elements/src/link-restriction.ts`
- `packages/packages/libs/editor-elements/src/sync/types.ts` (optional typing for preview container access where needed)
- `packages/packages/libs/editor-elements/src/__tests__/link-restriction.test.ts`

## Test instructions

Automated (from `packages/`):

```bash
npx jest --config='./jest.config.js' --testPathPattern='link-control.test' --no-coverage
npx jest --config='./jest.config.js' --testPathPattern='link-restriction.test' --no-coverage
npx jest --config='./jest.config.js' --testPathPattern='use-debounced-callback' --no-coverage
```

Manual:

1. Open the v4 editor with Components enabled.
2. Create a Component from a container with a link, add a child with a link, and expose both link properties.
3. Place an instance on a page and open the instance editing panel.
4. **Case 1** — Assign a link in the first exposed control. The second control's "+" should disable immediately without typing.
5. **Case 2** — Open one exposed control, type a partial URL, then trigger a restriction from a sibling. The control should close and the partial value clear.
6. **Case 3** — When a restriction is shown for a target outside the active document (or not navigable the same way), confirm **no** "Take me there" CTA; messaging still visible.
7. **Case 4** — From outside the component, give the wrapping container a link. Open the instance panel; inner exposed link controls should show restriction on first render.

## Quality assurance

- [x] I have added unit tests to verify the code works as intended

Fixes ED-23746

[ED-23746]: https://elementor.atlassian.net/browse/ED-23746?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## 1. Problem & Context

Link-in-link restrictions weren't re-evaluating when sibling/ancestor elements changed via V1 commands, causing stale UI state. The "Take me there" button also appeared for cross-document targets where navigation would fail.

## 2. What Changed (Where)

| File | Change |
|------|--------|
| `link-control.tsx` | Replaced `useMemo` + `debounce` with new `useDebouncedCallback` hook; added V1 command listener for `document/elements/set-settings` |
| `restricted-link-infotip.tsx` | Conditional render of "Take me there" button only when target in current document via new `isTargetInCurrentDocument()` check |
| `link-restriction.ts` | Enhanced `getElementDOM()` fallback for inner component elements without V1 views; added `queryPreviewDOMByElementId()` |
| `use-debounced-callback.ts` | New hook wrapping debounce with ref-based callback stability and cleanup |
| Test files | Added 3 new integration tests + 4 unit tests covering restriction re-evaluation and cross-document scenarios |

## 3. How It Works

Control listens to V1 `set-settings` command → debounced check re-runs `getLinkInLinkRestriction()` → comparison via `isSameRestriction()` prevents unnecessary updates → infotip button only renders if target's `data-elementor-id` matches current document. For inner component elements (no V1 view), falls back to querying preview iframe DOM directly.

## 4. Risks

Debouncing window (300ms) could mask rapid consecutive element changes, though tests show command batching mitigates this. The preview DOM fallback query is direct and unoptimized—could be slow with large DOMs, but only hits on missing V1 container (rare).


packages/packages/libs/editor-elements/src/link-restriction.ts
```diff
@@ -144,12 +145,29 @@ function checkForInlineLink( elementId: string, resolvedValue?: LinkValue ): boo
...
 function getElementDOM( id: string ) {
 	try {
-		return getContainer( id )?.view?.el || null;
+		const fromContainer = getContainer( id )?.view?.el;
+
+		if ( fromContainer ) {
+			return fromContainer;
+		}
+
+		return queryPreviewDOMByElementId( id );
```
Omerisra6, line 158: maybe we should add a comment explaining the component case?

packages/packages/libs/editor-elements/src/link-restriction.ts
```diff
@@ -144,12 +145,29 @@ function checkForInlineLink( elementId: string, resolvedValue?: LinkValue ): boo
...
 	} catch {
 		return null;
 	}
 }
 
+function queryPreviewDOMByElementId( id: string ): HTMLElement | null {
```
Omerisra6, line 164: dont we have a util for that?

packages/packages/libs/editor-controls/src/context/link-navigation-context.tsx
```diff
@@ -0,0 +1,18 @@
...
+	showTakeMeThereButton: boolean;
+};
+
+const LinkNavigationContext = createContext< LinkNavigationContextValue >( {
+	showTakeMeThereButton: true,
+} );
+
+export const useLinkNavigationContext = () => useContext( LinkNavigationContext );
+
+export const HideTakeMeThereProvider = ( { children }: { children: React.ReactNode } ) => (
```
IshayMaya: whats the use for it
and where does it change

packages/packages/core/editor-components/src/components/instance-editing-panel/override-prop-control.tsx
```diff
@@ -193,35 +194,39 @@ function OverrideControl( { overridableProp }: InternalProps ) {
...
 	return (
-		<OverridablePropProvider
-			value={ componentOverridablePropTypeUtil.extract( matchingOverride ) ?? undefined }
-			componentInstanceElement={ componentInstanceElement }
-		>
-			<ElementProvider
-				element={ { id: elementId, type: elementType.key } }
-				elementType={ elementType }
-				settings={ resolvedElementSettings }
+		<HideTakeMeThereProvider>
```
IshayMaya: what is is doing?

packages/packages/libs/editor-controls/src/controls/link-control.tsx
```diff
@@ -69,22 +70,30 @@ export const LinkControl = createControl( ( props: Props ) => {
...
+
+					if ( value !== null ) {
+						setValue( null );
+					}
 				}
 
 				setLinkInLinkRestriction( newRestriction );
 			}, 300 ),
-		[ elementId, isActive, value, linkPlaceholder ]
+		[ elementId, isActive, value, linkPlaceholder, setValue ]
```
IshayMaya: setValue change on every render, maybe we should use useDebounce

packages/packages/libs/editor-controls/src/controls/link-control.tsx
```diff
@@ -69,22 +70,30 @@ export const LinkControl = createControl( ( props: Props ) => {
 
 				if ( newRestriction.shouldRestrict && isActive && ! linkPlaceholder ) {
 					setIsActive( false );
+
+					if ( value !== null ) {
```
IshayMaya: did we not remove the value so far?

packages/packages/libs/editor-controls/src/controls/link-control.tsx
```diff
@@ -69,22 +70,30 @@ export const LinkControl = createControl( ( props: Props ) => {
...
-		[ elementId, isActive, value, linkPlaceholder ]
+		[ elementId, isActive, value, linkPlaceholder, setValue ]
+	);
+
+	useListenTo(
+		commandEndEvent( 'document/elements/set-settings' ),
+		() => {
+			debouncedCheckRestriction();
+		},
+		[ debouncedCheckRestriction ]
```
IshayMaya, line 85: i think we're running it too many times

packages/packages/libs/editor-elements/src/link-restriction.ts
```diff
@@ -144,12 +145,29 @@ function checkForInlineLink( elementId: string, resolvedValue?: LinkValue ): boo
 
 function getElementDOM( id: string ) {
 	try {
-		return getContainer( id )?.view?.el || null;
+		const fromContainer = getContainer( id )?.view?.el;
```
IshayMaya, line 148: wasn't the element change effect enough? we do we need this one as well?

packages/packages/libs/editor-controls/src/context/link-navigation-context.tsx
Omerisra6: This provider is the way the instance editing panel suppresses the "Take me there" CTA on the link-in-link infotip.

Why we need it:
- "Take me there" calls `selectElement(restriction.elementId)`, which selects the *ancestor* element with the conflicting link.
- When that ancestor is outside the component instance (e.g. the wrapping `div-block` with the link), the editor switches out of `InstanceEditingPanel` to the standard panel — the exposed link controls disappear and the user is now editing a completely different element. From the user's perspective the panel breaks.

Where it changes:
- Default: button visible (`showTakeMeThereButton: true`).
- Only flipped to `false` by `HideTakeMeThereProvider`, which wraps each rendered exposed control in `OverridePropControl` (`packages/packages/core/editor-components/src/components/instance-editing-panel/override-prop-control.tsx`). `RestrictedLinkInfotip` consumes the context and conditionally renders the action.

That's the only consumer/provider pair right now — the context exists so that any other panel that mounts a `LinkControl` keeps the default behavior unless it explicitly opts out.

packages/packages/core/editor-components/src/components/instance-editing-panel/override-prop-control.tsx
Omerisra6: This wraps every exposed control rendered by the instance editing panel with `HideTakeMeThereProvider`, which suppresses the link-in-link infotip's "Take me there" CTA on any `LinkControl` rendered inside.

Without it, clicking "Take me there" would `selectElement(ancestorId)` — typically the wrapping `div-block` with the conflicting link, which lives outside the component instance. That selection unmounts `InstanceEditingPanel` and switches the editor to the standard panel, so the user loses the exposed-prop controls they were editing and finds themselves on a different element.

Wrapping at this layer (`OverridePropControl`) means we cover *all* exposed link props automatically, with no need to teach `LinkControl` about its render context.

packages/packages/libs/editor-controls/src/controls/link-control.tsx
Omerisra6: No — before this PR the value was only cleared on a manual toggle (`onEnabledChange`), not when the control auto-closes due to a newly-detected restriction.

Repro on `main`:
1. Open an exposed link control, type a partial URL.
2. Make a sibling change that introduces a link-in-link restriction.
3. `debouncedCheckRestriction` fires, sets `isActive = false` and closes the collapsible — but the partial URL stays in the prop value.
4. The next time the user re-opens the control, the stale value is still there, and because the restriction view is computed from `value ?? linkPlaceholder` it can also produce secondary / misleading restrictions.

The new branch mirrors what `onEnabledChange` already does on close (`if (! newState && value !== null) setValue(null)`). It's covered by the new test `should clear partial value when restriction forcibly closes the active control`.

IshayMaya (dismissed):

IshayMaya (approved):

PR #35871, opened 2026-05-13
Fix: Atomic Forms not working inside Components [ED-23721]

MiryamOren:
https://www.loom.com/share/f51c66255e334b0e9d5213775b5a2af6

## 1. Problem & Context

Atomic Forms (Component_Instance widgets) couldn't be searched recursively because `find_element_recursive()` didn't know how to traverse into component's origin elements. This breaks element selection/targeting in components nested within other elements.

## 2. What Changed (Where)

- **utils.php**: Refactored `find_element_recursive()` to use a filterable hook for retrieving inner elements, enabling custom behavior per element type.
- **components/module.php**: Registered filter hook; new methods extract component ID and delegate search to Component_Instance.
- **component-instance.php**: Added methods to fetch component's origin elements formatted for recursive search with proper ID prefixes.
- **test-find-element-recursive.php**: New test suite covering raw element search and component element discovery.

## 3. How It Works

When `find_element_recursive()` encounters a Component_Instance widget, the filter hook fires, triggering `get_inner_elements_data_for_search()`. This retrieves the component's stored elements from the repository, formats their IDs to reflect the instance context (e.g., prefixing), and returns them for continued recursion. Raw elements bypass this and search directly.

## 4. Risks

Filter hook adds indirection and instantiation overhead per recursive level containing components—potential performance impact on deeply nested component hierarchies. Mitigation: filter only fires for elements with defined inner_elements, and Components_Repository likely caches lookups.


includes/base/element-base.php
```diff
@@ -327,6 +327,22 @@ public function get_children() {
 		return $this->children;
 	}
 
+	/**
+	 * Get inner elements data for recursive search.
```
eyalmrejen-elementor: Redundant comments - the function name is great and informative

includes/utils.php
```diff
@@ -730,18 +730,30 @@ public static function find_element_recursive( $elements, $id ) {
 				return $element;
 			}
 
-			if ( ! empty( $element['elements'] ) ) {
-				$element = self::find_element_recursive( $element['elements'], $id );
+			$inner_elements = self::get_element_inner_elements_for_search( $element );
```
Omerisra6: What’s the difference? Could this cause issues for other callers?

modules/components/widgets/component-instance.php
```diff
@@ -100,4 +102,33 @@ private function get_merged_overrides( $value ): array {
...
+	 *
+	 * Overrides the parent method to return the origin component's inner elements
+	 * instead of direct children, since Component_Instance stores a reference
+	 * to the origin component rather than containing elements directly.
+	 *
+	 * @return array Inner elements data from origin component.
+	 */
+	public function get_inner_elements_data_for_search(): array {
+		$settings = $this->get_settings();
+		$component_id = $settings['component_instance']['value']['component_id']['value'] ?? null;
```
Omerisra6: Don’t we have a `get_component_id` method? I believe we should have one to avoid relying on that path.

includes/utils.php
Omerisra6: I think that could lead to a performance issue on big pages, we are now creating an instance for each element inside the tree just for search, not sure how many times we are using this method.

ronenboxer: Should we suggest caching?

includes/utils.php
```diff
@@ -730,18 +730,30 @@ public static function find_element_recursive( $elements, $id ) {
...
+				if ( $found ) {
+					return $found;
 				}
 			}
 		}
 
 		return false;
 	}
 
+	private static function get_element_inner_elements_for_search( array $element_data ): array {
```
ronenboxer: Why not the same name as the element base's one?

includes/utils.php
```diff
@@ -730,18 +730,30 @@ public static function find_element_recursive( $elements, $id ) {
...
+		$element_instance = Plugin::$instance->elements_manager->create_element_instance( $element_data );
```
ronenboxer: I would also use this opportunity to create a util function that does this logic to be used in the future without having the entire `Plugin::$instance->elements_manager->` nightmare. your decision

modules/components/widgets/component-instance.php
ronenboxer: I also remember php errors thrown when one of the paths is unset, despite the ?? operator. would suggest to wrap it with isset just to be on the safe side

includes/utils.php
```diff
@@ -730,18 +731,38 @@ public static function find_element_recursive( $elements, $id ) {
...
+		}
+
+		return $element_instance->get_inner_elements_data_for_search();
+	}
+
+	private static function is_component_instance( array $element_data ): bool {
+		return isset( $element_data['elType'], $element_data['widgetType'] )
+			&& 'widget' === $element_data['elType']
+			&& Component_Instance::get_element_type() === $element_data['widgetType'];
+	}
```
Omerisra6: lets move this to the component module

includes/utils.php
```diff
@@ -730,18 +731,38 @@ public static function find_element_recursive( $elements, $id ) {
...
+				: ( $element['elements'] ?? [] );
 
-				if ( $element ) {
-					return $element;
+			if ( ! empty( $inner_elements ) ) {
+				$found = self::find_element_recursive( $inner_elements, $id );
+
+				if ( $found ) {
+					return $found;
 				}
```
Omerisra6: expose a filter that the component module will use

Omerisra6 (approved):

ronenboxer (approved):

PR #35883, opened 2026-05-14
POC: Scoped dynamic global class placeholders

Omerisra6:
## Summary

POC for **scoped dynamic global classes** using CSS variable interleaving (per HLD).

**Architecture (consumer-delegated resolution):**
- **Core `Dynamic_Styles_Manager`** — definition catalog only: registers `--e-dyn-*` var names + dynamic tag nodes when scoped global class CSS is built. Provides `get_definitions()`, `build_inline_style()`, `wrap_scoped_html()`. Does **not** resolve tag values.
- **`Dynamic_Tags_Style_Resolver`** — optional Core helper consumers may call (`get_tag_data_content`).
- **Pro Collection Loop** — reads definitions, resolves values per item after `the_post()`, applies inline CSS vars on a scope wrapper so `var()` in the static stylesheet resolves per loop item.

**Pipeline:**
1. Parser tags variants with `is_scoped` when props contain dynamic values
2. Renderer emits `z-index: var(--e-dyn-...)` and registers definitions on DSM
3. Static CSS cached with `.placeholders.json` sidecar (definitions only)
4. Loop consumer resolves tags and wraps each item: `style="--e-dyn-...: [value];"`

**Fix included:** Scoped resolution no longer routes through `Render_Props_Resolver` with `String_Prop_Type` (which returned null for `$$type: dynamic`).

Companion PR: elementor/elementor-pro#6923

## Test plan

- [ ] Enable `e_atomic_elements`, `e_pro_collection_loop`, `e_v4_page`
- [ ] Seed via `/wp-admin/?poc_seed_loop=1` (mu-plugin) or apply global class with dynamic prop in loop template
- [ ] Frontend: each loop item shows distinct resolved values (e.g. post ID in z-index POC)
- [ ] Verify static CSS contains `var(--e-dyn-...)` and item wrappers carry inline `--e-dyn-*` values
- [ ] `phpunit` — `Test_Dynamic_Styles_Manager`, existing atomic styles manager tests


modules/atomic-widgets/styles/atomic-styles-manager.php
```diff
@@ -31,6 +32,8 @@ class Atomic_Styles_Manager {
 
 	private array $fonts = [];
 
+	private array $dynamic_styles_decisions = [];
```
Omerisra6: maybe the dynamic tags module should own that logic?

modules/atomic-widgets/styles/atomic-styles-manager.php
```diff
@@ -149,18 +171,112 @@ private function render( array $styles_by_key ) {
...
-	private function render_css( array $styles, string $style_key ) {
+	private function has_dynamic_styles( string $style_key, callable $get_styles ): bool {
+		if ( ! array_key_exists( $style_key, $this->dynamic_styles_decisions ) ) {
+			$this->dynamic_styles_decisions[ $style_key ] = self::style_definitions_require_inline_dynamic_path( $get_styles() );
+		}
+
+		return $this->dynamic_styles_decisions[ $style_key ];
+	}
+
+	private function enqueue_inline_style( array $breakpoint_path, string $breakpoint_media, string $css, ?string $version ): void {
```
Omerisra6: what inline styles do we need here?

modules/atomic-widgets/styles/atomic-styles-manager.php
```diff
@@ -149,18 +171,112 @@ private function render( array $styles_by_key ) {
 		}
 	}
 
-	private function render_css( array $styles, string $style_key ) {
+	private function has_dynamic_styles( string $style_key, callable $get_styles ): bool {
```
Omerisra6: this is a pricey operation, we should run it once, only on render (when cache is invalid)

modules/atomic-widgets/styles/css-files-manager.php
```diff
@@ -5,10 +5,11 @@
 class CSS_Files_Manager {
 	const DEFAULT_CSS_DIR = 'elementor/css/';
 	const FILE_EXTENSION = '.css';
+	const PLACEHOLDERS_EXTENSION = '.placeholders.json';
```
Omerisra6: I think it doesn't make sense that the CSS manager owns the placeholders management, while the dynamic styles manager owns all the other logic.

Maybe we need to make the placeholder mechanism more generic

modules/atomic-widgets/styles/styles-renderer.php
```diff
@@ -152,6 +172,43 @@ private function props_to_css_string( array $props ): string {
 			->implode( '' );
 	}
 
+	/**
+	 * @return array{0: array, 1: string}
+	 */
+	private function split_dynamic_props( array $props, string $class_id, int $variant_index ): array {
```
Omerisra6: I think we should have more obvious boundaries between each layer and each entity

PR #35932, opened 2026-05-18
Fix: Gracefully handle externally deleted CSS classes in the editing panel [ED-24067]

MiryamOren:
image

This is NOT a change to the normal class deletion flow (which already unapplies the class from all elements referencing it).

This targets bug mitigation only: cases where a class is removed from the DB outside the normal flow (e.g. a now-fixed Kit import bug, direct DB deletion by the user, or a third-party plugin), leaving elements with ghost class references.

Before this fix, opening the style tab on such an element threw
`StylesProviderNotFoundError` and crashed the whole editing panel.
Now, we filter out missing classes, and show an alert under the class selector when missing class references exist, with a dismiss button that removes the dead references from the element.

## 1. Problem & Context

Externally deleted CSS classes (references to styles no longer in any provider) leave the UI in an inconsistent state. This PR adds detection, alerts, and batch removal to gracefully handle orphaned class IDs.

## 2. What Changed (Where)

| File | Change |
|------|--------|
| `use-apply-and-unapply-class.ts` | Renamed `useApplyClass`/`useUnapplyClass` → `useUndoableApplyClass`/`useUndoableUnapplyClass`; refactored `useUnapply()` → `useUnapplyClasses()` (array-based, batch removal with validation) |
| `use-missing-classes.ts` | New hook: detects applied class IDs absent from all providers |
| `css-class-selector.tsx` | Added missing classes detection, alert UI, dismiss callback that batch-unapplies orphaned IDs |
| `use-active-style-def-id.ts` | Added `useValidClassIds()` to filter out missing IDs before setting active style |
| `missing-classes-alert.tsx` | New component: warning alert with dismissal |
| `css-class-menu.tsx`, `duplicate-class-menu-item.tsx` | Updated imports to use new undoable function names |
| `css-class-selector.test.tsx` | Added 4 test cases for missing class detection and removal logic |

## 3. How It Works

On render, `useMissingClassesIds()` compares applied IDs against all known provider styles; orphaned IDs trigger the alert. Dismissing calls `clearMissingClasses()` → `useUnapplyClasses(missingClassesIds)`, which batch-removes only the orphaned IDs while preserving valid ones. `useValidClassIds()` prevents orphaned IDs from becoming the active style. Undo/redo uses the new batch-removal function, fixing single-ID assumption.

## 4. Risks

`useUnapplyClasses()` now validates all IDs exist before removal; fails if called with partially missing IDs (the validation logic filters first, so error path is defensive). Batch removal changes undo semantics slightly (single action vs. per-ID), but consistent with intent. Test coverage added.


Omerisra6 (commented):
nice, please add tests

Omerisra6 (approved):

PR #35941, opened 2026-05-19
Internal: Add mention support for newly added form field types [ED-24059]

marcin-el:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #

## 1. Problem & Context

Adding mention/field reference support for newly introduced form field types (select, date-picker, time-picker). Previous implementation used basic type checking; now needs schema-aware extraction and deduplication by CSS ID.

## 2. What Changed (Where)

- `use-form-field-suggestions.ts`: Refactored to traverse form hierarchy, extract string props via schema defaults, deduplicate CSS IDs, support new field types.
- `form-handlers.js`: Changed grouped field ID from `name` to `interaction-id` for mention support.
- `use-form-field-suggestions.test.ts`: Added 11 test cases covering schema defaults, deduplication, nested resolution.

## 3. How It Works

Entry point: hook listens for form structure changes. On trigger, retrieves selected element, walks parent chain to find form container, recursively collects form field children. For each field, extracts CSS ID via `getSettingWithDefault()` (checks settings first, falls back to widget schema default), deduplicates with Set, returns suggestions. Field type validation now uses `isFormFieldWidgetType()` instead of direct array check for extensibility.

## 4. Risks

Minor: `stringPropTypeUtil.extract()` behavior change from `isTransformable()` pattern—verify prop extraction handles all edge cases (null, undefined, non-string types). Test coverage is solid; main risk is schema cache miss returning null cascading to missing defaults.


packages/packages/libs/editor-controls/src/hooks/use-form-field-suggestions.ts
```diff
@@ -15,12 +15,61 @@ const FORM_FIELD_WIDGET_TYPES = [
...
+
+	while ( container ) {
+		if ( container.model.get( 'elType' ) === FORM_ELEMENT_TYPE ) {
+			return container;
+		}
+
+		container = container.parent ?? null;
+	}
+
+	return null;
```
ronenboxer, line 62: will it work with something like this?
`containerElement = container.el`
and then `formElement = containerElement.closest('form[data-el-type="e-form"])`
`id = formElement.dataset.id`
`getContainer(id)`

marcin-el: we resolve parent form with the container parent chain from the selected element which matches how we read field settings for mentions and avoids coupling to DOM, I would stick to the current model only approach
if we hit case when parent is not set we could add that DOM fallback, wdyt?

ronenboxer (dismissed):

packages/packages/libs/editor-controls/src/hooks/use-form-field-suggestions.ts
```diff
@@ -15,12 +15,61 @@ const FORM_FIELD_WIDGET_TYPES = [
...
+}
+
+function getSettingWithDefault( child: V1Element, widgetType: string, key: string ): unknown {
+	const fromJson = child.settings.toJSON?.()?.[ key ];
+
+	if ( fromJson !== null && fromJson !== undefined ) {
+		return fromJson;
+	}
+
+	const fromGet = child.settings.get( key );
```
mykytamurzin, line 36: is this returns the same data `child.settings.get( key )` and `child.settings.toJSON?.()?.[ key ]` ?

packages/packages/libs/editor-controls/src/hooks/use-form-field-suggestions.ts
```diff
@@ -15,12 +15,61 @@ const FORM_FIELD_WIDGET_TYPES = [
...
 };
 
+function extractStringPropValue( value: unknown ): string | undefined {
+	const extracted = stringPropTypeUtil.extract( value );
+
+	if ( extracted ) {
+		return extracted;
+	}
+
+	return typeof value === 'string' && value ? value : undefined;
```
mykytamurzin: do we need this check? especially undefined return, looks over-defensive. 

`extract` is returning string | null - this should be enough I guess

mykytamurzin:
```
return extractStringPropValue( getSettingWithDefault( child, widgetType, CSS_ID_PROP_KEY ) ) ?? null;
```
looking at how it's used - we can just safely return string | null here and remove condition

packages/packages/libs/editor-controls/src/hooks/use-form-field-suggestions.ts
```diff
@@ -31,42 +80,47 @@ export function useFormFieldSuggestions( options?: Options ): Suggestion[] {
...
+			const seenCssIds = new Set< string >();
 
-			container.children.forEachRecursive?.( ( child ) => {
+			formContainer.children.forEachRecursive?.( ( child ) => {
 				const widgetType = child.model.get( 'widgetType' ) as string | undefined;
 
-				if ( ! widgetType || ! FORM_FIELD_WIDGET_TYPES.includes( widgetType ) ) {
+				if (
+					! widgetType ||
+					! FORM_FIELD_WIDGET_TYPES.includes( widgetType as ( typeof FORM_FIELD_WIDGET_TYPES )[ number ] )
```
mykytamurzin: `widgetType as ( typeof FORM_FIELD_WIDGET_TYPES )[ number ]`
do we need this typing? looks weird

modules/atomic-widgets/assets/js/frontend/form-handlers.js
```diff
@@ -138,12 +138,18 @@ function getAtomicFormFields( form ) {
...
+	return primaryInput?.dataset.interactionId ?? name;
+}
+
 function getGroupedFields( name, inputs, type, form ) {
 	const checkedInputs = inputs.filter( ( input ) => input.checked );
 	const value = getAtomicFormFieldValue( checkedInputs, type );
 	const options = getAtomicFieldOptions( inputs, type, form );
 	return {
-		id: name,
+		id: getGroupedFieldId( name, inputs, checkedInputs ),
```
mykytamurzin, line 152: are we breaking BC by this? name <> interactionId

forms created before with mentions will work?

packages/packages/libs/editor-controls/src/hooks/use-form-field-suggestions.ts
marcin-el: yup, good call, simplified

modules/atomic-widgets/assets/js/frontend/form-handlers.js
marcin-el: mentions use _cssid which maps to element id so form_fields.id must be interactionId

packages/packages/libs/editor-controls/src/hooks/use-form-field-suggestions.ts
```diff
@@ -31,42 +72,44 @@ export function useFormFieldSuggestions( options?: Options ): Suggestion[] {
...
 
-				if ( ! widgetType || ! FORM_FIELD_WIDGET_TYPES.includes( widgetType ) ) {
+				if ( ! widgetType || ! isFormFieldWidgetType( widgetType ) ) {
 					return;
 				}
 
 				if ( options?.inputType ) {
-					const typeProp = child.settings.get( 'type' );
-					const typeValue = isTransformable( typeProp ) ? typeProp.value : typeProp;
+					const typeValue = extractStringPropValue( getSettingWithDefault( child, widgetType, 'type' ) );
```
mykytamurzin, line 98: you have a helper for this `getFieldCssId`

packages/packages/libs/editor-controls/src/hooks/use-form-field-suggestions.ts
```diff
@@ -15,12 +15,53 @@ const FORM_FIELD_WIDGET_TYPES = [
...
 	'e-form-time-picker',
-];
+] as const;
+
+function isFormFieldWidgetType( widgetType: string ): boolean {
+	return ( FORM_FIELD_WIDGET_TYPES as readonly string[] ).includes( widgetType );
+}
+
+const FORM_ELEMENT_TYPE = 'e-form';
+const CSS_ID_PROP_KEY = '_cssid';
```
mykytamurzin: move consts to top

packages/packages/libs/editor-controls/src/hooks/use-form-field-suggestions.ts
marcin-el: nope, getFieldCssId is for cssid prop, here we read type filter by inputType

packages/packages/libs/editor-controls/src/hooks/use-form-field-suggestions.ts
marcin-el: done

packages/packages/libs/editor-controls/src/hooks/use-form-field-suggestions.ts
mykytamurzin: it can be generalised maybe into a way 
```
function getFieldStringSetting( child: V1Element, widgetType: string, key: string ): string | null {
      return extractStringPropValue( getSettingWithDefault( child, widgetType, key ) );
  }
```

murzin-nikita (approved):

PR #35947, opened 2026-05-19
Internal: Add render_atomic_element AJAX action for loop edit mode [ED-23852]

IshayMaya:
## Summary

- Adds `render_atomic_element` AJAX action to render an atomic element server-side in the editor context
- Wires the action into the atomic widgets module and exposes it to the legacy dynamic tags manager for editor use
- Includes PHPUnit coverage for the new action

## Context

Supports collection loop edit mode ([ED-23852](https://elementor.atlassian.net/browse/ED-23852)) — Pro needs to fetch full loop HTML to display static preview items alongside the editable template item.

## Test plan

- [ ] Run PHPUnit: `tests/phpunit/elementor/modules/atomic-widgets/ajax/test-render-element-action.php`
- [ ] In editor with Pro branch, verify loop static items render after entering edit mode
- [ ] Confirm AJAX returns rendered HTML for a valid collection loop element payload

Fixes ED-23852

[ED-23852]: https://elementor.atlassian.net/browse/ED-23852?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## 1. Problem & Context

PR ED-23852 enables server-side rendering of atomic elements during loop edit mode by adding a new AJAX action that renders elements with proper permissions context, and improves dynamic tag caching to support cache-busting for fresh renders.

## 2. What Changed (Where)

- **New AJAX action** (`modules/atomic-widgets/ajax/render-element-action.php`): Renders atomic elements with edit mode context and permissions checks.
- **Module registration** (`modules/atomic-widgets/module.php`): Hooks new action into `elementor/ajax/register_actions`.
- **Dynamic tag caching** (`assets/dev/js/editor/components/dynamic-tags/manager.js` + types): Adds `disableCache` option to `refreshCacheFromServer()` for cache-busting via `unique_id`.
- **Type refinements** (`packages/core/editor-canvas/src/legacy/types.ts` + related): Exports `NestedTemplatedElementViewClass` type, marks `_renderChildren()` async, adds Backbone `on/off` methods.
- **Test coverage** (`tests/phpunit/elementor/modules/atomic-widgets/ajax/test-render-element-action.php`): Full handler validation (render output, payload validation, permissions, element instantiation).

## 3. How It Works

Request flows: client sends `{editor_post_id, data}` → handler validates permissions via `get_with_permissions()` → temporarily enables edit mode → switches document context → creates element instance → captures `print_element()` output via ob_start → restores original state in finally block → returns HTML. Dynamic tags bypass cache when `disableCache: true` by appending `unique_id` to prevent stale renders.

## 4. Risks

**Permission bypass risk**: Handler relies on `get_with_permissions()` — verify it rejects unauthenticated requests (test confirms this). **State pollution**: Finally block restores edit mode and query, but nested exceptions could leak state if uncaught. **Cache stampede**: Disabling cache on every refresh could hammer server; unclear if this is per-request or session-wide.


modules/atomic-widgets/ajax/render-element-action.php
```diff
@@ -0,0 +1,63 @@
...
+namespace Elementor\Modules\AtomicWidgets\Ajax;
+
+use Elementor\Core\Common\Modules\Ajax\Module as Ajax;
+use Elementor\Plugin;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+class Render_Element_Action {
```
Omerisra6, line 12: I think we should add a render mechanism in the editor, as we have for other elements and prop types.

That way, we will hold the query in the settings, resolve it when needed, and re-render the templates on change.

assets/dev/js/editor/components/dynamic-tags/manager.js
```diff
@@ -40,7 +40,11 @@ module.exports = elementorModules.Module.extend( {
 
 		this.cacheCallbacks = [];
 
+		// The ajax batcher keys pending requests by `unique_id` and defaults it to the action name.
+		// Concurrent `render_tags` batches (common with v4 awaited renders) would otherwise overwrite
+		// each other in the pending map, dropping the earlier batch's callbacks. A per-batch id keeps them distinct.
 		elementorCommon.ajax.addRequest( 'render_tags', {
+			unique_id: 'render_tags-' + elementorCommon.helpers.getUniqueId(),
```
Omerisra6: if we need to skip cache for specific calls, lets add a flag in the dynamic transformer (something like `skipCache`) and skip it only for v4.

assets/dev/js/editor/components/dynamic-tags/manager.js
```diff
@@ -51,15 +51,34 @@ module.exports = elementorModules.Module.extend( {
...
 				} );
 			},
-		} );
+		};
+
+		var uniqueId = cacheCallbacks.reduce( function( found, entry ) {
+			return found || ( entry.unique_id || null );
+		}, null );
+
+		if ( uniqueId ) {
```
Omerisra6: `uniqueId` has no meaning, Its hard to understand what it does. Maybe we should use name like `shouldCache` or add a comment

packages/packages/core/editor-canvas/src/legacy/create-pro-promotion-nested-type.ts
```diff
@@ -57,7 +57,7 @@ function createPromotionView( BaseView: typeof ElementView ): typeof ElementView
 			} );
 		}
 
-		_renderChildren() {}
+		async _renderChildren() {}
```
Omerisra6, line 60: what is it for?

packages/packages/core/editor-editing-panel/src/dynamics/dynamic-transformer.ts
```diff
@@ -53,8 +53,13 @@ function getDynamicValue( name: string, settings: Record< string, unknown > ) {
 	}
 
 	return new Promise( ( resolve ) => {
+		const extendedWindow = window as unknown as LegacyWindow;
+		const getUniqueId = extendedWindow.elementorCommon?.helpers?.getUniqueId;
+
+		const uniqueId = getUniqueId ? `render_tags-${ getUniqueId() }` : undefined;
```
Omerisra6: the call shouldn't know the method internals, dont use `render_tags` prefix

packages/packages/core/editor-canvas/src/legacy/create-pro-promotion-nested-type.ts
IshayMaya: its an async method, i've updated its type so needed to update hese as well. the type was wrong

assets/dev/js/editor/components/dynamic-tags/manager.js
IshayMaya: i initially passed just a boolean and implemented the unique id from here but then moved it outside, mauybe it should indeed just be a boolean?

Omerisra6: I think it should be a name with meaning. I'm fine with passing a bool, but how would it be called?

IshayMaya: updated

packages/packages/core/editor-editing-panel/src/dynamics/dynamic-transformer.ts
IshayMaya: changed to boolean

Omerisra6 (approved):

PR #35999, opened 2026-05-25
Fix: Angie to manage insertion of orphan widgets [ED-24123]

ronros-elementor:
## 1. Problem & Context

ED-24123: MCP tool now needs to handle orphan widgets (leaf widgets as root elements) by wrapping them in a container, plus validate that only LLM-compatible widgets are used in compositions.

## 2. What Changed (Where)

- **tool.ts**: Handler refactored to call `adaptLeafRootParams()` before processing; added LLM availability check via `isWidgetAvailableForLLM()`.
- **xml-leaf-wrapper.ts**: New module exports `adaptLeafRootParams()` to wrap leaf widget roots in div-block containers with zero spacing.
- **size.ts**: Added clarification comment about schema divergence from PHP implementation.
- **xml-leaf-wrapper.test.ts**: Comprehensive test suite (5 scenarios) covering wrapping logic and edge cases.

## 3. How It Works

When build-compositions handler receives params: (1) validates all widgets support V4 + LLM, (2) checks if root is leaf widget via `adaptLeafRootParams()`, (3) if yes, wraps in div-block with config ID + zero margin/padding, (4) returns modified xmlStructure and stylesConfig. Non-leaf roots and unknown elements bypass wrapping and return unchanged.

## 4. Risks

**Serialization round-trip**: DOMParser → cloneNode → XMLSerializer cycle could lose whitespace or attributes. Mitigate: Test covers attribute preservation; schema validation catches malformed output. **Config ID mapping**: Assumes `widgetsCache[DIV_BLOCK_TAG].title` exists; falls back to tag name if missing — safe but worth verifying div-block always in cache.


eavichay-elementor (changes requested):
Wrapping with a simple div-block (and setting the style to 0 padding/margin) should be executed only if the root element in the AI-provided tree is NOT a container (i.e. heading, buttons, etc). If the root element is a container - no need to add another layer.

# Suggested refactor and isolation of concerns
## Composition Builder
Code should remain as-is unchanged. It's a functional utility that is not aware where the data came from.

## Build Compositions Tool
This is where the validation/adaptation of the data happens - so this is where the "divblock" wrapper injection should occur. I suggest instead of creating business logic - pass the incoming tool params via an adapter ("filter") that does the check. If a root element in the XML is a "leaf" element, it must be inside a container. This is where the XML can be changed and the wrapping node can be added. The wrapping node `configurationId` should also be added to the "styleProperties" param with zero padding/margin values - So it would not affect the final result layout.

-----

Using this SOC approach keeps the complex logic of building as-is, no extra potential-bug holes. The XML adapter is a standalone utility - therefore can be tested in isolation. The tool is a flow function, adding the adapter in the beginning of the flow should work out clean.

-----

### Forward-Compatibility side note
In the future, the "build-composition" would support adding constructs into any valid container, not just in the root document, therefore simply by disabling the "filter" when a root container is given other than the document, the support would be a simple one-line change.

eavichay-elementor (approved):

PR #36014, opened 2026-05-26
Feature: Element-aware props resolver, bound panel sections, and nested dependencies [ED-24051]

Omerisra6:
## Summary

Enables element-aware prop transformers and PHP-bound panel sections for nested object props in the atomic editor: transformers receive the owning element via `Props_Resolver_Context`, and sections with `bind` render nested controls with object-scoped dependency visibility and union-aware unwrap in the panel.

Companion Pro PR: https://github.com/elementor/elementor-pro/pull/6980

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

### Problem

1. **Transformers lacked element context.** `Props_Resolver` transformers could not access the element owning the props being resolved without global side-channels (e.g. `Render_Context`).
2. **No generic nested settings shell.** Complex object props (e.g. loop query settings) needed PHP-defined child controls in a bound section with correct binding and visibility.
3. **Dependencies evaluated at element scope.** Nested object fields with relative dependency paths (e.g. `source`, `template_type`) were evaluated against element-level settings instead of sibling keys within the object.
4. **Dynamic tags widen props to unions.** Section bind targets like `query` can become `Union_Prop_Type` (`loop-query` + `dynamic`); the panel must resolve the object variant for shape, disable rules, and envelope unwrap/rewrap.

### Solution

**Props resolver — element context**

- Optional `?Element_Base $element = null` on `Props_Resolver::resolve()`, `resolve_item()`, and `transform()`.
- `Props_Resolver_Context::set_element()` / `get_element()`.
- `Has_Atomic_Base` passes `$this` from `get_atomic_settings()` and `get_atomic_setting()`.
- `Render_Props_Resolver` and `Import_Export_Props_Resolver` updated; fully backward-compatible.

**Editor — bound panel sections**

- PHP: optional `bind` on `Section`; `Has_Atomic_Base` resolves section schema from object or union object variant.
- Editor: `BoundSettingsSection` + `TransformableObjectBridge` unwrap/rewrap transformable object envelopes; nested fields use `SettingsControl` with `dependencyScope` for object-scoped hide/disable.
- `resolveObjectPropType()` shared for union → object shape resolution (PHP validation + TS panel).

**Nested prop dependencies**

- `getObjectSettingsWithDefaults`, scoped `isDisabled`, and `extractDependencyEffect` evaluate against the inner object bag, not top-level element settings.
- Bridge fallback unwrap/create when `createPropUtils` is not registered for the object key.

### Key decisions

- Element is threaded as an explicit parameter (stateless, testable) — not global state.
- Bound **sections** (not a custom control type): Pro defines `Section::make()->bind('query')` and child controls in PHP; Core renders generically.
- Dependency clauses on schema fragments use relative paths within the object shape.

## Test instructions

1. Run atomic widget / props resolver tests — existing render paths unchanged when element is omitted.
2. **Bound section:** Open Collection Loop (with Pro PR); expand Query section; verify nested fields bind, persist edits, and show/hide per dependency clauses.
3. **Union / dynamic:** With dynamic tags enabled, confirm query section still renders when value is `loop-query` (not `dynamic`).
4. **Dependencies:** Confirm relative paths (e.g. `source` → filter fields) react to sibling object keys.
5. **Transformer context:** Assert a transformer reading `$context->get_element()` receives the correct element during render resolution.

## Quality assurance

Fixes ED-24051

## 1. Problem & Context

Enables sections to bind to object props, allowing nested controls to scope dependencies and transformations relative to parent object shape. Supports element context in props resolution for dynamic tag evaluation.

## 2. What Changed (Where)

| File | Change |
|------|--------|
| `Section.php` | Added `bind()` method; refactored `jsonSerialize()` to conditionally include bind in output |
| `Has_Atomic_Base.php` | Section bind resolves nested schema from Object/Union prop types; passes element context to resolver |
| `Props_Resolver*.php` | Added `$options` parameter threading through resolve chain; updated abstract signatures |
| `BoundSettingsSection.tsx` | New component wraps TransformableObjectBridge; renders nested controls with dependency scoping |
| `TransformableObjectBridge.tsx` | New component unwraps transformable objects; provides inner shape to nested prop context |
| `SettingsControl.tsx` | Now accepts optional `dependencyScope`; applies visibility rules and PropKeyProvider for nested context |
| `prop-dependency-utils.ts` | Added `resolveObjectPropType()`, `getObjectSettingsWithDefaults()`, `getDependencyEvaluationSettings()` for nested dependency evaluation |

## 3. How It Works

When a Section has `bind: "query"`, the trait resolves the bound prop's shape (Object or Union variant), then validates nested controls against that schema. During render, `BoundSettingsSection` wraps content in `TransformableObjectBridge`, which unwraps the transformable envelope (`{ $$type, value }`) and re-wraps edits. Nested controls run dependency checks against the inner object's shape, not the root element schema. Element reference flows through `resolve()` options, enabling dynamic tag resolution in transformers.

## 4. Risks

**Nested dependency evaluation path mismatch**: Dependencies with absolute paths (targeting root element) vs. relative paths (within object) could evaluate incorrectly if `getDependencyEvaluationSettings()` doesn't correctly merge scopes—mitigated by test coverage for both cases. **Breaking change to resolver signatures**: All `Prop_Type` implementations must handle new `$options` parameter—mitigated by default empty array in abstract method.


includes/managers/icons.php
```diff
@@ -431,6 +431,16 @@ public static function on_import_migration( array $element, $old_control = '', $
 		return $element;
 	}
 
+	public static function get_library_style_handle( string $library ): string {
```
Omerisra6: remove

packages/packages/core/editor-editing-panel/src/controls-registry/element-controls/registry.ts
```diff
@@ -1,10 +1,12 @@
 import { type ControlComponent } from '@elementor/editor-controls';
 
+import { ObjectSectionControl } from '../../controls/object-section-control';
 import { type ControlRegistry, controlsRegistry } from '../controls-registry';
 import { TabsControl } from './tabs-control/tabs-control';
 
 const controlTypes = {
 	tabs: { component: TabsControl as ControlComponent, layout: 'full' },
+	'object-section': { component: ObjectSectionControl as ControlComponent, layout: 'custom' },
```
Omerisra6: It shouldn't be inside the element controls registry. Move it to the regular registry.

packages/packages/core/editor-editing-panel/src/controls/object-section-control.tsx
```diff
@@ -0,0 +1,116 @@
...
+		topDivider?: boolean;
+	};
+};
+
+type ObjectSectionControlProps = {
+	label: string;
+	items: SerializedItem[];
+};
+
+export const ObjectSectionControl = createControl( ( { label, items }: ObjectSectionControlProps ) => {
```
Omerisra6: why arent we simply reusing the logic from sections?

packages/packages/core/editor-editing-panel/src/controls-registry/control-layout.tsx
Omerisra6, line 1: ensure the logic wasnt changes

modules/atomic-widgets/controls/types/object-section-control.php
```diff
@@ -0,0 +1,31 @@
+<?php
+namespace Elementor\Modules\AtomicWidgets\Controls\Types;
+
+use Elementor\Modules\AtomicWidgets\Controls\Base\Atomic_Control_Base;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+class Object_Section_Control extends Atomic_Control_Base {
```
IshayMaya: maybe just `Section_Control`

modules/atomic-widgets/props-resolver/props-resolver-context.php
```diff
@@ -15,6 +16,8 @@ class Props_Resolver_Context {
 
 	private bool $disabled = false;
 
+	private ?Element_Base $element = null;
```
IshayMaya: maybe it should be something more generic? and not element aware

modules/atomic-widgets/props-resolver/props-resolver.php
```diff
@@ -97,7 +99,8 @@ protected function transform( $value, $key, Prop_Type $prop_type ) {
 			$context = Props_Resolver_Context::make()
 				->set_key( $key )
 				->set_disabled( (bool) ( $value['disabled'] ?? false ) )
-				->set_prop_type( $prop_type );
+				->set_prop_type( $prop_type )
+				->set_element( $element );
```
IshayMaya: here as well

packages/packages/core/editor-editing-panel/src/controls-registry/object-prop-field.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+import { extractDependencyEffect } from '../utils/prop-dependency-utils';
+
+type ObjectPropFieldProps = React.PropsWithChildren< {
+	bind: string;
+	shape: PropsSchema;
+	settings: Props;
+} >;
+
+export const ObjectPropField = ( { bind, shape, settings, children }: ObjectPropFieldProps ) => {
```
IshayMaya: is it only object prop? cannot be in an array?

packages/packages/core/editor-editing-panel/src/controls/object-section-control.tsx
```diff
@@ -0,0 +1,86 @@
...
```
IshayMaya: does it not creating ui issues with stuff that register to the control apis ? (replacement actions etc.)

packages/packages/core/editor-editing-panel/src/hooks/use-transformable-object-prop.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export function useTransformableObjectProp() {
+	const {
+		propType: parentPropType,
+		value,
+		setValue,
+		...boundPropContext
+	} = useBoundProp< ObjectPropValue, ObjectPropType >();
+
+	const envelope = isTransformable( value ) ? value : null;
```
IshayMaya: why isTransformable? should be types from the prop type util

packages/packages/core/editor-editing-panel/src/hooks/use-transformable-object-prop.ts
```diff
@@ -0,0 +1,56 @@
...
+	const objectPropType = resolveObjectPropType( parentPropType, envelope?.$$type ?? '' );
+	const envelopeTypeKey = envelope?.$$type ?? objectPropType.key;
+	const innerValue = ( envelope?.value ?? {} ) as Props;
```
IshayMaya: feels too hard-coded.
check we cant we use get the extracted values and types from our utils like we usually do

packages/packages/core/editor-editing-panel/src/init.ts
```diff
@@ -28,6 +29,7 @@ export function init() {
 
 	// TODO: Move it from here once we have element-controls package.
 	registerElementControls();
+	registerObjectSectionControl();
```
IshayMaya: maybe we should address it from the sections and not from the controls.
meaning - a section that is aware to the schema (bind to) instead of control that behave as a section

PR #36015, opened 2026-05-26
Internal: Add query content search support [ED-24286]

IshayMaya:
## 1. Problem & Context

Adds query-time content search capability to Post_Query, allowing searches beyond post titles to include content and excerpts via optional `search_in_content` parameter (ED-24286).

## 2. What Changed (Where)

- `post-query-builder.php`: Reformatted query params array with proper alignment; passes through new `SEARCH_IN_CONTENT_KEY`
- `post-query.php`: New `SEARCH_IN_CONTENT_KEY` constant; conditional SQL injection of `post_content` and `post_excerpt` LIKE clauses; added to API param spec and allowed keys list
- Test files: Two new test cases validating content search behavior on/off

## 3. How It Works

When `search_in_content=true`, `customize_post_query()` conditionally appends `OR post_content LIKE` and `OR post_excerpt LIKE` clauses to the existing title search. Parameter flows from REST endpoint → query builder → `build_query_args()` → WP_Query filter hook. Default is `false` (backwards compatible).

## 4. Risks

SQL injection via `esc_sql()` is already applied to `$term` before interpolation—mitigation sufficient. Consider if `post_excerpt` should have higher weight than content in future ranking; no performance concerns for typical excerpt sizes.

### DIFF
--- a/modules/wp-rest/classes/post-query.php
+++ b/modules/wp-rest/classes/post-query.php
@@ -13,6 +13,7 @@ class Post_Query extends Base {
 	const ENDPOINT = 'post';
 	const SEARCH_FILTER_ACCEPTED_ARGS = 2;
 	const DEFAULT_FORBIDDEN_POST_TYPES = [ 'e-floating-buttons', 'e-landing-page', 'elementor_library', 'attachment', 'revision', 'nav_menu_item', 'custom_css', 'customize_changeset' ];
+	const SEARCH_IN_CONTENT_KEY = 'search_in_content';
 
 	/**
 	 * @param string    $search_term The original search query.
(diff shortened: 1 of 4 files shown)


modules/wp-rest/classes/post-query.php
```diff
@@ -25,8 +26,13 @@ public function customize_post_query( string $search_term, \WP_Query $wp_query )
 
 		if ( $is_custom_search && ! empty( $term ) ) {
 			$escaped = esc_sql( $term );
+			$search_in_content = $wp_query->get( self::SEARCH_IN_CONTENT_KEY ) ?? false;
 			$search_term .= ' AND (';
 			$search_term .= "post_title LIKE '%{$escaped}%'";
+			if ( $search_in_content ) {
```
IshayMaya, line 32: was not sure if we want to for the other queries as well

Omerisra6 (dismissed):

Omerisra6 (approved):

PR #36039, opened 2026-05-27
Fix: Control value of empty exposed further override in nested component [ED-24264]

MiryamOren:
[The bug](https://elementor.atlassian.net/browse/ED-24264?atlOrigin=eyJpIjoiOTY4ODAwNmFhY2ZkNDdjN2EyNzIzOWNmNmEzNWFmMzEiLCJwIjoiaiJ9)

When exposing further empty override:

| Before | After |
|--------|-------|
| The override's control shows default value (like "This is a title") as the value | The override's control shows resolved origin value as a placeholder |
| image | image |

## 1. Problem & Context

ED-24264: When exposing nested component overrides with empty values on dynamic inherited props, the control was incorrectly showing the default value instead of respecting the empty override. The fix removes `matchingOverride` parameter and derives the logic directly from `overrideValue`.

## 2. What Changed (Where)

- `override-prop-control.tsx`: Removed `matchingOverride` parameter from `resolveOverrideValues()` call and function signature; refactored baseValue logic to check `overrideValue` directly instead.

## 3. How It Works

The `resolveOverrideValues()` function now computes `shouldUseInheritedAsValue` based on whether the override is explicitly empty (`!overrideValue`) combined with inherited dynamic status. The `baseValue` calculation inverts the condition (`overrideValue ||`) to properly return `null` when an override exists (empty or not) or inherited is dynamic, avoiding fallback to inherited value when an explicit empty override is set.

## 4. Risks

None identified—this directly fixes the logical error where `matchingOverride` presence was conflated with `overrideValue` presence. Removal clarifies intent and eliminates redundant parameter passing.


packages/packages/core/editor-components/src/components/instance-editing-panel/override-prop-control.tsx
```diff
@@ -235,10 +233,10 @@ function resolveOverrideValues(
 	const inheritedValue = unwrappedSettings[ propKey ] ?? null;
 	const isInheritedDynamic = isDynamicPropValue( inheritedValue );
 
-	const shouldUseInheritedAsValue = isInheritedDynamic && ! matchingOverride;
+	const shouldUseInheritedAsValue = isInheritedDynamic && ! overrideValue;
 
 	const propValue = shouldUseInheritedAsValue ? inheritedValue : overrideValue;
-	const baseValue = matchingOverride || isInheritedDynamic ? null : inheritedValue;
+	const baseValue = overrideValue || isInheritedDynamic ? null : inheritedValue;
```
IshayMaya, line 239: does it mean we're passing the base value forward now?

IshayMaya (dismissed):

IshayMaya (approved):

PR #36046, opened 2026-05-28
Internal:  add registerElementPanelDefaults API [ED-24331]

IshayMaya:
## 1. Problem & Context

Adds a public API for third-party code to register custom element panel defaults, replacing the previous React Context pattern with a simple registration mechanism (ED-24331).

## 2. What Changed (Where)

- `use-default-panel-settings.ts`: Converted `defaultPanelSettingsContext` from a React Context to a plain object; exported `registerElementPanelDefaults()` function and `ElementPanelDefaults` type.
- `index.ts`: Exposed the new registration API and type as public exports.

## 3. How It Works

`registerElementPanelDefaults(type, defaults)` mutates the `elementPanelDefaults` map; `useDefaultPanelSettings()` performs direct lookup instead of `useContext()`, returning the registered defaults or fallback. Registration happens at module load time, decoupling consumer logic from React Context infrastructure.

## 4. Risks

Mutation-based registry can cause ordering issues if plugins register conflicting types — consider documenting precedence or adding conflict detection.


packages/packages/core/editor-editing-panel/src/hooks/use-default-panel-settings.ts
```diff
@@ -32,10 +32,13 @@ const defaultPanelSettingsContext = createContext< Record< string, Defaults | un
 		defaultSectionsExpanded: fallbackEditorSettings.defaultSectionsExpanded,
 		defaultTab: 'style',
 	},
-} );
+};
+
+export function registerElementPanelDefaults( type: string, defaults: Defaults ): void {
```
Omerisra6, line 37: Let's make the api better and more self explained:
`elementDefaultPanel( 'loop' )->tab('styles')->section('grid');`

packages/packages/core/editor-editing-panel/src/hooks/use-default-panel-settings.ts
```diff
@@ -32,10 +32,13 @@ const defaultPanelSettingsContext = createContext< Record< string, Defaults | un
...
+	elementPanelDefaults[ type ] = defaults;
+}
 
 export const useDefaultPanelSettings = () => {
 	const { element } = useElement();
-	const defaults = useContext( defaultPanelSettingsContext )[ element.type ];
```
Omerisra6, line 39: no one used the context?

IshayMaya: no, there's no context provider usage.
we can rely on the element's reactivity here as the source of trhuth

Omerisra6 (approved):

PR #36062, opened 2026-05-31
Feature: Support Dynamic in LLM flow [ED-24275]

eavichay-elementor:
## 1. Problem & Context

LLM integration needs to support `dynamic` prop types with bindable content tags (ED-24275). Currently, prop validation and conversion only handled static values; now adds dialect layer to bridge LLM `bindTo` wire format ↔ canonical dynamic PropValues with fallbacks.

## 2. What Changed (Where)

| Module | Change |
|--------|--------|
| **llm-dialect/** | New adapter registry system: `dynamic`, `html-v3`, `image-src`, `size`, `overridable` adapters handle bidirectional conversion between dialect (LLM wire) and canonical forms |
| **editor-props** | Exports `propFromLlm`/`propToLlm` (uses walk.ts tree traversal), `validateLlmJson` (jsonschema validation), `initLlmDialect` for setup |
| **do-update-element-property.ts** | Changed validation from `validatePropValue` → `validateLlmJson`; resolution from `adjustLlmPropValueSchema` → `propFromLlm` with propType context |
| **composition-builder.ts** | Added prop schema validation before applying config; validates against tagName's atomic_props_schema |
| **get-element-config/tool.ts** | Now converts raw element settings to LLM dialect via `propToLlm` for resource export |
| **dynamic-tags-resource.ts** | New MCP resource exposing available dynamic tags by category; consumed by LLM for bindTo completions |
| **Tests** | Round-trip tests for dialect adapters, image/html-v3/size nested conversions, LLM validation |

## 3. How It Works

**Dialect flow:** `propFromLlm(value, {propType})` → `walkPropValueWithPropType(..., 'toProp')` → adapter chain converts bindTo markers & nested structures → canonical PropValue.
Reverse: `propToLlm(value, {propType})` → walk direction='toDialect' → expand fallbacks, add bindTo markers for LLM consumption.

**Schema generation:** `propTypeToLlmJsonSchema` passes `LLMDialectAdapter` through recursion; adapters inject `allowBind`/`bindTo` schema properties for dynamic unions, strip binding from nested contexts (html-v3 content, image-src url).

**Adapters handle:** `dynamic` (union branch replacement + fallback reconciliation), `html-v3` (flatten/expand content bindTo), `image-src` (hoist nested url bindTo), `size` (compact/expand unit/size fields), `overridable` (skip schema).

Edge cases: nested bindTo suppression (prevent deep binding), fallback type reconciliation (e.g., control type lookup for dynamic tag), default value coercion (empty image-src → default).

## 4. Risks

- **Silent schema mismatches:** `propTypeToLlmJsonSchema` differs from canonical schema via dialect; LLM sees different structure. Mitigated by explicit validation call + test coverage.
- **Adapter ordering:** Registry order matters (dynamic before html-v3); duplicate registration throws. Mitigation: init guards (`builtInAdaptersRegistered` flag), tests verify order.
- **Fallback type safety:** Dynamic tags reference atomic_controls for fallback control type; missing metadata silently degrades (strips fallback). Acceptable for MVP; should add warnings if tag not found.
- **Performance:** Walk recursion on every conversion; no caching. Acceptable volume; propTypes are typically shallow.


packages/packages/libs/editor-props/src/llm-dialect/dynamic-tag-metadata-registry.ts
```diff
@@ -0,0 +1,56 @@
...
+		return;
+	}
+
+	const resolvedTags = dynamicTagsResolver();
+
+	if ( resolvedTags ) {
+		mergeDynamicTags( resolvedTags );
+	}
+
+	dynamicTagsResolved = true;
```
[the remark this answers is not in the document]
eavichay-elementor: Initialization is lazy, happens after dynamics module is loaded.

packages/packages/core/editor-canvas/src/composition-builder/composition-builder.ts
```diff
@@ -253,13 +261,15 @@ export class CompositionBuilder {
 				}
 			}
 
-			const customCSS = this.elementCustomCSS[ configId ];
-			if ( customCSS ) {
+			const intentionCss = typeof styleConfig?.$intention === 'string' ? styleConfig.$intention.trim() : '';
+			const fallbackCss = hasInvalidStyles && intentionCss ? intentionCss : '';
```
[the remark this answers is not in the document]
eavichay-elementor: $intention testing is missing because tool testing infra is not there yet. Under jest, all mcp's are not being initialized.

packages/packages/core/editor-canvas/src/composition-builder/composition-builder.ts
```diff
@@ -209,6 +209,13 @@ export class CompositionBuilder {
 			const config = this.elementConfig[ configId ];
 			if ( config ) {
 				for ( const [ propertyName, propertyValue ] of Object.entries( config ) ) {
+					const { valid, errors: validationErrors } = validateInput.validatePropSchema( node.tagName, {
```
eavichay-elementor, line 212: Validate and exit early

packages/packages/core/editor-canvas/src/composition-builder/composition-builder.ts
```diff
@@ -12,6 +12,7 @@ import {
 import { type z } from '@elementor/schema';
 
 import { doUpdateElementProperty } from '../mcp/utils/do-update-element-property';
+import { mergeCustomCssText } from '../mcp/utils/merge-custom-css';
```
eavichay-elementor, line 15: Bug fix "on the way" -> the error recovery in the tool for  bad schemas using the "$intent" was DELETING intentional custom CSS

packages/packages/core/editor-canvas/src/composition-builder/composition-builder.ts
```diff
@@ -223,17 +230,18 @@ export class CompositionBuilder {
 			}
 
 			const styleConfig = this.elementStylesConfig[ configId ];
+			let hasInvalidStyles = false;
 			if ( styleConfig ) {
 				const validStylesPropValues: Record< string, AnyValue > = {};
 				for ( const [ styleName, stylePropValue ] of Object.entries( styleConfig ) ) {
+					if ( styleName === '$intention' ) {
```
eavichay-elementor, line 237: Sometime angie thinks "intention" should be top level property

packages/packages/core/editor-canvas/src/mcp/resources/build-dynamic-tags-resource-payload.ts
```diff
@@ -0,0 +1,99 @@
+type DynamicTagSource = {
```
eavichay-elementor: Resource is required for AI to understand what can be bound and which values allowed

packages/packages/core/editor-canvas/src/mcp/resources/widgets-schema-resource.ts
```diff
@@ -119,7 +119,7 @@ Variables from the user context ARE NOT SUPPORTED AND WILL RESOLVE IN ERROR.
 			if ( ! stylesSchema ) {
 				throw new Error( `No styles schema found for category: ${ category }` );
 			}
-			const asJson = Schema.propTypeToJsonSchema( stylesSchema as PropType );
+			const asJson = Schema.propTypeToLlmJsonSchema( stylesSchema as PropType );
```
eavichay-elementor, line 122: Semantic change - to work with the actual LLM dialect

packages/packages/libs/editor-props/src/llm-dialect/adapters/html-v3.ts
```diff
@@ -0,0 +1,132 @@
+import { stringPropTypeUtil, type StringPropValue } from '../../prop-types';
```
eavichay-elementor, line 1: "Special" prop type needs a special adapter

packages/packages/libs/editor-props/src/llm-dialect/adapters/image-src.ts
```diff
@@ -0,0 +1,285 @@
+import { type PropType, type PropValue } from '../../types';
```
eavichay-elementor: The image/image-src combo breaks the contract of the propType schema. fallback is treated via custom transformers - out of this library's reach.

packages/packages/libs/editor-props/src/llm-dialect/adapters/overridable.ts
```diff
@@ -0,0 +1,8 @@
+import { type PropDialectAdapter } from '../registry';
```
eavichay-elementor: This one ensures LLM are not aware of overridables. Once ready - implementation details goes here.

packages/packages/libs/editor-props/src/llm-dialect/adapters/size.ts
```diff
@@ -0,0 +1,85 @@
+import { type PropType, type TransformablePropValue } from '../../types';
```
eavichay-elementor, line 1: Flattened/Expanded conversion

packages/packages/core/editor-canvas/src/mcp/tools/build-composition/tool.ts
Omerisra6: move to a different PR

packages/packages/libs/editor-props/src/llm-dialect/adapters/overridable.ts
Omerisra6: This adapter is dead code. In `utils/props-to-llm-schema.ts` line 105 we already have the hardcoded skip:

```
if ( typeKey === 'overridable' || ( typeKey === 'dynamic' && dialect ) ) { continue; }
```

The `continue` short-circuits before any adapter ever runs on the `overridable` branch — so registering `overridableLlmDialectAdapter` changes nothing observable. The PR comment "Once ready - implementation details goes here" is speculative.

Suggest deleting this file and the corresponding line in `register-built-in-adapters.ts`. If we ever do need overridable-specific behavior, we add it then; right now the file is a phantom extension point.

packages/packages/libs/editor-props/src/llm-dialect/adapters/image-src.ts
```diff
@@ -0,0 +1,285 @@
+import { type AnyTransformable, type PropType, type PropValue } from '../../types';
```
Omerisra6, line 1: This is fixing a bug in the wrong layer. The root cause is in `modules/atomic-widgets/prop-types/image-src-prop-type.php`:

```php
protected function define_shape(): array {
    return [
        'id'  => Image_Attachment_Id_Prop_Type::make()->...,
        'url' => Url_Prop_Type::make(),
        'alt' => String_Prop_Type::make()->...,
    ];
}

protected function validate_value( $value ): bool {
    $has_id  = ! empty( $value['id'] );
    $has_url = ! empty( $value['url'] );
    return ( $has_id xor $has_url ) && parent::validate_value( $value );
}
```

`define_shape()` says "id, url, alt are siblings" but `validate_value` enforces XOR(id, url). The schema lies to **every** consumer (LLM, editor, third-party tools) — it's not an LLM-dialect problem. You acknowledge this yourself in the PR comment: *"the image/image-src combo breaks the contract of the propType schema"*.

**Proposal:** fix it in PHP by emitting a real `oneOf`/two-branch shape from `Image_Src_Prop_Type` (e.g. branch A = `{ id, alt? }`, branch B = `{ url, alt? }`). That single PHP change makes the entire 285-line `image-src.ts` adapter — `coerceToDefaultWhenNoValidIdOrUrl`, `hoistImageSrcDynamicChild`, `migrateImageSrc`, `ensureAltNull`, `findDynamicChildField`, `withoutDynamicFallback`, `extractImageSrcForDialect`, `wrapImageSrcInImage`, all of it — disappear. The TS side just trusts the schema.

Happy to take the PHP change as a prerequisite ticket so this PR doesn't grow, but I don't think we should land 285 lines of TS coercion that we know is papering over a wrong PHP schema.

packages/packages/libs/editor-props/src/llm-dialect/adapters/html-v3.ts
```diff
@@ -0,0 +1,132 @@
...
+
+	return {
+		$$type: 'html-v3',
+		value: {
+			content: structuredClone( fallback ),
+			children: [],
+		},
+	} as PropValue;
+};
+
```
Omerisra6, line 74: This adapter is mostly redundant with logic already in `dynamic.ts`. In `adapters/dynamic.ts` lines 153–172, `findControlTypeByBind(tag?.atomic_controls, 'fallback')` already detects whether the dynamic-tag's `fallback` control is a `text` (string) or something else, and `reconcileFallbackToProp` / `reconcileFallbackToDialect` (lines 210, 230) already shape the fallback accordingly.

That generic mechanism subsumes 90% of what `html-v3.ts` does. Specifically:

- **`toPropValue` branch at lines 101–123** — "if the union has html-v3 and the value is dynamic, ensure fallback is a string": this is exactly what `reconcileFallbackToProp` does generically when the dynamic tag's `fallback` control type is `text`. The html-v3 widgets that use this union (atomic-paragraph, atomic-button, atomic-tab, atomic-form fields) all bind to dynamic tags whose `fallback` control is `text`. So the generic path already handles it.
- **`toDialectValue` at lines 125–131** + the `stringBindToToHtmlV3Wire` / `dynamicStringFallbackToHtmlV3Wire` helpers — same story; the `dynamic` adapter's outbound path produces `{$$type: 'string', value, bindTo, allowBind}` which is exactly the canonical wire shape. There's no actual html-v3-specific reshaping needed.
- **`toDialectSchema` at lines 78–84** (the `stripDynamicBinding` for `content` shape key) — this is the only piece that's html-v3-specific (preventing nested `bindTo` on the inner `content` field). But this is a *property of the html-v3 propType* (its `content` field is non-bindable), not a property of the dialect. It belongs as metadata on the propType definition (e.g. `String_Prop_Type::make()->non_bindable()` in PHP, or a `meta.allowBind: false` flag), where the schema generator picks it up generically — same as how we already handle `nonConfigurablePropKeys` in `props-to-llm-schema.ts`.

**Net:** delete this adapter. Lean on `findControlTypeByBind` for the fallback shape. Add a `non_bindable` (or `meta.allowBind: false`) flag for html-v3's `content` field if not already there.

packages/packages/libs/editor-props/src/llm-dialect/registry.ts
```diff
@@ -0,0 +1,95 @@
...
+export type PropDialectAdapter = {
+	id: string;
+	matches: ( ctx: PropDialectContext ) => boolean;
+	toDialectSchema?: ( schema: JsonSchema7, ctx: PropDialectContext ) => JsonSchema7 | LlmDialectSkip;
+	toPropValue?: ( value: AnyTransformable, ctx: PropDialectContext ) => PropValue | LlmDialectSkip;
+	toDialectValue?: ( value: AnyTransformable, ctx: PropDialectContext ) => PropValue | LlmDialectSkip;
+};
+
+type SchemaFinalizeHook = ( schema: JsonSchema7 ) => JsonSchema7;
+
```
Omerisra6, line 26: Once `overridable`, `image-src`, and `html-v3` adapters go away, we have exactly two remaining transforms: `dynamic` (the `bindTo` ↔ canonical conversion) and `size` (the unit/size compaction). Neither needs a registry:

- `dynamic` is the *one* truly cross-cutting transform — it applies to any union containing `dynamic`. Inline its logic directly into `walk.ts` (or keep it as one module imported by the walker), no registry needed.
- `size` is a per-propType shape mismatch (PHP wire is nested, TS canonical is flat — see the `NOTE` in `prop-types/size.ts` line 7). It belongs as `toLlm` / `fromLlm` methods on `sizePropTypeUtil`, co-located with the schema definition. The walker dispatches by `$$type` using the existing `getPropSchemaFromCache` infrastructure — no separate registry.

Removing the registry also removes: `registerLlmDialectAdapter`, `registerLlmDialectSchemaFinalize`, `getLlmDialectAdapters`, `getMatchingLlmDialectAdapters`, `applyLlmDialectSchema`, `applyLlmDialectPropValue`, `finalizeLlmDialectSchema`, `resetLlmDialectRegistryForTests`, `ensureBuiltInLlmDialectAdapters`, `resetBuiltInLlmDialectAdaptersForTests`, the duplicate-id check, the registration-order risk you flagged in the PR description, and the `LLM_DIALECT_SKIP` sentinel.

The trade-off is: editor-props core becomes aware of `dynamic`-handling and prop-types co-locate their LLM hooks. I think that's the right trade given there are exactly 2 real cases — coupling beats speculative-extensibility complexity here.

packages/packages/libs/editor-props/src/utils/props-to-llm-schema.ts
```diff
@@ -72,25 +90,22 @@ function convertPlainPropType(
...
+): JsonSchema7 {
 	const schema = structuredClone( baseSchema );
 
 	const propTypes = propType.prop_types || {};
 	const schemas: JsonSchema7[] = [];
 
-	// Convert each prop type in the union
 	for ( const [ typeKey, subPropType ] of Object.entries( propTypes ) ) {
-		if ( typeKey === 'dynamic' || typeKey === 'overridable' ) {
+		if ( typeKey === 'overridable' || ( typeKey === 'dynamic' && dialect ) ) {
```
Omerisra6, line 105: Heads-up: the new `overridableLlmDialectAdapter` is registered to handle exactly this case via `LLM_DIALECT_SKIP`, but this hardcoded `continue` runs first inside the union loop and short-circuits before adapters are consulted. Two paths exist for the same behavior.

Pick one: either delete the adapter (my preference — see comment on `adapters/overridable.ts`) or remove this hardcoded skip and let the adapter handle it. Having both is confusing.

packages/packages/libs/editor-props/src/utils/adjust-llm-prop-value-schema.ts
```diff
@@ -1,98 +1,55 @@
-import { numberPropTypeUtil, type NumberPropValue, stringPropTypeUtil, type StringPropValue } from '../prop-types';
-import { type ObjectPropValue, type PropValue, type TransformablePropValue } from '../types';
-
-const ensureNotNull = ( v: unknown, fallback: unknown ) => ( v === null ? fallback : v );
+import { type PropValue, type TransformablePropValue } from '../types';
```
Omerisra6, line 1: Great cleanup here. Worth renaming the file too — it's now just `prop-value-helpers.ts` or split into `apply-global-variable-resolvers.ts` + `strip-intention.ts`. The old `adjust-llm-prop-value-schema` name describes behavior that no longer lives here.

packages/packages/libs/editor-props/src/llm-dialect/dynamic-tag-metadata-registry.ts
```diff
@@ -0,0 +1,56 @@
...
+let dynamicTagsResolved = false;
+
+const mergeDynamicTags = ( tags: Record< string, LlmDynamicTagMetadata > ) => {
+	dynamicTagsRegistry = {
+		...dynamicTagsRegistry,
+		...tags,
+	};
+};
+
+const ensureDynamicTagsLoaded = () => {
```
Omerisra6, line 18: Already flagged by the auto-reviewer (dynamic tag registry race); sanity-check the fix is in place. The `dynamicTagsResolved` flag + `setLlmDialectDynamicTagsResolver` resetting it to `false` looks like the right direction — just confirming this is addressed before merge. Not blocking.

docs/v4/PropTypes.kb.md
Omerisra6, line 1: move to a different PR

packages/packages/core/editor-canvas/src/composition-builder/composition-builder.ts
Omerisra6, line 15: move to a different PR

packages/packages/core/editor-canvas/src/mcp/resources/build-dynamic-tags-resource-payload.ts
Omerisra6, line 1: I think you can add that config on the PHP side, we already manipulate the V3 data there (in `Dynamic_Prop_Types_Mapping`), and I think the original shape is as you need here

packages/packages/core/editor-canvas/src/mcp/utils/do-update-element-property.ts
Omerisra6, line 1: move to a different PR

packages/packages/libs/editor-props/src/llm-dialect/__tests__/cleanup-llm-json-schema.test.ts
Omerisra6, line 1: please create one test suite with the flows you want to cover instead of multiple unit tests

packages/packages/libs/editor-props/src/llm-dialect/adapters/overridable.ts
eavichay-elementor: That should have been removed from the code (the if-else). The "overridable" adapter SHOULD be there to hide it from LLMs.

packages/packages/libs/editor-props/src/llm-dialect/adapters/image-src.ts
eavichay-elementor: Modifying the PHP can have unknown side effects.

packages/packages/libs/editor-props/src/llm-dialect/adapters/html-v3.ts
eavichay-elementor: Try locally - don't change the tests. See what happens.

packages/packages/libs/editor-props/src/llm-dialect/registry.ts
eavichay-elementor: Size adapter is required. The propType parses size to "value:object{unit:stringPropType,size:numberPropType|stringPropType}" full shape but in-code "performance improvements" reduce the shape to simple object with two strings. The value-conversion must remain or the apps break - or trade it with schema-conversion. Either way - the adaptor needs to stay.

packages/packages/core/editor-canvas/src/mcp/resources/build-dynamic-tags-resource-payload.ts
eavichay-elementor: Will look into it - might be a good point

packages/packages/libs/editor-props/src/llm-dialect/__tests__/cleanup-llm-json-schema.test.ts
eavichay-elementor: There's is a "roundtrip" test

packages/packages/core/editor-canvas/src/composition-builder/composition-builder.ts
```diff
@@ -209,6 +209,13 @@ export class CompositionBuilder {
 			const config = this.elementConfig[ configId ];
 			if ( config ) {
 				for ( const [ propertyName, propertyValue ] of Object.entries( config ) ) {
+					const { valid, errors: validationErrors } = validateInput.validatePropSchema( node.tagName, {
+						[ propertyName ]: propertyValue,
+					} );
+					if ( ! valid ) {
+						configErrors.push( ...( validationErrors || [] ) );
```
Omerisra6, line 216: Wouldn't it override previous errors for the same key?

packages/packages/core/editor-canvas/src/mcp/resources/build-dynamic-tags-resource-payload.ts
Omerisra6: Did you check?

packages/packages/core/editor-canvas/src/mcp/utils/do-update-element-property.ts
Omerisra6: What about that?

packages/packages/libs/editor-props/src/llm-dialect/adapters/dynamic.ts
```diff
@@ -0,0 +1,249 @@
+import { type PropType, type PropValue, type TransformablePropType, type TransformablePropValue } from '../../types';
```
Omerisra6, line 1: Please explain why we need special treatment for each one of the cases.

packages/packages/libs/editor-props/src/llm-dialect/adapters/dynamic.ts
Omerisra6, line 1: move the dialect into a different PR

PR #36066, opened 2026-05-31
Internal: Global classes order to have preview context [ED-24345]

ronros-elementor:
## 1. Problem & Context

Global classes need separate preview and frontend orderings to support divergent class priority states across draft/published contexts. Currently, both contexts share a single order, causing published pages to break when preview-only edits reorder classes.

## 2. What Changed (Where)

| File | Change |
|------|--------|
| `global-classes-order.php` | Added `Has_Preview_Context` trait; dual meta keys (`META_KEY`, `META_KEY_PREVIEW`); context-aware get/set via `get_context_key()` |
| `global-classes-repository.php` | All order operations now call `.set_preview()` before get/set; non-preview updates now sync to preview context too |
| `global-classes-labels.php` | Order retrieval respects preview context |
| `atomic-global-styles.php` | Cache invalidation includes `affected_post_ids` for granular updates |
| Export/import/migration files | All callsites updated to `.set_preview(false)` for frontend order |
| REST API, tests, Playwright E2E | Order validation relaxed (non-existent IDs stripped, missing IDs appended); robust test coverage for order divergence |

## 3. How It Works

Order operations now branch by context: `Global_Classes_Order::make($kit)->set_preview($is_preview)->get_order()` reads from either `_elementor_global_classes_order` (frontend) or `_elementor_global_classes_order_preview` (preview). On non-preview saves, both contexts are updated to maintain eventual consistency. Preview falls back to frontend order if empty (migration safety). Cache invalidation now includes affected post IDs to trigger style recomputation on dependent documents.

## 4. Risks

**None significant.** Fallback logic ensures preview orders never permanently diverge from frontend. REST API changes (stripping/appending order IDs) are backward-compatible and defensive. E2E tests validate cross-page consistency under reorder/rename/delete scenarios. Only risk: existing code calling `get_order()` without `.set_preview()` defaults to frontend—mitigated by exhaustive callsite updates.


packages/packages/core/editor-global-classes/src/components/class-manager/global-classes-list.tsx
```diff
@@ -133,14 +134,15 @@ export const GlobalClassesList = ( {
...
-										renameClass={ ( newLabel: string ) => {
+										renameClass={ async ( newLabel: string ) => {
 											trackGlobalClasses( {
 												event: 'classRenamed',
 												classId: cssClass.id,
 												oldValue: cssClass.label,
 												newValue: newLabel,
 												source: 'class-manager',
 											} );
+											await loadExistingClasses( [ cssClass.id ] );
```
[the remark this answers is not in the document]
ronros-elementor: The goal is to make ensure the renamed class is loaded, before updating the store, otherwise it would be considered as a new style instead of an existing one

modules/global-classes/import-export-customization/runners/export.php
```diff
@@ -92,7 +92,7 @@ static function ( array $class_data ) use ( &$files, &$labels_by_id ) {
 	}
 
 	private function build_order_entries( $kit, array $labels_by_id ): array {
-		$repository_order = Global_Classes_Order::make( $kit )->get_order();
+		$repository_order = Global_Classes_Order::make( $kit )->set_preview( false )->get_order();
```
IshayMaya, line 95: might be too much but maybe we can have a function or some more clear separation logic that make sure all those places has preview(false) ? just thinking if someone will need to add some more functionality

modules/global-classes/atomic-global-styles.php
```diff
@@ -128,6 +128,12 @@ private function invalidate_cache_for_updated_classes( string $context, array $c
 		$document_ids = [];
 		$is_preview = Global_Classes_Repository::CONTEXT_PREVIEW === $context;
 
+		if ( ! empty( $changes['affected_post_ids'] ) ) {
```
IshayMaya, line 131: this is the param for order change?

packages/packages/core/editor-global-classes/src/components/class-manager/class-manager-panel.tsx
```diff
@@ -291,7 +291,9 @@ const usePreventUnload = () => {
 
 const usePublish = () => {
 	return useMutation( {
-		mutationFn: () => saveGlobalClasses( { context: 'frontend' } ),
+		mutationFn: async () => {
```
IshayMaya: revert to just return the function return value

packages/packages/core/editor-global-classes/src/components/class-manager/global-classes-list.tsx
```diff
@@ -133,14 +134,15 @@ export const GlobalClassesList = ( {
 									<ClassItem
 										id={ cssClass.id }
 										label={ cssClass.label }
-										renameClass={ ( newLabel: string ) => {
+										renameClass={ async ( newLabel: string ) => {
```
IshayMaya, line 150: should we have some loading state?

tests/playwright/sanity/modules/v4-tests/global-classes/utils.ts
```diff
@@ -113,6 +113,8 @@ export async function openClassManager( page: Page, editor: EditorPage, divBlock
 	}
 
 	await dismissClassManagerIntro( page );
+
+	// Await page.locator( 'li[role="listitem"]' ).first().waitFor( { state: 'visible', timeout: timeouts.expect } );
```
IshayMaya: leftovers?

modules/global-classes/atomic-global-styles.php
ronros-elementor: actually yet another thing noticed while working on this fix, was the potential miss on cache invalidation when deleting styles
I prefer having as many of these done together solely because it should also be cherry-picked as a patch

modules/global-classes/import-export-customization/runners/export.php
ronros-elementor: I'm not sure I follow, like to throw an error if this method was not called?
basically this is anyway the default behavior, preview is considered false if was not explicitly clarified, but in this case I just wanted to highlight the fact I only take the frontend version

what did you have in mind?

packages/packages/core/editor-global-classes/src/components/class-manager/global-classes-list.tsx
```diff
@@ -133,25 +147,34 @@ export const GlobalClassesList = ( {
...
+														},
+													} )
+												);
+											} finally {
+												removeLoadingClass( cssClass.id );
+											}
 										} }
 										selected={ isDragged }
-										disabled={ disabled || isDragPlaceholder }
+										disabled={ disabled || isDragPlaceholder || loading[ cssClass.id ] }
```
IshayMaya, line 177: since the rename class returns the promise we can handle it from there, will be cleaner IMO but can be done in a separated pr since we want to merge it

IshayMaya (approved):

PR #36074, opened 2026-06-01
Internal: Add element to resolver context [ED-24051]

Omerisra6:
## Summary
- Adds optional bind on atomic Section controls and renders bound sections through SettingsField in the editing panel.
- Resolves object shape from union-bound section props and scopes dependency evaluation for nested controls.
- Inlines bound section rendering into SettingsTab / SettingsControl, removing BoundSettingsSection, TransformableObjectBridge, and object-section plumbing.

## Test plan
- [ ] With a build that includes bound sections, open an atomic widget with a bound query section and verify child controls render.
- [ ] Change a union-bound query prop and confirm nested fields stay editable.
- [ ] Run cd packages && npm test -- --testPathPattern=editor-editing-panel

**Related:** https://github.com/elementor/elementor-pro/pull/7025

## 1. Problem & Context

Resolver context needs access to the element instance during prop transformation (ED-24051). Currently, transformers receive only key, disabled state, and prop type—insufficient for element-aware transformations.

## 2. What Changed (Where)

| File | Change |
|------|--------|
| `Props_Resolver_Context` | Added `$options` array field with getter/setter for passing arbitrary context data |
| `Props_Resolver` abstract | Updated method signatures to accept `array $options` parameter through resolve chain |
| `Render_Props_Resolver`, `Import_Export_Props_Resolver` | Threaded `$options` through `resolve()` and `resolve_item()` calls |
| `Has_Atomic_Base` | Pass `[ 'element' => $this ]` when calling resolver |
| Tests | New unit test validates options storage and fluent setter chaining |
| TypeScript exports | Added `ControlLabel` export (unrelated cleanup) |

## 3. How It Works

Options propagate through the resolver chain: `resolve($schema, $props, ['element' => $this])` → `resolve_item()` → `transform()` → `Props_Resolver_Context::set_options()`. Transformers access via `$context->get_options()['element']` during `transform()` execution. Backward compatible—defaults to empty array.

## 4. Risks

None identified. Additive changes with default empty array; existing callers unaffected. Test coverage validates round-trip behavior.


packages/packages/core/editor-editing-panel/src/utils/prop-dependency-utils.ts
```diff
@@ -6,9 +6,11 @@ import {
 	isDependencyMet,
```
IshayMaya, line 6: This is the only file im not sure about here, Iwould expect that we will get this behavior for free.
What we have in the infra today is a synthetic top-level object schema that mimic an object prop type that wrap all the element settings, and the deps are handled on its level via the settings field.
now in query control we want to introduce the same thing but with an actual prop type,should'nt it work with the existing infra if we render the correct tree?

PR #36075, opened 2026-06-01
Fix: Font Family option doesn't work on Custom Font names with spaces/special characters [ED-24110]

nami-p:
# Font-family in atomic/v4 styles

## Summary

Atomic/v4 styles treated `font-family` as a plain string end-to-end. That broke two separate concerns: valid CSS output and correct font asset loading. This work introduces a dedicated `font-family` prop type and separates **CSS transformation** from **font enqueue** in the style render pipeline.

---

### 1. `font-family` is not a plain string

In v4 atomic styles, `font-family` was stored and emitted like any other string prop (e.g. `{ $$type: 'string', value: 'Open Sans' }` or raw concatenation into CSS). That is incorrect because font names with spaces or special characters need different handling at render time than a generic string value.

v1 typography already avoids this: it stores the bare catalog key (`Open Sans`) and applies quoting only in the CSS template (`font-family: "{{VALUE}}", sans-serif;`). Quoting in stored JSON breaks `enqueueFont` / `elementor.config.controls.font.options` lookup, which expects the bare catalog key.

### 2. Style renderer conflated CSS output with enqueue input

The style render pipeline has a single transform path: stored prop bag → CSS string. Font loading is a separate concern — it needs the **stored catalog key**, not the CSS-formatted value.

Using the same transformed output for both caused breakage: fixing CSS (e.g. quoting) would break enqueue lookup, and using the raw CSS string for enqueue would fail against the font catalog.

---

## Solution

Split responsibilities into three layers, matching the v1 mental model:

| Layer | Value | Consumer |
|---|---|---|
| **Stored prop bag** | `{ $$type: 'font-family', value: 'Open Sans' }` | UI, JSON, validation |
| **CSS transform** | Output of `format_for_css` / transformer | Stylesheet |
| **Enqueue** | Output of `get_enqueue_font_family` / `getEnqueueFontFamily` on stored value | `enqueueFont` (JS), `Style_Fonts` (PHP) |

Enqueue reads the **original prop bag** alongside resolve output — it does not reverse the CSS transformer.

---

### Dedicated `font-family` prop type

**PHP:** `Font_Family_Prop_Type` extends `String_Prop_Type`, implements `Font_Enqueueable`, exposes `format_for_css()` and `get_enqueue_font_family()`.

**JS:** `fontFamilyPropTypeUtil` with `formatForCss` and `getEnqueueFontFamily`, registered in the prop schema cache via `Object.assign` on the base util so cache lookups receive the full API.

**Schema:** `font-family` registered in `Style_Schema` as `Font_Family_Prop_Type`, not a generic string.

### CSS-only transformers

**PHP:** `Font_Family_Transformer` delegates to `Font_Family_Prop_Type::format_for_css()`.

**JS:** `fontFamilyTransformer` delegates to `fontFamilyPropTypeUtil.formatForCss`.

Both registered for `$$type: 'font-family'` in the style transformer registries.

### Enqueue from stored value (not transformed CSS)

**PHP (`Styles_Renderer`):**

- `on_font_enqueue` callback during `props_to_css_string`
- After resolve, for each prop: schema lookup → `Font_Enqueueable` → `get_enqueue_font_family( $prop_value['value'] )`
- `Atomic_Styles_Manager::render_css()` wires `on_font_enqueue` → `Style_Fonts::add()`
- Union props resolve `$$type` from the prop bag for schema lookup

**JS (`editor-canvas`):**

- `createPropsResolver` `onPropResolve` receives `{ key, value, propValue, propType }` — `value` = transformed CSS, `propValue` = stored bag
- `useStylePropResolver` wires `onPropResolve` → `maybeEnqueueFontFromStyleProp`
- `enqueue-font-from-style-prop.ts` duck-types `getEnqueueFontFamily` via `getPropSchemaFromCache`

**Removed:** `Font_Enqueue_Collector`, `collect-enqueueable-fonts.ts`, `on_prop_transform` + `raw_value` plumbing, separate enqueue pass in `create-styles-renderer`.

### Unchanged / separate flows

- Control `onDebounce={ enqueueFont }` — immediate editor preview while browsing fonts (independent of render pipeline)
- `useStylePropResolver` — still builds the memoized resolver with transformers + schema (CSS only)

---

## Example flow

User picks **Open Sans**:

1. Control writes `{ $$type: 'font-family', value: 'Open Sans' }` to the style variant
2. **Editor:** `useStyleItems` → `createStylesRenderer` → `resolve()` → CSS via transformer + enqueue via `onPropResolve` → `<style>` in iframe
3. **Frontend:** `Atomic_Styles_Manager` → `Styles_Renderer::resolve()` + `on_font_enqueue` → cached CSS file + `frontend->enqueue_font()`

---

## Known gaps / follow-up

- CSS quoting in `format_for_css` is not yet implemented — output may still be `font-family:Open Sans;` until quoting is added in the transform layer without changing stored/enqueue shape
- No kit fallback stack (`, sans-serif`) in v4 — v1 adds this via typography `selector_value`
- Font stacks / variables (e.g. `Open Sans, sans-serif` as a single stored value) not fully modeled

---

## Key files

| Area | Path |
|---|---|
| PHP prop type | `modules/atomic-widgets/prop-types/font-family-prop-type.php` |
| PHP contract | `modules/atomic-widgets/prop-types/contracts/font-enqueueable.php` |
| PHP transformer | `modules/atomic-widgets/props-resolver/transformers/styles/font-family-transformer.php` |
| PHP renderer | `modules/atomic-widgets/styles/styles-renderer.php` |
| PHP manager | `modules/atomic-widgets/styles/atomic-styles-manager.php` |
| JS prop type | `packages/.../editor-props/src/prop-types/font-family.ts` |
| JS value utils | `packages/.../editor-props/src/utils/font-family-value.ts` |
| JS resolver hook | `packages/.../editor-canvas/src/hooks/use-style-prop-resolver.ts` |
| JS enqueue helper | `packages/.../editor-canvas/src/renderers/enqueue-font-from-style-prop.ts` |
| JS transformer | `packages/.../editor-canvas/src/transformers/styles/font-family-transformer.ts` |

## 1. Problem & Context

Font family values with spaces/special characters weren't being properly quoted in CSS output, breaking custom font names. PR introduces dedicated `Font_Family` prop type and transformer to handle quote normalization across PHP and TypeScript layers.

## 2. What Changed (Where)

| Component | Change |
|-----------|--------|
| **PHP** | New `Font_Family_Prop_Type` + `Font_Family_Transformer` for quote wrapping; registered in module and style schema |
| **PHP Styles** | `Atomic_Styles_Manager` now uses `on_font_enqueue()` callback instead of generic `on_prop_transform()` |
| **TypeScript** | New `fontFamilyTransformer` + `fontFamilyPropTypeUtil` with `getEnqueueFontFamily()` method; updated control to use new util |
| **Font Enqueueing** | Introduced `Font_Enqueueable` interface; abstracted logic into `maybeEnqueueFontFromStyleProp()` |
| **Migrations** | Added `string-to-font-family` operation to convert existing string type props |
| **Tests** | Updated mocks and tests to use `$type: 'font-family'` instead of `string` |

## 3. How It Works

**PHP flow**: `Font_Family_Transformer.transform()` wraps unquoted font names in double quotes, strips existing quotes before re-wrapping. Value flows to CSS renderer which calls `on_font_enqueue()` callback for enqueuing via `Font_Family_Prop_Type.get_enqueue_font_family()`.

**TypeScript flow**: `fontFamilyTransformer` mirrors PHP logic (quote wrapping). When styles render, `maybeEnqueueFontFromStyleProp()` checks if prop type implements `getEnqueueFontFamily()`, extracts the raw font name by removing quotes, and enqueues it.

**Edge cases**: Non-string values return null; already-quoted values pass through; union prop types resolved via `$type` discriminator.

## 4. Risks

Type system change (`string` → `font-family`) requires migration for existing data. Backward compatibility handled via migration operation. Test coverage is solid but FontFamilyControl integration assumes `fontFamilyPropTypeUtil` export availability.


modules/atomic-widgets/styles/styles-renderer.php
```diff
@@ -147,7 +147,9 @@ private function props_to_css_string( array $props ): string {
 					call_user_func( $this->on_prop_transform, $prop, $value );
 				}
 
-				return $prop . ':' . $value . ';';
+				$css_value = 'font-family' === $prop && is_string( $value ) && ! str_starts_with( trim( $value ), 'var(' ) ? '"' . $value . '"' : $value;
```
asafdl: i'm not sure i like this general fix. 
every style that adds font-family should add quotes around it, would this not solve the issue aswell?

modules/atomic-widgets/props-resolver/transformers/styles/font-family-transformer.php
```diff
@@ -0,0 +1,16 @@
...
+use Elementor\Modules\AtomicWidgets\PropsResolver\Props_Resolver_Context;
+use Elementor\Modules\AtomicWidgets\PropsResolver\Transformer_Base;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+class Font_Family_Transformer extends Transformer_Base {
+	public function transform( $value, Props_Resolver_Context $context ) {
+		return is_string( $value ) ? '"' . $value . '"' : $value;
```
asafdl: it has to be string here

modules/atomic-widgets/styles/styles-renderer.php
```diff
@@ -142,16 +144,44 @@ private function props_to_css_string( array $props ): string {
...
+			->map( function ( $value, $prop ) use ( $props, $schema ) {
+				$this->maybe_enqueue_font( $schema, $prop, $props[ $prop ] ?? null );
 
 				return $prop . ':' . $value . ';';
 			} )
 			->implode( '' );
 	}
 
+	private function maybe_enqueue_font( array $schema, string $prop_key, $prop_value ): void {
+		if ( ! $this->on_font_enqueue || ! is_array( $prop_value ) || empty( $prop_value['value'] ) || ! empty( $prop_value['disabled'] ) ) {
```
mykytamurzin: should be `$prop_value` nullable type then?

packages/packages/libs/editor-props/src/utils/font-family-value.ts
```diff
@@ -0,0 +1,9 @@
+export const normalizeFontFamilyValue = ( storedValue: string ): string | null => {
+	const normalized = storedValue.trim();
+
+	return normalized || null;
+};
+
+export const formatFontFamilyForCss = normalizeFontFamilyValue;
+
+export const getEnqueueFontFamily = normalizeFontFamilyValue;
```
mykytamurzin: do we need all 3 exports here?

tests/phpunit/elementor/modules/atomic-widgets/styles/test-styles-renderer.php
```diff
@@ -27,6 +27,50 @@ public function tear_down() {
...
+
+		// Assert.
+		$this->assertStringContainsString( $expected_css, $css );
+	}
+
+	public function font_family_css_quoting_data_provider() {
+		return [
+			'multi-word with digit' => [
+				[ '$$type' => 'font-family', 'value' => 'My Font 3' ],
+				'font-family:My Font 3;',
```
mykytamurzin: it's invalid css - font-family can't contain a number token without quoting.

tests/phpunit/elementor/modules/atomic-widgets/props-resolver/transformers/styles/test-font-family-transformer.php
```diff
@@ -0,0 +1,60 @@
...
+
+	public function test_transform__wraps_string_in_quotes() {
+		// Arrange.
+		$context = $this->create_context();
+
+		// Act.
+		$result = $this->transformer->transform( 'Open Sans', $context );
+
+		// Assert.
+		$this->assertSame( 'Open Sans', $result );
```
mykytamurzin: `test_transform__wraps_string_in_quotes` - there is no quotes in asserted result

modules/atomic-widgets/styles/styles-renderer.php
mykytamurzin: btw ai said that `! empty( $prop_value['disabled'] )` is "dead". guard. Disabled prop types are filtered earlier

modules/atomic-widgets/styles/styles-renderer.php
```diff
@@ -62,12 +65,11 @@ public function render( array $styles ): string {
...
 	}
 
-	public function on_prop_transform( callable $callback ): self {
-		$this->on_prop_transform = $callback;
+	public function on_font_enqueue( callable $callback ): self {
+		$this->on_font_enqueue = $callback;
 
 		return $this;
 	}
-
```
mykytamurzin, line 70: let's keep empty line

murzin-nikita (dismissed):

packages/packages/libs/editor-props/src/prop-types/font-family.ts
```diff
@@ -0,0 +1,22 @@
...
+		if (
+			( trimmed.startsWith( '"' ) && trimmed.endsWith( '"' ) ) ||
+			( trimmed.startsWith( "'" ) && trimmed.endsWith( "'" ) )
+		) {
+			return trimmed.slice( 1, -1 ).trim();
+		}
+
+		return trimmed;
+	},
+} );
```
marcin-el, line 20: we have like very similar functions here and in packages/packages/core/editor-canvas/src/transformers/styles/font-family-transformer.ts
maybe we can do something like 
```
const hasWrappingQuotes = (value: string) =>
	(value.startsWith('"') && value.endsWith('"')) ||
	(value.startsWith("'") && value.endsWith("'"));

const stripWrappingQuotes = (value: string) =>
	hasWrappingQuotes(value)
		? value.slice(1, -1).trim()
		: value;

const ensureWrappingQuotes = (value: string) =>
	hasWrappingQuotes(value)
		? value
		: `"${value}"`;
```

murzin-nikita:
Looks like the feature is not working correctly, as I see in regression tests in screenshots fonts were changed

packages/packages/libs/editor-props/src/prop-types/font-family.ts
mykytamurzin: agree, same can be done for PHP - like quote/unquote helpers

murzin-nikita (approved):

PR #36079, opened 2026-06-02
Internal: Update popup logic [ED-24302]

RobiFerentz:


ronenboxer (dismissed):

modules/promotions/conversion-banner.php
```diff
@@ -0,0 +1,278 @@
...
+
+		?>
+		<div class="e-conversion-banner__paper">
+			<?php if ( $dismissable ) : ?>
+				<button type="button" class="e-conversion-banner__dismiss notice-dismiss">
+					<span class="screen-reader-text"><?php echo esc_html__( 'Dismiss this notice.', 'elementor' ); ?></span>
+				</button>
+			<?php endif; ?>
+			<div class="e-conversion-banner__content">
+				<h2 class="e-conversion-banner__title"><?php echo esc_html( $banner['title'] ); ?></h2>
```
mykytamurzin, line 133:
```
'title' => esc_html__( 'Go Pro, Go Limitless', 'elementor' )
```
double escape?

modules/promotions/conversion-banner.php
```diff
@@ -0,0 +1,278 @@
...
+
+			if ( empty( $placement ) ) {
+				return;
+			}
+
+			add_action( 'in_admin_header', [ $this, 'render_banner_container' ], 11 );
+			add_action( 'admin_enqueue_scripts', function () use ( $placement ) {
+				$this->enqueue_assets( $placement );
+			} );
+		} );
```
mykytamurzin: nested add_aciton is kinda of smell

if you need to have a screen inside action - let's just do checks for this inside every action

modules/promotions/conversion-banner.php
```diff
@@ -0,0 +1,278 @@
+<?php
+
+namespace Elementor\Modules\Promotions;
```
mykytamurzin, line 3: should be inside pointers folder?

modules/promotions/conversion-banner.php
```diff
@@ -0,0 +1,278 @@
...
+		if ( Utils::has_pro() ) {
+			return $config;
+		}
+
+		$config['welcome'] = [];
+
+		return $config;
+	}
+
+	public function ajax_dismiss_banner(): void {
```
mykytamurzin, line 70: it's better to split responsibility between classes - rendering, handling ajax, registering actions/scripts

modules/promotions/conversion-banner.php
```diff
@@ -0,0 +1,278 @@
...
+			],
+			'image' => [
+				'src' => ELEMENTOR_ASSETS_URL . 'images/decade-birthday.png',
+				'alt' => esc_html__( 'Celebrate 10 years of Elementor', 'elementor' ),
+			],
+		];
+	}
+
+	private function get_active_placement(): array {
+		if ( ! $this->is_user_allowed() || ! $this->should_display() ) {
```
mykytamurzin: should be && maybe?

modules/promotions/module.php
```diff
@@ -75,6 +76,8 @@ public function __construct() {
 			new Black_Friday();
 		}
 
+		new Conversion_Banner();
```
mykytamurzin: can it be the same style like
```
		if ( Birthday::should_display_notice() ) {
			new Birthday();
		}
```

then some of the checks can be dropped from Conversion_Banner class

modules/promotions/conversion-banner.php
```diff
@@ -0,0 +1,278 @@
...
+	public function ajax_dismiss_banner(): void {
+		check_ajax_referer( self::NONCE_ACTION, 'nonce' );
+
+		if ( ! $this->is_user_allowed() ) {
+			wp_send_json_error( 'Permission denied', 403 );
+		}
+
+		User::set_introduction_viewed( [ 'introductionKey' => self::DISMISS_KEY ] );
+
+		wp_send_json_success();
```
mykytamurzin: i would suggest to wrap it into try/catch in order not to disturb user with errors if something went wrong here

modules/promotions/conversion-banner.php
```diff
@@ -0,0 +1,278 @@
...
+	}
+
+	private function is_dismissed(): bool {
+		return (bool) User::get_introduction_meta( self::DISMISS_KEY );
+	}
+
+	private function get_allowed_admin_pages(): array {
+		$default = [ 'selector' => self::DEFAULT_SELECTOR ];
+
+		return [
```
mykytamurzin, line 227: is it a manually crafted list? looks like it should be added everywhere...

maybe we can just put some placeholder hook into general template, instead of checking each page like this

modules/promotions/conversion-banner.php
RobiFerentz: It's not a pointer, it's a banner. Pointer appear as popups on the side of the menu.

modules/promotions/conversion-banner.php
RobiFerentz: fixed.

modules/promotions/conversion-banner.php
RobiFerentz: Fixed.

modules/promotions/conversion-banner.php
RobiFerentz: Removed fix. Lint fails for failing to escape :(

modules/promotions/module.php
RobiFerentz: The banner is not a pointer, I prefer keeping this one seperate.

modules/promotions/conversion-banner.php
RobiFerentz: This is how it is done in the hello-theme,  which implements the current banner. It's weird, and fragile. Not sure how else to do it and get the same placement. There seems to be no way to add this UNDER the h1 element of the wp-admin pages.

modules/promotions/conversion-banner.php
```diff
@@ -0,0 +1,291 @@
...
+		$placement = $this->get_active_placement();
+
+		if ( empty( $placement ) ) {
+			return;
+		}
+
+		$this->placement = $placement;
+
+		add_action( 'in_admin_header', [ $this, 'render_banner_container' ], 11 );
+		add_action( 'admin_enqueue_scripts', [ $this, 'enqueue_banner_assets' ] );
```
mykytamurzin:
```
add_action( 'admin_enqueue_scripts', function () use ( $placement ) {
      $this->enqueue_assets( $placement );
} );
```

mykytamurzin: we can drop $this->placement then

modules/promotions/conversion-banner.php
mykytamurzin: @RobiFerentz wdyt?

modules/promotions/module.php
mykytamurzin: @RobiFerentz it's not related to pointer - you just do one single check to register banner or not, instead of multiple checks within conversion-banner.php

RobiFerentz: done.

modules/promotions/conversion-banner.php
RobiFerentz: Not sure we need this much separation of concerns for showing one banner

modules/promotions/conversion-banner.php
RobiFerentz: done

modules/promotions/module.php
```diff
@@ -76,6 +77,10 @@ public function __construct() {
 			new Black_Friday();
 		}
 
+		if ( Conversion_Banner::should_display_banner() ) {
```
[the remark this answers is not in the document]
RobiFerentz: different methods of display.

murzin-nikita (approved):

PR #36093, opened 2026-06-03
Internal: Track classes/variables events via design system [ED-24348]

ronros-elementor:
## 1. Problem & Context

PR tracks design system interactions (tab opens, exports) via Mixpanel for analytics. Supports both classes and variables managers with source attribution to distinguish panel-triggered vs. popover-triggered events.

## 2. What Changed (Where)

| File | Change |
|------|--------|
| `design-system-panel-content.tsx` | Added `trackDesignSystemTabOpened()` function; tracks tab switches with source='system-panel' |
| `design-system-header-menu.tsx` | Track export event |
| `use-open-design-system-toolbar.ts` | Track opened event; removed unused `useCallback` |
| `editor-global-classes/utils/tracking.ts` | Extended `classManagerOpened` event schema with source field |
| `editor-variables/utils/tracking.ts` | Added `VariablesManagerOpenSource` type; extended `trackVariablesManagerEvent()` with optional source param |
| Test file | Added mocks and test cases for tracking calls |
| `events-config.js` | Registered two new event names (design_system_export, design_system_opened) |

## 3. How It Works

Tab switches dispatch `trackDesignSystemTabOpened()` which routes to appropriate tracker: `trackGlobalClasses()` for classes tab, `trackVariablesManagerEvent()` for variables tab. Both receive `source: 'system-panel'` to differentiate from other entry points (e.g., 'vars-popover'). Toolbar onClick now tracks panel opens before dispatching toggle event.

## 4. Risks

None identified. Schema extensions are backward-compatible (source is optional). Tests cover tab tracking paths. Event names registered in config prevents Mixpanel silencing.

### DIFF
--- a/packages/packages/core/editor-design-system/src/components/design-system-header-menu.tsx
+++ b/packages/packages/core/editor-design-system/src/components/design-system-header-menu.tsx
@@ -59,7 +59,7 @@ export const DesignSystemHeaderMenu = () => {
 	const handleExport = () => {
 		popupState.close();
 		trackDesignSystem( { event: 'export' } );
-		runExport();
+		void runExport();
 	};
 
 	const triggerLabel = __( 'Design system actions', 'elementor' );
--- a/packages/packages/core/editor-design-system/src/components/design-system-header-menu.tsx
+++ b/packages/packages/core/editor-design-system/src/components/design-system-header-menu.tsx
@@ -58,7 +58,8 @@ export const DesignSystemHeaderMenu = () => {
 
 	const handleExport = () => {
 		popupState.close();
-		void runExport();
+		trackDesignSystem( { event: 'export' } );
+		runExport();
 	};
 
 	const triggerLabel = __( 'Design system actions', 'elementor' );
--- a/packages/packages/core/editor-design-system/src/components/design-system-panel-content.tsx
+++ b/packages/packages/core/editor-design-system/src/components/design-system-panel-content.tsx
@@ -31,12 +31,14 @@ export type DesignSystemPanelContentProps = {
 const EVENT_SET_TAB = 'elementor/design-system/set-tab';
 
 const trackDesignSystemTabOpened = ( tab: DesignSystemTab ) => {
-	if ( tab === 'classes' ) {
-		trackGlobalClasses( { event: 'classManagerOpened', source: 'system-panel' } );
-		return;
+	switch ( tab ) {
+		case 'classes':
+			trackGlobalClasses( { event: 'classManagerOpened', source: 'system-panel' } );
+			break;
+		case 'variables':
+			trackVariablesManagerEvent( { action: 'openManager', source: 'system-panel' } );
+			break;
 	}
-
-	trackVariablesManagerEvent( { action: 'openManager', source: 'system-panel' } );
 };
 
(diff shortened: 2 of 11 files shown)


packages/packages/core/editor-design-system/src/components/design-system-header-menu.tsx
```diff
@@ -58,7 +58,8 @@ export const DesignSystemHeaderMenu = () => {
 
 	const handleExport = () => {
 		popupState.close();
-		void runExport();
+		trackDesignSystem( { event: 'export' } );
+		runExport();
```
annab1:
``` suggestion
		void runExport();
```

packages/packages/core/editor-design-system/src/components/design-system-panel-content.tsx
```diff
@@ -30,6 +30,15 @@ export type DesignSystemPanelContentProps = {
 
 const EVENT_SET_TAB = 'elementor/design-system/set-tab';
 
+const trackDesignSystemTabOpened = ( tab: DesignSystemTab ) => {
+	if ( tab === 'classes' ) {
+		trackGlobalClasses( { event: 'classManagerOpened', source: 'system-panel' } );
+		return;
+	}
```
annab1, line 41: Let's add `else if ( tab === 'variables')` just in case we'll have other tabs in the future

packages/packages/core/editor-variables/src/utils/tracking.ts
```diff
@@ -42,6 +45,10 @@ export const trackVariablesManagerEvent = ( { action, varType, controlPath }: Va
 		action_type: name,
 	};
 
+	if ( action === 'openManager' && source ) {
```
annab1: Why do we need this condition? let's always send the source if it is defined

annab1 (changes requested):

packages/packages/core/editor-design-system/src/components/design-system-header-menu.tsx
ronros-elementor: I think it's just some redundant addition common to agentic code writing, in practice it doesn't serve any purpose, the fact we don't assign the value to a variable is explicit enough
I never saw it being intentionally added by humans up until now 😄
is there a reason you'd like it to remain there which I may have missed?

packages/packages/core/editor-design-system/src/components/design-system-panel-content.tsx
ronros-elementor: good point, will structure it as a switch-case to make it even stricter

packages/packages/core/editor-design-system/src/components/design-system-header-menu.tsx
annab1: It just lets everyone know you're consciously choosing not to await it. Since skipping await can sometimes introduce tricky bugs, using void is a helpful way to signal that it's intentional

annab1: It should come with an eslint rule that enforces either `await` or `void` on async functions, we can consider adding it later on

annab1 (dismissed):

annab1 (approved):

PR #36128, opened 2026-06-07
Internal: Fix race condition in legacy v1Ready listener registration [ED-24426]

IshayMaya:
## 1. Problem & Context

Race condition when `registerElementType()` is called after v1 initialization completes—listeners registered via `initLegacyViews()` miss late-registered elements. ED-24426 fixes this by checking `isV1Ready()` synchronously and invoking microtask-deferred callbacks to prevent missed registrations.

## 2. What Changed (Where)

- **init-legacy-views.ts**: Extract element registration logic into `registerElementInLegacyManager()`, call it immediately if v1 ready, defer to listener otherwise
- **listeners.ts**: Export `isReady()`, add microtask-deferred callback invocation for v1ReadyEvent if already ready
- **listeners/index.ts**: Export `__privateIsReady` 
- **Tests**: Add coverage for late registration, callback invocation timing, and cleanup behavior

## 3. How It Works

When `registerElementType(type, generator)` is invoked: if `isV1Ready()` returns true, immediately register via `registerElementInLegacyManager()`; otherwise, rely on the `initLegacyViews()` listener. For listeners registering *after* v1 ready, `registerWindowEventListener()` detects this and schedules callback via `Promise.resolve().then()` to avoid sync invocation but fire before next event loop, with a cleanup guard to handle unregistration before the microtask runs.

## 4. Risks

Microtask-deferred callbacks could mask logic errors if tests don't account for async timing. Mitigated by comprehensive tests checking non-sync firing, post-cleanup no-op behavior, and integration test validating end-to-end registration flow. State mutation via `__privateSetReady()` in test setup requires careful isolation—tests properly reset via beforeEach/afterEach.


packages/packages/core/editor-canvas/src/legacy/init-legacy-views.ts
```diff
@@ -21,6 +21,8 @@ export const elementsLegacyTypes: ElementLegacyType = {};
 
 const modelExtensionsRegistry: Record< string, ModelExtensions > = {};
 
+let postInitRenderer: DomRenderer | null = null;
```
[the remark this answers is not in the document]
ronros-elementor: @IshayMaya this seems like a valid concern

packages/packages/core/editor-components/src/index.ts
```diff
@@ -1,6 +1,7 @@
 export { init } from './init';
 
 export { apiClient } from './api';
+export { componentInstanceTransformer } from './component-instance-transformer';
```
ronros-elementor: is it related to the PR?

packages/packages/libs/editor-v1-adapters/src/listeners/__tests__/index.test.ts
```diff
@@ -300,4 +300,46 @@ describe( '@elementor/editor-v1-adapters/listeners', () => {
...
+		// Arrange
+		const callback = jest.fn();
+
+		// Act
+		const cleanup = listenTo( v1ReadyEvent(), callback );
+		cleanup();
+		await Promise.resolve();
+
+		// Assert
+		expect( callback ).not.toHaveBeenCalled();
```
ronros-elementor, line 336: Since we don't have a local state, it seems like it's relying on prior test executions, hence the test is inconsistent, and if it will be moved in execution path - the test may fail

ronros-elementor (changes requested):

packages/packages/core/editor-canvas/src/legacy/init-legacy-views.ts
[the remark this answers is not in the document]
IshayMaya: yeah, removed it completely to rely on the actual ready event

packages/packages/core/editor-components/src/index.ts
IshayMaya: no, removed

ronros-elementor (dismissed):

ronros-elementor (approved):

PR #36133, opened 2026-06-08
Internal: CSS to style schema conversion infra [ED-24441]

eavichay-elementor:
Part of [ED-24401]

[ED-24401]: https://elementor.atlassian.net/browse/ED-24401?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## 1. Problem & Context

Establishes foundation infrastructure for converting arbitrary CSS declarations into Elementor's atomic widget prop system (ED-24441). Enables LLM-generated styles to map faithfully to schema-backed properties instead of falling back to raw customCss.

## 2. What Changed (Where)

| Module | Purpose |
|--------|---------|
| `css-converter/*.php` | Core conversion pipeline: registry, context, dispatcher, variable resolution |
| `converters/*.php` | ~20 property converters (size, color, transform, filter, background layers, etc.) |
| `expanders/*.php` | Shorthand decomposition (background, border, physical→logical insets) |
| `value-parsers/*.php` | Tokenizers and leaf parsers (Size, Filter, Background_Image, Box_Shorthand) |
| `css-converter-rest-api.php` | POST `/elementor/v1/css-to-atomic` REST endpoint |
| `variable-prop-value-transformer.php` | Promotes known global variables to variable PropValues; ejects unresolved/type-mismatched |
| `background-gradient-overlay-prop-type.php` | Added `positions` field (regex) for radial-gradient centers; removed hardcoded enum |
| `position-prop-type.php` | Added `get_radial_position_regex()`, `is_valid_radial_position()` helpers |
| `variables-service.php` | Added `find_by_label_or_id()` for variable lookup by ID or label |
| Tests | 40+ unit & integration tests (DB-less + REST) covering converters, expanders, variables |
| `unit-bootstrap.php` | Fast local PHPUnit runner (no WP/MySQL) for prop-types & converters |
| `editor-canvas/index.ts` | Exported `convertCssToAtomic()` and types; removed unused schema exports |

## 3. How It Works

**Entry point:** REST API receives CSS text or property→value maps. `Css_Converter` parses CSS into rules, expands shorthands via `Expander_Registry` (e.g., `border: 1px solid red` → `border-width`, `border-style`, `border-color`), then dispatches to `Converter_Registry` in order until first match succeeds or falls to customCss.

**Happy path (e.g., `padding: 10px 20px`):**
1. Parser emits `{property: 'padding', value: '10px 20px'}`
2. `Dimensions_Property_Converter` tokenizes via `Box_Shorthand_Parser` → `[10px, 20px, 10px, 20px]`
3. Maps to logical sides: `{block-start: Size(10px), inline-end: Size(20px), block-end: Size(10px), inline-start: Size(20px)}`
4. Wraps in `Dimensions_Prop_Type::generate()` → canonical PropValue

**Edge cases handled:**
- Shorthand + longhand cascade (e.g., `padding: 2rem; padding-left: 1rem` seeds all sides, then overrides one)
- Variable promotion: `padding-bottom: var(--spacing-md)` → if variable is known size type, emits `Size_Variable_Prop_Type` instead of raw leaf
- Unresolved/type-mismatched vars ejected to `rejected` (not customCss) so client can surface type errors
- Rejected properties (animation, @keyframes-dependent constructs) separated from fallback customCss
- Layer merging (background image + repeat + size accumulate into one overlay object across declarations)

## 4. Risks

**Variable resolution dependency:** Transformer ejects props with unresolved `var(--X)` references back to customCss/rejected. If WP or kit loading fails silently, vars won't be found and declarations silently downgrade. Mitigation: explicit logging on finder miss; REST endpoint returns status 200 regardless (by design), so client must inspect props/rejected/customCss buckets.

**Schema drift:** Hardcoded enum lists (repeat, attachment, size keywords) and property families (SIZE_PROPERTIES, COLOR_PROPERTIES, etc.) must stay in sync with live `Style_Schema`. Coverage test exists but only runs during CI. Mitigation: test will fail loudly if schema properties added without converter coverage.

**Paren-aware tokenizer assumes balanced parens:** `Css_Token_Splitter` doesn't validate depth==0 at end (could silently drop malformed input like `rgb(0,0,0`. Low risk for LLM output (usually well-formed) but edge case for hostile input. No mitigation needed for Phase 1.


packages/packages/core/editor-canvas/src/mcp/tools/build-composition/prompt.ts
```diff
@@ -35,8 +35,9 @@ Some elements have internal tree structures (nesting). When using these elements
 - \`allowed_parents\` lists which element types this element can be placed inside
 
 # CONFIGURATION
-- Map configuration-id → elementConfig (props) + stylesConfig (layout only)
-- All PropValues require \`$$type\` matching schema
+- Map configuration-id → elementConfig (props) + style (raw CSS declarations)
```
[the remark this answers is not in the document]
eavichay-elementor: Known - This change is by design

packages/packages/core/editor-canvas/src/mcp/utils/convert-css-to-atomic.ts
```diff
@@ -0,0 +1,42 @@
...
+
+const convertCssBlocks = async (
+	blocks: Record< string, string >
+): Promise< Record< string, CssConversionResult > > => {
+	const { data } = await httpService().post< HttpResponse< Record< string, CssConversionResult > > >(
+		CSS_TO_ATOMIC_URL,
+		{ blocks }
+	);
+
+	return data.data;
```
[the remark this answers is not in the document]
eavichay-elementor: Conversion failure should cause tool failure. No graceful failure by-design

packages/packages/core/editor-canvas/src/mcp/utils/convert-css-to-atomic.ts
```diff
@@ -0,0 +1,42 @@
...
+		Object.entries( styleByName ).map( ( [ name, style ] ) => [ name, styleRecordToCssText( style ) ] )
+	);
+
+	return convertCssBlocks( blocks );
+};
+
+export const convertCssToAtomic = async ( style: Record< string, string > ): Promise< CssConversionResult > => {
+	const results = await convertStyleBlocksToAtomic( { [ SINGLE_BLOCK_KEY ]: style } );
+
+	return results[ SINGLE_BLOCK_KEY ];
```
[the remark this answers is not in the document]
eavichay-elementor: Hard tool failure by design - no graceful handling

modules/atomic-widgets/css-converter/converter-registry.php
```diff
@@ -0,0 +1,30 @@
...
+
+	public function register( Property_Converter $converter ): self {
+		$this->converters[] = $converter;
+
+		return $this;
+	}
+
+	/**
+	 * Converters in registration order. The dispatcher iterates these and applies
+	 * the try-until-success flow (is_supported -> convert -> fallthrough on failure).
```
Ntnelbaba, line 23: For future phases, what is the error failure? 
If we are anyway going to fallback to custom CSS, we can return what failed, but do not fail the all process.
Maybe even "max errors" when passing this threshold, cancel the conversion.

eavichay-elementor: Failed is going to metrics TBD with product team - to understand what is missing/not working

modules/atomic-widgets/css-converter/css-converter.php
```diff
@@ -0,0 +1,129 @@
...
+	 * Naive top-level split on ';' (breaks on values containing ';'); acceptable for clean
+	 * LLM input. Splits each declaration on the first ':' so values keep colons (e.g. url()).
+	 *
+	 * @return array<int, array{property: string, value: string}>
+	 */
+	private function parse( string $css ): array {
+		$rules = [];
+
+		foreach ( explode( ';', $css ) as $declaration ) {
+			$declaration = trim( $declaration );
```
Ntnelbaba, line 176: Save decleration as original css and get it back to custom css if needed
It will remove the render rule function

modules/atomic-widgets/css-converter/css-converter.php
```diff
@@ -0,0 +1,129 @@
...
+
+			$separator = strpos( $declaration, ':' );
+
+			if ( false === $separator ) {
+				continue;
+			}
+
+			$property = strtolower( trim( substr( $declaration, 0, $separator ) ) );
```
Ntnelbaba, line 184: The declaration to key-value here is the most critical path
Please check that we are sanitizing all we can

modules/atomic-widgets/css-converter/css-converter-rest-api.php
```diff
@@ -0,0 +1,74 @@
...
+
+class Css_Converter_REST_API {
+	const API_NAMESPACE = 'elementor/v1';
+	const API_BASE = 'css-to-atomic';
+
+	public function register_hooks() {
+		add_action( 'rest_api_init', fn() => $this->register_routes() );
+	}
+
+	private function register_routes() {
```
Ntnelbaba, line 27: Use case for stress test, how much css can we give this route until the API collapsing the server
For future purposes, we might need to limit it to known limitations.

modules/atomic-widgets/css-converter/css-converter-rest-api.php
```diff
@@ -0,0 +1,74 @@
...
+		register_rest_route( self::API_NAMESPACE, '/' . self::API_BASE, [
+			[
+				'methods' => 'POST',
+				'callback' => fn( $request ) => $this->route_wrapper( fn() => $this->convert( $request ) ),
+				'permission_callback' => fn() => current_user_can( 'edit_posts' ),
+				'args' => [
+					'blocks' => [
```
Ntnelbaba, line 34: Will external LLM understand that?

packages/packages/core/editor-canvas/src/mcp/tools/build-composition/prompt.ts
```diff
@@ -134,13 +134,12 @@ Section with heading + button (NO explicit heights - content sizes naturally):
   elementConfig: {
     "section1": { "tag": { "$$type": "string", "value": "section" } }
   },
-  customCSS: {
-    "Section Title": "padding: 6rem 4rem; background: linear-gradient(135deg, #faf8f5 0%, #f0ebe4 100%);"
-  },
-  stylesConfig: {
+  style: {
```
Ntnelbaba, line 137: This is not the correct example 🫨 
Explained: `**style**: configuration-id → raw CSS declarations (property → value strings; no selectors)`
Actual example: `style:{ property: value }`

packages/packages/core/editor-canvas/src/mcp/tools/build-composition/schema.ts
```diff
@@ -13,25 +13,18 @@ export const inputSchema = {
...
 		)
 		.describe( 'A record mapping element IDs to their configuration objects. REQUIRED' ),
-	stylesConfig: z
+	style: z
 		.record(
 			z.string().describe( 'The configuration id' ),
 			z.record(
-				z.string().describe( 'StyleSchema property name' ),
-				z.any().describe( `The PropValue for the style property. MANDATORY, refer to [${ STYLE_SCHEMA_URI }]` )
+				z.string().describe( 'A CSS property name, e.g. "color", "padding".' ),
```
Ntnelbaba, line 20: Same here, your LLM got drifted 🚗

packages/packages/core/editor-canvas/src/mcp/tools/configure-element/prompt.ts
```diff
@@ -108,20 +102,9 @@ Do NOT send "group" (it is resolved automatically). Use { "settings": {} } only
...
-          $$type: 'number',
-          value: 20
-        },
-        unit: {
-          $$type: 'string',
-          value: 'px'
-        }
-      }
-    }
+  style: {
```
Ntnelbaba, line 100: Here again
Cursor/Claude - Please do a proper check on this error all across the PR, make sure all of it follows the same base rule of "key is the configuration ID and the value is a CSS string with no selectors."

packages/packages/core/editor-canvas/src/mcp/tools/configure-element/schema.ts
```diff
@@ -12,13 +12,14 @@ export const inputSchema = {
 			z.any()
 		)
 		.describe( 'An object record containing property names and their new values to be set on the element' ),
-	stylePropertiesToChange: z
+	style: z
 		.record(
-			z.string().describe( 'The style property name' ),
-			z.any().describe( `The style PropValue, refer to [${ STYLE_SCHEMA_URI }] how to generate values` ),
-			z.any()
+			z.string().describe( 'A CSS property name, e.g. "color", "margin-top".' ),
```
Ntnelbaba, line 17: Same, configuration ID

packages/packages/core/editor-canvas/src/mcp/tools/configure-element/tool.ts
```diff
@@ -2,7 +2,8 @@ import { getWidgetsCache } from '@elementor/editor-elements';
 import { type MCPRegistryEntry } from '@elementor/editor-mcp';
 
 import { DYNAMIC_TAGS_URI } from '../../resources/dynamic-tags-resource';
-import { STYLE_SCHEMA_URI, WIDGET_SCHEMA_URI } from '../../resources/widgets-schema-resource';
+import { WIDGET_SCHEMA_URI } from '../../resources/widgets-schema-resource';
+import { convertCssToAtomic } from '../../utils/convert-css-to-atomic';
```
Ntnelbaba, line 6: Should be in other package as it is super common function
Can be used as well to parse the custom CSS block (FYI - @asafdl @davseve )

Ntnelbaba (dismissed):

packages/packages/core/editor-canvas/src/mcp/mcp-description.ts
```diff
@@ -70,27 +69,26 @@ Map each configuration-id to its widget properties using PropValues:
...
-- Example using global variable:
+### 6. Create style
+Map each configuration-id to raw CSS declarations (property → value strings). The server converts them to native styles and stores any unconvertible declarations as the element custom CSS.
+- Example:
 \`\`\`json
 {
   "heading-1": {
-    "color": { "$$type": "global-color-variable", "value": "primary-color-id" },
-    "font-size": { "$$type": "size", "value": "2rem" }
+    "color": "#1a1a1a",
```
Ntnelbaba: Here as well

modules/atomic-widgets/css-converter/css-converter-rest-api.php
eavichay-elementor: Already fixed in the PR of this PR

packages/packages/core/editor-canvas/src/mcp/tools/build-composition/schema.ts
ronros-elementor: Isn't it's set like that to group by the style classes? i.e. each configuration id targets one class

packages/packages/core/editor-canvas/src/mcp/tools/configure-element/tool.ts
```diff
@@ -83,34 +74,47 @@ export const initConfigureElementTool = ( reg: MCPRegistryEntry ) => {
...
+	}
+	if ( Object.keys( styleValue ).length === 0 ) {
+		return;
+	}
+	try {
+		doUpdateElementProperty( {
+			elementId,
+			elementType,
+			propertyName: '_styles',
+			propertyValue: styleValue,
```
ronros-elementor, line 107: asking since I'm not too aware of the flow going deeper into the `doUpdateElementProperty` flow, but it won't override prior-existing custom_css?

packages/packages/core/editor-canvas/src/mcp/tools/build-composition/schema.ts
```diff
@@ -13,25 +13,18 @@ export const inputSchema = {
...
+				z.string().describe( 'A CSS value, e.g. "6rem 4rem", "#2d2a26".' )
```
ronros-elementor, line 21: if it would like to reset a value, how would the data should look like?
because in theeditor we usually use `null` value for resetting purposes, however type-wise we expect this to always be a string, so was wondering

tests/phpunit/elementor/modules/atomic-widgets/css-converter/test-css-converter-rest-api.php
```diff
@@ -0,0 +1,119 @@
...
+		$data = $response->get_data()['data'];
+
+		// Assert.
+		$this->assertSame( 200, $response->get_status() );
+		$this->assertSame( [ 'el-1', 'el-2' ], array_keys( $data ) );
+		$this->assertSame( 'color: red;', $data['el-1']['customCss'] );
+		$this->assertSame( 'gap: 4px;', $data['el-2']['customCss'] );
+	}
+
+	public function test_post__requires_authentication() {
```
ronros-elementor: could you also add a test for a non-privileged user? (I believe 401 are for non-logged users)

modules/atomic-widgets/css-converter/css-converter-rest-api.php
eavichay: It's handled in the tool - LLM unaware of the HTTP conversion. If we expose a server-side LLM capability - it should wrap the HTTP call in similar manner

packages/packages/core/editor-canvas/src/mcp/tools/build-composition/prompt.ts
eavichay: Fixed in the second PR

packages/packages/core/editor-canvas/src/mcp/tools/build-composition/schema.ts
eavichay: Fixed in the second PR

packages/packages/core/editor-canvas/src/mcp/tools/configure-element/tool.ts
eavichay: Delegate to tech debt?

packages/packages/core/editor-canvas/src/mcp/tools/build-composition/prompt.ts
eavichay-elementor: In build composition we need the "selector" which is the target element.

packages/packages/core/editor-canvas/src/mcp/tools/configure-element/tool.ts
eavichay-elementor: See here: https://elementor.atlassian.net/browse/ED-24497

ronros-elementor (approved):

PR #36135, opened 2026-06-08
Fix: Order of global classes not consistent when used in embeddings [ED-24311]

ronros-elementor:
## 1. Problem & Context

Global class CSS ordering was inconsistent when components embed other components—embedded post styles weren't being aggregated into parent CSS bundles, causing ordering conflicts when the same class appeared across posts (ED-24311). Fix ensures parent posts collect styles from all transitively embedded children, with proper cache invalidation when relationships change.

## 2. What Changed (Where)

- **component-styles.php**: Refactored `render_post()` to delegate traversal logic to cached `get_component_ids_from_post_cached()`, added filter handler for `elementor/document/related_posts` to expose component IDs.
- **atomic-global-styles.php**: Rewrote `register_styles()` to resolve embedding hierarchy transitively, skip embedded posts from producing separate CSS entries, aggregate child styles into parent bundles. Added relation persistence (forward/reverse maps), ancestor traversal for cascading cache invalidation, `on_document_save()` hook to invalidate parent caches when children update.
- **test files**: Added 6 new tests covering related-posts filter, embedding aggregation, cache reuse, parent invalidation on child changes, multi-level nesting, and stale relation cleanup.

## 3. How It Works

Entry point: `elementor/atomic-widgets/styles/register` calls `register_styles()` with post IDs. New logic (1) transitively resolves embeddings via `elementor/document/related_posts` filter, building forward/reverse relation maps cached per context, (2) skips embedded posts from producing separate CSS entries, (3) aggregates all transitive descendants' styles into parent's global CSS callback ordered by kit order. When child embedded in parent is saved, `on_document_save()` invalidates parent's cache by traversing reverse map to find all ancestors. Component traversal (used by both render hook and filter handler) delegates to shared `get_component_ids_from_post_cached()` to avoid duplication.

## 4. Risks

Reverse relation cleanup relies on diff detection (added vs removed posts); false negatives if `elementor/document/related_posts` filter is non-deterministic risk stale parent entries. Mitigated: full invalidation on class order changes, explicit cleanup when embeddings removed. Cache path proliferation (`_related`, `_related-reverse` keys) adds storage; acceptable given per-post-per-context scoping. Transitive resolution could be O(n²) with deep nesting; mitigated by visited set and cycle guards.


modules/global-classes/atomic-global-styles.php
```diff
@@ -43,24 +56,113 @@ public function register_hooks() {
...
 
+		// Build parent→embedded map by applying the filter transitively for every
+		// requested post id.  We guard against cycles with a visited set.
+		$parent_to_embedded = [];
+		$visited = [];
+
+		$resolve = null;
+		$resolve = function( int $pid ) use ( &$parent_to_embedded, &$visited, &$resolve ): array {
+			if ( isset( $visited[ $pid ] ) ) {
+				if ( isset( $visited[ $pid ] ) ) {
```
Omerisra6: remove

modules/global-classes/atomic-global-styles.php
```diff
@@ -43,24 +56,113 @@ public function register_hooks() {
...
+					return $parent_to_embedded[ $pid ] ?? [];
+				}
+			}
+
+			$visited[ $pid ] = true;
```
Omerisra6: move to bottom

modules/global-classes/atomic-global-styles.php
```diff
@@ -43,24 +56,113 @@ public function register_hooks() {
...
+			 *
+			 * Allows registering embedded post ids whose global classes should be
+			 * merged into the parent post's global CSS file instead of producing
+			 * a separate file per embedded post.
+			 *
+			 * @param int[]  $related_post_ids Accumulated child post ids (initially empty).
+			 * @param int    $post_id          Parent post id being inspected.
+			 * @return int[] Embedded post ids to merge into $post_id's global styles.
+			 */
+			$children = (array) apply_filters( 'elementor/document/related_posts', [], $pid );
```
Omerisra6:
```suggestion
			$related = (array) apply_filters( 'elementor/document/related_posts', [], $pid );
```

modules/global-classes/atomic-global-styles.php
```diff
@@ -43,24 +56,113 @@ public function register_hooks() {
...
 	private function register_styles( Atomic_Styles_Manager $styles_manager, array $post_ids ) {
 		$context = $this->get_context();
 
+		// Build parent→embedded map by applying the filter transitively for every
+		// requested post id.  We guard against cycles with a visited set.
+		$parent_to_embedded = [];
+		$visited = [];
+
+		$resolve = null;
+		$resolve = function( int $pid ) use ( &$parent_to_embedded, &$visited, &$resolve ): array {
```
Omerisra6: maybe we should move it into a dedicated method

modules/global-classes/atomic-global-styles.php
```diff
@@ -100,6 +202,111 @@ private function get_document_global_styles( int $post_id, string $context ): ar
...
+	 * routines can find parent posts when a child is saved or when a class
+	 * definition changes.
+	 *
+	 * The reverse map is kept in sync edge-by-edge from the forward map diff:
+	 * newly embedded children gain a parent entry; dropped children lose it.
+	 * This ensures stale parent references are pruned without losing entries
+	 * for children that appear in other parents not present in this batch.
+	 *
+	 * @param array<int,int[]> $parent_to_embedded Forward map: parent_id => child_ids[].
+	 */
```
Omerisra6, line 187: not sure we need all comments

modules/global-classes/atomic-global-styles.php
```diff
@@ -100,6 +202,111 @@ private function get_document_global_styles( int $post_id, string $context ): ar
...
+			$forward_path = [ $this->get_cache_root_key( self::RELATED_KEY ), $parent, $context ];
+
+			$old_children = array_map( 'intval', (array) ( $cache_validity->get_meta( $forward_path ) ?? [] ) );
+			$new_children = array_map( 'intval', $new_children );
+
+			$cache_validity->validate( $forward_path, $new_children );
+
+			foreach ( array_diff( $new_children, $old_children ) as $added_child ) {
+				$this->add_reverse_relation( $cache_validity, $added_child, $parent, $context );
+			}
```
Omerisra6: Maybe move to a method

ronros-elementor: Not sure extracting lines 223-230 to a new function, would be too valuable, as we'll be left with a very slim wrapping, which would call this new external function, which calls internally to two other  functions (add/remove_reverse_relation), and by that time anyone who'd try to track the code execution might get a bit disoriented 😄

Omerisra6 (dismissed):

Omerisra6 (approved):

PR #36136, opened 2026-06-08
Feature: Supporting dynamic in LLM tools [ED-24275]

eavichay-elementor:
## 1. Problem & Context

LLM tools need to support dynamic tags—runtime-resolved values from post data, site config, etc.—as an alternative to static values. This enables richer, context-aware composition and element configuration without requiring the model to know the full tag catalog upfront.

## 2. What Changed (Where)

| File | Change |
|------|--------|
| `canvas-mcp.ts` | Initialize dynamic tags resource and inject resolver into Schema |
| `dynamic-tags-resource.ts` | New resource exposing available tags with categories and settings schemas |
| `resolve-dynamic-tag.ts` | Tag registry queries, LLM output validation/wrapping, settings reconstruction from schema |
| `props-to-llm-schema.ts` | Dynamic union member generation with category-filtered tag name enums; suppress nested duplicates |
| `do-update-element-property.ts` | Wire dynamic tag resolver into LLM prop transformers pipeline |
| `build-composition/prompt.ts`, `configure-element/prompt.ts` | Guidance on dynamic syntax and resource reference |
| Test stubs & test suites | Coverage for nested dynamic (e.g., image `src`), tag filtering, settings wrapping |

## 3. How It Works

**Entry:** Schema injection sets resolver; tools reference `DYNAMIC_TAGS_URI` in prompts.

**Happy path:** (1) LLM sees union with `dynamic` variant constrained to allowed tags via `getDynamicTagNamesByCategories(categories)`. (2) Model outputs `{$$type: "dynamic", value: {name: "tag-name", settings: {...}}}`. (3) `dynamicTagLLMResolver` fills authoritative `group` from registry, wraps scalar settings per schema, omits `fallback`. (4) Resolver runs in transformers chain during prop validation.

**Edge cases:** Unknown tag names return structurally valid but empty object. Nested dynamic (e.g., image `src`) suppresses duplication at root. Overridable union members skipped. Scalar settings auto-wrapped with `$$type` from schema.

## 4. Risks

**Schema drift:** If tag registry changes without prompt updates, model may reference stale tags—mitigate by versioning dynamic tags resource or adding test coverage for resolver robustness to unknown names (already done).

**Nested dynamic suppression logic:** Outermost-only offering prevents duplicates but requires careful tree traversal—test suite covers image/html nesting scenarios; watch for untested nesting patterns.


packages/packages/core/editor-canvas/src/mcp/utils/resolve-dynamic-tag.ts
```diff
@@ -0,0 +1,102 @@
...
+	const wanted = new Set( categories );
+	return Object.values( getAtomicDynamicTags() )
+		.filter( ( tag ) => tag.categories?.some( ( category ) => wanted.has( category ) ) )
+		.map( ( tag ) => tag.name );
+};
+
+// Reconstructs an intact dynamic PropValue from whatever the LLM produced: the authoritative `group`
+// is taken from the registry (never trusted from the model) and `settings` are rebuilt strictly from
+// the tag's props schema (provided values are wrapped, omitted ones fall back to their defaults).
+export const dynamicTagLLMResolver = ( value: unknown ): PropValue => {
```
Omerisra6, line 42:
```suggestion
export const dynamicTagLLMResolver = ( value: LlmDynamicValue ): PropValue => {
```

packages/packages/libs/editor-props/src/utils/props-to-llm-schema.ts
Omerisra6, line 1: Please open a task so that we can think about how we can allow developers to register "resolvers" or to define an "edge case" for a prop type so that we can support changing values from outside the UI

Omerisra6 (approved):

packages/packages/core/editor-canvas/src/mcp/utils/resolve-dynamic-tag.ts
eavichay-elementor: type PropConverter must accept `unknown` input

PR #36145, opened 2026-06-08
Internal: Centralize styles handling for embeddings [ED-23729]

ronros-elementor:
## 1. Problem & Context

Centralizes embedded/related post styles handling by extracting it from the editor-components Redux store into a dedicated manager. Simplifies state management and enables reuse across packages (global-classes, components).

## 2. What Changed (Where)

| File | Change |
|------|--------|
| `editor-embedded-documents-manager/` | New package: document registry + styles provider for embedded posts |
| `editor-components/init.ts` | Removed local styles registration; delegates to embedded-documents-manager |
| `editor-components/store/` | Removed `styles` state, `removeStyles`/`addStyles` reducers, `selectStyles` selectors |
| `editor-components/utils/get-component-documents.ts` | Added `isRecursive` parameter to control nested component traversal |
| `editor-global-classes/init.ts` | Now subscribes to `embeddedDocumentsManager.onDocumentLoad()` for class loading |
| `editor-v2-loader.php` | Added `editor-embedded-documents-manager` to module list |

## 3. How It Works

**Entry**: `editor-embedded-documents-manager/init()` registers styles provider and attaches to `attach-preview` hook.

**Happy path**: Components fetch embedded documents → call `embeddedDocumentsManager.setDocument(id, data)` → manager extracts styles, notifies listeners, updates centralized styles provider. Global classes subscribes via `onDocumentLoad()` callback.

**Edge cases**: Current document IDs skipped (no duplicate styles); `isRecursive=false` stops nested traversal; already-loaded documents skip listener notifications; unsubscribe removes callbacks cleanly.

## 4. Risks

**Behavioral change**: `loadComponentsAssets()` no longer awaits async operations—now fire-and-forget with `void`. Verify cascade timing against save operations doesn't race. Mitigation: document-aware reset on `attach-preview` ensures clean state transitions.

**Package dependency**: New cross-package dependency (components, global-classes → embedded-documents-manager). Circular risk low—manager is leaf dependency. Ensure version alignment in package.json.


packages/packages/core/editor-components/src/store/actions/load-components-assets.ts
```diff
@@ -1,15 +1,17 @@
 import { isDocumentDirty, setDocumentModifiedStatus } from '@elementor/editor-documents';
 import { type V1ElementData } from '@elementor/editor-elements';
+import { setPost } from '@elementor/editor-related-posts-manager';
 
 import { type ComponentDocumentsMap, getComponentDocuments } from '../../utils/get-component-documents';
 import { loadComponentsOverridableProps } from './load-components-overridable-props';
-import { loadComponentsStyles } from './load-components-styles';
 
 export async function loadComponentsAssets( elements: V1ElementData[] ) {
```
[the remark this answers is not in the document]
ronros-elementor: added

packages/packages/core/editor-related-posts-manager/src/api.ts
```diff
@@ -0,0 +1,140 @@
...
+
+		if ( ! result ) {
+			return;
+		}
+
+		const { data } = result;
+		loadedPosts.set( id, data );
+		notifyRelatedListeners( id, data );
+	} );
+}
```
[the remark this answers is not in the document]
ronros-elementor: agree about the early return if not in pending, will not check for loadedPosts though, as I might want to override existing data

packages/packages/core/editor-embedded-documents-manager/src/styles-provider.ts
```diff
@@ -0,0 +1,61 @@
...
+import { type StyleDefinition } from '@elementor/editor-styles';
+import { createStylesProvider } from '@elementor/editor-styles-repository';
+
+let styles: StyleDefinition[] = [];
+let postStyles = new Map< number, StyleDefinition[] >();
+const styleListeners = new Set< () => void >();
+
+export const relatedPostsStylesProvider = createStylesProvider( {
+	key: 'related-posts-styles',
+	priority: 75,
```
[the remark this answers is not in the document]
ronros-elementor: that's how it's set in all other style providers

packages/packages/core/editor-components/package.json
ronros-elementor: check if this could be removed

packages/packages/core/editor-components/src/init.ts
```diff
@@ -59,14 +56,20 @@ export function init() {
...
-		if ( id ) {
-			removeComponentStyles( id );
+		if ( ! id ) {
+			return;
 		}
 
-		await loadComponentsAssets( ( config?.elements as V1ElementData[] ) ?? [] );
+		removeComponentStyles( id );
+
+		void loadComponentsAssets( ( config?.elements as V1ElementData[] ) ?? [] );
```
MiryamOren, line 68:
```suggestion
		loadComponentsAssets( ( config?.elements as V1ElementData[] ) ?? [] );
```

packages/packages/core/editor-components/src/init.ts
```diff
@@ -59,14 +56,20 @@ export function init() {
...
+	} );
+
+	onPostLoad( ( _postId, data ) => {
+		void loadComponentsAssets( data.elements ?? [] );
```
MiryamOren, line 72: In `loadComponentsAssets` we loop over nested components, do we loop over them here too?

packages/packages/core/editor-embedded-documents-manager/src/init.ts
```diff
@@ -0,0 +1,17 @@
...
+import { stylesRepository } from '@elementor/editor-styles-repository';
+import { registerDataHook } from '@elementor/editor-v1-adapters';
+
+import { reset, setCurrentDocumentId } from './api';
+import { relatedPostsStylesProvider } from './styles-provider';
+
+export function init() {
+	stylesRepository.register( relatedPostsStylesProvider );
+
+	registerDataHook( 'after', 'editor/documents/attach-preview', () => {
```
MiryamOren, line 11: both the reset here and the loading in the components package are after hooks of 'editor/documents/attach-preview', how can we ensure the order of reset -> reload?

packages/packages/core/editor-related-posts-manager/src/api.ts
```diff
@@ -0,0 +1,145 @@
...
+
+/**
+ * Registers a callback that is invoked whenever a related (non-current) post
+ * finishes loading. Already-loaded related posts are delivered to the callback
+ * asynchronously on subscribe. Returns an unsubscribe function.
+ *
+ * @param {RelatedPostLoadCallback} callback Function called with (postId, data) on each load.
+ */
+export function onPostLoad( callback: RelatedPostLoadCallback ): () => void {
```
MiryamOren: maybe?
```suggestion
export function onNestedPostLoad( callback: RelatedPostLoadCallback ): () => void {
```

packages/packages/core/editor-related-posts-manager/src/index.ts
```diff
@@ -0,0 +1,3 @@
+export { init } from './init';
```
MiryamOren: Naming question - I think "related posts" is a bit vague - related to what? The first thing I think of is maybe the post parent thing we have in WP

How about "nested posts"?

packages/packages/core/editor-embedded-documents-manager/src/manager.ts
```diff
@@ -0,0 +1,145 @@
+import { type Document } from '@elementor/editor-documents';
```
MiryamOren, line 1: Why `api`? Shouldn't it be `manager`?

packages/packages/core/editor-components/src/init.ts
ronros-elementor: So actually @annab1 educated me about this -
this is to emphasize we won't be waiting for any returned value back
so in the context of async functions - this convey's we don't care of the function's outcome

> Prefixing the call with void tells both TypeScript and the linter: "I am aware this returns a promise, but I intentionally want it to run in the background without awaiting its completion." [void in TypeScript](youtube.com/watch?v=YpT0rCWdCaE&t=395)

packages/packages/core/editor-related-posts-manager/src/api.ts
ronros-elementor: for now - will change to `onRelatedPostLoad` and will generally use the related wording in all exported functions, such as `resetRelatedPosts`, etc.. so it would be clearer indeed, and we've got ourselves another thread regarding the "related" wording [here](https://github.com/elementor/elementor/pull/36145/changes#r3397160656) for the final decision about the core naming itself 😄

packages/packages/core/editor-related-posts-manager/src/index.ts
ronros-elementor: the thing is that it's not only nesting/embedding of templates/components, it could also be the header/footer in pro, which are configured via the theme builder, and has nothing to do with one post or the other (as these templates won't appear in any document's data, but are rather managed via the locations-manager in elementor-pro)
so went with related, but could also be... I don't know - external maybe? as in - external posts? that may be a bit ambiguous too, but nested would certainly be confusing, as a document may not necessarily have these posts nested in it, but could also be "accompanied" by them

packages/packages/core/editor-components/src/init.ts
ronros-elementor: it's a good point I'll take offline with you tomorrow to discuss further, as it may be a good idea really to cancel the recursion at [get-component-documents.ts](https://github.com/elementor/elementor/blob/main/packages/packages/core/editor-components/src/utils/get-component-documents.ts#L26-L47), and let the `onRelatedPostLoad` handle it, as we're anyway firing the loadComponentAssets in that callback...

packages/packages/core/editor-embedded-documents-manager/src/init.ts
ronros-elementor: I assume that based on its order in the [php loader](https://github.com/elementor/elementor/blob/5c5dfcd47239cab0deb680efcb265df03e199ebc/core/editor/loader/v2/editor-v2-loader.php) it works as expected, as all other consumers are conditionally loaded later on, sort of a more "core"-ish than other modues (components, global classes and templates are all loaded as part of their own modules lifecycles)

MiryamOren (approved):

PR #36159, opened 2026-06-09
Internal: Collection Loop Pro promotion [ED-24366]

IshayMaya:
## Summary
- Add Collection Loop promotion element and widget for free Core users when the atomic elements experiment is active
- Show upgrade CTA in the elements panel and promotion modal, following the existing atomic form promotion pattern
- Register promotion data via the promotions module and add PHPUnit coverage for the promotion element

## Test plan
- [ ] With Pro deactivated and `e_atomic_elements` active, verify Collection Loop appears in the v4 elements panel with promotion styling
- [ ] Click the widget and confirm the promotion modal opens with correct CTA links
- [ ] Drag the promotion element to the canvas and verify the placeholder with upgrade link renders
- [ ] Confirm existing elements panel behavior is unchanged for other widgets
- [ ] Run `vendor/bin/phpunit tests/phpunit/elementor/modules/atomic-widgets/elements/test-collection-loop-promotion.php`

## 1. Problem & Context

Adding Collection Loop widget promotion for free users, mirroring existing Atomic Form promotion infrastructure (ED-24366). Generalizes atomic widget promotions to support multiple widget types.

## 2. What Changed (Where)

- **Promotions module**: New `Collection_Loop_Widget_Promotion` class registers promotion data into `atomicWidgetPromotions` array (replaces form-specific config keys).
- **Atomic widgets module**: New `Collection_Loop_Promotion` element registered for non-Pro users; refactored checkbox logic to use generic `promotionType` field.
- **Editor UI**: Generalized promotion handling to iterate `atomicWidgetPromotions` array; renamed CSS classes from `.e-form-placeholder` to `.e-pro-promotion-placeholder`; event dispatch now dynamic (`{type}-promotion:open`).
- **Frontend**: `AppManager` refactored to generic `mountCustomPromotion()` method; event listeners loop over promotion types instead of hardcoding Atomic Form.

## 3. How It Works

Event flow: User clicks locked widget → `element.js` dispatches `{promotionType}-promotion:open` → `AppManager` listener routes to `mountCustomPromotion()` with promotion data from `elementor.config.atomicWidgetPromotions[]`. Lock icon shows when `!editable && !promotionType && !birthdayEasterEgg`. Twig templates use shared `.e-pro-promotion-placeholder` class; styles decoupled from specific widget type.

## 4. Risks

Promotion data keying change (`atomicFormPromotionWidgets`/`atomicFormPromotion` → `atomicWidgetPromotions[]`) requires consumers to migrate. Event naming convention (`{type}-promotion:open`) assumes type slugs don't contain special characters. No backward compatibility shim provided — existing integrations break.


assets/dev/js/editor/regions/panel/pages/elements/views/element.js
```diff
@@ -58,6 +58,10 @@ module.exports = Marionette.ItemView.extend( {
 		return !! this.model.get( 'atomicFormPromotion' );
 	},
 
+	isCollectionLoopPromotion() {
```
Omerisra6: Is there a way to extend it from the collection module/package?

modules/atomic-widgets/elements/atomic-collection-loop/collection-loop-promotion.php
```diff
@@ -0,0 +1,80 @@
+<?php
```
Omerisra6, line 1: why do we need to register as an element type, what if angie uses that element?

modules/promotions/assets/js/react/app-manager.js
```diff
@@ -82,6 +88,40 @@ export class AppManager {
 		} );
 	}
 
+	mountCollectionLoopPromotion( targetEl, ctaUrl ) {
```
Omerisra6: cant we find a better way?

assets/dev/js/editor/regions/panel/pages/elements/views/element.js
```diff
@@ -54,8 +54,8 @@ module.exports = Marionette.ItemView.extend( {
 		return !! this.model.get( 'integration' );
 	},
 
-	isAtomicFormPromotion() {
-		return !! this.model.get( 'atomicFormPromotion' );
+	isAtomicWidgetPromotion() {
+		return !! this.model.get( 'promotionType' );
```
Omerisra6, line 58: could it cause BC problems?

assets/dev/js/editor/regions/panel/pages/elements/elements.js
```diff
@@ -113,7 +113,18 @@ PanelElementsLayoutView = Marionette.LayoutView.extend( {
 				icon: widget.icon,
 				categories: JSON.parse( widget.categories ),
 				editable: false,
-				atomicFormPromotion: true,
+				promotionType: 'atomic-form',
+			} );
+		} );
+
+		jQuery.each( elementor.config.collectionLoopPromotionWidgets || [], ( index, widget ) => {
```
Omerisra6: maybe we should create a promotions config array so that it would work for all promotions

modules/atomic-widgets/elements/atomic-collection-loop/collection-loop-promotion.html.twig
```diff
@@ -0,0 +1,17 @@
...
+     data-interaction-id="{{ interaction_id }}"
+     {{ editor_attributes | default('') | raw }}>
+    <div class="e-form-placeholder">
+        <i class="eicon-upgrade-crown e-form-placeholder__icon"></i>
+        <div class="e-form-placeholder__title">Collection Loop is a Pro feature</div>
+        <div class="e-form-placeholder__description">Upgrade now to display dynamic content in a repeating layout with full query control.</div>
+        <div class="e-form-placeholder__actions">
+            <button type="button" class="e-form-placeholder__remove-btn">Remove</button>
+            <a href="https://go.elementor.com/go-pro-collection-loop-canvas-upgrade/" target="_blank" rel="noopener noreferrer" class="e-form-placeholder__unlock-btn">Unlock with Pro</a>
+        </div>
```
Omerisra6: You are using "form" prefix + do we really need all that attributes?

modules/promotions/assets/js/react/app-manager.js
Omerisra6, line 1: Dont you think all of this logic should be come from the promotion-widgets module?

Omerisra6 (dismissed):

modules/promotions/assets/js/react/app-manager.js
IshayMaya: This is how we register the atomic pro widgets, created a utils function for both

assets/dev/js/editor/regions/panel/pages/elements/views/element.js
IshayMaya: changed it to a generic `promotionType` param to pass so we won't need to touch this file again

modules/atomic-widgets/elements/atomic-collection-loop/collection-loop-promotion.php
IshayMaya: valid concern, i think we need to align it with forms, we've already opened a ticket to fix the empty element after re-acativation, i'll ask to add it to it as well.

assets/dev/js/editor/regions/panel/pages/elements/views/element.js
IshayMaya: safe, we only use it internally

Omerisra6 (approved):

PR #36203, opened 2026-06-15
Internal: Ensure loop items are rerendered after creation [ED-24480]

Omerisra6:
## 1. Problem & Context

Loop items weren't rerendering after creation because consumers had no signal when composition build completed. This event enables downstream systems (e.g., loop handlers) to trigger necessary rerenders post-build.

## 2. What Changed (Where)

- **composition-builder.ts**: Dispatches `elementor/composition/built` CustomEvent with root container IDs after properties applied
- **composition-builder.test.ts**: Added test suite verifying event dispatch with correct payload and timing

## 3. How It Works

After `applyProperties()` completes, the builder dispatches a CustomEvent to `window.top` (or `window` as fallback) containing mapped root container IDs. Listeners can then perform post-build operations like rerendering loop items. Window check prevents SSR breaks.

## 4. Risks

None identified. Event is fire-and-forget, non-blocking, and isolated to browser context. Existing consumers unaffected since this is new functionality.

### DIFF
--- a/packages/packages/core/editor-canvas/src/composition-builder/composition-builder.ts
+++ b/packages/packages/core/editor-canvas/src/composition-builder/composition-builder.ts
@@ -320,6 +320,15 @@ export class CompositionBuilder {
 
 		const { configErrors, styleErrors } = await this.applyProperties();
 
+		// Ensure all containers are rendered after applying properties
+		for ( const container of this.rootContainers ) {
+			const view = container.view as Record< string, unknown > | undefined;
+			if ( typeof view?.render === 'function' ) {
+				( view.render as () => void )();
+			}
+			await this.awaitViewRender( container );
+		}
+
 		return {
 			configErrors,
 			styleErrors,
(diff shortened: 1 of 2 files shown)


packages/packages/core/editor-canvas/src/composition-builder/composition-builder.ts
```diff
@@ -320,6 +320,15 @@ export class CompositionBuilder {
...
 		const { configErrors, styleErrors } = await this.applyProperties();
 
+		// Ensure all containers are rendered after applying properties
+		for ( const container of this.rootContainers ) {
+			const view = container.view as Record< string, unknown > | undefined;
+			if ( typeof view?.render === 'function' ) {
+				( view.render as () => void )();
+			}
+			await this.awaitViewRender( container );
+		}
```
Omerisra6: maybe we could send an event and let the loop handle the rerender instead of applying it for all

IshayMaya (approved):

PR #36225, opened 2026-06-16
Internal: Add an api for adding a control decoration [ED-24541]

Omerisra6:
## 1. Problem & Context

Adds extensibility API for injecting custom controls into the grid layout section without modifying core layout code. Enables third-party extensions via location-based injection pattern.

## 2. What Changed (Where)

- `layout-section.tsx`: Creates `GridFieldsSlot` location and injects it into `GridFields` component render output
- `index.ts`: Exports `injectIntoGridFields` function for public API consumption

## 3. How It Works

Uses `createLocation()` to establish an injection point in the GridFields component. External code calls `injectIntoGridFields()` to register controls that render at the `GridFieldsSlot` placeholder, positioned after `GridAutoFlowField` in the layout panel.

## 4. Risks

None identified. Pattern mirrors existing `injectIntoStyleTab` precedent. Slot placement non-breaking since it's additive-only in a flex container.


packages/packages/core/editor-editing-panel/src/components/settings-tab.tsx
```diff
@@ -46,7 +50,13 @@ export const SettingsTab = () => {
 								key={ type + '.' + index }
 								defaultExpanded={ isDefaultExpanded( value.id ) }
 							>
+								{ value.id ? (
```
IshayMaya: why not always have the slot?

IshayMaya (approved):

PR #36231, opened 2026-06-17
Fix: Each site kit should have its own set of global classes CPTs [ED-24438]

ronros-elementor:
## 1. Problem & Context

Previously, global class CPT posts were shared across all kits on a site via lazy backfill, causing data isolation violations. This PR ensures each kit owns an exclusive, complete set of class-to-post mappings via a three-pass migration (ED-24438).

## 2. What Changed (Where)

- **Database updater**: Bumped DB_VERSION to 4, added `Migrate_All_Kits_Post_IDs` migration.
- **Migrate_All_Kits_Post_IDs**: New migration handling Pass A (restructure unmigrated kits), Pass B (fill gaps conflict-aware), Pass C (deduplicate shared posts).
- **Migrate_To_Posts**: Refactored to process all kits instead of just active kit; made `migrate_kit()` public & static.
- **Global_Classes_Post_IDs**: Removed backfill logic entirely; `get_post_id()` now returns null for unmapped entries instead of lazy-resolving.
- **Global_Classes_Repository**: Removed unused `meta_key` context config; updated `find_by_class_id()` calls to pass kit parameter.
- **Kit_Utils**: New utility extracting kit-fetching logic (replaces duplicated queries).
- **Tests**: Added comprehensive test suite for new migration; updated existing tests to match no-backfill semantics.

## 3. How It Works

Migration runs in three sequential passes over all site kits:

**Pass A**: Kits with old aggregate meta but no order meta run the original `migrate_kit()` logic to create initial CPT posts.

**Pass B**: For each kit, builds union of claimed post_ids across all kits, then fills missing map entries by (1) reusing unclaimed CPT posts, or (2) creating fresh posts from aggregate data—never double-claiming.

**Pass C**: Identifies shared post_ids across kit maps; "loser" kits (higher ID) get fresh cloned posts, "owner" kits (lowest ID) retain originals. Guarantees one-to-one post_id ownership.

Runtime no longer backfills unmapped classes—all mappings must exist in kit meta or they resolve to null.

## 4. Risks

**Data loss if migration fails mid-pass**: Pass A→B→C has no transaction wrapper. Partial failures could leave some kits without complete mappings. Mitigation: migration is idempotent (rerun safe); monitoring needed on large sites with many kits.

**Performance on high kit counts**: Iterates all kits multiple times with database queries per kit. Mitigated by batching within passes, but large multisite deployments should test.


modules/global-classes/database/migrations/migrate-all-kits-post-ids.php
```diff
@@ -0,0 +1,313 @@
...
+	 * @return array<int, true> post_id => true
+	 */
+	private function build_claimed_post_ids( array $kits ): array {
+		$claimed = [];
+
+		foreach ( $kits as $kit ) {
+			$map = $this->read_post_ids_map( $kit );
+
+			foreach ( $map as $post_id ) {
+				$claimed[ (int) $post_id ] = true;
```
ronros-elementor: change to style_id

modules/global-classes/database/migrations/migrate-all-kits-post-ids.php
```diff
@@ -0,0 +1,313 @@
...
+
+	// -------------------------------------------------------------------------
+	// Pass A
+	// -------------------------------------------------------------------------
+
+	/**
+	 * @param Kit[] $kits
+	 */
+	private function restructure_unmigrated_kits( array $kits ): void {
+		foreach ( $kits as $kit ) {
```
MiryamOren, line 64: Isn't it the same logic we have in `Migrate_To_Posts::migrate_kit`? Why do we need to run it here too?

modules/global-classes/database/migrations/migrate-all-kits-post-ids.php
```diff
@@ -0,0 +1,313 @@
...
+				$post_id = $this->resolve_post_id_for_class( $class_id, $kit, $aggregate_items, $claimed_post_ids );
+
+				if ( null !== $post_id ) {
+					$resolved[ $class_id ] = $post_id;
+					$claimed_post_ids[ $post_id ]   = true;
+				}
+			}
+
+			// Persist reused CPT entries. Entries created via Global_Class_Post::create() are
+			// already written by that call, but set_many() is idempotent so it's safe to include them.
```
MiryamOren: I didn't get this comment?

modules/global-classes/database/migrations/migrate-all-kits-post-ids.php
ronros-elementor: I was wrong, it is indeed the post_id, in foreach we iterate over the values, which are post ids

modules/global-classes/database/migrations/migrate-all-kits-post-ids.php
```diff
@@ -0,0 +1,313 @@
...
+		}
+	}
+
+	/**
+	 * Find or create a post_id for a class_id that belongs exclusively to $kit.
+	 *
+	 * Priority:
+	 *  1. Find an existing CPT post for this class_id that is not yet claimed by any kit.
+	 *  2. Create a fresh CPT post from the kit's aggregate data.
+	 *  3. If no aggregate data exists, skip (returns null).
```
MiryamOren, line 136: If the first two options failed, maybe try to search claimed posts too?

modules/global-classes/database/migrations/migrate-all-kits-post-ids.php
```diff
@@ -0,0 +1,313 @@
...
+
+			if ( ! $source_post ) {
+				continue;
+			}
+
+			$source_data  = $source_post->get_data( true );
+			$source_label = $source_post->get_label();
+
+			foreach ( $losers as [ 'kit' => $loser_kit, 'class_id' => $class_id ] ) {
+				$forked = Global_Class_Post::create( $class_id, $source_label, $source_data, null );
```
MiryamOren: Why is the fourth parameter null and not the kit?

modules/global-classes/database/migrations/migrate-all-kits-post-ids.php
ronros-elementor: we do need to run it again, but I could indeed just re-use the migrate_to_posts migration, as the current implementation _does_ duplicate this logic

as to the why - we take into account we have users who were going through the `migrate_to_posts` migration already, back when it targeted only the active kit, so this migration (which would be invoked for all users regardless) - just makes sure all kits ran through this migrator

it would skip if we already have the `Global_Classes_Order` posteta

modules/global-classes/database/migrations/migrate-all-kits-post-ids.php
ronros-elementor: me neither 🙈😅 sorry was probably just a leftover that was left there accidentally

modules/global-classes/database/migrations/migrate-all-kits-post-ids.php
ronros-elementor: claimed posts are posts that are already associated to a particular kit, re-using them in another kit is what I'm trying to prevent

MiryamOren: I meant find + duplicate

MiryamOren (approved):

PR #36255, opened 2026-06-18
Fix: Some CSS rules fails to convert [ED-24584]

eavichay-elementor:
- **Initial fix**
- **stronger flow**
- **adding kb**

## 1. Problem & Context

CSS rules with `null` values (property resets) and certain shorthand properties (outline, background) were failing to convert properly. The system needed to handle null resets throughout the pipeline—parse, expand, convert, validate, and cleanup—while collapsing aggregate objects where all sub-values become null.

## 2. What Changed (Where)

- **Converters & Expanders**: Template method pattern—`convert()`/`expand()` now dispatch `null → convert_null()/expand_null()`, `non-null → do_convert()/do_expand()`. All 20+ converters updated.
- **Css_Converter**: Added `dedupe()` to drop superseded declarations before conversion; `cleanup_props()` recursively collapses object props with all-null leaves; `validate_props()` bypasses schema validation for partial-null objects.
- **Parsers**: `parse()` normalizes `'null'` string → PHP `null`; `is_blocked()` filters dangerous values.
- **Expanders**: `Outline_Shorthand_Expander` (new); `Background_Shorthand_Expander` / `Border_Shorthand_Expander` now handle null expansion to all longhands.
- **REST API & Tests**: Simplified test boilerplate via `get_data()` helper; added 80+ test cases for null semantics and multi-rule interactions.

## 3. How It Works

**Null flows through the pipeline:**

```
'null' string (input) → parse(): PHP null
  ↓ expand_null(): fans out to all longhands (e.g., border: null → border-width: null, border-style: null, border-color: null)
  ↓ dedupe(): last declaration wins (later nulls override earlier values)
  ↓ convert_null(): merge converters slot null into aggregate fields
  ↓ validate_props(): skips Props_Parser if any leaf is null
  ↓ cleanup_props(): if all sub-values of {field1: null, field2: null} are null → collapse to top-level null
```

**Key insight**: Context is a shared build surface—`background-color: red; background-image: url(img)` accumulates fields into one `background` object. Nulls participate identically: each converter reads existing prop, merges its slot (null or value), writes back. Last declaration wins (CSS cascade).

**Deduplication happens post-expansion**: `padding: 10px; padding-top: null; padding: 5px` → dedupe keeps only the final `padding: 5px` (physical becomes logical after expansion). Invalid early declarations are dropped entirely before conversion—they never reach customCss.

## 4. Risks

- **Null collapse aggressiveness**: An object where every child is null collapses to `null` at the parent level. If client code expects `{ $type: 'dimensions', value: { block-start: null, … } }` vs. bare `null`, breakage. Mitigated by comprehensive test coverage (80+ null cases).
- **Deduplication before conversion**: Invalid declarations now disappear before reaching converters (not in customCss). Intentional—CSS cascade says last wins, so an invalid first declaration is superseded anyway. Matches browser semantics.
- **Outline expander new surface**: Outline is a new shorthand expander (not previously converted). Low risk—outline-offset inclusion is explicit and tested.
- **Props now arrays not objects**: REST API returns `{ padding: null }` as array key not `props->padding`. Minor serialization shift; `(array)` casts handle it.


modules/atomic-widgets/css-converter/css-converter.php
```diff
@@ -182,12 +275,14 @@ private function parse( string $css ): array {
 			}
 
 			$property = strtolower( trim( substr( $declaration, 0, $separator ) ) );
-			$value = trim( substr( $declaration, $separator + 1 ) );
+			$raw_value = trim( substr( $declaration, $separator + 1 ) );
 
-			if ( '' === $property || '' === $value || $this->is_blocked( $property, $value ) ) {
+			if ( '' === $property || '' === $raw_value || $this->is_blocked( $property, $raw_value ) ) {
```
ronros-elementor, line 280: shouldn't an empty string be considered as null too?

ronros-elementor (approved):

PR #36258, opened 2026-06-19
Fix: Tools page break when Cloud Kits API unavailable [ED-24714]

abolfazl-moeini:
## Summary

When the cloud kits quota endpoint is unreachable, `Cloud_Kits::get_quota()` triggers `Base_App::http_request()`, which called `wp_die()` on transport errors — breaking admin page rendering mid-response (e.g. Elementor → Tools).

This adds `with_error_data => true` to the `get_quota()` HTTP request so transport failures return a `WP_Error` that existing callers (such as `check_eligibility()`) already handle gracefully.

## Changes

- `modules/cloud-kit-library/connect/cloud-kits.php` — pass `with_error_data => true` in `get_quota()` `http_request()` options

## Notes

Keeps the existing `http_request()` contract unchanged; surgical fix scoped to the Tools page quota check path as suggested in review.

## 1. Problem & Context

Tools page crashes when Cloud Kits API is unavailable due to unhandled HTTP errors. Adding error data tracking prevents fatal wp_die() calls and gracefully handles timeout/connection failures (ED-24714).

## 2. What Changed (Where)

- **cloud-kits.php**: Added `'with_error_data' => true` flag to http_request() call for quota/kits endpoint.
- **test-common-app.php**: New test validates WP_Error handling with error_data flag on connection failures.

## 3. How It Works

The `with_error_data` flag signals the HTTP layer to wrap API failures (timeouts, connection errors) as structured WP_Error objects with code 500/"No Response" rather than triggering fatal errors. Calling code then handles gracefully instead of page breaking.

## 4. Risks

None — backward compatible flag addition. Existing error handling paths unaffected; new flag only activates structured error wrapping for this specific quota endpoint.

### DIFF
--- a/core/common/modules/connect/apps/base-app.php
+++ b/core/common/modules/connect/apps/base-app.php
@@ -466,9 +466,8 @@ abstract class Base_App {
 			$args
 		);
 
-		if ( is_wp_error( $response ) && empty( $options['with_error_data'] ) ) {
-			// PHPCS - the variable $response does not contain a user input value.
-			wp_die( $response, [ 'back_link' => true ] ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+		if ( is_wp_error( $response ) ) {
+			return $response;
 		}
 
 		$body = wp_remote_retrieve_body( $response );
--- a/core/common/modules/connect/apps/base-app.php
+++ b/core/common/modules/connect/apps/base-app.php
@@ -466,8 +466,9 @@ abstract class Base_App {
 			$args
 		);
 
-		if ( is_wp_error( $response ) ) {
-			return $response;
+		if ( is_wp_error( $response ) && empty( $options['with_error_data'] ) ) {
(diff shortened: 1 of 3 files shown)


louiswol94 (commented):
Hi @abolfazl-moeini, thank you for the contribution. The issue you found on the Tools page is a legitimate issue worth fixing.

Right now, with the removal of `wp_die()`, it is risky to change the contract of a base method that has existed for 7+ years in the code (unknown blast radius).

We prefer to add a surgical fix for this scenario, so can I ask you to please add `with_error_data => true` to the `http_request()` in `get_quota()` in `cloud-kits.php`.

Thanks!

abolfazl-moeini:
Thanks for the feedback. I've updated the PR to keep `http_request()` unchanged and added `with_error_data => true` to `Cloud_Kits::get_quota()` only. This fixes the Tools page without changing the base method contract.

louiswol94 (approved):

PR #36282, opened 2026-06-22
CI: Two step release tag creation [ED-24451]

davseve:
## 1. Problem & Context

Implements a two-step release workflow to safely orchestrate tag creation with validation gates, addressing ED-24451.

## 2. What Changed (Where)

- `.github/workflows/release-tag-creation.yml`: New CI workflow for manual release tag creation with dry-run capability.

## 3. How It Works

Triggered manually with version input, validates deployment permissions, then delegates to external action (`elementor/elementor-editor-github-actions/actions/release-tag-creation@main`) with git credentials and dry-run flag.

## 4. Risks

Dry-run defaults to `true` but operators must explicitly disable it—potential for accidental non-deployment. External action dependency on `main` branch introduces drift risk; consider pinning to tag.


.github/workflows/release-tag-creation.yml
```diff
@@ -0,0 +1,129 @@
+name: Release Tag Creation
+
+on:
+  workflow_dispatch:
+    inputs:
+      branch:
+        required: true
+        type: string
+        description: 'Target branch to release from (e.g. main or 3.33)'
+        default: main
```
davseve: infered from version

.github/workflows/release-tag-creation.yml
```diff
@@ -0,0 +1,129 @@
...
+        with:
+          filename: "./.github/workflows/config.json"
+          prefix: config
+
+      - name: Test permissions
+        uses: ./.github/workflows/permissions
+        with:
+          ENVIRONMENT: ${{ env.config_deployment_environment }}
+          DEPLOYMENT_PERMITTED: ${{ env.config_deployment_permitted }}
+          DEPLOYMENT_REPOSITORY_OWNER: ${{ env.config_deployment_repository_owner }}
```
davseve: Verify purpose

.github/workflows/release-tag-creation.yml
```diff
@@ -0,0 +1,129 @@
...
+
+      - name: Validate version format
+        run: |
+          if ! [[ "${DESIRED_VERSION}" =~ ^[0-9]+\.[0-9]+\.[0-9]+(-beta[0-9]+)?$ ]]; then
```
davseve: use semver

.github/workflows/release-tag-creation.yml
```diff
@@ -0,0 +1,129 @@
...
+              echo "Version '${DESIRED_VERSION}' is not in the correct format."
+              echo "Version format is not valid, expected: X.Y.Z or X.Y.Z-betaN (e.g. 4.1.0, 5.20.15, 4.1.0-beta1, 5.20.0-beta3)"
+              exit 1
+          fi
+          echo "Version format is valid: ${DESIRED_VERSION}"
+
+      - name: Check if version already exists
+        run: |
+          if git tag -l | grep -q "^v${DESIRED_VERSION}$"; then
+              echo "Version ${DESIRED_VERSION} already exists as a GitHub Release."
```
davseve: Check github releases instead of tags

davseve: Check if prev release version exists.

.github/workflows/release-tag-creation.yml
```diff
@@ -0,0 +1,129 @@
...
+        uses: actions/setup-node@v6
+        with:
+          node-version: 24.15.0
+          cache: npm
+      - name: Install dependencies
+        run: npm ci --ignore-scripts
+      - name: Calculate next version
+        id: version
+        env:
+          CHANNEL: ${{ env.CHANNEL }}
```
davseve: Use input and output instead of ENV

.github/scripts/calculate-next-release-version.js
```diff
@@ -0,0 +1,147 @@
...
+	return { calculatedReleaseVersion: next, latestTag: latest };
+}
+
+// --- main ---
+
+function main() {
+	if ( ! CHANNEL ) {
+		console.error( 'CHANNEL env var is required (stable or beta).' );
+		process.exit( 1 );
+	}
```
davseve: redundant

.github/workflows/release-tag-creation.yml
```diff
@@ -0,0 +1,129 @@
...
+          echo "CHANNEL=${CHANNEL}" >> $GITHUB_ENV
+
+      - name: Install Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version: 24.15.0
+          cache: npm
+      - name: Install dependencies
+        run: npm ci --ignore-scripts
+      - name: Calculate next version
```
davseve: Remove step

.github/workflows/release-tag-creation.yml
```diff
@@ -0,0 +1,129 @@
...
+
+      - name: Verify Release Version
+        id: verify_release_version
+        run: |
+          CALCULATED="${{ steps.version.outputs.CALCULATED_RELEASE_VERSION }}"
+          if [[ "${DESIRED_VERSION}" != "${CALCULATED}" ]]; then
+            echo "Version mismatch: desired=${DESIRED_VERSION}, calculated=${CALCULATED} (channel=${CHANNEL})"
+            exit 1
+          fi
+          echo "Version verified: ${DESIRED_VERSION} matches calculated version (channel=${CHANNEL})."
```
davseve: remove

.github/workflows/update-version-in-tag-files/action.yml
```diff
@@ -0,0 +1,68 @@
...
+      run: |
+        VERSION="${{ inputs.VERSION }}"
+        DRY_RUN="${{ inputs.DRY_RUN }}"
+
+        if [[ "${DRY_RUN}" == "true" ]]; then
+          echo "Dry run — skipping commit."
+          exit 0
+        fi
+
+        git add elementor.php package.json readme.txt
```
davseve:
```suggestion
        git add elementor.php readme.txt
```

.github/workflows/update-version-in-tag-files/action.yml
```diff
@@ -0,0 +1,68 @@
...
+    - name: Update Beta tag in readme.txt
+      shell: bash
+      if: inputs.CHANNEL == 'beta'
+      run: |
+        VERSION="${{ inputs.VERSION }}"
+        sed -i -E "s/Beta tag: .*/Beta tag: ${VERSION}/" readme.txt
+
+        echo "readme.txt beta tag updated:"
+        grep -E "Stable tag:|Beta tag:" readme.txt
+
```
davseve: Should always update beta and stable in readme.txt

.github/workflows/release-tag-creation.yml
```diff
@@ -0,0 +1,129 @@
...
+          VERSION: ${{ env.DESIRED_VERSION }}
+          CHANNEL: ${{ env.CHANNEL }}
+          DRY_RUN: ${{ env.DRY_RUN }}
+
+      - name: Create release tag
+        run: |
+          # Tag the local version-bump commit.
+          # Only the tag is pushed — the version-bump commit itself stays local
+          # so the source branch history remains unchanged.
+          git tag -a "v${DESIRED_VERSION}" -m "Release version ${DESIRED_VERSION}"
```
davseve: should not create a tag

.github/workflows/handle-version-input/action.yml
```diff
@@ -0,0 +1,25 @@
+name: Handle Version Input
```
asafdl: i feel like theres some confusion here around action vs workflow. this isn't a workflow

.github/workflows/handle-version-input/index.mjs
```diff
@@ -0,0 +1,113 @@
...
+function checkTagDoesNotExist(version) {
+  let existingTags;
+
+  try {
+    existingTags = execSync('git tag -l', { encoding: 'utf8' });
+  } catch ( err ) {
+    throw new Error( `Failed to list git tags: ${err.message}` );
+  }
+
+  const tagName = `${version}`;
```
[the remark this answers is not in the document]
asafdl: +1

.github/workflows/handle-version-input/index.mjs
```diff
@@ -0,0 +1,113 @@
...
+    throw new Error(
+      `Version "${version}" is not in the correct format.\n` +
+      'Expected: X.Y.Z or X.Y.Z-betaN  (e.g. 4.1.0, 5.20.15, 4.1.0-beta1, 5.20.0-beta3)'
+    );
+  }
+
+  console.log(`✅ Version format is valid: ${version}`);
+}
+
+function checkTagDoesNotExist(version) {
```
asafdl: tag will fail to create if we have another one with the same name so im not sure how important this is

.github/workflows/handle-version-input/index.mjs
```diff
@@ -0,0 +1,113 @@
...
+  let existingTags;
+
+  try {
+    existingTags = execSync('git tag -l', { encoding: 'utf8' });
```
asafdl: `git ls-remote --tags origin refs/tags/v1.0.0` if youre going to check, iterating over the entire list is redundant

.github/workflows/handle-version-input/index.mjs
```diff
@@ -0,0 +1,113 @@
...
+
+  throw new Error(
+    `Could not determine channel from version "${version}".\n` +
+    'Pre-release identifier must be "beta" (e.g. 4.1.0-beta1).'
+  );
+}
+
+function deriveBranch(version) {
+  const { major, minor } = semver.parse(version);
+  const paddedMinor = String(minor).padStart(2, '0');
```
asafdl: we shouldnt always pad, its going away next version i believe

.github/workflows/handle-version-input/index.mjs
```diff
@@ -0,0 +1,113 @@
...
+  try {
+    const version = getVersion();
+
+    validateFormat(version);
+    checkTagDoesNotExist(version);
+
+    const channel = extractChannel(version);
+    console.log(`✅ Channel resolved to: ${channel}`);
+
+    const branch = deriveBranch(version, channel);
```
asafdl: id actually verify that branch exists here

.github/workflows/update-version-in-tag-files/action.yml
```diff
@@ -0,0 +1,94 @@
...
+    description: 'The Beta tag value written to readme.txt.'
+    value: ${{ steps.capture-readme.outputs.beta_tag }}
+  elementor_php_version:
+    description: 'The version string written to elementor.php.'
+    value: ${{ steps.capture-php.outputs.version }}
+
+runs:
+  using: "composite"
+  steps:
+    - name: Update version in elementor.php
```
asafdl: i feel like, this code is more complex than the code written above, and maybe instead of complex regex patterns and bash scripts we can have a simple js or python script to handle file alterations. but not critical

.github/workflows/update-version-in-tag-files/action.yml
```diff
@@ -0,0 +1,94 @@
...
+        grep -E "Version:|ELEMENTOR_VERSION" elementor.php
+
+        echo "version=${VERSION}" >> "$GITHUB_OUTPUT"
+
+    - name: Update Stable tag in readme.txt
+      shell: bash
+      if: inputs.CHANNEL == 'stable'
+      run: |
+        VERSION="${{ inputs.VERSION }}"
+        LAST_BETA_TAG=$(git ls-remote --tags | awk '{print $2}' | sed 's/^refs\/tags\/v//' | sed 's/^refs\/tags\///' | grep -E '^[0-9]+\.[0-9]+\.[0-9]-beta[1-9]$' | sort -V | tail -n 1)
```
asafdl: we should be looking at git releases, using [gh cli](https://github.com/cli/cli)
not going by tags

.github/workflows/jest.yml
```diff
@@ -45,3 +48,21 @@ jobs:
         run: npm run install:ci
       - name: "Run Jest"
         run: npm run test
+
+  handle-version-input-tests:
```
asafdl: this, should not be commited to repo. we don't need workflow to run action tests.
if you're interested. actions can be wrapped into docker container and tests run before the docker is built.

.github/workflows/release-tag-creation.yml
```diff
@@ -0,0 +1,212 @@
...
+        type: boolean
+        default: true
+        description: Dry run — validate everything but do not push commits or create tag.
+jobs:
+  release-tag-creation:
+    runs-on: ubuntu-22.04
+    permissions:
+      contents: write 
+
+    steps:
```
asafdl, line 22: we can skip a lot of code troubles, if we validate here, that the input is a correct semver

```
 - name: Validate SemVer Format
        run: |
          # Official SemVer 2.0.0 RegEx
          SEMVER_REGEX="^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$"
          
          if [[ ! "${{ inputs.version }}" =~ $SEMVER_REGEX ]]; then
            echo "::error::Invalid SemVer format: '${{ inputs.version }}'. Expected format like 1.2.3 or 1.0.0-beta.1"
            exit 1
          fi
          echo "Input version '${{ inputs.version }}' is valid SemVer."
```

asafdl: we can also enforce v or no v depending on what we want

.github/workflows/release-tag-creation.yml
```diff
@@ -0,0 +1,212 @@
...
+        uses: ./.github/workflows/handle-version-input
+        with:
+          version: ${{ inputs.version }}
+
+      - name: Switch to release branch
+        run: |
+          # Move HEAD to the release branch so version-bump commits land on it.
+          # --no-overlay keeps files from the current tree that don't exist on the
+          # release branch (e.g. .github/workflows/ actions used later in this job).
+          git checkout ${{ steps.handle-version-input.outputs.checkout_branch }}
```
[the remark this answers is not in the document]
asafdl: +1

.github/workflows/handle-version-input/action.yml
davseve: I am not sure, this is an action. AFAIU workflows have event triggers and here we do not have. but lets talk about it

.github/workflows/release-tag-creation.yml
davseve: We don't need the `v` it was a mistake.
I think you wanted me to use SEMVER and not regex...
Using SEMVER does adds some overhead as it needs to be installed.
But I am happy with both ways.

.github/workflows/handle-version-input/index.mjs
davseve: This is the point in which it fails

.github/workflows/handle-version-input/index.mjs
davseve: Good comment, fixed it.

.github/workflows/handle-version-input/index.mjs
davseve: Will the next branch be named 4.3 and not 4.03?

.github/workflows/handle-version-input/index.mjs
davseve: remove function

.github/workflows/handle-version-input/index.mjs
davseve: Check with netanel is relevant

Ntnelbaba:
Let's convert the config file to a GitHub variable please
It will be easier to modify it

.github/workflows/release-tag-creation.yml
```diff
@@ -0,0 +1,48 @@
...
+        description: 'Version to release (e.g. 4.1.0-beta1 or 4.0.11)'
+      dry_run:
+        required: false
+        type: boolean
+        default: true
+        description: Dry run — validate everything but do not push commits or create tag.
+
+jobs:
+  release-tag-creation:
+    runs-on: ubuntu-22.04
```
Ntnelbaba:
```
    runs-on: ubuntu-24.16
```

.github/workflows/release-tag-creation.yml
Ntnelbaba: V not needed
Semver parser is always a good thing, not inventing the wheel
@davseve check if there is npx call directly and use it

.github/workflows/release-tag-creation.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Get Configuration
+        id: configuration
+        uses: rgarcia-phi/json-to-variables@9835d537368468c4e4de5254dc3efeadda183793
+        with:
+          filename: "./.github/workflows/config.json"
```
Ntnelbaba: Change this step to read the json from GitHub variable please

.github/workflows/release-tag-creation.yml
```diff
@@ -0,0 +1,48 @@
...
+    permissions:
+      contents: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Get Configuration
```
Ntnelbaba: All this step is overhead
Did we blocked the deployment from config at some point? No one will push false to code
Move it all to clean GitHub variables, that way we might be able to really control it

.github/workflows/release-tag-creation.yml
```diff
@@ -0,0 +1,48 @@
...
+          ENVIRONMENT: ${{ env.config_deployment_environment }}
+          DEPLOYMENT_PERMITTED: ${{ env.config_deployment_permitted }}
+          DEPLOYMENT_REPOSITORY_OWNER: ${{ env.config_deployment_repository_owner }}
+
+      - name: Release tag creation
+        uses: elementor/elementor-editor-github-actions/actions/release-tag-creation@main
+        with:
+          version: ${{ inputs.version }}
+          dry_run: ${{ inputs.dry_run }}
+          maintain_username: ${{ secrets.MAINTAIN_USERNAME }}
```
Ntnelbaba, line 40: Is this the same as one click? Maybe we need to change it to the executor? WDYT?
A bit of credit 😜

davseve: I really don't want to start changing stuff in other actions...

.github/workflows/release-tag-creation.yml
davseve: Changed to 24.04, Ubuntu-24.16 is not a valid GitHub-hosted runner label

Ntnelbaba (approved):

PR #36291, opened 2026-06-23
Internal: Improve What's New display logic [ED-24716]

avivu:
## Summary
- Adds support for YouTube video embed in What's New notification items via `youtubeEmbedId` field
- Adds `youtubeAutoplay` boolean field to control autoplay (muted, per browser policy)
- Adds `gifSrc` field for GIF media, rendered via the existing thumbnail component
- All fields are optional — existing `imageSrc` behavior is fully unchanged

## How it works
- `youtubeEmbedId` → renders a 16:9 YouTube iframe (full panel width)
- `youtubeAutoplay: true` → appends `?autoplay=1&mute=1` to the embed URL and grants `autoplay` feature policy
- `gifSrc` → renders as `` (GIFs auto-play natively)
- Priority: YouTube > GIF > static image

## Backward compatibility
If `imageSrc` is also provided alongside `youtubeEmbedId` or `gifSrc`, older plugin versions will show it as a static image fallback. If only the new media fields are set with no `imageSrc`, older versions show no media.

## Assets side
Fields can be added to any notification entry in `editor-assets` without a schema change (schema does not enforce `additionalProperties: false`). A separate editor-assets PR will add the fields to the schema with documentation.

## Test plan
- [ ] Notification with `youtubeEmbedId` renders YouTube iframe at correct aspect ratio
- [ ] Notification with `youtubeAutoplay: true` autoplays muted
- [ ] Notification with `youtubeAutoplay: false` requires user click to play
- [ ] Notification with `gifSrc` renders animated GIF
- [ ] Notification with only `imageSrc` (existing behavior) renders unchanged
- [ ] Notification with no media fields renders unchanged

## 1. Problem & Context

Refactor What's New display logic to handle multiple media types (YouTube embeds, GIFs, images) with centralized rendering logic. Ticket: ED-24716.

## 2. What Changed (Where)

- New component `whats-new-item-media.js`: Encapsulates media rendering (YouTube iframe, GIF, image fallback).
- Updated `whats-new-item.js`: Replaced inline thumbnail conditional with `WhatsNewItemMedia` component; removed direct `WhatsNewItemThumbnail` import.

## 3. How It Works

`WhatsNewItemMedia` receives item data and renders: (1) YouTube embed if `youtubeEmbedId` exists (with optional autoplay), (2) GIF/image via `WhatsNewItemThumbnail` if `gifSrc` or `imageSrc` available, (3) null otherwise. Parent component delegates media rendering instead of handling it inline.

## 4. Risks

YouTube embeds now execute autoplay logic client-side—verify autoplay+mute flags work cross-browser. Component lacks prop-types validation (eslint-disable suggests temporary), should add schema before merge if not intended.

[ED-24716]: https://elementor.atlassian.net/browse/ED-24716?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/notifications/assets/js/components/whats-new-item.js
```diff
@@ -30,9 +30,20 @@ export const WhatsNewItem = ( { item, itemIndex, itemsLength, setIsOpen } ) => {
 					{ item.title }
 				</Typography>
 			</WrapperWithLink>
-			{ item.imageSrc && (
+			{ item.youtubeEmbedId ? (
```
Ntnelbaba: Why not to split this condition to 2 different sections and enforce one of them is optional but not both together?
As well as gif?
Like, what is the expected behavior and handle it differently (even a new component)

modules/notifications/assets/js/components/whats-new-item.js
```diff
@@ -30,9 +30,20 @@ export const WhatsNewItem = ( { item, itemIndex, itemsLength, setIsOpen } ) => {
...
+				<Box sx={ { pb: 2 } }>
+					<Box
+						component="iframe"
+						src={ `https://www.youtube.com/embed/${ item.youtubeEmbedId.split( '?' )[ 0 ] }${ item.youtubeAutoplay ? '?autoplay=1&mute=1' : '' }` }
```
Ntnelbaba: Why? I don't get this split logic
Maybe YouTube video ID? Maybe a full link (can use something from other providers or even direct drive link)

modules/notifications/assets/js/components/whats-new-item.js
```diff
@@ -30,9 +30,20 @@ export const WhatsNewItem = ( { item, itemIndex, itemsLength, setIsOpen } ) => {
 					{ item.title }
 				</Typography>
 			</WrapperWithLink>
-			{ item.imageSrc && (
+			{ item.youtubeEmbedId ? (
+				<Box sx={ { pb: 2 } }>
+					<Box
```
Ntnelbaba: Why not a video tag?

modules/notifications/assets/js/components/whats-new-item.js
```diff
@@ -30,9 +30,20 @@ export const WhatsNewItem = ( { item, itemIndex, itemsLength, setIsOpen } ) => {
...
+						component="iframe"
+						src={ `https://www.youtube.com/embed/${ item.youtubeEmbedId.split( '?' )[ 0 ] }${ item.youtubeAutoplay ? '?autoplay=1&mute=1' : '' }` }
+						title={ item.title }
+						allow={ `accelerometer; ${ item.youtubeAutoplay ? 'autoplay; ' : '' }clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share` }
```
Ntnelbaba: Well, many allow list here 😅
See if all of them relevant, better to keep it minimum as possible

modules/notifications/assets/js/components/whats-new-item.js
avivu: Good call — extracted to a dedicated `WhatsNewItemMedia` component. YouTube and image/GIF are now two clear `if` blocks with no nested ternary. Pushed.

modules/notifications/assets/js/components/whats-new-item.js
avivu: YouTube doesn't support the `<video>` tag — it only works via iframe embed. The `<video>` tag is for direct file URLs (mp4/webm), not YouTube streams. Iframe is the only correct option here.

modules/notifications/assets/js/components/whats-new-item.js
avivu: The field is intentionally an embed ID (not a full URL) — we keep the base URL (`youtube.com/embed/`) hardcoded so we fully control what gets embedded. The `.split('?')[0]` strips query params that YouTube appends to share links (e.g. `?si=...`), so copy-pasting a share link ID still works. Added a comment explaining this in the new component.

modules/notifications/assets/js/components/whats-new-item.js
avivu: Trimmed to the minimum needed: `encrypted-media; picture-in-picture` (+ `autoplay` only when `youtubeAutoplay` is true). Removed `accelerometer`, `gyroscope`, `clipboard-write`, and `web-share` — those were from YouTube's full boilerplate but are not required for basic embed playback.

modules/notifications/assets/js/components/whats-new-item-media.js
```diff
@@ -0,0 +1,43 @@
+import { Box } from '@elementor/ui';
+import { WhatsNewItemThumbnail } from './whats-new-item-thumbnail';
+
+export const WhatsNewItemMedia = ( { item } ) => {
+	if ( item.youtubeEmbedId ) {
+		// YouTube does not support the <video> tag — iframe embed is required.
```
Ntnelbaba: Remove comments

modules/notifications/assets/js/components/whats-new-item-media.js
```diff
@@ -0,0 +1,43 @@
...
+	return (
+		<WhatsNewItemThumbnail
+			imageSrc={ mediaSrc }
+			link={ item.link }
+			title={ item.title }
+		/>
+	);
+};
+
+WhatsNewItemMedia.propTypes = {
```
Ntnelbaba: Remove this check, not really helpful, be protective at the component

modules/notifications/assets/js/components/whats-new-item-media.js
avivu: Done — removed both inline comments.

modules/notifications/assets/js/components/whats-new-item-media.js
avivu: Removed the `PropTypes` declaration. Added `/* eslint-disable react/prop-types */` at the top — ESLint enforces this rule and blocks the commit without it. Since `item` is already validated as `PropTypes.object.isRequired` at the parent `WhatsNewItem` level, the check here is redundant.

Ntnelbaba (approved):

PR #36303, opened 2026-06-23
Fix: Loop manual selection shows empty options until user start typing [ED-24594]

MiryamOren:
## 1. Problem & Context

Loop manual selection (ED-24594) displayed empty options until user typed—fix fetches initial results when `minInputLength` is 0, eliminating the empty state UX gap.

## 2. What Changed (Where)

- **post-query.php**: Removed early-exit for empty search term; moved search params into conditional block; changed sort from ID/ASC to modified/DESC.
- **use-query-autocomplete.ts**: Added `useEffect` hook to fetch on mount when `minInputLength === 0`; made `term` optional in `FetchOptionsParams`; removed `setOptions([])` from `updateOptions`.
- **use-query-autocomplete.test.ts**: Added two test cases for initial fetch behavior.

## 3. How It Works

On component mount, if `minInputLength` is 0, `useEffect` triggers an initial fetch with empty term—bypassing the old guard clause. Results populate immediately. Subsequent user input still flows through `updateOptions`. Backend now returns all posts sorted by modification date when no search term provided, instead of requiring a search term.

## 4. Risks

Empty search requests could hit performance if no result caching exists backend-side. Verify `get_posts()` with empty search_term and large datasets behaves acceptably. Test that modified/DESC ordering doesn't break existing filtering assumptions (meta_query, tax_query still work as expected).


modules/wp-rest/classes/post-query.php
```diff
@@ -80,14 +71,17 @@ protected function get( \WP_REST_Request $request ) {
...
-			self::SEARCH_IN_CONTENT_KEY    => $params[ self::SEARCH_IN_CONTENT_KEY ] ?? false,
 			'post_status'                  => $is_public_only ? 'publish' : 'any',
-			'orderby'                      => 'ID',
-			'order'                        => 'ASC',
+			'orderby'                      => 'modified',
+			'order'                        => 'DESC',
 		];
 
+		if ( ! empty( $term ) ) {
+			$query_args['custom_search'] = true;
```
IshayMaya: TBH im not sure what is it mean exactly, is it safe to now have it only true when not empty?

modules/wp-rest/classes/post-query.php
```diff
@@ -80,14 +71,17 @@ protected function get( \WP_REST_Request $request ) {
...
 			'numberposts'                  => $post_count,
 			'suppress_filters'             => false,
-			'custom_search'                => true,
-			'search_term'                  => $term,
-			self::SEARCH_IN_CONTENT_KEY    => $params[ self::SEARCH_IN_CONTENT_KEY ] ?? false,
 			'post_status'                  => $is_public_only ? 'publish' : 'any',
-			'orderby'                      => 'ID',
-			'order'                        => 'ASC',
+			'orderby'                      => 'modified',
+			'order'                        => 'DESC',
```
IshayMaya, line 77: seems like this is the only place that we also change for other queries as well, might be ok but worth checking if its not an issue

packages/packages/libs/editor-controls/src/hooks/use-query-autocomplete.ts
```diff
@@ -47,14 +47,22 @@ export function useQueryAutocomplete( {
...
+		const termStr = term ?? '';
 
-		if ( ! term || ! url || term.length < minInputLength ) {
+		if ( ! url || termStr.length < minInputLength ) {
 			return;
 		}
 
-		debounceFetch( { ...params, term } );
+		debounceFetch( { ...params, term: termStr } );
 	};
```
IshayMaya, line 66:
```suggestion

	const debouncedFetch = useMemo(
		() =>
			debounce( ( term: string ) => {
				fetchOptions( url, { ...params, term } ).then( ( newOptions ) => {
					setOptions( formatOptions( filterExcludedOptions( newOptions, excludeIdSet ) ) );
				} );
			}, 400 ),
		[ url, params, excludeIdSet ]
	);

	const updateOptions = ( term: string | null ) => {
		const termStr = term ?? '';

		if ( ! url || termStr.length < minInputLength ) {
			return;
		}

		if ( termStr === '' ) {
			fetchOptions( url, { ...params, term: '' } ).then( ( newOptions ) => {
				setOptions( formatOptions( filterExcludedOptions( newOptions, excludeIdSet ) ) );
			} );
			return;
		}

		debouncedFetch( termStr );
	};
```

and then in `query-chip-control` pass it to the autocomplete prop already supported)

```
			<Autocomplete
				onOpen={ () => {
					if ( minInputLength === 0 ) {
						updateOptions( '' );
					}
				} }
```

modules/wp-rest/classes/post-query.php
MiryamOren: Yes, it's a change for all query controls - I asked Elli about it and he approved

packages/packages/libs/editor-controls/src/hooks/use-query-autocomplete.ts
MiryamOren: Why not support it for any control with `minInputLength === 0`?

modules/wp-rest/classes/post-query.php
MiryamOren: It's just means we won't customize the query (see `modules/wp-rest/classes/post-query.php:32`) - I can actually revert this specific line, as we check there if the term is empty

IshayMaya (approved):

PR #36334, opened 2026-06-25
Post query search ordering breaks after empty-term browse change

Svitlana-Dykun:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #


modules/wp-rest/classes/post-query.php
```diff
@@ -73,13 +73,17 @@ protected function get( \WP_REST_Request $request ) {
 			'suppress_filters'             => false,
 			'custom_search'                => true,
 			'post_status'                  => $is_public_only ? 'publish' : 'any',
-			'orderby'                      => 'modified',
-			'order'                        => 'DESC',
 		];
```
ronros-elementor, line 76: We can also consider doing it like that -
```suggestion
		'orderby'                      => [ 'modified' => 'DESC', 'ID' => 'DESC' ],
```

Because prior to [this change](https://github.com/elementor/elementor/pull/36303) we had IDs to rely on, and your fix brings back the use of IDs only in the absence of a search term

Logic-wise - we'd probably like to show the user the newest posts first, and the post ID is a good proxy of this data, as we use auto-incremented integer IDs, so highest to lowest == newest to oldest (in terms of creation only)

I think we should not mix the existence of a search term with the use of IDs, but instead maintain the "last-modified is first" approach, with a fallback to the IDs

wdyt?

PR #36392, opened 2026-07-01
Fix: Add widget content preserving when pro is disabled [ED-24489]

murzin-nikita:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #

## 1. Problem & Context

When Pro is disabled, promotion widgets (e-form, collection-loop) lose their child elements and settings on save because the editor strips them. This PR preserves that content via a data restoration hook that runs when Pro is absent.

## 2. What Changed (Where)

- **Promotion classes** (`atomic-form-promotion.php`, `collection-loop-promotion.php`): Added `Preserves_Children_Subtree` trait to use `Preserved_Element` as fallback child type during saves.
- **Atomic_Form**: Exposed `get_base_props_schema()` method for promotion to reuse schema.
- **Pro_Promotion_Data_Preservation**: New class that intercepts document saves, maps stored promotion elements, and restores stripped children/settings.
- **Preserved_Element**: New passthrough element type that preserves nested structure and data verbatim.
- **Module**: Registered preservation hooks in `register_hooks()`.
- **Tests**: Added coverage for promotion schema reuse, preserved element fidelity, and cross-save restoration.
- **Editor styles**: Fixed hover state contrast (minor).

## 3. How It Works

On document save, `Pro_Promotion_Data_Preservation` identifies promotion-type elements in current DB state, stores their children and settings. After the stripped save completes, it restores those stored values into empty promotions. `Preserves_Children_Subtree` trait makes promotion widgets return `Preserved_Element` during saves (when Pro absent), which serializes data unchanged instead of filtering it.

## 4. Risks

Race condition if multiple saves occur before first restore completes—mitigated by keying storage on element ID. Preservation data only exists after first Pro-enabled save; fresh promotions created without Pro stay empty (acceptable, documented in test).

### DIFF
--- a/modules/atomic-widgets/elements/promotions/preserved-element.php
+++ b/modules/atomic-widgets/elements/promotions/preserved-element.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace Elementor\Modules\AtomicWidgets\Elements\Base;
+namespace Elementor\Modules\AtomicWidgets\Elements\Promotions;
 
 use Elementor\Element_Base;
 
--- a/modules/atomic-widgets/elements/promotions/pro-promotion-data-preservation.php
+++ b/modules/atomic-widgets/elements/promotions/pro-promotion-data-preservation.php
@@ -11,12 +11,15 @@ if ( ! defined( 'ABSPATH' ) ) {
 
 class Pro_Promotion_Data_Preservation {
 
+	const EMAIL_ACTION_COUNT = 2;
+
 	public function register_hooks(): void {
 		if ( Utils::has_pro() ) {
 			return;
 		}
 
 		add_filter( 'elementor/document/save/data', fn ( $data, $document ) => $this->preserve( $data, $document ), 10, 2 );
+		add_filter( 'elementor/atomic/form/email_action_count', fn ( $count ) => max( $count, self::EMAIL_ACTION_COUNT ) );
 	}
 
 	public function preserve( $data, $document ) {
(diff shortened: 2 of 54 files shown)


modules/atomic-widgets/elements/promotions/preserved-element.php
```diff
@@ -0,0 +1,40 @@
...
+		if ( ! empty( $widget_type ) ) {
+			return $widget_type;
+		}
+
+		$el_type = $this->get_data( 'elType' );
+
+		return empty( $el_type ) ? 'e-preserved-element' : $el_type;
+	}
+
+	protected function _get_default_child_type( array $element_data ) {
```
RobiFerentz, line 25: I'll add pretty much the only thing Cursor found 😄 :
nit — The trait name implies “unregistered” only, but it overrides _get_default_child_type() for all children, including registered core containers (e.g. e-flexbox). That still works (the test covers it) and favors verbatim preservation, but a parent-delegate-then-fallback pattern would be more precise if registered children should go through normal atomic save parsing.

modules/atomic-widgets/elements/promotions/pro-promotion-data-preservation.php
```diff
@@ -0,0 +1,108 @@
...
+			if ( $id && $is_promotion && ! empty( $element['elements'] ) ) {
+				$map[ $id ] = [
+					'settings' => $element['settings'] ?? [],
+					'elements' => $element['elements'],
+				];
+
+				continue;
+			}
+
+			$this->map_promotion_elements( $element['elements'] ?? [], $promotion_types, $map );
```
ronenboxer, line 82: Wouldn't `$element['elements']` be empty by this point for sure?

mykytamurzin: We need it for first iteration of recursion

ronenboxer (approved):

PR #36421, opened 2026-07-05
Feature: Add children dependencies reconciler to editor-elements [ED-24815]

IshayMaya:
## Summary
- Adds generic `children-dependencies` module to `@elementor/editor-elements`
- Implements pre-super `reconcileInitialChildren` and post-super `bindSettingsReconcile` with session-backed stash
- Dispatches `elementor/navigator/refresh-children` after silent child mutations
- Includes jest coverage for reconcile, bind, stash, insert-index, and evaluate-when

## Test plan
- [ ] `cd packages && npx tsc --noEmit`
- [ ] `cd packages && npx jest packages/libs/editor-elements/src/children-dependencies`

## 1. Problem & Context

Introduces schema-driven reconciliation for conditional child elements—when settings change, dependent children attach/detach and persist via session storage. Enables declarative UI composition without imperative wiring.

## 2. What Changed (Where)

| File | Change |
|------|--------|
| `bind-settings-reconcile.ts` | Entry point: subscribes to settings changes, attaches/detaches children matching `ChildDependencyRule` conditions |
| `reconcile-initial-children.ts` | Initial sync: ensures element tree matches rules at load time |
| `stash.ts` | Session storage layer for preserving child state across attach/detach cycles |
| `resolve-insert-index.ts` | Positioning logic: handles `first`, `last`, `index`, `after_type`, `before_type` insertion strategies |
| `types.ts` | Adds `ElementPosition`, `ChildDependencyRule` types and `children_dependencies` config field |
| `index.ts` | Public exports; `package.json` adds `@elementor/session` dependency |

## 3. How It Works

**Entry:** `bindSettingsReconcile()` attaches listener to element settings model.

**Happy path:** On setting change, `evaluateWhen()` checks if dependency is met. If previous state differs from current, call `attachChildFromRule()` or `detachChildFromRule()`. Attach uses stashed data (restored from session) or `default_model` or minimal `{elType}`, inserted at position resolved by `resolveInsertIndex()`. Detach stores snapshot if `rule.stash=true`. Both trigger `elementor/navigator/refresh-children` event to sync Structure panel.

**Edge cases:** Handles missing models gracefully (returns no-op), fails gracefully if add/remove ops fail (preserves stash), idempotent (checks if child already present before attaching).

**Initialization:** `reconcileInitialChildren()` applies same logic to initial element tree before reactive subscriptions activate.

## 4. Risks

**Session storage collisions:** Stash key format `elementor/editor-state/{elementId}/children-deps/{childType}` could collide if element IDs aren't globally unique—mitigated by reliance on existing ID uniqueness assumption.

**Memory leak potential:** `on('change')` listener never unsubscribed if detach not called—mitigated by caller responsibility; tests verify cleanup works.

**Navigator refresh reliability:** Relies on custom window event; fails silently in non-browser environments but guards with `typeof window` checks.


packages/packages/libs/editor-elements/src/sync/resolve-insert-index.ts
ronros-elementor, line 1: could this be a util that's not dependency-specific, like my comment for the PHP implementation?

packages/packages/libs/editor-elements/src/children-dependencies/utils.ts
ronros-elementor, line 1: is this wrapper really necessary? it seems longer then the actual logic

IshayMaya: not a must, but i also have some other utils in next pr, moved it to local utils file (wil be populated with more next)

ronros-elementor (approved):

PR #36440, opened 2026-07-07
Fix: Moving the parent of an atomic form shows invalid message [ED-23858]

RobiFerentz:
## 1. Problem & Context

Moving an atomic form parent was incorrectly blocked. Logic was checking if containers held form fields, but not distinguishing between loose fields (invalid to move outside forms) and fields within form containers (valid moves). ED-23858 fixes this.

## 2. What Changed (Where)

`enforce-form-ancestor-commands.ts`: Updated `blockFormFieldMove()` to check if containers are form fields *not* within a form, then validates target context.

## 3. How It Works

Changed validation from checking `hasElementTypes(container, FORM_FIELD_ELEMENT_TYPES)` to `!hasElementType(container, FORM_ELEMENT_TYPE) && hasElementTypes(container, FORM_FIELD_ELEMENT_TYPES)`. Now only blocks moves when loose field elements target non-form destinations, allowing atomic form parents with field children to move freely.

## 4. Risks

Low. Conditional logic is more specific—reduces false positives without relaxing safety constraints on genuinely orphaned form fields.

### DIFF
--- a/packages/packages/core/editor-canvas/src/form-structure/enforce-form-ancestor-commands.ts
+++ b/packages/packages/core/editor-canvas/src/form-structure/enforce-form-ancestor-commands.ts
@@ -5,9 +5,11 @@ import { __ } from '@wordpress/i18n';
 import {
 	clipboardRootsAreAtomicForms,
 	type CreateArgs,
+	FORM_ELEMENT_TYPE,
 	FORM_FIELD_ELEMENT_TYPES,
 	getArgsElementType,
 	hasClipboardElementTypes,
+	hasElementType,
 	hasElementTypes,
 	isWithinForm,
 	type MoveArgs,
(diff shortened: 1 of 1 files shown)


packages/packages/core/editor-canvas/src/form-structure/enforce-form-ancestor-commands.ts
```diff
@@ -59,7 +61,9 @@ function blockFormFieldMove( args: MoveArgs ): boolean {
 	const { containers = [ args.container ], target } = args;
 
 	const hasFormFieldElement = containers.some( ( container ) =>
-		container ? hasElementTypes( container, FORM_FIELD_ELEMENT_TYPES ) : false
+		container
+			? ! hasElementType( container, FORM_ELEMENT_TYPE ) && hasElementTypes( container, FORM_FIELD_ELEMENT_TYPES )
```
Ntnelbaba, line 66: I would add here a test cases to justify the BL and a comment of the why

RobiFerentz: @Ntnelbaba  comment added, but testing will be done in pro playwright tests, since the atomic form tests are there, once this is merged

Ntnelbaba (approved):

PR #36425, opened 2026-07-05
Internal: Static resources + move to server - style best practices [ED-24840]

eavichay-elementor:
merge after #36416 !

## 1. Problem & Context

Moving style best practices from client-side hardcoded prompt to server-side static resource, enabling server registration and MCP protocol exposure per ED-24840. Decouples design guidelines from codebase for easier updates.

## 2. What Changed (Where)

| File | Change |
|------|--------|
| `modules/mcp/abilities/simple-resource-ability.php` | New ability class for serving static markdown resources via MCP protocol |
| `modules/mcp/module.php` | Register Simple_Resource_Ability for style best practices; add to server abilities list |
| `modules/mcp/rest-api/mcp-proxy-rest-api.php` | Split tool/resource handling; add GET endpoint for resource retrieval; instantiate best-practices resource |
| `modules/mcp/static-resources/style/best-practices.md` | Extract from prompt constant into standalone markdown file |
| `packages/.../canvas-mcp.ts` | Import and initialize best-practices resource |
| `packages/.../best-practices-resource.ts` | New client-side resource wrapper; proxies GET requests to server REST endpoint |
| `packages/.../widgets-schema-resource.ts` | Remove duplicate best-practices logic; update URI reference |
| `packages/.../web-mcp-adapter.ts` | Fix URI example in description (styles → style) |

## 3. How It Works

Server creates `Simple_Resource_Ability` instance serving markdown from disk at `elementor://style/best-practices`. MCP module registers ability and exposes in server capabilities. Client-side `initBestPracticesResource` registers resource handler that GETs from `mcp-proxy` REST endpoint (passing `uri` param), receives markdown response. `Mcp_Proxy_REST_API` routes GET requests to `handle_resource()`, looking up URI in resource map, executing handler, returning result.

## 4. Risks

**Duplication risk**: `BEST_PRACTICES_URI` now defined in both `best-practices-resource.ts` and `widgets-schema-resource.ts` (line 62). Consider single source of truth. **File permission**: Markdown file served without additional validation—relies on existing file-exists check in ability constructor and `edit_posts` capability check.

_Generated by LinearB AI and added by gitStream._
<sub>AI-generated content may contain inaccuracies. Please verify before using.
💡 **Tip:** You can customize your AI Description using **Guidelines** [Learn how](https://docs.gitstream.cm/automation-actions/#describe-changes)</sub>

PR #36444, opened 2026-07-07
Internal: Migrate variables client-side MCP to server via mcp-proxy [ED-24836]

Omerisra6:
## Summary

Migrates the variables MCP integration (list resource + `manage-global-variable` tool + guide resource) from the browser to server-side abilities exposed via both `mcp-proxy` (for the editor) and WP MCP (for headless agents). The client becomes a thin fetch wrapper — same pattern as `style-best-practices` in #36425 (ED-24840).

Jira: [ED-24836](https://elementor.atlassian.net/browse/ED-24836)

Depends on: #36425 (ED-24840) — shares the `mcp-proxy` REST infrastructure.

## Motivation

Before this PR, the variables MCP surface lived entirely FE-side (`packages/core/editor-variables/src/mcp/*`). That meant:

- Headless agents (no editor open) couldn't list or mutate variables.
- Editor-open agents ran validation, cache reads, and mutation logic in the browser.

This aligns variables with the broader "MCP abilities live on the server, client is a thin proxy" direction.

### Server (PHP) — new abilities in `modules/mcp/abilities/`

- **`List_Variables_Ability`** — resource `elementor://global-variables`, returns the active kit's non-deleted v4 variables.
- **`Manage_Variable_Ability`** — tool `manage-global-variable` (create/update/delete). Delegates to `Variables_Service`; maps its exceptions to the same error codes/statuses as `modules/variables/classes/rest-api.php`. Permission: `manage_options` (parity with the variables REST API).
- **`Manage_Variable_Guide_Ability`** — resource `elementor://variables/tools/manage-global-variable-guide`. Serves the guide with Pro/non-Pro branching resolved server-side.

All three are registered with WP MCP (`module.php`) and routed through the mcp-proxy REST endpoint (`mcp-proxy-rest-api.php`).

### Client (TS) — reduced to fetch wrappers

- `packages/core/editor-variables/src/mcp/variables-resource.ts` — `GET mcp-proxy?uri=…`. **Kept** the `STORAGE_UPDATED_EVENT` + `document/save/update` listeners so editor-connected MCP clients still get `sendResourceUpdated` pings.
- `packages/core/editor-variables/src/mcp/manage-variable-tool.ts` — `POST mcp-proxy` with `{ tool, input }`; guide handler `GETs` from proxy. `isProActive()` still gates the tool schema enum (UX concern for the tool declaration only).
- `variable-tool-prompt.ts` — reduced to a single URI constant (guide text now lives server-side).

### Behavior parity note (see PR comment below)

Client-side MCP validation for value/type formats, font availability, and Pro gating was **removed** rather than ported. Details in a review comment.

## Test plan

- [x] `npx jest packages/core/editor-variables` — **297/297 passing** (39 suites).
- [x] `npx eslint packages/core/editor-variables/src/mcp` — clean.
- [x] `php -l` on all changed PHP files — clean.
- [x] New PHPUnit: `test-list-variables-ability.php`, `test-manage-variable-ability.php` covering happy paths, required-field errors, and `Variables_Service` exception mapping.
- [x] New Jest: `variables-resource.test.ts` covers proxy fetch shape + notification wiring. Rewritten `manage-variable-tool.test.ts` covers proxy POST payload for each action, guide GET, error propagation, and Pro enum gating.
- [ ] Manual: with the editor open, MCP client fetches `elementor://global-variables` → matches previous shape (minus the never-used `version: 'v4'` field).
- [ ] Manual: headless MCP client (no browser) can call `manage-global-variable` and see the mutation reflected in the editor after reload.

## Files

- New: `modules/mcp/abilities/{list-variables-ability,manage-variable-ability,manage-variable-guide-ability}.php`
- New tests: `tests/phpunit/elementor/modules/mcp/test-{list,manage}-variable-ability.php`, `packages/packages/core/editor-variables/src/mcp/__tests__/variables-resource.test.ts`
- Modified: `modules/mcp/module.php`, `modules/mcp/rest-api/mcp-proxy-rest-api.php`, `packages/packages/core/editor-variables/src/mcp/{variables-resource,manage-variable-tool,variable-tool-prompt}.ts` and its test.

Ref: ED-24836

[ED-24836]: https://elementor.atlassian.net/browse/ED-24836?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## 1. Problem & Context

Migrating client-side MCP variable operations to server-side execution via mcp-proxy. Centralizes validation and business logic on the server, improving security and maintainability (ED-24836).

## 2. What Changed (Where)

| Component | Change |
|-----------|--------|
| **PHP Backend** | Three new abilities: `List_Variables_Ability`, `Manage_Variable_Ability`, `Manage_Variable_Guide_Ability`. Guide content moved from client prompt generation to server. |
| **REST API** | `mcp-proxy-rest-api.php` wired new abilities as tools/resources; removed hardcoded style guide. |
| **Client (TS)** | `manage-variable-tool.ts` stripped validation/business logic; now thin HTTP proxy to server. `variables-resource.ts` fetches data via mcp-proxy instead of local state. |
| **Tests** | Rewrote test suites: client tests verify proxy behavior only; added server-side ability tests. |
| **Module** | Registered three new abilities in MCP module. |

## 3. How It Works

Client MCP tool/resource handlers now POST/GET to `elementor/v1/mcp-proxy` endpoint instead of executing logic locally. Server abilities validate input, perform operations, and return results. Guide content generation and variable list filtering happen server-side. Validation errors map to `WP_Error` responses with appropriate HTTP status codes.

## 4. Risks

**State sync:** Client resource cache invalidation via `STORAGE_UPDATED_EVENT` works only if server mutations trigger this event reliably—verify fire-and-forget POST doesn't lose sync signals. **Pro gating:** Size variable Pro checks now server-side; inconsistency if license changes mid-request is low-risk but worth documenting. **Error transparency:** HTTP failures in proxy layer could mask service exceptions; add logging.


Omerisra6:
### Note on removed client-side validation

The deleted `manage-variable-tool.ts` performed four categories of validation before calling the FE `service.create/update/delete`:

| Category | Where it lived | Kept? |
|---|---|---|
| Label allowed chars (regex `[a-zA-Z0-9_-]+`) + all-dashes rejection | FE tool + FE UI (`validateLabel` in `utils/validations.ts`) | **Removed** |
| Value/type format checks (`COLOR_PATTERN`, `LENGTH_UNIT_PATTERN` — e.g. reject `24px` as font, `Roboto` as color) | FE tool only | **Removed** |
| Font-family availability against `getFontConfigs()` | FE tool only | **Removed** |
| Pro gating for `global-size-variable` / `global-custom-size-variable` | FE tool only | **Removed** |

#### Why removed rather than ported

I initially ported them to `Manage_Variable_Ability`. On review, we noticed these checks **never existed on the server**: hitting `POST /elementor/v1/variables/create` directly with `{ type: 'global-color-variable', value: 'Roboto' }` has always succeeded and created a garbage variable. Porting them to the MCP ability would make the MCP path stricter than the direct REST path — perpetuating an inconsistency rather than a migration.

`Variables_Service` + `Variable::validate()` still enforces:
- Label has no spaces
- Label length ≤ 50
- Duplicate labels (`assert_label_is_unique`)
- Variables count limit
- Type-change restrictions (`Type_Mismatch`)

Everything else is now guidance in the tool description + guide resource (which the LLM reads before calling), not hard enforcement — same behavior as the existing REST API surface.

#### Follow-up

The correct home for the value/type/font checks is `Variable::validate()` (or a value-object per type) so both MCP and REST agree. That's a separate ticket — this PR stays a pure migration.

modules/mcp/abilities/manage-variable-guide-ability.php
```diff
@@ -0,0 +1,91 @@
...
+			[
+				'mcp' => [
+					'type' => 'resource',
+					'uri' => self::URI,
+					'public' => true,
+					'mimeType' => 'text/plain',
+					'description' => __( 'Detailed guide for using the manage-global-variable tool', 'elementor' ),
+				],
+			],
+			fn() => current_user_can( 'edit_posts' )
```
ronros-elementor: I think we should match with the actual permissions for managing variables, shouldn't we?

in [variables/classes/rest-api](https://github.com/elementor/elementor/blob/27a14ca1c520ea586d1bda25c9653012669f4a6a/modules/variables/classes/rest-api.php#L46) they use

```
current_user_can( 'manage_options' );
```

ronros-elementor (approved):

PR #36468, opened 2026-07-12
Tweak: Add markdown rendering support for pro widgets [ED-24724]

Ntnelbaba:
## Summary
- Add a markdown-rendering flag so CSS generation and enqueue are skipped while serving markdown output
- Update widget markdown rendering to avoid leaking HTML from default widget content and nested elements
- Harden post CSS enqueue in the editor by detecting editor requests reliably and filtering unregistered style dependencies (fixes WP 6.9.1 notices)
- Skip caching/restoring `elementor-post-*` styles when `elementor-frontend` is not registered

## Test plan
- [ ] Enable the markdown rendering experiment and request a page with `?format=markdown` — verify markdown output is returned without frontend CSS side effects
- [ ] Open the Elementor editor for a built page — confirm no `elementor-frontend` dependency notice appears
- [ ] Open the editor preview iframe — confirm the page renders with correct kit and post styles
- [ ] Verify nested element widgets do not inject HTML into markdown output

## Jira
[ED-24724](https://elementor.atlassian.net/browse/ED-24724)

[ED-24724]: https://elementor.atlassian.net/browse/ED-24724?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## 1. Problem & Context

Markdown rendering for pro widgets needs conditional style enqueuing and render-mode awareness. Previously, all cached styles were enqueued blindly and markdown rendering bypassed widget render hooks, breaking third-party widget support and filter chains.

## 2. What Changed (Where)

- **Widget_Content_Render_Mode** (new): Static context manager for markdown vs. normal rendering modes with execute-as wrapper
- **Document.php**: Filter cached styles via `should_enqueue_cached_style()` to skip elementor-post handles without frontend dependency
- **Editor.php**: Added `is_editor_request()` to detect editor context independent of edit_mode overrides
- **CSS Base/Post**: Skip enqueue when rendering markdown; filter dependencies to registered handles only
- **Widget_Base**: Enhanced `render_markdown()` to run full widget pipeline with render hooks, mode-aware filters
- **Markdown_Renderer**: Wrap render calls in `execute_while_rendering_markdown()` to propagate mode context

## 3. How It Works

Rendering sets `Widget_Content_Render_Mode::MARKDOWN` via `execute_as()`, which wraps callback execution and guarantees reset even on exception. CSS enqueue checks this flag and skips (no inline styles during markdown). Widgets receive before/render hooks with mode context, allowing third-party widgets to adapt. Dependencies filter to only registered handles to avoid broken link errors.

## 4. Risks

**Dependency registration gaps**: If third-party code registers styles conditionally post-init, they won't enqueue. Mitigation: filter already documented in code comments.

**Exception safety**: `execute_as()` uses try-finally—solid. Mode persists if callback doesn't throw but crashes elsewhere, but surface area is now contained.


Ntnelbaba:
Addressed CI and review feedback in 8328593:

- Fixed PHPCS `self::` violations in `markdown-render/module.php`
- Removed unused `render_elements_data()` (no callers in repo)
- Added test verifying the markdown rendering flag resets after a render exception

Playwright shard 5 failures (`checklist`, `link-dependencies`) look like editor-load timeouts unrelated to this diff; re-ran failed jobs and merged latest `main`.

includes/base/widget-base.php
```diff
@@ -698,11 +698,7 @@ public function render_plain_content() {
 	}
 
 	public function render_markdown(): string {
-		ob_start();
-		$this->render_content();
-		$html = ob_get_clean();
-
-		return wp_strip_all_tags( $html );
+		return '';
```
Ntnelbaba: Convert to MD to support 3rd party

modules/markdown-render/markdown-helpers.php
Ntnelbaba, line 1: Why? simplify

modules/floating-buttons/base/widget-floating-bars-base.php
```diff
@@ -1551,4 +1551,5 @@ protected function render(): void {
 
 		$render_strategy->render();
 	}
+
```
Ntnelbaba: Remove change

modules/markdown-render/module.php
```diff
@@ -87,23 +97,29 @@ public function maybe_serve_markdown() {
...
-		status_header( 200 );
-		header( 'Content-Type: text/markdown; charset=utf-8' );
-		header( 'X-Content-Type-Options: nosniff' );
-		echo $markdown; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
-		exit;
+			nocache_headers();
+			status_header( 200 );
+			header( 'Content-Type: text/markdown; charset=utf-8' );
+			header( 'X-Content-Type-Options: nosniff' );
+			echo $markdown; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
Ntnelbaba: We have wrapper util for that

modules/markdown-render/module.php
```diff
@@ -87,23 +97,29 @@ public function maybe_serve_markdown() {
...
 			}
-		}
 
-		nocache_headers();
-		status_header( 200 );
-		header( 'Content-Type: text/markdown; charset=utf-8' );
-		header( 'X-Content-Type-Options: nosniff' );
-		echo $markdown; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
-		exit;
+			nocache_headers();
```
Ntnelbaba: Add explanation to all of this block about the why

Ntnelbaba (commented):
Make sure all tests covers the new behaviors (when doing this work, do not change the BL)

core/editor/editor.php
```diff
@@ -212,6 +212,43 @@ public function redirect_to_new_url() {
...
+	 * Whether the current request targets the Elementor editor.
+	 *
+	 * Unlike `is_edit_mode()`, this is not affected by temporary `set_edit_mode()` overrides.
+	 *
+	 * @since 4.1.0
+	 * @access public
+	 *
+	 * @return bool Whether the current request targets the Elementor editor.
+	 */
+	public function is_editor_request() {
```
matipojo, line 225: It's not a reliable way because if someone adds more actions, they will not appear automatically in this list. Also, it breaks the modularity. 

Maybe use https://developer.wordpress.org/reference/classes/wp_screen/

matipojo: Okay, I see. It's an extraction of old code.

core/files/css/base.php
```diff
@@ -606,6 +612,29 @@ public function add_controls_stack_style_rules( Controls_Stack $controls_stack,
 	 */
 	abstract protected function get_file_handle_id();
 
+	protected function should_skip_enqueue(): bool {
+		if ( Markdown_Render_Module::is_rendering_markdown() ) {
+			return true;
+		}
```
matipojo: It would be better to introduce a render mode enum and check the current render mode.

includes/base/widget-base.php
```diff
@@ -698,11 +698,36 @@ public function render_plain_content() {
...
+
+		if ( '' === $content ) {
+			return '';
+		}
+
+		return \Elementor\Modules\MarkdownRender\Html_To_Markdown::convert( $content );
+	}
+
+	protected function get_render_content_for_markdown(): string {
+		do_action( 'elementor/widget/before_render_content', $this );
```
matipojo:
```suggestion
		do_action( 'elementor/widget/before_render_content', $this, [ 'mode' => 'markdown' ] );
```

includes/base/widget-base.php
```diff
@@ -698,11 +698,36 @@ public function render_plain_content() {
...
+			$this->render_by_mode();
+		}
+
+		$widget_content = ob_get_clean();
+
+		if ( '' === $widget_content ) {
+			return '';
+		}
+
+		return apply_filters( 'elementor/widget/render_content', $widget_content, $this );
```
matipojo:
```suggestion
		return apply_filters( 'elementor/widget/render_content', $widget_content, $this, [ 'mode' => 'markdown' ] );
```

core/editor/editor.php
```diff
@@ -258,13 +266,40 @@ public function is_edit_mode( $post_id = null ) {
...
+	 *
+	 * @since 1.0.0
+	 * @access public
+	 *
+	 * @param int $post_id Optional. Post ID. Default is `null`, the current
+	 *                     post ID.
+	 *
+	 * @return bool Whether the edit mode is active.
+	 */
+	public function is_edit_mode( $post_id = null ) {
```
matipojo, line 287: Let's return to the original code lines to reduce the diff.

matipojo (approved):

PR #36473, opened 2026-07-12
Fix: Component overridable props are not updated on enter to edit mode [ED-23678]

ronros-elementor:
## 1. Problem & Context

Component overridable props cache stale during editing sessions when other users modify shared components. Force-refresh on component edit entry ensures current state.

## 2. What Changed (Where)

- `load-components-overridable-props.ts`: Added `Options` type with `force` flag to bypass cache guard; filters IDs conditionally.
- `switch-to-component.ts`: Calls `loadComponentsOverridableProps` with `force: true` after document switch.
- Test coverage: New tests for force-refresh behavior; mocked integration test in switch handler.

## 3. How It Works

When entering component edit mode, `switchToComponent` invalidates document data, then explicitly calls `loadComponentsOverridableProps([componentId], { force: true })` to bypass the existing cache check. Default behavior (no force) filters unloaded IDs; force flag bypasses this filter entirely, ensuring fresh props fetch regardless of cached state.

## 4. Risks

None identified. Cache bypass is scoped to intentional edit-mode entry; normal loads retain caching. Tests validate both paths.


packages/packages/core/editor-components/src/store/actions/load-components-overridable-props.ts
```diff
@@ -3,14 +3,23 @@ import { __dispatch as dispatch, __getState as getState } from '@elementor/store
 import { apiClient } from '../../api';
 import { selectIsOverridablePropsLoaded, slice } from '../store';
 
-export async function loadComponentsOverridableProps( componentIds: number[] ) {
-	const unloadedIds = componentIds.filter( ( id ) => ! selectIsOverridablePropsLoaded( getState(), id ) );
+type Options = {
+	// Bypasses the "already loaded" guard, forcing a re-fetch for the given ids.
+	// Used when we know the cached data may be stale (e.g. entering a component's edit mode),
+	// since other users may have changed the component's overridable props during the session.
+	force?: boolean;
```
Omerisra6, line 10: when does the cache is relevant?

packages/packages/core/editor-components/src/utils/switch-to-component.ts
```diff
@@ -11,12 +13,18 @@ export async function switchToComponent(
 
 	invalidateDocumentData( componentId );
 
-	await switchToDocument( componentId, {
-		selector,
-		mode: 'autosave',
-		setAsInitial: false,
-		shouldScroll: false,
-	} );
+	await Promise.all( [
```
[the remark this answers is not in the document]
Omerisra6: check that case

packages/packages/core/editor-components/src/store/actions/load-components-overridable-props.ts
ronros-elementor: when you drag and drop a component onto the canvas, which was already loaded - there's no need to re-fetch it

Omerisra6 (approved):

PR #36480, opened 2026-07-12
Internal: Add to_json_schema() method to prop types [ED-24826]

Omerisra6:
## Summary

Add JSON Schema generation capability to all prop types via the new `Has_Json_Schema_Meta` concern. This allows prop types to be serialized as JSON Schema for MCP/LLM consumption.

## Changes

- Add `Has_Json_Schema_Meta` trait with `to_json_schema()` method
- Add `to_json_schema()` to `Prop_Type` contract
- Implement `to_json_schema()` in all prop type classes:
  - `Array_Prop_Type`, `Object_Prop_Type`, `Plain_Prop_Type`, `Unknown_Prop_Type`
  - `Boolean_Prop_Type`, `Number_Prop_Type`, `String_Prop_Type`
  - `Union_Prop_Type`
- Update `Mock_Prop_Type` for testing
- Add comprehensive tests for JSON schema generation

## Test plan

- [x] Unit tests added for to_json_schema() on all prop types
- [ ] Verify existing prop type functionality is not affected

Ref: ED-24826

Made with [Cursor](https://cursor.com)

## 1. Problem & Context

Adds JSON Schema serialization to prop types to enable LLM integration. Enables programmatic schema generation for atomic widgets' type system without frontend code duplication.

## 2. What Changed (Where)

| Component | Change |
|-----------|--------|
| Base prop types (Plain, Array, Object, Union) | Added `to_json_schema()` implementing enveloped `{$type, value}` structure |
| Primitives (String, Number, Boolean) | Implemented schema generation with enum support for String |
| Special types | Unknown returns `[]`; Classes returns array schema; Overridable wraps origin schema |
| Concerns | New `Has_Json_Schema_Meta` trait extracts description from meta |
| Contract & mocks | Added `to_json_schema()` and `get_aliases()` to Prop_Type interface; updated implementations |
| Tests | 250 lines covering all prop type schemas with fixture helpers |

## 3. How It Works

Each prop type wraps its value schema in a discriminated union: `{$type: <key>, value: <schema>}`. Base Plain_Prop_Type provides `wrap_json_schema()` helper for primitives. Container types (Array, Object, Union) recursively call `to_json_schema()` on children. `Has_Json_Schema_Meta` enriches schemas with stored description metadata. Unknown_Prop_Type short-circuits to empty object for graceful degradation.

## 4. Risks

[...]

PR #36481, opened 2026-07-12
Internal: Introduce Dynamic_Prop_Type for dynamic tag references [ED-24826]

Omerisra6:
## Summary

Add `Dynamic_Prop_Type` class for representing dynamic tag references in prop types. This enables widgets to declare which props support dynamic content.

## Changes

- Add `to_json_schema()` to `Dynamic_Prop_Type`
- Register `Dynamic_Prop_Type` in dynamic-tags-module
- Update `dynamic-tags-resource.ts` to consume server-side data
- Update tests

## Test plan

- [x] Unit tests updated for dynamic-tags-resource
- [ ] Verify dynamic tags work correctly in editor

Ref: ED-24826

## 1. Problem & Context

Dynamic tags need metadata filtering for LLM schemas and stricter tag filtering at the prop level. This enables smarter schema generation and prevents invalid tag references in composed types.

## 2. What Changed (Where)

| File | Change |
|------|--------|
| `dynamic-prop-type.php` | Added `allowed_tag_names()` setter and `to_json_schema()` with enum constraints on permitted tags |
| `dynamic-prop-types-mapping.php` | Filters tag names by category intersection; removed stale comments |
| `dynamic-tags-module.php` | New `get_dynamic_tag_names_by_categories()` method; registered LLM schema deduplication filter |
| `llm-schema-*.php` (new) | Abstract filter walker + concrete `LLM_Schema_Dedupe_Filter` to strip redundant dynamic branches in nested schemas |
| `overridable-llm-filter.php` (new) | Removes overridable branches from LLM schemas (components module) |
| `dynamic-tags-resource.ts` | Moved tag hydration to backend MCP proxy endpoint; removed client-side schema building |
| Tests | Comprehensive coverage for tag filtering, schema deduplication, and resource fetching |

## 3. How It Works

Backend now controls the dynamic tag catalog: `dynamic-tags-module` collects registered tags by category, `dynamic-prop-types-mapping` constrains each prop's allowed tags via `enum`, and filters walk the schema tree to deduplicate dynamic/overridable branches when ancestors already offer them. Frontend calls MCP proxy endpoint instead of building schemas locally.

## 4. Risks

Schema filter walks deeply nested structures—performance impact negligible on typical schemas but worth profiling if field counts spike. Filter ordering matters (deduplication then overridable); if hooks register out of order, incorrect branches may survive. Backward compat: old schemas without `enum` constraints still validate, but new frontend expecting narrower tag lists might encounter stale data if backend tags aren't properly synced.

### DIFF
--- a/modules/atomic-widgets/dynamic-tags/dynamic-prop-types-mapping.php
+++ b/modules/atomic-widgets/dynamic-tags/dynamic-prop-types-mapping.php
@@ -25,11 +25,6 @@ class Dynamic_Prop_Types_Mapping extends Prop_Types_Schema_Extender {
 		return new static();
 	}
 
-	/**
-	 * Get the dynamic prop type to add to the prop type
-	 *
-	 * @param Prop_Type $prop_type
-	 */
 	protected function get_prop_types_to_add( Prop_Type $prop_type ): array {
 		$categories = [];
 
@@ -39,8 +34,6 @@ class Dynamic_Prop_Types_Mapping extends Prop_Types_Schema_Extender {
 
 		foreach ( $transformable_prop_types as $transformable_prop_type ) {
 			if ( $transformable_prop_type instanceof Transformable_Prop_Type ) {
-				// When the prop type is originally a union, we need to merge all the categories
-				// of each prop type in the union and create one dynamic prop type with all the categories.
 				$categories = array_merge( $categories, $this->get_related_categories( $transformable_prop_type ) );
 			}
 		}
@@ -49,7 +42,13 @@ class Dynamic_Prop_Types_Mapping extends Prop_Types_Schema_Extender {
 			return [];
 		}
 
-		return [ Dynamic_Prop_Type::make()->categories( $categories ) ];
+		$allowed_tag_names = Dynamic_Tags_Module::instance()->get_dynamic_tag_names_by_categories( $categories );
+
+		return [
+			Dynamic_Prop_Type::make()
+				->categories( $categories )
+				->allowed_tag_names( $allowed_tag_names ),
+		];
 	}
 
 	private function get_related_categories( Transformable_Prop_Type $prop_type ): array {
--- a/modules/components/module.php
+++ b/modules/components/module.php
@@ -17,6 +17,7 @@ use Elementor\Core\Base\Document;
 use Elementor\Modules\Components\PropTypes\Override_Prop_Type;
 use Elementor\Modules\Components\Transformers\Override_Transformer;
 use Elementor\Modules\Components\Widgets\Component_Instance;
+use Elementor\Modules\Components\Schema\Overridable_LLM_Filter;
 
 if ( ! defined( 'ABSPATH' ) ) {
 	exit; // Exit if accessed directly.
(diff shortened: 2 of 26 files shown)


Omerisra6:
Rebased onto the latest `ED-24826-prop-type-json-schema` (#36480) and added the union prop type support discussed for this PR:

- `Union_Prop_Type` gained a static registry (`register_omitted_variant_key()` / `register_exclusive_variant()`) so it can omit a variant (e.g. `overridable`, wired in a later PR) or advertise an "exclusive" variant like `dynamic` at most once per branch — mirroring the frontend's `propTypeToJsonSchema`/`suppressDynamic` behavior — without knowing about those prop types by name.
- Threaded the new `suppress_dynamic` flag through `Prop_Type::to_json_schema()`, `Object_Prop_Type`, and `Array_Prop_Type` so nested unions inside object/array shapes are covered too.
- Wired `Dynamic_Prop_Type` as the `dynamic` exclusive variant and added a server-side `get_dynamic_tag_names_by_categories()` resolver (mirrors the frontend's `getDynamicTagNamesByCategories`) so the emitted `name` enum is constrained to tags matching the prop's categories.

Note: since #36480 is still open, this PR's diff against `main` currently also includes its commits — that'll resolve once #36480 merges and this branch is rebased.

Tests: I added coverage for the new union suppression logic (fast unit tests) and for the dynamic-tag wiring/resolver (full WP test), but haven't been able to run the full WP-backed suite in this environment yet (local test DB isn't provisioned) — flagging so this gets verified before merge.

modules/atomic-widgets/dynamic-tags/dynamic-prop-types-mapping.php
```diff
@@ -49,7 +42,13 @@ protected function get_prop_types_to_add( Prop_Type $prop_type ): array {
...
 		}
 
-		return [ Dynamic_Prop_Type::make()->categories( $categories ) ];
+		$allowed_tag_names = Dynamic_Tags_Module::instance()->get_dynamic_tag_names_by_categories( $categories );
+
+		return [
+			Dynamic_Prop_Type::make()
+				->categories( $categories )
+				->allowed_tag_names( $allowed_tag_names ),
+		];
```
Omerisra6, line 51: why do we need to add it here?

modules/components/schema/overridable-llm-filter.php
```diff
@@ -0,0 +1,53 @@
...
+}
+
+class Overridable_LLM_Filter {
+	private const OVERRIDABLE_KEY = 'overridable';
+
+	public function apply( array $schema ): array {
+		return $this->walk( $schema );
+	}
+
+	private function walk( array $schema ): array {
```
Omerisra6: maybe we could create an abstraction for this? both `Overridable_LLM_Filter` and the dynamic dedup filter do the same thing

modules/components/module.php
```diff
@@ -50,6 +51,11 @@ public function __construct() {
 		add_action( 'elementor/atomic-widgets/settings/transformers/register', fn ( $transformers ) => $this->register_settings_transformers( $transformers ) );
 		add_action( 'elementor/document/after_migrate', fn( Document $document, array $data ) => $this->after_component_migrate( $document, $data ), 10, 2 );
 
+		add_filter(
+			'elementor/atomic-widgets/llm-json-schema',
```
Omerisra6, line 55: why do we need the dedup here as well and not just for dynamics? does it exist in the editor build composition tool?

PR #36482, opened 2026-07-12
Internal: CSS converter and font-family transformer fixes [ED-24826]

Omerisra6:
## Summary

Fix and improve CSS converter components for better variable handling.

## Changes

- Fix `string-property-converter` to handle edge cases
- Improve `variable-prop-value-transformer` logic
- Update `converter-registry-factory`
- Fix `font-family-transformer` for consistent behavior
- Add/update tests for all changes

## Test plan

- [x] Unit tests added/updated
- [ ] Verify CSS conversion works correctly for widgets with variables

Ref: ED-24826

## 1. Problem & Context

Font-family values were being unconditionally quoted, breaking CSS variables, generic families, and keywords. This fixes CSS converter to respect prop-type subclasses and transforms to quote only multi-word font names per CSS spec.

## 2. What Changed (Where)

• `String_Property_Converter`: Added `$type_key` parameter to emit schema's canonical `$type` (supports `Font_Family_Prop_Type` subclasses)
• `Font_Family_Transformer`: Replaced blanket quoting with whitespace detection; only quotes multi-word fonts
• `Variable_Prop_Value_Transformer`: Expanded string-backed type recognition to include `Font_Family_Prop_Type`
• Tests: Updated assertions to expect unquoted single-word fonts, generic families, CSS variables, and keywords

## 3. How It Works

`String_Property_Converter` now receives the actual prop-type's `$type` key from schema (e.g., `font-family` instead of `string`), allowing validators like `Font_Family_Prop_Type` to recognize and validate their own subclass instances. `Font_Family_Transformer` checks for whitespace—only wrapping values containing spaces—leaving `var(--x)`, `sans-serif`, `inherit` unquoted. Variable transformer recognizes font-family props as string-backed to extract var references correctly.

## 4. Risks

Low. Changes are spec-compliant and localized to font-family handling. Test coverage expanded significantly (new test cases for all edge cases). Single risk: external code directly instantiating `String_Property_Converter` without the new parameter will default to `String_Prop_Type::get_key()`, which is backwards-compatible.

### DIFF
--- a/modules/atomic-widgets/css-converter/converters/string-property-converter.php
+++ b/modules/atomic-widgets/css-converter/converters/string-property-converter.php
@@ -11,9 +11,12 @@ if ( ! defined( 'ABSPATH' ) ) {
 }
 
 /**
- * Reusable converter for properties backed by a plain String_Prop_Type. One instance per property.
- * When an allowlist is provided the value must be one of it (enum-backed props); otherwise any
- * non-empty value is accepted (free-string props). Emits the canonical PropValue from generate().
+ * Reusable converter for properties backed by a String_Prop_Type (or a subclass with its own
+ * `$$type` key, e.g. Font_Family_Prop_Type). One instance per property. When an allowlist is
+ * provided the value must be one of it (enum-backed props); otherwise any non-empty value is
+ * accepted (free-string props). Emits the canonical PropValue enveloped with the schema's own
+ * `$$type` key, sourced from the live schema rather than hardcoded, so subclasses that override
+ * get_key() (like Font_Family_Prop_Type) still validate against Props_Parser.
  */
 class String_Property_Converter extends Property_Converter_Base {
 	private string $property;
@@ -23,13 +26,19 @@ class String_Property_Converter extends Property_Converter_Base {
 	 */
 	private ?array $allowed_values;
 
+	private string $type_key;
+
 	/**
 	 * @param string        $property       The schema property this converter owns.
 	 * @param string[]|null $allowed_values Enum allowlist, or null for a free-string property.
+	 * @param string|null   $type_key       The `$$type` key to envelope the value with, sourced from
+	 *                                      the schema's prop type (e.g. `font-family`). Defaults to
+	 *                                      the plain String_Prop_Type key.
 	 */
-	public function __construct( string $property, ?array $allowed_values = null ) {
+	public function __construct( string $property, ?array $allowed_values = null, ?string $type_key = null ) {
 		$this->property = $property;
 		$this->allowed_values = $allowed_values;
+		$this->type_key = $type_key ?? String_Prop_Type::get_key();
 	}
 
 	protected function get_supported_properties(): array {
@@ -47,7 +56,10 @@ class String_Property_Converter extends Property_Converter_Base {
 			return false;
 		}
 
-		$context->set_prop( $this->property, String_Prop_Type::generate( $value ) );
+		$context->set_prop( $this->property, [
+			'$$type' => $this->type_key,
+			'value' => $value,
+		] );
 
 		return true;
 	}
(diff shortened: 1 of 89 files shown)


modules/atomic-widgets/css-converter/variable-prop-value-transformer.php
```diff
@@ -193,7 +195,13 @@ private function extract_var_reference( array $prop_value ): ?string {
 		$type = $prop_value['$$type'] ?? '';
 		$value = $prop_value['value'] ?? null;
 
-		if ( in_array( $type, [ 'color', 'string' ], true ) && is_string( $value ) ) {
+		$string_backed_types = [
+			'color',
```
Omerisra6: why cant we use the color prop type key as well?

modules/atomic-widgets/props-resolver/transformers/styles/font-family-transformer.php
```diff
@@ -10,22 +10,34 @@
...
-			( str_starts_with( $trimmed, "'" ) && str_ends_with( $trimmed, "'" ) )
-		);
-
-		if ( $is_quoted ) {
+		if ( '' === $trimmed || $this->is_already_quoted( $trimmed ) ) {
 			return $trimmed;
 		}
 
-		return '"' . $trimmed . '"';
+		return preg_match( '/\s/', $trimmed )
```
Omerisra6, line 36: add comment explaining why we added that

modules/atomic-widgets/css-converter/variable-prop-value-transformer.php
Omerisra6: Good call — switched to `Color_Prop_Type::get_key()` for consistency with the other keys in the list (9001601). Thanks!

modules/atomic-widgets/props-resolver/transformers/styles/font-family-transformer.php
Omerisra6: Added an inline comment above the `preg_match` explaining that only multi-word names need quoting and that single-token values (`var(--x)`, `sans-serif`, `inherit`) must stay unquoted to remain valid CSS (9001601).

modules/atomic-widgets/props-resolver/transformers/styles/font-family-transformer.php
```diff
@@ -10,22 +10,36 @@
...
-
-		if ( $is_quoted ) {
+		if ( '' === $trimmed || $this->is_already_quoted( $trimmed ) ) {
 			return $trimmed;
 		}
 
-		return '"' . $trimmed . '"';
+		// Only multi-word font names (e.g. `Open Sans`) need to be wrapped in quotes.
+		// Single-token values like `var(--x)`, `sans-serif` or `inherit` must stay unquoted to remain valid CSS.
+		return preg_match( '/\s/', $trimmed )
```
Omerisra6, line 36: not sure if thats the best option

modules/atomic-widgets/css-converter/converters/string-property-converter.php
```diff
@@ -47,7 +56,10 @@ protected function do_convert( Conversion_Context $context, array $rule ): bool
 			return false;
 		}
 
-		$context->set_prop( $this->property, String_Prop_Type::generate( $value ) );
+		$context->set_prop( $this->property, [
```
IshayMaya, line 59: maybe get the prop type by key and then run generate?

modules/atomic-widgets/props-resolver/transformers/styles/font-family-transformer.php
```diff
@@ -10,22 +10,36 @@
...
-		);
-
-		if ( $is_quoted ) {
+		if ( '' === $trimmed || $this->is_already_quoted( $trimmed ) ) {
 			return $trimmed;
 		}
 
-		return '"' . $trimmed . '"';
+		// Only multi-word font names (e.g. `Open Sans`) need to be wrapped in quotes.
+		// Single-token values like `var(--x)`, `sans-serif` or `inherit` must stay unquoted to remain valid CSS.
```
IshayMaya, line 35: how does the var passed into it?

Omerisra6: it might get passed if the var isnt recognized in our system (but a user might have a custom var)

IshayMaya (approved):

PR #36488, opened 2026-07-12
Fix: Atomic Form Angie support > inputs are not wrapped with an atomic form container [ED-24881]

davseve:
## 1. Problem & Context

Form fields weren't being validated as wrapped within atomic form containers during composition building, allowing invalid structure. Adding structural validation catches orphaned fields, missing/duplicate submit buttons, and empty message elements at build time.

## 2. What Changed (Where)

- **form-structure/utils.ts**: Added three validation functions (`collectFormAncestorErrors`, `collectSubmitButtonErrors`, `collectEmptyMessageErrors`) to detect form structure violations
- **composition-builder.ts**: Integrated form validation into build pipeline, collects errors alongside existing config/style checks
- **build-composition/tool.ts**: Surfaces form errors to error handler; refactored destructuring to include new `formErrors` field
- **Test files**: Added utility helper and comprehensive test suite covering all validation scenarios

## 3. How It Works

Build process now executes three validation passes on the XML document:
1. **Form ancestor validation**: Traverses all form field elements, ensures each has `e-form` ancestor (any depth)
2. **Submit button validation**: Per-form checks that exactly one submit button exists
3. **Message validation**: Ensures success/error message elements contain children

All errors collected and returned alongside existing `configErrors`, then mapped to Error objects and reported through standard error handling pipeline.

## 4. Risks

None identified. Validation is purely additive—only detects structural problems, doesn't modify behavior for valid structures. Comprehensive test coverage (18 test cases across three functions) de-risks edge cases.

### DIFF
--- a/packages/packages/core/editor-canvas/src/composition-builder/composition-builder.ts
+++ b/packages/packages/core/editor-canvas/src/composition-builder/composition-builder.ts
@@ -15,6 +15,7 @@ import { doUpdateElementProperty } from '../mcp/utils/do-update-element-property
 import { mergeCustomCssText } from '../mcp/utils/merge-custom-css';
 import { RequiredChildrenEnforcer } from './utils/required-children-enforcer';
 import { getRequiredDefaultChildTemplates } from './utils/required-default-child-tags';
+import { collectFormAncestorErrors, collectSubmitButtonErrors, collectEmptyMessageErrors } from '../form-structure/utils';
 
 type AnyValue = z.infer< z.ZodTypeAny >;
 type AnyConfig = Record< string, Record< string, AnyValue > >;
@@ -292,6 +293,25 @@ export class CompositionBuilder {
 		}
 		if ( childTypeErrors.length ) {
 			throw new Error( `Invalid element structure:\n${ childTypeErrors.join( '\n' ) }` );
+		}
+
+		const formAncestorErrors = collectFormAncestorErrors(this.xml);
+
+		if (formAncestorErrors.length) {
+		  throw new Error(`Invalid form structure:\n${formAncestorErrors.join('\n')}`);
+		}
+
+
+		const submitErrors = collectSubmitButtonErrors(this.xml);
+
+		if (submitErrors.length) {
+			throw new Error(`Invalid form structure:\n${submitErrors.join('\n')}`);
+		}
+
+		const emptyMessageErrors = collectEmptyMessageErrors(this.xml);
+
+		if (emptyMessageErrors.length) {
+			throw new Error(`Invalid form structure:\n${emptyMessageErrors.join('\n')}`);
 		}
 
 		const children = Array.from( this.xml.children );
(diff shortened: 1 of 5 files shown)


packages/packages/core/editor-canvas/src/composition-builder/composition-builder.ts
```diff
@@ -294,6 +295,17 @@ export class CompositionBuilder {
...
 		}
 
+		
+		const formErrors = [
+			...collectFormAncestorErrors( this.xml ),
+			...collectSubmitButtonErrors( this.xml ),
+			...collectEmptyMessageErrors( this.xml ),
+		];
+
+		if ( formErrors.length ) {
```
Ntnelbaba: This is a big failure to do, it means even one issue result to not doing anything
Make sure this is the way at this tool

packages/packages/core/editor-canvas/src/composition-builder/composition-builder.ts
Ntnelbaba, line 1: Make sure with @Omerisra6 that this fix is valid once moving it to the BE

packages/packages/core/editor-canvas/src/form-structure/__tests__/form-structure-utils.test.ts
```diff
@@ -23,6 +26,10 @@ function mockElement( widgetType?: string, elType?: string ): V1Element {
 	} as V1Element;
 }
 
+function parseXml( xml: string ): Document {
+    return new DOMParser().parseFromString( `<root>${ xml }</root>`, 'text/xml' );
```
Ntnelbaba: Don't we have a function like that?

packages/packages/core/editor-canvas/src/form-structure/utils.ts
```diff
@@ -115,3 +115,49 @@ export function clipboardRootsAreAtomicForms( elements: ClipboardElement[] ): bo
 
 	return elements.every( ( el ) => getClipboardElementType( el ) === FORM_ELEMENT_TYPE );
 }
+
+export function hasFormAncestor(node: Element): boolean {
```
Ntnelbaba: Wow this is so wasty
Why iterating like that? Simple search on the string can be better

packages/packages/core/editor-canvas/src/form-structure/utils.ts
Ntnelbaba, line 1: All of the file can be simplified to a better structure 
Maybe a class with constructor and simple grep if there is even a form

Ntnelbaba (changes requested):
Moving this logic to BE, wait for job done

packages/packages/core/editor-canvas/src/form-structure/utils.ts
davseve: This code will be replaces in a few days.

packages/packages/core/editor-canvas/src/composition-builder/composition-builder.ts
davseve: This change will not pass to the BE implementation.
Opened a new ticket for it
https://elementor.atlassian.net/browse/ED-24891

Ntnelbaba (approved):

PR #36497, opened 2026-07-13
Internal: Use local migrations manifest by default and remote on rollback [ED-24892]

Ntnelbaba:
## Summary
- Use the bundled local migrations manifest (`migrations/`) for normal prop-type migration runs instead of always fetching from the remote URL
- On detected plugin downgrade (`ELEMENTOR_VERSION` < stored `elementor_version`), use the remote manifest with fallback to the bundled local copy when the fetch fails
- Add `Migrations_Loader` remote-to-local fallback support and PHPUnit coverage for both scenarios

## Test plan
- [ ] `vendor/bin/phpunit --filter 'Test_Migrations_Loader|Test_Migrations_Orchestrator' tests/phpunit/elementor/modules/atomic-widgets/prop-type-migrations/`
- [ ] Open a document with atomic widgets on a normal (non-rollback) install — confirm no remote manifest HTTP request is made
- [ ] Simulate rollback (`elementor_version` option higher than running version) — confirm remote manifest is fetched
- [ ] With rollback + remote unreachable — confirm migrations still work via bundled local manifest

Ref: ED-24892

## 1. Problem & Context

ED-24892: Switch migrations manifest loading strategy—use bundled local by default, fall back to remote during rollback. Eliminates singleton pattern, enabling flexible loader composition for different environments.

## 2. What Changed (Where)

- **migrations-loader.php**: Removed static singleton instance; added `fallback_base_path` parameter; split `read_source()` into `read_remote_source()` and `read_local_source()` with fallback resolution logic.
- **migrations-orchestrator.php**: Decoupled loader initialization; added `is_rollback()` detection; routing to local/remote loaders based on version state; added new test coverage validating rollback behavior.

## 3. How It Works

Entry point: `Migrations_Orchestrator::migrate()` calls `get_active_loader()` which:
- Returns custom loader if `migrations_path` set
- Returns remote loader with local fallback if `is_rollback()` (stored version > current version)
- Returns local loader otherwise (upgrade path)

Remote fetch attempts transient caching; on HTTP failure or non-2xx response, resolves fallback path and loads locally. This ensures migrations function even when CDN is unavailable during rollbacks.

## 4. Risks

**Fallback path resolution assumptions**: `resolve_fallback_path()` uses `basename()` for non-URL paths—assumes manifest structure symmetry between remote and local. If naming diverges, fallback silently loads wrong manifest. Mitigation: tests validate both HTTP error and timeout scenarios.

**Version detection reliability**: `is_rollback()` compares `ELEMENTOR_VERSION` against DB-stored version—edge case if option doesn't exist (returns false, defaults to local). Safe but worth documenting.

### DIFF
--- a/modules/atomic-widgets/prop-type-migrations/migrations-orchestrator.php
+++ b/modules/atomic-widgets/prop-type-migrations/migrations-orchestrator.php
@@ -21,6 +21,7 @@ if ( ! defined( 'ABSPATH' ) ) {
 class Migrations_Orchestrator {
 	const EXPERIMENT_BC_MIGRATIONS = 'e_bc_migrations';
 	const MIGRATIONS_URL = 'https://editor.elementor.com/v1/migrations/';
+	const BUNDLED_MIGRATIONS_DIRECTORY = 'migrations/';
 
 	private static ?self $instance = null;
 
@@ -398,7 +399,7 @@ class Migrations_Orchestrator {
 			return constant( 'ELEMENTOR_MIGRATIONS_PATH' );
 		}
 
-		return ELEMENTOR_PATH . 'migrations/';
+		return ELEMENTOR_PATH . self::BUNDLED_MIGRATIONS_DIRECTORY;
 	}
 
 	private function migrate_doc( array $data, Document $document ): array {
--- a/modules/atomic-widgets/prop-type-migrations/migrations-orchestrator.php
+++ b/modules/atomic-widgets/prop-type-migrations/migrations-orchestrator.php
@@ -3,6 +3,7 @@
 namespace Elementor\Modules\AtomicWidgets\PropTypeMigrations;
 
 use Elementor\Core\Base\Document;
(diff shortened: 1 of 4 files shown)


modules/atomic-widgets/prop-type-migrations/migrations-orchestrator.php
```diff
@@ -335,6 +361,46 @@ private function execute_prop_migration( $prop_value, array $migrations, string
...
+				'manifest.json',
+				self::get_local_migrations_path()
+			);
+		}
+
+		return $this->remote_loader;
+	}
+
+	private static function get_local_migrations_path(): string {
+		if ( defined( 'ELEMENTOR_MIGRATIONS_PATH' ) ) {
```
Ntnelbaba, line 398: But this is not defined now so how can he tell where is the migrations now?

RobiFerentz (approved):

PR #36558, opened 2026-07-17
Internal: Twig check [ED-24230]

marcin-el:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #

## 1. Problem & Context

Twig has multiple CVEs with advisories ignored in composer.json (PKSA-*). This adds automated guardrails to prevent unsafe Twig APIs from being introduced—scanning on CI and pre-commit to catch violations early.

## 2. What Changed (Where)

| File | Change |
|------|--------|
| `bin/strip-comments.php` | New utility: strips PHP/Twig comments while preserving line numbers for accurate matching |
| `bin/check-twig-safety.sh` | New bash script: runs pattern checks against stripped files for 6 different Twig security policies |
| `composer.json` | Added `check-twig-safety` script, integrated into `lint-staged` for `.php` and `.twig` files |
| `.github/workflows/lint-php-changes-only.yml` | Added Twig file detection, conditional Twig safety check in CI pipeline |
| `.github/workflows/lint.yml` | Integrated `check-twig-safety` and `composer audit` into main lint job |
| `ruleset.xml` | Excluded `strip-comments.php` from phpcs analysis |
| `package.json` | Added `check-twig-safety.sh` to lint-staged for `.php` and `.twig` files |

## 3. How It Works

Pattern matching on security-sensitive Twig APIs (sandbox, SourcePolicy, spaceless, HtmlDumper, template_from_string) runs against comment-stripped file contents—eliminating false positives from docblock mentions. Script runs pre-commit (lint-staged) and in CI, blocking PRs that violate policy.

## 4. Risks

Comment stripping via regex/tokenizer could miss edge cases (nested comments, heredoc strings with `{#`); mitigated by conservative patterns and pre-existing CVE mitigations in Twig 3.11.3. Performance negligible for typical repos.


bin/check-twig-safety.sh
```diff
@@ -0,0 +1,115 @@
...
+}
+
+check "PKSA-8zx5-v2nz-58pb, PKSA-kvv6-36cr-fkzb, PKSA-n14z-jjjg-g8vd, PKSA-3mcc-k66d-pydb, PKSA-dpx1-78wg-1kqs, PKSA-g9zw-qxh8-pq8w, PKSA-yd6k-t2gh-1m43, PKSA-1tmc-rt7x-12w6, PKSA-xx6c-6d96-db2w" \
+	"Twig sandbox must stay disabled" \
+	'enableSandbox|SandboxExtension|\{%[[:space:]]*sandbox' \
+	php twig
+
+check "PKSA-gw7n-z4yx-7xjt, PKSA-21g2-dzjv-sky5, PKSA-fbvq-z33h-r2np" \
+	"Twig SourcePolicy must stay unconfigured" \
+	'SourcePolicyInterface|setSourcePolicy|new[[:space:]]+SourcePolicy' \
```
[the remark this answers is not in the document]
marcin-el: fixed

bin/check-twig-safety.sh
```diff
@@ -0,0 +1,132 @@
...
+cd "$(dirname "$0")/.."
+
+FILES=("$@")
+
+EXCLUDES=(
+	--exclude-dir=vendor
+	--exclude-dir=vendor_prefixed
+	--exclude-dir=node_modules
+	--exclude-dir=.git
+	--exclude-dir=build
```
mykytamurzin, line 29: i think we are using twig only inside atomic-widgets folder? maybe we can limit it in such way?

marcin-el: yeah but this is deliberately broad as we want to catch usage anywhere not just where it lives today, check takes like 1-2seconds

.github/workflows/lint-php-changes-only.yml
```diff
@@ -54,3 +54,5 @@ jobs:
         run: |
           #composer lint $(echo "${PHP_CHANGED_FILES}" | tr '\n' ' ')
           vendor/bin/phpcs --extensions=php --standard=./ruleset.xml --warning-severity=0 $(echo "${PHP_CHANGED_FILES}" | tr '\n' ' ')
+      - name: Twig safety check
```
mykytamurzin, line 66: is this job running only on *.php files change? Our template files have *.twig extension, so maybe it won't run on only-twig changes

bin/check-twig-safety.sh
```diff
@@ -0,0 +1,132 @@
...
+
+check "PKSA-gw7n-z4yx-7xjt, PKSA-21g2-dzjv-sky5, PKSA-fbvq-z33h-r2np" \
+	"Twig SourcePolicy must stay unconfigured" \
+	'SourcePolicyInterface|setSourcePolicy|new[[:space:]]+SourcePolicy' \
+	php
+
+check "PKSA-sjvz-tbbr-vwth" \
+	"Twig 'spaceless' filter/tag must not be used in templates" \
+	'\{%[[:space:]]*spaceless|\bspaceless\(' \
+	twig php
```
mykytamurzin, line 109:
```
The spaceless check can't match a real Twig 3 usage

Twig 3 removed the {% spaceless %} tag — there's no Spaceless node in your vendored 3.11.3 (find for *Spaceless* returns nothing). It only exists as a filter. The pattern is \{%\s*spaceless|\bspaceless\(, which matches the deleted tag and a call syntax the filter doesn't have. The two forms that actually work both sail through:

{% apply spaceless %}hi{% endapply %}
{{ foo|spaceless }}
→ Twig safety check passed.  EXIT=0

Since composer.json justifies PKSA-sjvz-tbbr-vwth with "Our templates do not use the spaceless filter", this is the one check that most directly backs a stated assumption, and it's the one that detects nothing. Needs \|[[:space:]]*spaceless\b and \{%[[:space:]]*apply[^%]*\bspaceless\b.
```

**not sure if it's relevant?**

bin/check-twig-safety.sh
```diff
@@ -0,0 +1,132 @@
...
+
+check "PKSA-h8hf-ytnd-5t9q" \
+	"Twig {% use %} must only reference hardcoded, quoted template names" \
+	'\{%[[:space:]]*use[[:space:]]+[a-zA-Z_][a-zA-Z0-9_]*[[:space:]]*%\}' \
+	twig
```
mykytamurzin:
```
The {% use %} check only matches templates Twig already rejects

UseTokenParser hard-requires a constant:

if (!$template instanceof ConstantExpression) {
    throw new SyntaxError('The template references in a "use" statement must be a string.', ...);
}

So {% use layout %} — a bare unquoted name, the only thing the pattern matches — is a parse error already; it can't exist in a working template. Meanwhile the assumption in composer.json is "template names passed to Twig are hardcoded. No user can inject twig source" — that's a PHP-side property, and nothing here checks it. No check for ArrayLoader/StringLoader, or a variable template name reaching ->render()/->load(). That's where the guard would earn its keep.
```

**not sure if it's relevant**

murzin-nikita (approved):

PR #36577, opened 2026-07-19
Internal: Migrate get-globals tool to global-classes and global-variables MCP resources [ED-24919]

Omerisra6:
## Why

`build-composition` and other prompts reference these MCP resources:

- `elementor://global-classes`
- `elementor://global-variables`

The editor already exposes them as MCP resources (`classes-resource.ts`, `variables-resource.ts`), but the PHP MCP server only offered a combined `elementor/get-globals` **tool** — so MCP clients hitting the server couldn't resolve the referenced URIs.

## What

- Add `Global_Classes_Resource_Ability` (URI `elementor://global-classes`) and `Global_Variables_Resource_Ability` (URI `elementor://global-variables`), each registered as an MCP resource with `application/json` mimeType.
- Register both in `module.php` under the resources argument of `create_server`.
- Add both URIs to `list-resources` catalog and `read-resource` executors.
- Add both URIs to the REST proxy resource map so editor GETs to `elementor/v1/mcp-proxy?uri=...` work.
- Remove the now-obsolete `Get_Globals_Ability` (no other callers).

## Payload

Same shape the previous tool exposed, split per resource:

- `elementor://global-classes` → `Global_Classes_Repository::make($kit)->all_labels()`
- `elementor://global-variables` → `{ variables, total, watermark }`

## 1. Problem & Context

Migrating from a monolithic `Get_Globals_Ability` tool to granular MCP resources following resource-oriented architecture; enables clients to query global classes and variables independently via standard MCP URIs.

## 2. What Changed (Where)

- Created two new resource abilities: `Global_Classes_Resource_Ability` and `Global_Variables_Resource_Ability` with separate URIs and JSON payloads
- Removed legacy `Get_Globals_Ability` (74 lines); replaced single tool with two resource endpoints
- Registered new abilities in `Module::register_abilities()` and wired into `List_Resources_Ability`, `Read_Resource_Ability`, and REST proxy

## 3. How It Works

Each resource ability extends `Abstract_Ability`, defines a URI and metadata, and executes independently: global-classes queries `Global_Classes_Repository` for CSS class labels; global-variables instantiates `Variables_Service` with repository/processor to load and format design tokens. `Read_Resource_Ability` maps URIs to execution lambdas; clients fetch resources by URI rather than calling a single composite tool.

## 4. Risks

Clients relying on the old single `get-globals` tool will break—confirm deprecation/migration period communicated. Payload structure changed (previously nested under `global_classes`/`variables` keys, now direct JSON per resource)—verify client parsers handle new format.


modules/mcp/abilities/global-classes-resource-ability.php
```diff
@@ -0,0 +1,45 @@
...
+			[
+				'mcp' => [
+					'type'        => 'resource',
+					'uri'         => self::URI,
+					'public'      => true,
+					'mimeType'    => 'application/json',
+					'description' => __( 'Global class definitions and order from the active kit.', 'elementor' ),
+				],
+			],
+			fn() => current_user_can( 'edit_posts' )
```
IshayMaya, line 34: is it the default check for all the resource abilities? maybe we should have a shared util for it

Omerisra6: Think it's better that the developer will declare it and be aware of it

IshayMaya (approved):

PR #36582, opened 2026-07-20
Feature: Add mode parameter to build-composition ability

Omerisra6:
## Description
Add a `mode` parameter to the `build-composition` ability with values:
- `append` (default): Insert new elements under `parent_id`, preserving existing children
- `replace_children`: Remove all direct children of `parent_id` first, then insert new elements

This enables "Redesign this section" to be a single call instead of separate delete + build operations.

## Changes
- Add `MODE_APPEND` and `MODE_REPLACE_CHILDREN` constants
- Update input schema with `mode` parameter
- Update output schema with `removed_element_ids` (returned when mode is `replace_children`)
- Add mode validation in `validate_input()`
- Implement `replace_children` logic in `Composition_Persister`
- Update TypeScript editor tool with `mode` parameter
- Update `build-composition.md` documentation with MODE section
- Add PHPUnit tests for all mode scenarios

## Testing
- `mode` omitted → identical to current behavior (regression guard)
- `mode: 'append'` → new elements appended, `removed_element_ids` not in response
- `mode: 'replace_children'` → existing children removed, new elements inserted, `removed_element_ids` returned
- `mode: 'replace_children'` with empty parent → behaves as append, `removed_element_ids: []`
- `mode: 'replace_children'` with `parent_id: 'document'` → all top-level elements wiped
- `mode: 'replace_children'` with non-existent parent → returns 404 error
- Invalid mode value → returns 400 error

## 1. Problem & Context

Adds `mode` parameter to enable replacing existing children instead of always appending. Supports redesigning parent containers or entire pages in single operation.

## 2. What Changed (Where)

- **build-composition-ability.php**: Added MODE constants, extract `mode` parameter, pass to persister, validate input, include `removed_ids` in response
- **composition-persister.php**: Added `replace_children()` method; `insert_and_save()` now accepts mode parameter, deletes children before insertion
- **Tests**: 7 new tests covering append (default), replace_children with/without children, document root replacement, nonexistent parent, invalid mode
- **TypeScript tool**: Updated schema to accept mode enum, pass through handler, return `removed_element_ids`
- **Documentation**: Added MODE section explaining behavior, trade-offs, use cases

## 3. How It Works

Default `mode: 'append'` preserves existing behavior. When `mode: 'replace_children'`, `replace_children()` collects all direct child IDs, removes them via mutator, tracks removed IDs. Special case: `parent_id: 'document'` clears entire root. Removed IDs only appear in response when mode is `replace_children`; validation rejects invalid modes.

## 4. Risks

None significant. Mode parameter is optional with safe default. Validation prevents invalid modes. Replace operation is atomic per child (fails early on error). Test coverage includes edge cases (empty parent, document root, nonexistent parent).


modules/mcp/abilities/build-composition/composition-persister.php
```diff
@@ -57,6 +70,54 @@ public function insert_and_save( Document $document, array $subtrees, string $pa
...
+			'removed_ids' => $removed_ids,
+		];
+	}
+
+	/**
+	 * @return array{tree: array, removed_ids: string[]}|\WP_Error
+	 */
+	private function replace_children( array $tree, string $parent_id ) {
+		if ( self::DOCUMENT_ROOT === $parent_id ) {
+			$removed_ids = array_map(
```
ronros-elementor, line 82: should we also care about the recursive children ids?

ronros-elementor (dismissed):

modules/mcp/abilities/build-composition/composition-persister.php
Omerisra6: Not realy that's semantic data to allow the LLM to communicate the changes

IshayMaya (approved):

PR #36594, opened 2026-07-20
Fix: Cache validity flags to check against actual files [ED-24903]

ronros-elementor:
## 1. Problem & Context

Cache validity wasn't checking actual filesystem state, causing stale files to be served after external eviction. This centralizes cache-validity logic into `CSS_Files_Manager` so it owns the write-and-validate pipeline, including a `should_exist` meta flag to distinguish legitimate empty results from broken cache.

## 2. What Changed (Where)

- **atomic-styles-manager.php**: Removed inline cache checks; now passes `$breakpoint_path` to manager.
- **css-files-manager.php**: Refactored `get()` signature to accept cache path; added atomic write with temp file, `should_exist` meta tracking, and file-existence validation on cache hits.
- **test-atomic-styles-manager.php**: Added mocks for `move()` and filesystem checks; added self-heal tests (missing files, legitimate empty breakpoints, write failures).
- **test-css-files-manager.php**: Completely rewritten with 9 tests covering cache hits, empty CSS, write failures, self-heal scenarios, and legacy leaf fallback.

## 3. How It Works

On `get(handle, media, get_css, cache_path)`: check if cache is valid and `should_exist=true`, then verify file exists and is non-empty before returning cached file. On miss/invalid, call `get_css()`, write atomically (tmp → move), and validate cache with `should_exist=true` or `false`. On write failure, leave cache invalid to retry next request. This self-heals when files are externally deleted (exists=false but should_exist=true).

## 4. Risks

Minor: atomic move fallback to direct write may lose some safety on unsupported filesystems, but tests confirm fallback works. Metadata schema change (`should_exist`) is backward-compatible—missing meta treats as null and rechecks disk.


modules/atomic-widgets/styles/css-files-manager.php
```diff
@@ -2,55 +2,119 @@
...
+	 * The manager regenerates when the cache is invalid or when a should-be-present file
+	 * cannot be found on disk. On a hard write failure it deliberately leaves the leaf
+	 * invalid so the next request retries instead of getting stuck.
+	 *
+	 * @param string        $handle
+	 * @param string        $media
+	 * @param callable      $get_css
+	 * @param array<string> $cache_path Cache-validity leaf path for this breakpoint.
+	 * @return Style_File|null
+	 */
```
Omerisra6, line 31: Make the description not specific to breakpoints (we do support any kind of CSS) file)

Omerisra6 (dismissed):

Omerisra6 (approved):

PR #36598, opened 2026-07-21
Internal: Avoid persist of empty values. [ED-24037]

RobiFerentz:
## 1. Problem & Context

Prevents persistence of empty composite values (objects, arrays, HTML) while preserving falsy primitives (0, false, empty strings). Reduces database bloat from scaffold data that doesn't carry semantic meaning.

## 2. What Changed (Where)

- **Prop_Type interface & implementations**: Added `should_persist($value): bool` contract; composite types (Array, Object, Html_V2/V3, Key_Value) return false for empty values; primitives & Union delegate appropriately
- **Props_Parser**: Now filters fields post-sanitization via `should_persist()` before persisting
- **Has_Atomic_Base::get_data_for_save()**: Refactored to use `set_data_field_for_save()` helper; unsets empty settings/styles/editor_settings/interactions
- **Html_V2/V3**: Guarantees `children` array exists after sanitize; `should_persist()` checks for non-empty content or children
- **Key_Value_Prop_Type**: Persists only if key is non-empty (value alone doesn't justify storage)
- **Override_Prop_Type**: Added validation & always persists (including null values for component overrides)
- **Test coverage**: New tests verify empty fields removed while zero/false preserved; component overrides persist null

## 3. How It Works

Entry: Props_Parser sanitizes each field, then checks `should_persist()` before storing. Composite types recursively filter nested children using same logic. Array items and Object properties skip persistence if their type's `should_persist()` returns false. Plain types always persist. Result: database contains only meaningful data; scaffolding (empty arrays/objects) stripped away automatically.

## 4. Risks

Minor: Mock/test fixtures updated to match new schema (empty fields removed). Override_Prop_Type always persists—verify component override null-handling doesn't create unintended state. Package.json engine bump to Node 24.15+ may affect CI/tooling without backward compatibility warning.

### DIFF
--- a/modules/atomic-widgets/prop-types/base/object-prop-type.php
+++ b/modules/atomic-widgets/prop-types/base/object-prop-type.php
@@ -117,6 +117,10 @@ abstract class Object_Prop_Type implements Transformable_Prop_Type {
 		return $value;
 	}
 
+	public function should_persist( $value ): bool {
+		return ! empty( $value['value'] );
+	}
+
 	public function sanitize_value( $value ) {
 		foreach ( $this->get_shape() as $key => $prop_type ) {
 			if ( ! isset( $value[ $key ] ) ) {
@@ -124,6 +128,11 @@ abstract class Object_Prop_Type implements Transformable_Prop_Type {
 			}
 
 			$sanitized_value = $prop_type->sanitize( $value[ $key ] );
+
+			if ( ! $prop_type->should_persist( $sanitized_value ) ) {
+				unset( $value[ $key ] );
+				continue;
+			}
 
 			$value[ $key ] = $sanitized_value;
 		}
--- a/modules/atomic-widgets/prop-types/contracts/prop-type.php
+++ b/modules/atomic-widgets/prop-types/contracts/prop-type.php
@@ -21,6 +21,7 @@ interface Prop_Type extends \JsonSerializable {
 	public function set_dependencies( ?array $dependencies ): self;
 	public function get_dependencies(): ?array;
 	public function get_initial_value();
+	public function should_persist( $value ): bool;
 
 	/**
 	 * Returns the list of alias names for this prop type.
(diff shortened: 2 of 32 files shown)


modules/atomic-widgets/prop-types/contracts/prop-type.php
```diff
@@ -22,6 +22,16 @@ public function set_dependencies( ?array $dependencies ): self;
...
 
+	/**
+	 * Whether a sanitized prop value should be written to persisted element data.
+	 *
+	 * Called after sanitize() by Props_Parser and composite prop types when filtering
+	 * nested values. Each prop type defines its own empty semantics.
+	 *
+	 * @param array $value Sanitized transformable prop value.
+	 */
+	public function should_persist( $value ): bool;
```
Ntnelbaba, line 33: Can you set a default return here?

packages/package.json
```diff
@@ -3,8 +3,8 @@
   "author": "Elementor Team",
   "license": "GPL-3.0-or-later",
   "engines": {
-    "node": ">=20.19.0",
-    "npm": ">=10.0.0"
+    "node": ">=24.15.0",
+    "npm": ">=11.12.1"
```
Ntnelbaba:
```suggestion
    "npm": ">=11.0.0"
```

Ntnelbaba (dismissed):

packages/package.json
RobiFerentz: fixed.

modules/atomic-widgets/prop-types/contracts/prop-type.php
RobiFerentz: Cannot be done in php interfaces.

modules/atomic-widgets/prop-types/base/object-prop-type.php
```diff
@@ -117,6 +117,10 @@ public function sanitize( $value ) {
 		return $value;
 	}
 
+	public function should_persist( $value ): bool {
+		return ! empty( $value['value'] );
```
Ntnelbaba, line 121: Should get deleted only if key is missing

Ntnelbaba (dismissed):

Ntnelbaba (approved):

PR #36600, opened 2026-07-21
Internal: Lean get-page-structure MCP tool with element_id filter [ED-24951]

Omerisra6:
## What

Reduce the `get-page-structure` MCP tool response by pruning to the essential skeleton (`id`, `elType`, `widgetType`, nested `elements`) instead of dumping full settings/styles per element. Adds an optional `element_id` input to scope the response to a single subtree, and an opt-in `include_content` flag (requires `element_id`) to also return each node's `settings` and `styles` in the exact shape that `build-composition` accepts as input.

## Why

The current tool returns `$document->get_elements_data()` verbatim — every setting, style variant, and `custom_css` blob per element. On non-trivial pages this dominates the MCP context window. The lean skeleton is enough for navigation; when the LLM needs to understand the content of a specific subtree, it can opt into `include_content=true` for that subtree only.

The `settings` are passed through as-is (already stored as `{$$type,value}` envelopes — the same shape `build-composition`'s `element_config` expects). The `styles` are projected from the element's local desktop/no-state style variant back to a flat CSS `{ prop: value }` map — the same shape `build-composition`'s `style` input accepts — using the existing `Render_Props_Resolver::for_styles()` and `Style_Schema::get()` (same path used by `Styles_Renderer`).

Part of ED-24951 (Optimize context for build composition and helper tools). This is PR 1 of 4.

## Changes

- [`Get_Structure_Ability`](modules/mcp/abilities/get-structure-ability.php): prune output via a private `prune_elements()` method that delegates recursion to the existing `Plugin::$instance->db->iterate_data()` util. Adds optional `element_id` input (missing id returns `404 element_not_found` via `\Elementor\Utils::find_element_recursive`). Adds optional `include_content` input (requires `element_id`, else `400 invalid_input`); when true, each pruned node also includes `settings` (envelope pass-through) and `styles` (flat CSS map). New private `project_styles_to_input_shape()` helper reverses the local desktop/no-state variant into the input shape; non-desktop variants are attached under `__variants` and decoded `custom_css` under `__custom_css`.
- TS wrapper [`get-page-structure/tool.ts`](packages/packages/core/editor-canvas/src/mcp/tools/get-page-structure/tool.ts): exposes `elementId` and `includeContent` params and updates descriptions/output.
- Tests: updated ability tests to assert pruned skeleton, element_id happy path, 404 not-found path; added cases for `include_content` validation, settings pass-through, style projection with `__custom_css`, class-only (empty styles), and `__variants` attachment.

No new classes or utilities added — reuses existing tree-walking and style-resolution utils.

## Testing

- `php -l` clean on all touched PHP files.
- Editor-side lint clean.
- PHPUnit suite requires the wp-lite-env / `npm run test:setup:playwright` setup (not run locally).

## Context-size impact

- Default (no `include_content`): on any non-trivial page the response drops ~80–95% depending on how many settings/styles per element.
- `include_content=true` scoped to a subtree: proportional to the subtree size, but never the whole document.

Ref: ED-24951

## 1. Problem & Context

ED-24951: Lean the get-page-structure MCP tool to return minimal element skeleton (id, elType, widgetType, nested only) instead of full data, with optional element_id filtering and include_content for selective full serialization.

## 2. What Changed (Where)

| File | Change |
|------|--------|
| `local-style-serializer.php` | New: extracts local style variant (desktop/no-state) into flat CSS map with style_id, custom_css, and non-desktop variants. |
| `style-props-to-css.php` | New: delegates PropValue→CSS conversion to reusable utility. |
| `get-structure-ability.php` | Refactored execute() to prune elements (skeleton-only by default), added element_id subtree lookup, include_content flag validation. |
| `llm-guidance-builder.php` | Delegated prop-to-CSS logic to Style_Props_To_Css. |
| `test-get-structure-ability.php` | Replaced generic test with 7 focused tests: pruning, subtree filtering, 404 on missing element, content inclusion, style serialization, variants handling. |
| `tool.ts` | Updated schema: added elementId, includeContent params; updated description and handler. |

## 3. How It Works

Request → validate post_id → fetch elements → [if element_id: find subtree recursively, else: use root] → prune to skeleton (id/elType/widgetType/elements) → [if include_content: attach resolved settings & serialized styles] → return. Style serialization: extract first local style entry, find desktop/no-state variant, project props via Props_Resolver, flatten to {prop: css_value}, attach style_id and custom_css (decoded) separately.

## 4. Risks

**Recursive element lookup**: Utils::find_element_recursive traverses entire tree — large documents could be slow. **Style serialization assumes first entry is local**: If storage order changes, wrong style serialized. **include_content + element_id coupling**: Reject include_content without element_id — API contract clear but asymmetric. All mitigated: tests cover happy/edge paths; schema docs flag the constraint.


modules/atomic-widgets/styles/local-style-serializer.php
```diff
@@ -0,0 +1,100 @@
...
+		$desktop_variant = null;
+		$other_variants  = [];
+
+		foreach ( $variants as $variant ) {
+			$meta        = $variant['meta'] ?? [];
+			$is_desktop  = ( $meta['breakpoint'] ?? self::DESKTOP_BREAKPOINT ) === self::DESKTOP_BREAKPOINT;
+			$is_no_state = null === ( $meta['state'] ?? null );
+
+			if ( $is_desktop && $is_no_state && null === $desktop_variant ) {
+				$desktop_variant = $variant;
```
IshayMaya: why? is it the default variant? it's a bit confusing, maybe just. the naming

modules/atomic-widgets/styles/local-style-serializer.php
```diff
@@ -0,0 +1,100 @@
...
+ *     '<cssProp>'    => '<value>',
+ *     ...
+ *     '__custom_css' => '<decoded raw>',   // only if present
+ *     '__variants'   => [ <non-desktop variants raw> ],  // only if present
+ *   ]
+ */
+class Local_Style_Serializer {
+
+	const LOCAL_STYLE_ID_PREFIX = 's-';
+	const LOCAL_STYLE_TYPE      = 'class';
```
IshayMaya: what other types do we have

Omerisra6: I will find a better way to identify the local style

modules/atomic-widgets/styles/local-style-serializer.php
Omerisra6: Good catch — the `null === $desktop_variant` guard was there to pick only the first desktop/no-state variant and push the rest into `__variants`. Refactored into a dedicated `find_desktop_variant_index()` helper so the intent is explicit: find the index, split it out, return the remainder.

modules/atomic-widgets/styles/local-style-serializer.php
Omerisra6: There aren't other types in an element's own `styles` map — global classes live in `Global_Classes_Repository` and are only referenced by ID in `settings.classes`. Every entry in `$node['styles']` is already local by construction.

Also, the `s-` prefix filter was wrong: real local style IDs are `e-{elementId}-...` (see `create-element-style.ts`), so the filter never matched production data. Removed the type/prefix constants entirely and just take the first (only) entry in the map.

IshayMaya (approved):

PR #36614, opened 2026-07-22
Internal: Plain-only settings contract for MCP [ED-24951]

Omerisra6:
## What & Why

Close the loop between what the LLM **sees** in widget schemas and what the MCP `build-composition` / `manage-elements` abilities **accept** as `element_config` / `settings`. Before this PR, schemas advertised the internal envelope shape (`{ "$$type": "...", "value": ... }`) but the server had already started accepting plain input in some places — an inconsistent, confusing contract.

This PR makes the contract **plain-only end to end**.

### Plain resolvers registry (input path)

New `modules/atomic-widgets/plain-resolvers/`:

- `Plain_Resolver_Base` + `Plain_Resolvers_Registry` mirror the transformers registry.
- `Plain_Values_Resolver` walks a `Prop_Type` tree (`Union → Object → Array → leaf`) and converts plain input into typed prop envelopes.
- Leaf resolvers: `String`, `Number`, `Boolean`, `Size`, `Identity`.
- Composite resolvers: `Dynamic`, `Global_Variable`, `Html_V3` (short-circuit the walker with fully-formed subtrees, mirroring CSS value-parsers).
- `Size_Value_Parser` refactored to delegate to `Size_Plain_Resolver`.
- Wired into `Element_Config_Applier` (build-composition + manage-elements). `Llm_Prop_Value_Adjuster` deleted.

### Plain LLM schema converter (output path)

New `Plain_Llm_Schema_Converter` (`modules/atomic-widgets/prop-types/utils/`):

- Recursive walker that recognises the envelope shape `{ type: object, properties: { $$type: {const: K}, value: V } }` and replaces it with the plain form of `V`.
- Dynamic branches hoist to `{ name, settings }` (no `$$type` / `group`).
- Union `anyOf` branches are recursed and deduped (so redundant collapses like `[string, global-color-variable]` don't clutter the schema).
- Wired into `Widget_Context_Helper::apply_llm_schema_filters()` (runs **after** the existing envelope-aware filters `LLM_Schema_Dedupe_Filter` / `Overridable_LLM_Filter`) and into `List_Dynamic_Tags_Ability::build_settings_schema()`.
- `Prop_Type::to_json_schema()` is unchanged — only the MCP/LLM-facing seam converts.

### Prompts

Updated `build-composition.md`, `get-widget-schema.md`, `list-widget-schemas.md`, `list-dynamic-tags.md` to describe the plain contract and its rendering per type.

### Tests

- `test-plain-llm-schema-converter.php` — 14 tests (scalars, enum, object, array, union+dedupe, dynamic hoist, html-v3 nested unwrap, deep nesting, passthrough).
- Full test coverage for the plain-resolvers layer.
- `test-element-config-applier.php` covers the new applier flow.
- `test-build-composition-ability.php` / `test-manage-elements-ability.php` updated to send plain values.
- Orphan test for the deleted `Llm_Prop_Value_Adjuster` removed.

DB-less runner (`tests/phpunit/run-unit.sh`) hardened; `unit-bootstrap.php` gains WP sanitizer stubs so `Props_Parser` can run outside WordPress.

## Testing

```
tests/phpunit/run-unit.sh \
  tests/phpunit/elementor/modules/atomic-widgets/prop-types/utils/test-plain-llm-schema-converter.php \
  tests/phpunit/elementor/modules/atomic-widgets/plain-resolvers/*.php \
  tests/phpunit/elementor/modules/mcp/build-composition/test-element-config-applier.php
```

→ 58 tests, 97 assertions, all passing.

## Backward compatibility

The change is **schema-facing**: the LLM was already the only real consumer of these schemas via `elementor/v1/mcp-proxy`. Internal `to_json_schema()` calls (used by editor packages) are untouched.

## Ticket

Ref: ED-24951

## 1. Problem & Context

LLM clients previously sent widget settings as fully-enveloped PropValues (`{ $$type, value }`); now they send plain JSON scalars/objects (HTML as `{ content, children }`, dynamics as `{ name, settings }`). Plain-only contract simplifies LLM prompts and reduces schema verbosity for external callers.

## 2. What Changed (Where)

| Component | Change |
|-----------|--------|
| **Plain Resolvers** | New system: `Plain_Resolver_Base` + registry-based type dispatch (8 new files) converts plain → envelopes |
| **Schema Converter** | `Plain_Llm_Schema_Converter` unwraps envelope schemas to plain shapes for LLM consumption |
| **Element_Config_Applier** | Replaced `Llm_Prop_Value_Adjuster` + transformers with `Plain_Values_Resolver`; no longer expects pre-enveloped input |
| **Dynamic_Tag_Llm_Resolver** | Now accepts optional `$settings_resolver` callback; validates dynamic tag settings recursively via resolver |
| **MCP Abilities** | `Build_Composition` and `Manage_Elements` inject resolver instead of Variables_Service |
| **Widget_Context_Helper** | Converts JSON schemas to plain form via converter before returning to LLM |
| **Documentation** | Updated prompts: no more `$$type` wrappers in `element_config`, clearer examples |
| **Tests** | 358-line comprehensive test suite covering leaf/composite resolvers, schema conversion, integration |

## 3. How It Works

**Entry**: LLM sends `element_config` with plain values (`"title": "Hello"`, `"count": 42`, `"html": { "content": "...", "children": [] }`).

**Dispatch**: `Element_Config_Applier` routes each setting through `Plain_Values_Resolver`, which:
- Looks up registered resolver for the target `Prop_Type`
- For leaf types (String, Number, Boolean), coerces via resolver, wraps via `Prop_Type::generate()`
- For composites (Object, Array, Union), recurses shape/items; for special types (Dynamic, Html_V3), delegates to composite resolvers
- Returns `null` if unresolvable (skipped with warning)

**Dynamic tags**: `Dynamic_Tag_Llm_Resolver::resolve()` now accepts settings-resolver callback; builds strict schema-validated settings by resolving each tag prop through the same resolver chain.

**Schema generation**: `Widget_Context_Helper` calls `Plain_Llm_Schema_Converter::convert()` on JSON schemas—unwraps all `{ $$type: K, value: V }` envelopes to `V`'s plain form, dedupes union branches, preserves metadata.

## 4. Risks

- **Assumption: LLM calls resolve() with plain values.** If old clients still send envelopes, they become nested `{ $$type, value: { $$type, value: ... } }` → validation fails silently as warnings. Mitigation: test against real LLM client; document breaking change in release notes.
- **Dynamic tag settings validation now recursive.** Malformed nested settings (e.g., `{ name: 'tag', settings: { length: ['array'] } }`) are silently dropped from settings object rather than rejected. Mitigation: callers see `null` result if all settings fail; tests cover this.
- **Registry initialization in Module.** `get_settings_plain_values_resolver()` creates registry on first call via `do_action`, wiring composite resolvers last (Dynamic, Html_V3 depend on walker). If action fires before initialization, composites won't resolve. Mitigation: action fires in `__construct`, guaranteed before any ability runs.


modules/atomic-widgets/css-converter/value-parsers/size-value-parser.php
```diff
@@ -2,97 +2,16 @@
```
Omerisra6, line 2: Why do we still need this? Let's call the resolver directly

modules/atomic-widgets/plain-resolvers/resolvers/boolean-plain-resolver.php
```diff
@@ -0,0 +1,15 @@
...
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+class Boolean_Plain_Resolver extends Plain_Resolver_Base {
+	public function resolve( $plain_value ) {
+		return (bool) $plain_value;
+	}
+}
```
Omerisra6, line 15: shouldn't we return the prop type as an object? `[ '$$type' => 'boolean', 'value' => (bool) $plain_value ]`?

modules/atomic-widgets/plain-resolvers/resolvers/dynamic-plain-resolver.php
```diff
@@ -0,0 +1,142 @@
...
+use Elementor\Modules\AtomicWidgets\DynamicTags\Dynamic_Tags_Module;
+use Elementor\Modules\AtomicWidgets\PlainResolvers\Plain_Resolver_Base;
+use Elementor\Modules\AtomicWidgets\PlainResolvers\Plain_Values_Resolver;
+use Elementor\Modules\AtomicWidgets\PropTypes\Contracts\Prop_Type;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+class Dynamic_Plain_Resolver extends Plain_Resolver_Base {
```
Omerisra6, line 14: Ensure the logic is the same as the config applier dynamic logic we had

modules/atomic-widgets/plain-resolvers/resolvers/global-variable-plain-resolver.php
```diff
@@ -0,0 +1,34 @@
...
+namespace Elementor\Modules\AtomicWidgets\PlainResolvers\Resolvers;
+
+use Elementor\Modules\AtomicWidgets\PlainResolvers\Plain_Resolver_Base;
+use Elementor\Modules\Variables\Services\Variables_Service;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+class Global_Variable_Plain_Resolver extends Plain_Resolver_Base {
```
Omerisra6: same, ensure we didnt change it

modules/atomic-widgets/plain-resolvers/resolvers/identity-plain-resolver.php
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Elementor\Modules\AtomicWidgets\PlainResolvers\Resolvers;
+
+use Elementor\Modules\AtomicWidgets\PlainResolvers\Plain_Resolver_Base;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+class Identity_Plain_Resolver extends Plain_Resolver_Base {
```
Omerisra6: ?

modules/mcp/abilities/list-dynamic-tags-ability.php
```diff
@@ -78,7 +79,7 @@ private function build_settings_schema( array $props_schema ): object {
 				continue;
 			}
 
-			$settings[ $key ] = $prop_type->to_json_schema();
+			$settings[ $key ] = Plain_Llm_Schema_Converter::convert( $prop_type->to_json_schema() );
```
Omerisra6: why do we need it here?

modules/mcp/abilities/manage-elements-ability.php
```diff
@@ -226,7 +226,8 @@ private function handle_update( Document $document, string $element_id, array $i
 		$warnings = [];
 
 		if ( ! empty( $settings ) ) {
-			$config_applier = new Element_Config_Applier( $type_resolver, $variables_service );
+			$plain_values_resolver = AtomicWidgetsModule::instance()->get_settings_plain_values_resolver();
```
Omerisra6: Check how we usually use other modules?

modules/mcp/static-resources/abilities/build-composition.md
```diff
@@ -28,15 +28,24 @@ Some elements have internal tree structures (nesting). When using these elements
...
-- LINKS: a `link` prop is valid only when the target widget's schema (via `elementor/get-widget-schema`) includes a `link` property. On widgets without it, `link` is skipped and reported in `warnings` (the composition still builds) — wrap the element in a linkable container instead. Link PropValue shape: `{ "$$type": "link", "value": { "destination": { "$$type": "url", "value": "https://example.com" }, "isTargetBlank": { "$$type": "boolean", "value": true }, "tag": { "$$type": "string", "value": "a" } } }`
+- LINKS: a `link` prop is supported ONLY on these widgets: {{LINKABLE_WIDGETS}}. On any other widget a `link` is skipped and reported in `warnings` (the composition still builds) — to link a non-linkable element, wrap it in a linkable container. Plain link shape: `{ "destination": "https://example.com", "isTargetBlank": true, "tag": "a" }`
 - Retry on errors up to 10x
 - Check `llm_guidance.default_settings` in widget schemas — omit only keys listed there from element_config unless the user explicitly asks to change them
 
+## PLAIN element_config FORMAT
+Scalars and objects only — match the widget schema shape without envelopes:
```
Omerisra6: LLM shouldn't know about the envelop format at all, check no prompt has reference to that format

tests/phpunit/elementor/modules/atomic-widgets/plain-resolvers/test-composite-resolvers.php
```diff
@@ -0,0 +1,119 @@
+<?php
```
Omerisra6, line 1: Check if we can do a one happy path integration test that will check the ability to execute method, and a few tests that will ensure edge cases. All testing the tool itself with an input-output format.

modules/atomic-widgets/plain-resolvers/resolvers/dynamic-plain-resolver.php
```diff
@@ -0,0 +1,40 @@
...
+use Elementor\Modules\AtomicWidgets\PlainResolvers\Plain_Resolver_Base;
+use Elementor\Modules\AtomicWidgets\PlainResolvers\Plain_Values_Resolver;
+use Elementor\Modules\AtomicWidgets\PropTypes\Contracts\Prop_Type;
+use Elementor\Modules\Mcp\Abilities\Dynamic_Tag_Llm_Resolver;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+class Dynamic_Plain_Resolver extends Plain_Resolver_Base {
```
IshayMaya, line 14: are we going to reuse them in styles?

modules/atomic-widgets/plain-resolvers/resolvers/dynamic-plain-resolver.php
```diff
@@ -0,0 +1,40 @@
...
+	private ?Plain_Values_Resolver $walker;
+
+	public function __construct( ?Plain_Values_Resolver $walker = null ) {
+		$this->walker = $walker;
+	}
+
+	public function set_walker( Plain_Values_Resolver $walker ): void {
```
IshayMaya: why do we need that? where are we setting it

modules/atomic-widgets/plain-resolvers/resolvers/dynamic-plain-resolver.php
```diff
@@ -0,0 +1,40 @@
...
+		$this->walker = $walker;
+	}
+
+	public function resolve( $plain_value ) {
+		if ( ! is_array( $plain_value ) ) {
+			return null;
+		}
+
+		if ( ! isset( $plain_value['name'] ) || ! is_string( $plain_value['name'] ) ) {
```
IshayMaya, line 26: is it the dynamic name? what about the settings

modules/atomic-widgets/plain-resolvers/resolvers/html-v3-plain-resolver.php
```diff
@@ -0,0 +1,54 @@
...
+
+use Elementor\Modules\AtomicWidgets\PlainResolvers\Plain_Resolver_Base;
+use Elementor\Modules\AtomicWidgets\PlainResolvers\Plain_Values_Resolver;
+use Elementor\Modules\AtomicWidgets\PropTypes\Primitives\String_Prop_Type;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+class Html_V3_Plain_Resolver extends Plain_Resolver_Base {
```
IshayMaya, line 13: what happens in case of migration (html v4 prop type)

modules/atomic-widgets/plain-resolvers/resolvers/identity-plain-resolver.php
```diff
@@ -0,0 +1,19 @@
...
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * Registry fallback for prop types without a dedicated leaf resolver.
+ * Passes the plain value through; Plain_Values_Resolver wraps it via Prop_Type::generate().
+ */
+class Identity_Plain_Resolver extends Plain_Resolver_Base {
```
IshayMaya: do we still need that? and the name is confusing

modules/atomic-widgets/prop-types/utils/plain-llm-schema-converter.php
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * Converts an envelope-shaped JSON schema (as produced by `Prop_Type::to_json_schema()`) into the
+ * plain shape the LLM should send in `element_config` / dynamic-tag `settings`.
+ *
+ * Rule: whenever a subschema looks like `{ type: object, properties: { $$type: {const: K}, value: V } }`,
+ * replace it with the plain form of `V`, preserving top-level metadata (`description`, `examples`).
+ * `anyOf` branches are recursed and duplicates are collapsed so redundant variants
+ * (e.g. `string` vs `global-color-variable`) don't clutter the schema.
+ */
+class Plain_Llm_Schema_Converter {
```
IshayMaya, line 18: isn't it part of the other functions above? why is is not there?

modules/atomic-widgets/module.php
```diff
@@ -432,6 +445,32 @@ public function register_export_transformers( Transformers_Registry $transformer
 		$transformers->register( Svg_Src_Prop_Type::get_key(), new Svg_Src_Export_Transformer() );
 	}
 
+	private function register_settings_resolvers( Plain_Resolvers_Registry $registry ): void {
```
IshayMaya, line 448: maybe these functions should come from outside?

modules/atomic-widgets/plain-resolvers/resolvers/dynamic-plain-resolver.php
Omerisra6: Good catch — `set_walker()` was dead code. Removed it; the walker is now required via the constructor (same pattern as `Html_V3_Plain_Resolver`), wired in `get_settings_plain_values_resolver()`.

modules/atomic-widgets/plain-resolvers/resolvers/dynamic-plain-resolver.php
Omerisra6: Yes — `name` is the dynamic tag identifier. `settings` are validated recursively inside `Dynamic_Tag_Llm_Resolver::resolve()` via the `$settings_resolver` walker callback we pass in. The early guard here only rejects missing/non-string `name`.

modules/atomic-widgets/plain-resolvers/resolvers/dynamic-plain-resolver.php
Omerisra6: Not in this PR — styles use the envelope-serializers + CSS value-parsers path. The plain-resolvers registry is settings-only (`register_settings_resolvers` in `module.php`). If styles ever need plain input we can lift the registry up a layer then.

modules/atomic-widgets/plain-resolvers/resolvers/html-v3-plain-resolver.php
Omerisra6: No Html V4 prop type exists yet. This resolver is keyed by `Html_V3_Prop_Type::get_key()`. When V4 lands we register a `Html_V4_Plain_Resolver` under the new key. Until then unknown shapes fall through the passthrough fallback — safe, no crash. Migrations run when the stored envelope is loaded, not inside the plain-resolvers path.

modules/atomic-widgets/plain-resolvers/resolvers/identity-plain-resolver.php
Omerisra6: Still needed — `Plain_Values_Resolver::resolve_with_fallback` requires a registered fallback for leaf prop types without a dedicated resolver. Renamed to `Passthrough_Plain_Resolver` to better reflect what it does (pass-through; wrapping happens in the walker via `Prop_Type::generate()`).

modules/atomic-widgets/prop-types/utils/plain-llm-schema-converter.php
Omerisra6: `LLM_Schema_Filter` subclasses decorate envelope-shaped schemas (dedupe, overridable hints). `Plain_Llm_Schema_Converter` is a final shape rewrite that runs after all filters in `Widget_Context_Helper::apply_llm_schema_filters()` — it unwraps `{$$type, value}` envelopes into plain JSON. Keeping it separate avoids mixing decoration with structural conversion.

modules/atomic-widgets/module.php
Omerisra6: Kept symmetric with `register_settings_transformers` and the other `register_*` methods in `module.php` — extraction would need to happen for the whole family (settings/styles/import/export transformers + resolvers) together. Happy to follow up in a separate ticket if we want that.

IshayMaya (approved):

PR #36617, opened 2026-07-22
Internal: Bulk manage global classes/variables MCP tools [ED-24951]

Omerisra6:
## Summary
- Extend MCP manage-classes and manage-variable abilities to support bulk create/update/delete operations with compact response shapes.
- Add partial-success handling so callers receive per-item outcomes instead of failing the entire batch on a single error.
- Align related tests and supporting services (variables service, document mutator utilities) with bulk MCP flows.

## Test plan
- [ ] Run PHPUnit: `tests/phpunit/elementor/modules/mcp/test-manage-classes-ability.php`
- [ ] Run PHPUnit: `tests/phpunit/elementor/modules/mcp/test-manage-variable-ability.php`
- [ ] Run PHPUnit: `tests/phpunit/elementor/modules/mcp/test-manage-elements-ability.php` (regression)
- [ ] Exercise MCP tools against a local site: bulk create/update/delete global classes and variables; verify compact responses and partial success payloads
- [ ] Confirm vendor directory is intact locally (not committed as deletion)

## 1. Problem & Context

Refactor single-item MCP tools (manage-classes, manage-variables) to support bulk operations (up to 50 items/request) with lenient error handling—allowing partial success instead of all-or-nothing transactions. ED-24951.

## 2. What Changed (Where)

- **manage-classes-ability.php**: Replaced single action handlers with bulk pipeline—operation translation → class item building → label deduplication → max-item enforcement → atomic commit.
- **manage-variable-ability.php**: Refactored to translate operations, delegate batch to `Variables_Service.process_batch(lenient=true)`, merge results with per-operation status tracking.
- **variables-service.php**: Added `$lenient` parameter to `process_batch()` for partial-success mode; returns individual result records instead of throwing on first error.
- **bulk-operations-result.php** (new): Result aggregator tracking per-operation status, computing overall status (ok/partial_error/error).
- **Tests**: Refactored to assert bulk response structure (`results[]` array, per-item status codes) and lenient error semantics.

## 3. How It Works

Input `operations[]` → validate count ≤50 → translate each (extract action/params) → build items (CSS parsing, ID generation) → dedup labels (auto-rename with `DUP_` prefix) → enforce max-items limit → batch commit if any succeeds → return `{status, results[], order/watermark}` with per-operation `{index, action, status, code?, id?, label?}`.

For variables, lenient mode catches exceptions per-operation and continues; classes pre-validate before any mutation.

## 4. Risks

**Label collision in concurrent requests**: Deduplication is in-memory; simultaneous bulk ops could both trigger rename logic. Mitigated by atomic repository commit (first writer wins). **Partial failure without rollback**: Classes atomically update touched items; variables save if ≥1 success. Consumer must retry failed ops. Acceptable for MCP use case—AI can retry with corrected input.

### DIFF
--- a/modules/mcp/abilities/manage-variable-ability.php
+++ b/modules/mcp/abilities/manage-variable-ability.php
@@ -257,6 +257,13 @@ class Manage_Variable_Ability extends Abstract_Ability {
 		}
 
 		Plugin::$instance->files_manager->clear_cache();
+		$this->flush_runtime_cache();
+	}
+
+	private function flush_runtime_cache(): void {
+		if ( function_exists( 'wp_cache_flush_runtime' ) ) {
+			wp_cache_flush_runtime();
+		}
 	}
 
 	private function get_service(): Variables_Service {
--- a/modules/mcp/abilities/manage-variable-ability.php
+++ b/modules/mcp/abilities/manage-variable-ability.php
@@ -4,10 +4,7 @@ namespace Elementor\Modules\Mcp\Abilities;
 
 use Elementor\Modules\Variables\Services\Batch_Operations\Batch_Processor;
 use Elementor\Modules\Variables\Services\Variables_Service;
-use Elementor\Modules\Variables\Storage\Exceptions\DuplicatedLabel;
-use Elementor\Modules\Variables\Storage\Exceptions\RecordNotFound;
-use Elementor\Modules\Variables\Storage\Exceptions\Type_Mismatch;
-use Elementor\Modules\Variables\Storage\Exceptions\VariablesLimitReached;
+use Elementor\Modules\Variables\Storage\Exceptions\FatalError;
 use Elementor\Modules\Variables\Storage\Variables_Repository;
 use Elementor\Plugin;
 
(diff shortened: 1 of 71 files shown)


modules/mcp/abilities/manage-variable-ability.php
```diff
@@ -86,142 +85,177 @@ protected function get_definition(): Ability_Definition {
...
-		}
-
-		try {
-			$result = $this->get_service()->update( $id, [
-				'label' => $label,
-				'value' => $value,
-			] );
-		} catch ( \Exception $e ) {
-			return $this->map_service_exception( $e );
+	private function handle_bulk( array $operations ) {
```
ronros-elementor, line 116: I would also add some flush cache calling at the end of the bulk, similar to how we handle batches in the [global-classes-repository](https://github.com/elementor/elementor/blob/6639799627ff47b6b4add005c3e10e75e5840a6b/modules/global-classes/global-classes-repository.php#L499-L503)

ronros-elementor (approved):

PR #36628, opened 2026-07-23
Internal: Plain values in MCP tool output [ED-24951]

Omerisra6:
## Summary

Add envelope→plain output serializers as the counterpart to `Plain_Values_Resolver`, so MCP tools return plain values instead of internal `{ $$type, value }` envelopes.

Rebased onto `main` after #36614 merged. This PR is the output-only delta (3 commits).

## What changes

| Surface | Before | After |
|---------|--------|-------|
| `get-page-structure` settings (`include_content=true`) | Raw envelopes | Plain values (atomic widgets) |
| `get-page-structure` style `__variants` | Enveloped props | Flat CSS map |
| `llm_guidance.default_settings` | Raw envelopes | Plain values |

## Test plan

- [ ] `tests/phpunit/run-unit.sh tests/phpunit/elementor/modules/atomic-widgets/envelope-serializers/*.php`
- [ ] `vendor/bin/phpunit --filter Test_Get_Structure_Ability`
- [ ] `vendor/bin/phpunit --filter Test_Llm_Guidance_Builder`
- [ ] MCP smoke: `get-page-structure` with `include_content=true` — confirm settings and `__variants` have no `$$type` keys

Ref: ED-24951

## 1. Problem & Context

ED-24951 changes MCP tool output to return plain resolved values instead of wrapped PropType objects. This simplifies LLM consumption by removing internal schema metadata (`$type`, `value` wrappers) from settings and variants while maintaining schema-driven resolution.

## 2. What Changed (Where)

- **Local_Style_Serializer**: Added `serialize_variant()` to normalize variant objects with resolved CSS props and custom CSS
- **Get_Structure_Ability**: Integrated `Render_Props_Resolver` to resolve settings through schema before output
- **Widget_Context_Helper**: Removed base_settings hints logic; schema filtering now produces plain LLM values only
- **Llm_Guidance_Builder**: Stripped default_settings collection and instruction (no longer needed)
- **Tests**: Updated assertions to expect plain values (`'tag' => 'section'`) instead of wrapped objects

## 3. How It Works

Settings flow: raw settings → schema intersect → `Render_Props_Resolver.resolve()` → plain values in skeleton. Variants similarly resolved through `serialize_variant()`. LLM guidance no longer emits base_settings hints since all props are now equally configurable (no special handling for defaults).

## 4. Risks

Settings resolution assumes schema availability; missing schema silently passes through raw settings. No validation that resolved values match LLM expectations—test coverage validates happy path but edge cases (malformed `$type` wrappers, schema mismatches) lack explicit handling.


Omerisra6:
Rebased onto latest `feat/ED-24951-plain-only-settings-contract` (#36614). Branch is 2 commits ahead of #36614:

- `09961dd` feat(mcp): return plain settings and styles in tool results
- `2199094` test(mcp): assert plain output for settings, variants, and guidance

GitHub cannot set #36614's branch as the PR base on `elementor/elementor` (branch only exists on the fork), so this PR still targets `main` and shows all 15 commits.

**For a clean 2-commit review**, use the stacked fork PR: https://github.com/Omerisra6/elementor/pull/8

After #36614 merges, this branch should be rebased onto `main` and the diff will shrink to the output-only commits.

modules/mcp/abilities/get-structure-ability.php
```diff
@@ -129,14 +132,40 @@ private function prune_elements( array $elements, bool $include_content = false
...
+				$skeleton['settings'] = $settings ? $settings : (object) [];
 				$skeleton['styles']   = Local_Style_Serializer::serialize( $node['styles'] ?? [] );
 			}
 
 			return $skeleton;
 		} );
 	}
 
+	private function resolve_props_schema( array $node ): ?array {
+		$type = $node['widgetType'] ?? $node['elType'] ?? null;
```
annab1: I think you can use `get_element_type` here

modules/mcp/abilities/get-structure-ability.php
```diff
@@ -129,14 +132,40 @@ private function prune_elements( array $elements, bool $include_content = false
...
+
+		if ( ! $type ) {
+			return null;
+		}
+
+		$instance = Atomic_Elements_Utils::get_element_instance( $type );
+
+		if ( ! Atomic_Elements_Utils::is_atomic_element( $instance ) ) {
+			return null;
+		}
```
annab1: Why is this check needed?

modules/mcp/abilities/utils/llm-guidance-builder.php
```diff
@@ -20,7 +21,10 @@ public static function build( array $config, string $widget_type, array $parents
 		];
 
 		$default_styles = self::collect_default_styles( $config['base_styles'] ?? [] );
-		$default_settings = $config['base_settings'] ?? [];
+		$default_settings = self::collect_default_settings(
```
annab1: redundant for now

annab1 (changes requested):

modules/mcp/abilities/get-structure-ability.php
Omerisra6: we are not showing v3 elements in the structure, maybe we should?

annab1 (approved):

PR #36635, opened 2026-07-23
Internal: MCP style applier returns actionable validation feedback [ED-24951]

Omerisra6:
## What
Enriches validation feedback from the MCP `manage-elements` and `build-composition` abilities so LLM callers can self-correct instead of guessing at a bare `status: ok` or an opaque `invalid_value` parser error.

## Why
Today, when an update-style call fails or silently drops declarations, the caller gets almost nothing useful:
- `Style_Applier` returns `{props: [], customCss: '', rejected: []}` for unhandled declarations → `status: ok`, no signal at all.
- `Style_Parser` rejections bubble up as `variants[0].props.background: invalid_value` — the LLM cannot tell that e.g. `background-image: url(/relative)` failed because the URL must be absolute.
- Unknown widget property in settings already links to `elementor://widgets/schema/<tag>`; style errors did not.

## Changes
All in the MCP adapter layer — no changes to `Style_Parser` / `Css_Converter` public APIs.

- **Silent-drop path** (empty props + empty custom_css) now emits a warning listing the input declarations, a URL-must-be-absolute hint when any value uses `url(...)`, and `See elementor://widgets/schema/<tag>`.
- **custom_css fallback** warning now echoes the leftover declaration text, not just a generic 'some properties fell back'.
- **Rejected-variable** error now lists the input declarations that referenced the unresolved variables.
- **Style_Parser errors** now include the input declarations, URL hint, and schema link.
- `Style_Applier::apply()` accepts an optional per-config-id tag map (backwards-compatible default `[]`). `Manage_Elements_Ability` passes the resolved widget/element tag; `Build_Composition_Ability` derives it from the resolved subtree index.

## Example outputs
Before: `{"status":"ok","post_id":37147,"element_id":"30f21a26","version":"..."}`

After (shorthand background):
```
[30f21a26] No style applied. Unhandled declarations: background: url(/wp-content/uploads/x.png) center/cover no-repeat. URLs must be absolute (include the http(s):// scheme, e.g. https://example.com/image.png). Use longhand properties (e.g. background-image, background-size) instead of shorthands. See elementor://widgets/schema/e-heading.
```

After (relative URL in longhand):
```
[30f21a26] No style applied. Unhandled declarations: background-image: url(/wp-content/uploads/x.png). URLs must be absolute (...). See elementor://widgets/schema/e-heading.
```

After (var fallback):
```
[30f21a26] Some declarations fell back to custom_css (may not render on Pro 3.35+): color: var(--e-gv-nonexistent);. Prefer atomic longhand properties. See elementor://widgets/schema/e-heading.
```

## Tests
Adds two integration tests to `test-manage-elements-ability.php` covering the silent-drop warning and the parse-error message shape.

Verified live via the local MCP against a real Elementor document (post 37147).

## Ref
No Jira ticket yet — internal DX improvement.

## 1. Problem & Context

MCP style applier now returns actionable validation feedback (ED-24951). Error/warning messages lacked context about what failed and why, making debugging difficult for users applying styles to elements.

## 2. What Changed (Where)

- **build-composition-ability.php**: Pass element types to `Style_Applier::apply()` via new `element_types_from_index()` helper.
- **style-applier.php**: Accept optional `$element_types` parameter; enriched error/warning messages with declarations, URL hints, schema links; extracted formatting logic into reusable methods.
- **manage-elements-ability.php**: Pass element type when calling `Style_Applier::apply()`.
- **test-manage-elements-ability.php**: Added test validating warnings include declarations, URL hints, and schema references.

## 3. How It Works

Style validation errors now include: (1) config ID, (2) input declarations formatted as CSS, (3) context-specific hints (URL validation for `url()` values, schema link for widget type), and (4) parser errors. New private methods (`format_parse_error`, `format_declarations`, `url_hint_for_declarations`, `schema_link_hint`) centralize message formatting. Element types flow from document index through applier to enable schema hints in error messages.

## 4. Risks

None identified. Changes are additive (new optional parameter defaults to empty array), error messages are more informative without changing logic, and test coverage validates new behavior.


modules/mcp/abilities/build-composition-ability.php
```diff
@@ -341,6 +341,17 @@ private function as_map( $value ): array {
 		return is_array( $value ) ? $value : [];
 	}
 
+	private function tags_from_index( array $index ): array {
+		$tags = [];
```
IshayMaya: why tag?

IshayMaya (approved):

PR #36663, opened 2026-07-26
Internal: Support interactions in mcp [ED-24951]

Omerisra6:
## Summary
- Add `interactions` support to `build-composition` with a flat scalar LLM input shape (e.g. `{ on, effect, type, for, after, ease }`).
- Introduce `Interaction_Converter` that normalizes flat input and delegates envelope building to the shared `Plain_Values_Resolver`.
- Add `Size_Plain_Resolver` so timing/size fields (`for`/`after`) resolve to valid size envelopes.
- Expose `elementor://interactions/schema` resource with Pro-aware allowed values and defaults.

## Test plan
- [x] `tests/phpunit/run-unit.sh tests/phpunit/elementor/modules/mcp/abilities/build-composition/test-interaction-converter.php tests/phpunit/elementor/modules/mcp/abilities/build-composition/test-interactions-applier.php`
- [ ] MCP `build-composition` dry run with flat interaction on a paragraph widget
- [ ] Verify `elementor://interactions/schema` lists expected enums and Pro-gated fields
- [ ] Confirm interaction with `for`/`after` no longer fails PropType validation

## 1. Problem & Context

MCP abilities now support interactions—the native event-trigger-animation system. Extends build-composition and manage-elements to accept/apply interactions alongside settings and styles; adds schema resource for LLM reference.

## 2. What Changed (Where)

| Module | Change |
|--------|--------|
| atomic-widgets | Register `Size_Plain_Resolver`; add plain transformers hook & `for_plain()` context |
| mcp/appliers | Move 3 appliers to new namespace; add `Interactions_Applier` (resolves & validates items) |
| mcp/abilities | Build_Composition & Manage_Elements now accept `interactions` param; new `Interactions_Schema_Resource_Ability` |
| mcp/utils | Enhance `Widget_Context_Helper.to_plain_llm_schema()` to strip pro fields & enrich enums from `meta()` |
| tests | New test suites for applier & helper; stub for Utils::has_pro() |

## 3. How It Works

Input interactions flow: build-composition/manage-elements receive `interactions: { config-id: [item, ...] }` → `Interactions_Applier.apply()` validates & resolves each item via `Plain_Values_Resolver` using `Interaction_Item_Prop_Type` → resolved items written to index with schema version. `Get_Structure_Ability` serializes stored interactions back to plain shape for output. Schema exposed via `Interactions_Schema_Resource_Ability` (URI: `elementor://interactions/schema`).

## 4. Risks

Minimal. Namespace refactor (Build_Composition → Appliers) is mechanical and test-verified. Pro/free schema filtering adds complexity but isolated in helper with tests. Interactions feature gated behind experiment flag with no-op warnings if inactive.


modules/mcp/abilities/build-composition/interactions-applier.php
```diff
@@ -0,0 +1,128 @@
...
+use Elementor\Modules\Interactions\Props\Interaction_Item_Prop_Type;
+use Elementor\Plugin;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+class Interactions_Applier {
+
+	const MAX_INTERACTIONS_PER_ELEMENT = 5;
```
Omerisra6: where does this limit come from?

modules/mcp/abilities/build-composition/interactions-applier.php
```diff
@@ -0,0 +1,128 @@
...
+	const SCHEMA_VERSION = 1;
```
Omerisra6: why?

modules/mcp/static-resources/abilities/build-composition.md
```diff
@@ -137,6 +138,39 @@ BAD: `<e-flexbox style="height:100vh"><e-div-block style="height:100vh">overflow
...
+  "breakpoints": { "excluded": [] }
+}
+```
+
+- Required: `trigger`, `animation.effect`, `animation.type`
+- Optional timing: `animation.timing_config.duration`, `animation.timing_config.delay` (`{ "size": <number>, "unit": "ms" }`)
+- Optional direction: `animation.direction` (relevant for slide)
+- Optional easing: `animation.config.easing`
+- Optional breakpoints: `breakpoints.excluded` (array of breakpoint IDs to disable)
+- Pro-only fields (see schema): `animation.config.repeat`, `animation.config.times`, `animation.config.replay`, `animation.config.relativeTo`, `animation.config.start`, `animation.config.end`, `animation.custom_effect` (when `animation.effect` is `custom`)
```
Omerisra6: why do we need that? it comes from the interactions resource

modules/atomic-widgets/props-resolver/render-props-resolver.php
```diff
@@ -33,6 +34,16 @@ public static function for_settings(): self {
 		return static::instance( self::CONTEXT_SETTINGS );
 	}
 
+	public static function for_plain(): self {
```
Omerisra6, line 37: interactions has no transformers so we just return the plain values

modules/mcp/abilities/appliers/interactions-applier.php
```diff
@@ -0,0 +1,125 @@
+<?php
```
annab1, line 1: I think the appliers should be in a separate folder `appliers` (on the same level as `build-composition`) -  since they are used in several tools - build_composition, manage_element, etc..

modules/mcp/abilities/build-composition/interactions-applier.php
```diff
@@ -0,0 +1,125 @@
...
+	exit;
+}
+
+class Interactions_Applier {
+
+	private bool $is_experiment_active;
+	private Plain_Values_Resolver $plain_values_resolver;
+
+	public function __construct( ?bool $is_experiment_active = null, ?Plain_Values_Resolver $plain_values_resolver = null ) {
+		$this->is_experiment_active = $is_experiment_active ?? $this->detect_experiment_active();
```
annab1: the class shouldn't be aware of experiments, it should be handled on a higher level - if the experiment is active, don't create an instance of the InterationsApplier

modules/mcp/abilities/appliers/interactions-applier.php
```diff
@@ -0,0 +1,125 @@
...
+		$prop_type = Interaction_Item_Prop_Type::make();
+
+		foreach ( $interactions as $config_id => $items ) {
+			if ( ! isset( $index[ $config_id ] ) ) {
+				continue;
+			}
+
+			if ( ! is_array( $items ) ) {
+				$errors[] = sprintf( '[%s] Interactions must be an array.', $config_id );
+				continue;
```
annab1, line 45: Don't we need to return in this case?

modules/mcp/abilities/build-composition/interactions-applier.php
```diff
@@ -0,0 +1,125 @@
...
+		if ( ! $this->is_experiment_active ) {
+			$warnings[] = __( 'Interactions experiment is not active. Interactions were not applied.', 'elementor' );
+			return [
+				'error' => null,
+				'warnings' => $warnings,
+			];
+		}
+
+		$errors = [];
+		$prop_type = Interaction_Item_Prop_Type::make();
```
annab1: this is built only to be passed to `resolve_items`. No need, resolve_items can use it directly

modules/mcp/abilities/appliers/interactions-applier.php
```diff
@@ -0,0 +1,125 @@
...
+
+		foreach ( $items as $item_index => $plain_item ) {
+			if ( ! is_array( $plain_item ) ) {
+				$errors[] = sprintf( '[%s] Interaction at index %d must be an object.', $config_id, $item_index );
+				continue;
+			}
+
+			$resolved = $this->plain_values_resolver->resolve( $plain_item, $prop_type );
+
+			if ( null === $resolved || ! $prop_type->validate( $resolved ) ) {
```
annab1, line 97: what if we want to unapply an interaction? wouldn't it be null in this case and should be applied regardless?

modules/mcp/abilities/build-composition-ability.php
```diff
@@ -139,7 +140,16 @@ public function execute( $input = [] ) {
 			return $style_result['error'];
 		}
 
-		$warnings = array_merge( $config_result['warnings'], $style_result['warnings'] );
+		$interactions_applier = new Interactions_Applier( null, $this->create_plain_values_resolver() );
```
annab1: 1. You don't pass the actual experiment, you pass null
2. `create_plain_values_resolver` should be `get_plain_values_resolver` since it doesn't create an instance on each call

modules/mcp/abilities/manage-elements-ability.php
```diff
@@ -192,10 +198,11 @@ private function handle_update( Document $document, string $element_id, array $i
 		$settings = $this->as_map( $input['settings'] ?? [] );
 		$style = $this->as_map( $input['style'] ?? [] );
 		$classes = $input['classes'] ?? null;
+		$interactions = $input['interactions'] ?? null;
 
-		$has_change = ! empty( $settings ) || ! empty( $style ) || ! empty( $classes );
+		$has_change = ! empty( $settings ) || ! empty( $style ) || ! empty( $classes ) || null !== $interactions;
```
annab1, line 311: why not `! empty( $interactions )` like we did with the rest?

modules/mcp/static-resources/abilities/build-composition.md
```diff
@@ -1,6 +1,7 @@
 # RESOURCES (Read before use)
 - [elementor://global-classes] - Reusable CSS classes from the active kit; check FIRST before adding inline styles
 - [elementor://global-variables] - Design tokens from the active kit; use labels in CSS as `var(--label)` or `var(--label, fallback)`; ONLY variables listed here are valid
+- [elementor://interactions/schema] - Native interaction item shape, allowed enums, and Pro-gated fields for `interactions`
```
annab1: I don't think that the fact that there are Pro-gated fields should be reflected in this description, WDYT?

modules/mcp/static-resources/abilities/build-composition.md
```diff
@@ -137,6 +138,11 @@ BAD: `<e-flexbox style="height:100vh"><e-div-block style="height:100vh">overflow
 - Use rem/em exclusively for responsive scaling
 - Generous padding on CTAs: min 1rem 2.5rem
 
+# INTERACTIONS
+Attach element interactions via the `interactions` parameter — a record mapping `configuration-id` → array of native-shape interaction items. Read [elementor://interactions/schema] for the full shape, allowed enum values, and Pro-gated fields.
+
+If the Interactions experiment is inactive, interactions are skipped and a warning is returned. Invalid values fail the call with `elementor_invalid_interactions`.
```
annab1: why is it relevant? it's just another error

modules/mcp/abilities/utils/interactions-llm-schema-builder.php
```diff
@@ -0,0 +1,100 @@
...
+		}
+
+		$properties = $schema['properties'];
+
+		foreach ( $prop_type->get_shape() as $key => $child_prop_type ) {
+			if ( ! isset( $properties[ $key ] ) ) {
+				continue;
+			}
+
+			if ( ! $is_pro_active && self::is_pro_only_field( $child_prop_type ) ) {
```
annab1: this logic should be incorporated into the base `to_plain_llm_schema` function so we don't have to repeat it for every type of PropType that has pro fields

modules/mcp/abilities/interactions-schema-resource-ability.php
```diff
@@ -0,0 +1,40 @@
...
+	const URI = 'elementor://interactions/schema';
+
+	protected function get_ability_id(): string {
+		return 'elementor/interactions-schema-resource';
+	}
+
+	protected function get_definition(): Ability_Definition {
+		return new Ability_Definition(
+			__( 'Interactions Schema', 'elementor' ),
+			__( 'Native interaction input schema derived from Interaction_Item PropType: shape, allowed enums, and Pro-gated fields for build-composition.', 'elementor' ),
```
annab1: why do we need to mention that it is derived from Interaction_Item PropType? don't we transform the prop type schema for the LLM?

modules/mcp/abilities/list-resources-ability.php
```diff
@@ -79,6 +79,12 @@ private function get_resource_catalog(): array {
 				'description' => 'Design tokens (colors, fonts, sizes) from the active kit; check before styling with variables.',
 				'mimeType' => 'application/json',
 			],
+			[
+				'uri' => Interactions_Schema_Resource_Ability::URI,
+				'name' => 'Interactions Schema',
+				'description' => 'Flat interaction keys, enums, defaults, and Pro-gated fields for build-composition.',
```
annab1: same about "Pro-gated fields"

annab1 (changes requested):

modules/mcp/abilities/appliers/interactions-applier.php
Omerisra6: Keeping the collect-and-aggregate pattern here to match sibling appliers like `Style_Applier`: bad shapes across configs still accumulate into a single `elementor_invalid_interactions` `WP_Error` at the end (via `$errors[]`), so callers get complete feedback instead of first-bad-config-wins. Happy to revisit if we standardize on hard-fail for all appliers.

modules/mcp/abilities/appliers/interactions-applier.php
Omerisra6: Good catch — the current per-item null-reject blocks in-array unapply. Since the write shape is a whole-record replace keyed by `configuration-id`, I implemented full-clear semantics instead: sending `[]` for a `configuration-id` now writes an empty items list (test: `test_apply__empty_items_array_clears_interactions_on_node`), and I've documented that in both the input schema and `build-composition.md`. Per-item null-as-remove would need a separate design pass since the items are unordered.

modules/mcp/abilities/manage-elements-ability.php
Omerisra6: Kept `null !== $interactions` here on purpose: the applier now supports `[]` as an explicit clear (writes empty items). Switching to `! empty($interactions)` would silently swallow the clear-intent case. `null` still means "no change to interactions on this op".

Omerisra6:
Pushed review follow-ups in 3926a2f.

**Kept as-is (pushback):**
- **Non-array `items`:** still collect into `$errors[]` and return one `elementor_invalid_interactions` at the end — same pattern as `Style_Applier`.
- **Unapply:** clear with `[]` for the config-id / element (now implemented + documented). Per-item `null` stays invalid.
- **`manage-elements` guard:** kept `null !== $interactions` so explicit `[]` clear is not skipped by `!empty()`.

**Addressed:** appliers folder, experiment gating at ability layer, inlined `Interaction_Item_Prop_Type::make()`, Pro stripping in `to_plain_llm_schema`, description cleanups, `get_plain_values_resolver` rename.

modules/mcp/abilities/utils/widget-context-helper.php
```diff
@@ -177,6 +185,87 @@ private static function to_plain_llm_schema_from_json( array $schema ): array {
 		return Plain_Llm_Schema_Converter::convert( $filtered );
 	}
 
+	private static function is_pro_active(): bool {
```
annab1: Use the existing function that checks if pro is active

modules/mcp/abilities/utils/widget-context-helper.php
```diff
@@ -167,8 +169,14 @@ private static function apply_llm_schema_filters( array $properties ): array {
 		return $properties;
 	}
 
-	public static function to_plain_llm_schema( Transformable_Prop_Type $prop_type ): array {
-		return self::to_plain_llm_schema_from_json( $prop_type->to_json_schema() );
+	public static function to_plain_llm_schema( Transformable_Prop_Type $prop_type, ?bool $is_pro_active = null ): array {
```
annab1: is_pro_active is a redundant param

modules/mcp/abilities/utils/widget-context-helper.php
```diff
@@ -177,6 +185,87 @@ private static function to_plain_llm_schema_from_json( array $schema ): array {
...
+			$schema['enum'] = $is_pro_active
+				? array_values( $enum_values )
+				: self::filter_pro_enum_values( $enum_values, $prop_type );
+		}
+
+		return $schema;
+	}
+
+	private static function filter_pro_enum_values( array $enum_values, Prop_Type $prop_type ): array {
+		$pro_values = $prop_type->get_meta_item( 'pro' );
```
annab1, line 253: use is_pro_only_field

modules/mcp/abilities/utils/widget-context-helper.php
```diff
@@ -167,8 +169,14 @@ private static function apply_llm_schema_filters( array $properties ): array {
 		return $properties;
 	}
 
-	public static function to_plain_llm_schema( Transformable_Prop_Type $prop_type ): array {
-		return self::to_plain_llm_schema_from_json( $prop_type->to_json_schema() );
+	public static function to_plain_llm_schema( Transformable_Prop_Type $prop_type, ?bool $is_pro_active = null ): array {
+		$schema = self::to_plain_llm_schema_from_json( $prop_type->to_json_schema() );
+
+		if ( $prop_type instanceof Prop_Type ) {
```
annab1: check if needed, since $prop_type is of type Transformable_Prop_Type

annab1 (changes requested):

annab1 (approved):

PR #36664, opened 2026-07-26
Feature: Use component instances [ED-24706]

ronros-elementor:
## 1. Problem & Context

Enable users to place reusable component instances in compositions via a new `<e-component>` tag with overridable props. Adds discovery/schema abilities for MCP clients and refactors component extraction logic into a reusable utility.

## 2. What Changed (Where)

| Module | Change |
|--------|--------|
| `circular-dependency-validator.php` | Extract component ID parsing into `Component_Instance_Prop_Type::extract_component_id()` |
| `component-instance.php` | Use extracted utility instead of inline logic |
| `component-styles.php` | Replace filter+map pattern with extracted utility |
| `module.php` | Convert `is_experiment_active()` to static method |
| `prop-types/component-instance-prop-type.php` | Add `extract_component_id()` static helper |
| `mcp/abilities/appliers/component-instance-applier.php` | **New** — full applier for building/updating component instance envelopes |
| `mcp/abilities/appliers/element-config-applier.php` | Route `<e-component>` entries to `Component_Instance_Applier`, add document context parameter |
| `mcp/abilities/list-components-ability.php` | **New** — discovery + schema endpoint for components |
| `mcp/abilities/manage-elements-ability.php` | Handle partial component updates via applier |
| `mcp/module.php` | Register `List_Components_Ability` when components active |
| Tests & mocks | Comprehensive test coverage + mock prop shape updates |
| Documentation | Add `list-components` guidance and placement examples to build-composition |

## 3. How It Works

**Discovery flow**: `list-components` (no args) → summary list; with `component_ids` → full schema including `overridable_props` per component.

**Placement**: `<e-component configuration-id="hero">` in XML + `element_config: { component_id: 42, overrides: {...} }` shorthand. `Component_Instance_Applier` validates circular deps, resolves override values against prop types, builds full PropValue envelope.

**Updates**: `manage-elements` with `action=update` on e-component nodes triggers `apply_partial()` for per-key override merging (null = delete, omitted = preserve).

Circular dep check runs on Component_Document targets. Override keys validated against component's declared overridable props; values resolved via `Plain_Values_Resolver`.

## 4. Risks

**Breaking change**: `Element_Config_Applier::apply()` now requires optional `Document` parameter for component validation. Callers passing 3 args will fail — mitigated by internal-only usage (MCP module, tests).

**Scope creep on overrides**: Per-key merge in `apply_partial()` adds complexity; nested list traversal by `override_key` could fail on malformed envelopes — mitigated by full validation in `apply()` and tests.

**Component repo queries per placement**: No caching of component lookups during batch apply; large compositions with repeated component IDs will query repo N times — acceptable trade-off for simplicity; can optimize later if profiling shows issue.


modules/mcp/module.php
```diff
@@ -118,4 +127,15 @@ public function register_server( $adapter ) {
 			return;
 		}
 	}
+
+	private function is_components_active(): bool {
```
Omerisra6, line 110: maybe move it to the module

modules/mcp/abilities/build-composition/component-instance-applier.php
```diff
@@ -0,0 +1,248 @@
...
+	 * full component_instance PropValue envelope that Props_Parser expects.
+	 *
+	 * @param array<string, array&> $config_id_index    Index of subtree refs (from Subtree_Builder).
+	 * @param array                 $element_config     Per-config-id settings (modified by reference).
+	 * @param Document              $document           Target document (used for circular-dep check).
+	 * @param array                 $rewritten_config_ids Config-ids whose settings were fully resolved here
+	 *                                                     and must bypass generic schema adjustment downstream.
+	 * @return \WP_Error|null
+	 */
+	public function rewrite( array $config_id_index, array &$element_config, Document $document, array &$rewritten_config_ids = [] ): ?\WP_Error {
```
Omerisra6: whats the usecase here?

modules/mcp/abilities/list-component-schemas-ability.php
```diff
@@ -0,0 +1,139 @@
...
+			'id'               => $component->get_main_id(),
+			'name'             => $component->get_post()->post_title,
+			'uid'              => $component->get_component_uid(),
+			'is_archived'      => $component->get_is_archived(),
+			'overridable_props' => $props_schema,
+			'example_shorthand' => $this->build_example_shorthand( $component_id, $props_schema ),
+		];
+	}
+
+	private function build_props_schema( array $props ): array {
```
Omerisra6: ensure the shape returned here is plain value json schema like we do for settings and styles (no $$type)

modules/mcp/abilities/get-component-schema-ability.php
```diff
@@ -0,0 +1,139 @@
...
+		} catch ( \Exception $e ) {
+			return null;
+		}
+	}
+
+	private function build_example_shorthand( int $component_id, array $props_schema ): array {
+		$overrides = [];
+		foreach ( array_keys( $props_schema ) as $override_key ) {
+			$overrides[ $override_key ] = '<PropValue per origin_prop_schema>';
+		}
```
Omerisra6: why?

modules/components/module.php
```diff
@@ -61,7 +61,7 @@ public function __construct() {
 		( new Components_REST_API() )->register_hooks();
 	}
 
-	public function is_experiment_active() {
```
Omerisra6, line 63: why?

modules/mcp/abilities/appliers/component-instance-applier.php
Omerisra6, line 1: we moved the appliers into a saparate folder

modules/mcp/abilities/appliers/component-instance-applier.php
```diff
@@ -0,0 +1,438 @@
...
+		}
+
+		return new \WP_Error(
+			'elementor_invalid_component_instance',
+			implode( ' ', $errors ),
+			[ 'status' => \WP_Http::BAD_REQUEST ]
+		);
+	}
+
+	private function build_envelope( string $config_id, array $shorthand, Document $document, array &$errors ): ?array {
```
Omerisra6, line 94: not sure if it fits, but maybe you could use the resolvers registry we added?

modules/mcp/abilities/build-composition/component-instance-applier.php
```diff
@@ -0,0 +1,438 @@
...
+				],
+				'overrides'    => [
+					'$$type' => 'overrides',
+					'value'  => $overrides_list,
+				],
+			],
+		];
+	}
+
+	private function resolve_component_id( array $shorthand, ?array $existing_envelope ): int {
```
Omerisra6: can we reuse that logic from somewhere?

modules/mcp/abilities/appliers/component-instance-applier.php
```diff
@@ -0,0 +1,438 @@
...
+					$override_key,
+					empty( $valid_keys ) ? '(none)' : implode( ', ', $valid_keys )
+				);
+			}
+		}
+
+		return $errors;
+	}
+
+	private function build_overrides_value( array $raw_overrides, array $overridable_props, int $component_id ): array {
```
Omerisra6, line 366: same for here (maybe we could move it to a resolver)

modules/mcp/abilities/build-composition/element-config-applier.php
```diff
@@ -63,20 +78,53 @@ public function apply( array &$config_id_index, array $element_config, array $wi
...
+		if ( ! empty( $component_entries ) ) {
+			$component_error = $this->apply_component_entries( $config_id_index, $component_entries, $document );
+
+			if ( $component_error ) {
+				return [
+					'error' => $component_error,
+					'warnings' => $warnings,
+				];
+			}
+		}
```
Omerisra6: I think we should return all errors at once so that the LLM gets what wrong on the first try

modules/mcp/abilities/list-component-schemas-ability.php
Omerisra6, line 1: Maybe we should merge both list components and list components schemas into the same tool, like the list widgets schema tool

modules/mcp/abilities/manage-elements-ability.php
```diff
@@ -87,7 +89,7 @@ protected function get_definition(): Ability_Definition {
 								'element_id' => [ 'type' => 'string' ],
 								'settings' => [
 									'type' => 'object',
-									'description' => 'update only: partial plain settings map merged onto existing settings.',
+									'description' => 'update only: partial plain settings map merged onto existing settings. For <e-component> targets, use { component_id?: int, overrides: { <override_key>: <plain value | null> } }; component_id is optional (inferred from the existing instance), each override_key is deep-merged onto existing overrides, and passing null for an override_key removes it. Unmentioned override_keys are preserved.',
```
Omerisra6: I dont think we should reference it here, are we referencing the widgets schema in the prompt?

modules/mcp/abilities/get-component-schema-ability.php
ronros-elementor: Gone along with the file. `get-component-schema` was replaced by the batch tool, and the synthetic `example_shorthand` went with it — the placement example now lives once in the `list-components` prompt instead of being generated per response.

modules/mcp/abilities/appliers/component-instance-applier.php
ronros-elementor: Moved. Merged latest `main`, so `Component_Instance_Applier` now sits in `modules/mcp/abilities/appliers/` under `Elementor\Modules\Mcp\Abilities\Appliers`, with its test relocated to `tests/phpunit/elementor/modules/mcp/abilities/appliers/`.

modules/mcp/abilities/appliers/component-instance-applier.php
ronros-elementor: I looked at moving this behind `elementor/atomic-widgets/settings-resolvers/register` and decided against it for this PR. Two things do not fit the resolver contract:

1. **A resolver only sees the incoming value, never the node's current state.** `apply_partial()` (the `manage-elements` `action=update` path) has to read the existing envelope to merge overrides per key, keep unmentioned keys, and delete on explicit `null`. That logic cannot move into a resolver, so the applier stays either way and we would end up with the shaping in one place and the merging in another.
2. **Resolvers signal failure by returning `null`.** `Element_Config_Applier` then emits `Property "component_instance" on "e-component" could not be resolved`. Today this applier returns `Component %d is archived and cannot be placed.` and `Unknown override key "%s". Valid keys: %s.` For an LLM-facing tool that difference decides whether the model self-corrects on retry or guesses again.

It would also force the input to nest under the prop key (`{ component_instance: { component_id, overrides } }`) instead of the flat shorthand the prompt documents.

I did take the reuse half of the point: no hardcoded `$$type` literals left here. Envelopes are built from `Component_Instance_Prop_Type::get_key()`, `Overrides_Prop_Type::get_key()`, `Override_Prop_Type::get_key()`, `Number_Prop_Type::get_key()` and `Component_Override_Parser::get_override_type()`, and `make_placeholder_element()` now goes through `assemble_envelope()`.

If you would rather accept the coarser error messages to get this into the registry, say so and I will open a follow-up.

modules/mcp/abilities/build-composition/component-instance-applier.php
ronros-elementor: There were four copies of that lookup: `Component_Instance::get_component_id()`, `Circular_Dependency_Validator::extract_component_id_from_settings()`, `Component_Styles::get_component_ids_from_post_cached()`, and this one.

Added `Component_Instance_Prop_Type::extract_component_id( array $settings )` on the prop type that owns the envelope shape, and routed all four through it.

modules/mcp/abilities/appliers/component-instance-applier.php
ronros-elementor: Same reasoning as the `build_envelope` thread above — the shaping half would fit a resolver, but `apply_partial()`'s per-key merge needs the node's existing envelope, and the `null`-return contract would collapse `Unknown override key "%s". Valid keys: %s.` into a generic "could not be resolved".

What I did change here: the `$$type` literals are gone in favour of `Override_Prop_Type::get_key()` / `Number_Prop_Type::get_key()` / `Component_Override_Parser::get_override_type()`, so the envelope shape is owned by the prop types rather than duplicated as strings.

modules/mcp/abilities/build-composition/element-config-applier.php
ronros-elementor: Fixed. The settings loop no longer returns before the component pass — both run, and `combine_errors()` merges them into a single response so a payload with a bad setting *and* a bad component instance reports both at once.

Covered by `test_apply__reports_settings_and_component_errors_together`.

modules/mcp/abilities/list-component-schemas-ability.php
ronros-elementor: Merged. There is one `elementor/list-components` now, two-step like `list-widget-schemas`:

- no arguments -> `id`, `name`, `uid`, `is_archived` only
- `component_ids` -> the same plus `overridable_props`

I kept the schemas opt-in via `component_ids` rather than adding a `summary` flag, so discovery on a site with hundreds of components does not pay to build every schema up front. `list-component-schemas` and its prompt are deleted, and its tests are folded into `test-list-components-ability.php`.

modules/mcp/abilities/manage-elements-ability.php
ronros-elementor: No, we do not reference the widget schema there — that was the inconsistency. Trimmed back to `update only: partial plain settings map merged onto existing settings.`

The component update semantics were already documented in the `list-components` prompt, so the inline paragraph was duplicating it.

Omerisra6 (dismissed):

Omerisra6 (approved):

PR #36672, opened 2026-07-27
Internal: Make Editor V4 the V4 umbrella experiment and remove three dead flags [ED-25066]

Ntnelbaba:
## PR Checklist

## PR Type

- [x] Other... Please describe: Internal — V4 experiment flag consolidation and dead flag removal.

## Summary

* Internal: Made "Editor V4" the umbrella experiment for V4 and merged the Variables, Variables Manager, Components and Interactions experiments into it
* Internal: Removed three V4 experiments that were registered but never checked

## Description

**Editor V4 is now the V4 umbrella.** `e_opt_in_v4` was titled "Editor V4" and written by the opt-in page, but nothing gated on it — turning it off left Variables, Components and Interactions running. The real umbrella was `e_atomic_elements`, which every V4 module ANDed against.

* Editor V4 is registered with the core default features instead of by the `atomic-opt-in` module. `Modules_Manager` registers and constructs modules in one interleaved loop, so registering at position 136 left the flag unavailable to `atomic-widgets` (128), `global-classes` (129) and `variables` (130). It now sits with `container`, the V3 umbrella.
* Editor V4 is no longer hidden, so it shows in the experiments UI.
* `e_atomic_elements` stays registered but hidden and is marked `@deprecated` on both the constant and `get_experimental_data()`, so existing callers, stored options and tests keep working.
* The two mirror each other via `on_state_change`, so turning either on or off moves the other. Writing an option that already holds the target value is a WordPress no-op that fires no hook, so the mirrors cannot recurse.
* `_v_4_3_0_align_v4_umbrella_experiment` adopts a stored Atomic Widgets value for sites that enabled V4 before Editor V4 existed.

**The V4 feature experiments are merged into it.** `e_variables`, `e_variables_manager`, `e_components` and `e_interactions` were separately gated while only ever being meaningful under V4 — each ANDed against the deprecated Atomic Widgets flag, and each had to be listed in the opt-in/opt-out sets to stay in step. They now declare Editor V4 as a `dependency` and are `mutable => false`, the same shape used to merge `nested-elements` into `container` in `d8d364b4a5`. `set_feature_default_state_to_match_dependencies()` forces a dependent inactive when its dependency is inactive, overriding a stored active state, so a merged feature cannot be switched on behind V4's back. The modules therefore no longer check Atomic Widgets themselves, and the editor-side consumers in `editor-editing-panel` and `editor-variables` read Editor V4 for the same reason.

This is also the real answer to the review question about aligning `e_variables_manager` and `e_interactions` in the upgrade routine: they no longer carry state of their own, so there is nothing left to align.

**Removed three flags with no reader anywhere in PHP, TS or build config**, so removal cannot change behaviour:

| Flag | Why dead |
| --- | --- |
| `editor_mcp` | `editor-mcp` loads unconditionally via `Editor_Loader::EXTENSIONS`, and is shared with the V3, kit and capabilities MCP features. Gating it would have disabled MCP on every V3 site. |
| `e_indications_popover` | Added in `ab3133ccfd` (ED-18488) as a registration-only commit; the consuming code never landed. |
| `atomic_widgets_should_enforce_capabilities` | Enforces nothing. The ticket lists it as a kill-switch to keep, but only the constant and `add_feature()` call reference it. Its live counterpart `global_classes_should_enforce_capabilities` is left alone. |

## Test instructions

1. Open **Elementor → Settings → Features**. "Editor V4" is listed. "Atomic Widgets" only appears with `ELEMENTOR_SHOW_HIDDEN_EXPERIMENTS`, and Variables, Variables Manager, Components and Interactions no longer appear at all.
2. Set **Editor V4** to `Inactive` and save. Confirm the Atomic Widgets option follows, and that Variables, Components and Interactions are off in the editor.
3. Set **Editor V4** back to `Active` and save. Confirm everything comes back on.
4. Set only **Atomic Widgets** to `Active` from a fully-off state and save. Confirm Editor V4 and all four merged features come on — this is the path Playwright's `setExperiments` uses.
5. Write `active` directly to `elementor_experiment-e_variables` while Editor V4 is off. Confirm Variables stays off.
6. Upgrade path: delete `elementor_experiment-e_opt_in_v4`, set `elementor_experiment-e_atomic_elements` to `active`, run `_v_4_3_0_align_v4_umbrella_experiment()`, confirm Editor V4 becomes `active`.

Verified on wp-lite-env (WordPress 6.9 / PHP 8.1).

Merged features follow the umbrella and cannot escape it — forcing all four options to `active` while V4 is off changes nothing:

```
Editor V4 ON                          Editor V4 OFF (+ options forced active)
  e_opt_in_v4         active=YES        e_opt_in_v4         active=no
  e_variables         active=YES        e_variables         active=no
  e_variables_manager active=YES        e_variables_manager active=no
  e_components        active=YES        e_components        active=no
  e_interactions      active=YES        e_interactions      active=no
```

Setting only Atomic Widgets active in the UI cascades through the mirror to V4 and on to all four merged features:

```
after setting ONLY Atomic Widgets active
  e_opt_in_v4 YES / e_atomic_elements YES
  e_variables YES / e_variables_manager YES / e_components YES / e_interactions YES
```

Editor V4 toggling, driven through the real admin UI:

[editor_v4_toggle_mirrors_atomic_widgets_experiment.mp4](https://cursor.com/agents/bc-236c4c4c-d887-4f8b-84de-93e2a791b0ce/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Feditor_v4_toggle_mirrors_atomic_widgets_experiment.mp4)

Experiments list after the merge — the four merged features are gone, and the three dead flags with them:

[Experiments page after merging the V4 features](https://cursor.com/agents/bc-236c4c4c-d887-4f8b-84de-93e2a791b0ce/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Fexperiments_after_v4_features_merged.webp)

Automated: `Test_Opt_In`, `Test_Upgrades` and the components/interactions/variables suites pass locally against WordPress 6.9 (200 tests, 595 assertions), plus the packages Jest run for the two changed editor packages (96 suites, 701 tests). The editor loads at HTTP 200 with zero PHP fatals and the `editor-variables`, `editor-components` and `editor-interactions` bundles present. `php -l`, `phpcs` and `eslint` clean on all changed files.

## Quality assurance

- [ ] Docs have been added / updated (for bug fixes / features) — the V4 experiments docs live on `feat/v4-docs-plan-revision-ee1d`, not on `main`

**Review feedback addressed**

* *Why 4.4?* — a wrong guess, now `_v_4_3_0_align_v4_umbrella_experiment`. `ELEMENTOR_VERSION` is 4.3.0 and 4.2.0 is the latest release. `get_upgrade_callbacks()` runs a callback whenever its version exceeds the version stored for the site, so `_v_4_4_0` would have re-run on every upgrade until a site reached 4.4.0.
* *Should the upgrade also align `e_variables_manager` and `e_interactions`?* — addressed by merging them instead. They now derive from Editor V4 and hold no state of their own, so the two flags are removed from `OPT_IN_FEATURES` / `OPT_OUT_FEATURES` and there is nothing for the upgrade to align.
* *Stale `test_init__registers_feature`* — `init()` no longer registers the experiment and the assertion only passed because `Experiments_Manager` registers it earlier. Replaced with `test_experiment_is_registered_with_the_default_features`.
* *Missing coverage for `mirror_state()` and the migration* — added, plus `test_merged_features_cannot_be_activated_while_editor_v4_is_off`. All were mutation-checked: dropping the `null` normalisation, and dropping a merged feature's `dependencies`, each fail their test. The unchanged-value test is deliberately not mutation-sensitive, since WordPress's `update_option` also skips unchanged writes and the explicit guard is defence-in-depth.

**Remaining reviewer notes**

* `Module::EXPERIMENT_EDITOR_MCP` and `Module::ENFORCE_CAPABILITIES_EXPERIMENT` were public constants and are removed. I could not check `elementor-pro` from this environment (GitHub code search returned 0 even for control queries), so please grep Pro for those before merge.
* `on_state_change` only fires when a feature is `mutable` and `is_admin()`, so the Editor V4 / Atomic Widgets mirror covers the admin UI and Playwright's `setExperiments`, but not raw WP-CLI toggles. The merged features are unaffected — their `dependencies` resolve in every context. CI blueprints set both umbrella flags explicitly.
* `e_classes` (Global Classes) is left as its own experiment. It was not in the ticket's merge list and, unlike the four merged features, is not gated behind Atomic Widgets in the same way. Happy to fold it in too if that is wanted.

Fixes [ED-25066](https://elementor.atlassian.net/browse/ED-25066)

<sub>To show artifacts inline, <a href="https://cursor.com/dashboard/cloud-agents#team-pull-requests">enable</a> in settings.</sub>

<div><a href="https://cursor.com/agents/bc-236c4c4c-d887-4f8b-84de-93e2a791b0ce"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-236c4c4c-d887-4f8b-84de-93e2a791b0ce"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

[ED-25066]: https://elementor.atlassian.net/browse/ED-25066?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ED-18488]: https://elementor.atlassian.net/browse/ED-18488?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## 1. Problem & Context

Editor V4 replaces Atomic Widgets as the V4 umbrella experiment (ED-25066). Sites using only Atomic Widgets never got Editor V4 options, risking silent downgrade. Consolidates V4 feature flags under single experiment with bidirectional mirroring to preserve existing option values.

## 2. What Changed (Where)

| File | Change |
|------|--------|
| `core/experiments/manager.php` | Register Editor V4 via `Opt_In::get_experimental_data()` at startup |
| `modules/atomic-widgets/module.php` | Deprecate constants; add state-change callback to mirror Editor V4; remove dead experiments (`e_indications_popover`, `enforce_capabilities`, `editor_mcp`) |
| `modules/atomic-widgets/opt-in/opt-in.php` | Move feature registration from `init()` to static `get_experimental_data()`; add `mirror_state()` to sync both experiments bidirectionally; expand OPT_IN/OUT lists with Variables Manager and Interactions |
| `core/upgrade/upgrades.php` | New migration `_v_4_3_0_align_v4_umbrella_experiment()` adopts stored Atomic Widgets state if Editor V4 unset |
| Tests | Add 7 new tests covering migration, mirroring logic, and state alignment |

## 3. How It Works

Editor V4 registers as default feature early (before modules construct), making it available as umbrella gate. State changes on either Editor V4 or Atomic Widgets trigger `on_state_change` callbacks that sync the other via `mirror_state()`. WordPress skips hooks when writing identical values, preventing recursion. Upgrade migration copies Atomic state → Editor V4 only if Editor V4 never set, preserving any explicit user choice.

## 4. Risks

**Bidirectional mirroring complexity**: Depends entirely on WordPress's `get_option() === update_value` skipping hooks. If that contract breaks, infinite loops. Mitigated by tests verifying no-op behavior (test_mirror_state__does_not_write_when_the_state_already_matches). **Migration idempotency**: If upgrade runs twice, no issue (returns early if Editor V4 already set), but fragile if logic changes. **Deprecated constant lingering**: `Atomic_Widgets_Module::EXPERIMENT_NAME` kept for BC but still referenced by stored options—unclear cleanup path.

_Generated by L

### DIFF
--- a/core/upgrade/upgrades.php
+++ b/core/upgrade/upgrades.php
@@ -876,7 +876,7 @@ class Upgrades {
 	 * stored value to keep the two switches aligned and avoid silently dropping those sites out
 	 * of V4.
 	 */
-	public static function _v_4_4_0_align_v4_umbrella_experiment() {
+	public static function _v_4_3_0_align_v4_umbrella_experiment() {
 		$opt_in_key = 'elementor_experiment-' . Opt_In::EXPERIMENT_NAME;
 
 		if ( false !== get_option( $opt_in_key, false ) ) {
--- a/core/upgrade/upgrades.php
+++ b/core/upgrade/upgrades.php
@@ -8,6 +8,8 @@ use Elementor\Core\Settings\Manager as SettingsManager;
 use Elementor\Core\Settings\Page\Manager as SettingsPageManager;
 use Elementor\Icons_Manager;
 use Elementor\Includes\Elements\Container;
+use Elementor\Modules\AtomicWidgets\Module as AtomicWidgetsModule;
+use Elementor\Modules\AtomicWidgets\OptIn\Opt_In;
 use Elementor\Modules\Usage\Module;
 use Elementor\Plugin;
 use Elementor\Tracker;
@@ -866,6 +868,28 @@ class Upgrades {
 		delete_option( 'elementor_experiment-page-transitions' );
 		delete_option( 'elementor_experiment-search' );
 		delete_option( 'elementor_experiment-taxonomy-filter' );
+	}
+
+	/**
+	 * Editor V4 replaced Atomic Widgets as the V4 umbrella experiment. Sites that turned V4 on
+	 * through the Atomic Widgets experiment alone never got an Editor V4 option, so adopt the
+	 * stored value to keep the two switches aligned and avoid silently dropping those sites out
+	 * of V4.
+	 */
+	public static function _v_4_4_0_align_v4_umbrella_experiment() {
+		$opt_in_key = 'elementor_experiment-' . Opt_In::EXPERIMENT_NAME;
+
+		if ( false !== get_option( $opt_in_key, false ) ) {
+			return;
+		}
+
+		$atomic_state = get_option( 'elementor_experiment-' . AtomicWidgetsModule::EXPERIMENT_NAME, false );
+
+		if ( false === $atomic_state ) {
+			return;
+		}
+
+		update_option( $opt_in_key, $atomic_state );
 	}
 
 	private static function maybe_add_gap_control_data( $option_name ) {
--- a/core/experiments/manager.php
+++ b/core/experiments/manager.php
@@ -5,6 +5,7 @@ use Elementor\Core\Base\Base_Object;
 use Elementor\Core\Experiments\Exceptions\Dependency_Exception;
 use Elementor\Core\Upgrade\Manager as Upgrade_Manager;
 use Elementor\Core\Utils\Collection;
+use Elementor\Modules\AtomicWidgets\OptIn\Opt_In;
 use Elementor\Modules\System_Info\Module as System_Info;
 use Elementor\Plugin;
 use Elementor\Settings;
@@ -309,6 +310,8 @@ class Manager extends Base_Object {
 	}
 
 	private function add_default_features() {
+		$this->add_feature( Opt_In::get_experimental_data() );
+
 		$this->add_feature( [
 			'name' => 'e_font_icon_svg',
 			'title' => esc_html__( 'Inline Font Icons', 'elementor' ),
--- a/modules/atomic-widgets/module.php
+++ b/modules/atomic-widgets/module.php
@@ -156,7 +156,6 @@ if ( ! defined( 'ABSPATH' ) ) {
 class Module extends BaseModule {
 	const EXPERIMENT_NAME = 'e_atomic_elements';
 	const ENFORCE_CAPABILITIES_EXPERIMENT = 'atomic_widgets_should_enforce_capabilities';
-	const EXPERIMENT_EDITOR_MCP = 'editor_mcp';
 
 	const PACKAGES = [
 		'editor-canvas',
@@ -240,15 +239,6 @@ class Module extends BaseModule {
 			'default' => Experiments_Manager::STATE_ACTIVE,
 			'release_status' => Experiments_Manager::RELEASE_STATUS_DEV,
 		] );
-
-		Plugin::$instance->experiments->add_feature([
-			'name' => self::EXPERIMENT_EDITOR_MCP,
-			'title' => esc_html__( 'Editor MCP for atomic widgets', 'elementor' ),
-			'description' => esc_html__( 'Editor MCP for atomic widgets.', 'elementor' ),
-			'hidden' => true,
-			'default' => Experiments_Manager::STATE_ACTIVE,
-			'release_status' => Experiments_Manager::RELEASE_STATUS_DEV,
-		]);
 
 		Plugin::$instance->experiments->add_feature([
 			'name' => Migrations_Orchestrator::EXPERIMENT_BC_MIGRATIONS,
--- a/modules/atomic-widgets/module.php
+++ b/modules/atomic-widgets/module.php
@@ -155,7 +155,6 @@ if ( ! defined( 'ABSPATH' ) ) {
 
 class Module extends BaseModule {
 	const EXPERIMENT_NAME = 'e_atomic_elements';
-	const ENFORCE_CAPABILITIES_EXPERIMENT = 'atomic_widgets_should_enforce_capabilities';
 
 	const PACKAGES = [
 		'editor-canvas',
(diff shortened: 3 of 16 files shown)


core/upgrade/upgrades.php
```diff
@@ -868,6 +870,28 @@ public static function _v_3_26_0() {
 		delete_option( 'elementor_experiment-taxonomy-filter' );
 	}
 
+	/**
+	 * Editor V4 replaced Atomic Widgets as the V4 umbrella experiment. Sites that turned V4 on
+	 * through the Atomic Widgets experiment alone never got an Editor V4 option, so adopt the
+	 * stored value to keep the two switches aligned and avoid silently dropping those sites out
+	 * of V4.
+	 */
+	public static function _v_4_4_0_align_v4_umbrella_experiment() {
```
IshayMaya: why 4.4?

core/upgrade/upgrades.php
```diff
@@ -868,6 +870,28 @@ public static function _v_3_26_0() {
...
+			return;
+		}
+
+		$atomic_state = get_option( 'elementor_experiment-' . AtomicWidgetsModule::EXPERIMENT_NAME, false );
+
+		if ( false === $atomic_state ) {
+			return;
+		}
+
+		update_option( $opt_in_key, $atomic_state );
```
IshayMaya, line 892: should this upgrade also align e_variables_manager and e_interactions (the two added to OPT_IN_FEATURES) ?

PR #36677, opened 2026-07-27
Internal: MCP V3 parity — flag V3 in get-page-structure, block V3 writes [ED-25072]

Omerisra6:
## PR Type
- [x] Feature

## Summary

Make the Elementor MCP PHP abilities V3-aware so LLMs can see V3 elements on a mixed V3/V4 page and know they can only be edited in the Elementor editor — while V4 write tools cleanly refuse V3 targets instead of silently corrupting them.

Ref: [ED-25072](https://elementor.atlassian.net/browse/ED-25072)

## Description

Before this change, `elementor/get-page-structure` returned V3 and V4 elements mixed with no marker; `elementor/manage-elements` silently accepted V3 element ids and injected atomic `_styles` classes into V3 nodes (broken render); `elementor/get-widget-schema` returned a misleading fallback envelope for V3 widgets that no write tool in this MCP can consume.

Changes:

- `get-page-structure`: each node is tagged with `version: "v3" | "v4"` (omitted when the element type can't be resolved). When `include_content=true`, V3 nodes are still returned in the skeleton but their `settings` and `styles` are empty objects — V3 shapes must not round-trip through V4 write tools.
- `get-widget-schema`: V3 widget types are rejected with `elementor_v3_not_supported` (HTTP 400, includes `widget_type` and `version: "v3"` in error data).
- `manage-elements`: every op (update / delete / move / duplicate) validates the target element; V3 element ids and V3 `new_parent_id` values return `elementor_v3_not_supported` per-op. Sibling valid V4 ops in the same batch still apply.
- `build-composition`: unchanged — V3 `parent_id` remains accepted (V3 containers can host V4 children).
- Error messages are neutral (no reference to any specific MCP server); they only tell the LLM to edit V3 elements directly in the Elementor editor.

Audit context: canvas MCP (`configure-element`, `document-structure`) and the elementor-v3-mcp package are already V3-aware; this PR closes the same gap on the PHP surface exposed via `elementor/v1/mcp-proxy`.

## Test instructions

- Run the MCP PHPUnit suites:
  - `vendor/bin/phpunit --filter \"Test_Get_Structure_Ability|Test_Get_Widget_Schema_Ability|Test_Manage_Elements_Ability|Test_Build_Composition_Ability\"`
- Manual smoke on a mixed V3/V4 page:
  - `elementor/get-page-structure` with `include_content=true` on a V3 subtree root → V3 nodes carry `version: \"v3\"` with empty `settings`/`styles`; V4 descendants keep their content.
  - `elementor/get-widget-schema` with a V3 `widget_type` → `elementor_v3_not_supported`, HTTP 400.
  - `elementor/manage-elements` with a V3 `element_id` on update/delete/move/duplicate → per-op `elementor_v3_not_supported`; a mixed batch with one V3 op and one V4 op → V3 op fails, V4 op still applies and the document saves.

## Quality assurance

Fixes ED-25072

[ED-25072]: https://elementor.atlassian.net/browse/ED-25072?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## 1. Problem & Context

MCP V3 parity requires distinguishing legacy V3 elements from atomic V4 elements in API responses, tagging each with version number, and blocking V3 modifications through manage-elements (ED-25072). Clients must edit V3 elements directly in the Elementor editor, V4 elements through MCP.

## 2. What Changed (Where)

| File | Change |
|------|--------|
| `get-structure-ability.php` | Adds `resolve_element_version()` to detect V3 vs V4 via `Atomic_Elements_Utils`; tags each element with `version` field; strips settings/styles for V3 nodes when `include_content=true` |
| `manage-elements-ability.php` | Imports `Atomic_Elements_Utils`; adds `reject_v3_target()` check in `apply_operation()` to return `elementor_v3_not_supported` error per-operation |
| `get-widget-schema-ability.php` | Rejects V3 widget types upfront with `elementor_v3_not_supported` error |
| Tests | 8 new tests validating V3 rejection, version tagging, content stripping, per-operation errors, and mixed V3/V4 bulk operations |

## 3. How It Works

Entry point: Client calls `get-page-structure` → each element is tagged with `version: 3` (legacy) or `version: 4` (atomic) by inspecting widget instance against `Atomic_Elements_Utils::is_atomic_element()`. For V3 nodes with `include_content=true`, settings/styles returned empty to prevent client-side assumptions. On `manage-elements` or `get-widget-schema`, V3 targets fail immediately with structured error (code, status, version) before any mutation; V4 operations succeed; bulk batches report per-operation status allowing partial success.

## 4. Risks

Moving V4 elements into V3 parents is allowed (test confirms), avoiding hard architectural constraints. Version detection relies on widget type lookup—unknown types gracefully omit version field rather than error. Per-operation V3 rejection means clients must parse error arrays; mitigated by clear error code (`elementor_v3_not_supported`) and documentation updates in ability descriptions.


modules/mcp/abilities/manage-elements-ability.php
```diff
@@ -214,23 +212,33 @@ private function handle_bulk( Document $document, array $operations ): array {
...
+
+		if ( 'move' === $action ) {
+			$new_parent_id = $operation['new_parent_id'] ?? '';
+			if ( is_string( $new_parent_id ) && '' !== $new_parent_id && 'document' !== $new_parent_id ) {
+				$v3_parent_error = $this->reject_v3_target( $tree, $new_parent_id );
+				if ( $v3_parent_error ) {
+					return $v3_parent_error;
+				}
+			}
+		}
```
Omerisra6: allow moving to v3 parent, only modifying v3 elements should be restricted

modules/mcp/abilities/get-structure-ability.php
```diff
@@ -150,6 +161,22 @@ private function prune_elements( array $elements, bool $include_content = false
...
+			return null;
+		}
+
+		$instance = Atomic_Elements_Utils::get_element_instance( (string) $type );
+
+		if ( ! $instance ) {
+			return null;
+		}
+
+		return Atomic_Elements_Utils::is_atomic_element( $instance ) ? 'v4' : 'v3';
```
ronros-elementor: wouldn't it be better to return a numeric value?
it's called `resolve_element_version`, if it returns 4 or 3, it's clear enough that's the version, and we could easily use numeric comparisons like < 4

ronros-elementor (approved):

PR #36678, opened 2026-07-27
Internal: Expose MCP tool/resource lists via WP filters (ED-25074)

IshayMaya:
## Summary
Extract the hardcoded MCP tool and resource ability slug arrays from `Module::register_server()` into two private methods (`get_server_tools()`, `get_server_resources()`) and wrap each in a documented `apply_filters()` call. Downstream modules (e.g. Elementor Pro's Theme Builder) attach their MCP abilities to the `elementor-mcp-server` by hooking these filters — no coupling to internal Core classes.

## Filters
- `elementor/mcp/server/tools` — string[] of tool ability slugs.
- `elementor/mcp/server/resources` — string[] of resource ability slugs.

Slugs must match the id returned by the ability's `get_ability_id()`; the ability itself is registered via `wp_register_ability()` on `wp_abilities_api_init`, per WP Abilities API convention.

## Behavior
- Zero hooks attached: byte-identical output to the previous hardcoded arrays (same slugs, same order).
- Filter defensively normalizes the result (`array_filter( is_string )` + `array_unique` + `array_values`); non-array returns fall back to the defaults.
- `apply_filters()` runs synchronously inside `register_server()` on the `mcp_adapter_init` action — no ordering dependency on when Core's `register_abilities()` fires.

## Design note
Originally scoped as an internal `Server_Registry` class (see closed exploration in this branch's history). Team decision: prefer the simpler, WordPress-native filter pattern over a bespoke internal registry — smaller surface area, no hook-order pitfalls, standard extensibility for other modules.

## Test plan
- [x] `php -l modules/mcp/module.php`
- [x] `vendor/bin/phpcs --standard=./ruleset.xml modules/mcp/module.php`
- Manual: WP Playground admin — MCP server responds with the same tool/resource list as before.

## Ripple
Downstream tickets ED-25075 – ED-25078 (Pro theme-builder MCP submodule) now register their slugs with `add_filter( 'elementor/mcp/server/tools', ... )` instead of the previously-planned `Server_Registry::add_tool()`. The plan doc has been updated locally to reflect this.

## 1. Problem & Context

Decouples hardcoded MCP tool/resource lists from module initialization by extracting them into filterable methods, enabling other modules to extend capabilities without modifying core code (ED-25074).

## 2. What Changed (Where)

- `modules/mcp/module.php`: Refactored inline tool/resource arrays into `get_server_tools()` and `get_server_resources()` methods; added WP filter hooks for extensibility and `normalize_slugs()` utility for deduplication.

## 3. How It Works

Server registration now calls `get_server_tools()` and `get_server_resources()` instead of inline arrays. Each method defines core defaults, applies a filter (`elementor/mcp/server/tools` or `elementor/mcp/server/resources`), then deduplicates and merges via `normalize_slugs()`. Type-safety is enforced by filtering non-string values from filter results.

## 4. Risks

Low risk. Filter defaults to empty array with documented contract; core abilities always included. Minor: if external code passes non-string values, they're silently filtered (acceptable given defensive implementation).


modules/mcp/module.php
```diff
@@ -118,4 +100,60 @@ public function register_server( $adapter ) {
...
+		 *
+		 * Use this filter to add tool abilities (registered via `wp_register_ability` on the
+		 * `wp_abilities_api_init` hook) to the `elementor-mcp-server`. Slugs must match the
+		 * ability id returned by the ability's `get_ability_id()`.
+		 *
+		 * @since 4.3.0
+		 *
+		 * @param string[] $tools List of tool ability slugs.
+		 */
+		$filtered = apply_filters( 'elementor/mcp/server/tools', $tools );
```
Omerisra6:
```suggestion
		$filtered = apply_filters( 'elementor/mcp/server/tools', [] );
		
		$tools = array_merge( $tools, $filtered );
```

so that you ensure your tools are present (same for resources)

IshayMaya: Agreed — more defensive. Applied in fe49c88274: filter now receives `[]` and results are merged into (and can't remove) Core's default slugs. Extracted the normalization into a shared `normalize_slugs()` helper so tools and resources stay consistent. Same treatment applied to the resources filter.

Omerisra6 (dismissed):

annab1 (approved):

PR #36681, opened 2026-07-27
Fix: Angie tooltip anchor span affecting top bar layout [ED-24890]

Ntnelbaba:
## Summary
- Follow-up to #36680 (Editor top bar responsiveness).
- The `Infotip` anchor `<span />` in `AngieGuideLocation` was rendering with the `MuiInfotip-root` class even when closed, contributing an empty box to the top bar's flex layout.
- Kept the span in place (still needed as the tooltip anchor/ref) but added `display: contents` inline so it produces no box while it has no children, removing its layout footprint.

## Test plan
- [ ] Open the editor top bar and verify no extra empty gap/element appears near the tools menu.
- [ ] Confirm the Angie guide tooltip still opens/closes correctly via its toggle event.

## 1. Problem & Context

The Angie tooltip anchor span was causing unwanted layout shifts in the top bar. Reordering toolbar menu items and wrapping the anchor with `display: contents` eliminates the layout impact while preserving functionality.

## 2. What Changed (Where)

- **utilities-menu-location.tsx**: Moved toolbar items rendering after popover items to control render order
- **angie-guide-location.tsx**: Changed anchor span from empty to `display: contents` to prevent box model contribution
- **menus.test.tsx**: Added `popoverPosition` parameter to handle different menu positions (left/right) in assertions

## 3. How It Works

Utilities menu now renders popover first (positioned left), then toolbar items. The Angie tooltip anchor uses `display: contents` to participate in layout calculations without creating a box layer—preventing the span from affecting app bar metrics. Tests verify popover button position dynamically based on menu alignment.

## 4. Risks

Minimal. `display: contents` has solid browser support and the render order change only affects visual positioning—no behavioral impact. Test parametrization ensures both menu locations are validated.


Ntnelbaba:
Confirmed intentional — added a follow-up commit (536f05d5de) that reorders tools-menu-location.tsx to render `More` first as well, matching utilities-menu-location.tsx, and updated menus.test.tsx expectations. Both menus are now consistent.

ronenboxer (approved):

PR #36695, opened 2026-07-28
Fix: Atomic media widgets fail to save when URL is not strict-valid [ED-25101]

Omerisra6:
## Summary

Fixes a regression where saving an atomic image (or svg/video) with a user-typed URL like `https://google.c` fails with:

```
Settings validation failed. image: invalid_value
```

### Root cause

On save, `Has_Atomic_Base::parse_atomic_settings()` validates the `image` prop through `Image_Src_Prop_Type` -> `Url_Prop_Type::validate_value()`, which calls `wp_http_validate_url()`. That WP helper is stricter than users expect (rejects `https://google.c`, relative paths, fragments, etc.).

`Link_Prop_Type` already sidesteps this same problem by calling `->skip_validation()` on its inner `Url_Prop_Type`. Media source prop types did not — they should, since users author them the same way.

### Change

Add `->skip_validation()` to `Url_Prop_Type::make()` in:
- `Image_Src_Prop_Type`
- `Svg_Src_Prop_Type`
- `Video_Src_Prop_Type`

Sanitization (`esc_url_raw`) still runs on save, so security posture is unchanged. Transformable-shape validation (`{\$\$type: 'url', value: string}`) still runs via `Plain_Prop_Type::validate`.

### Existing tests updated

`test-image-prop-type.php`:
- Removed `test_validate__fail_when_passing_src_with_invalid_url` and `test_validate__fail_when_passing_src_with_non_string_url` (they asserted behavior we intentionally no longer guarantee, matching `Link_Prop_Type`).
- Added data-provider-backed `test_validate__passes_for_non_strict_url` covering `https://google.c`, `invalid-url`, `/relative/path`, `#anchor`.

## Test plan

- [x] Manual smoke test via PHP CLI: `Image_Prop_Type::validate()` now returns true for `https://google.c` (previously false).
- [ ] In editor: add atomic image, "Insert from URL", type `https://google.c`, save -> no validation error.
- [ ] In editor: same flow with a valid URL (`https://google.com`) -> still saves.
- [ ] In editor: save an image with an attachment id (no URL) -> unchanged behavior.
- [ ] CI PHPUnit passes on updated `test-image-prop-type.php`.

## Related

- Reported alongside ED-25101 (which also mentions Add-to-Cart dynamic tag as link destination — that's a **separate** bug in `Link_Prop_Type` / dynamic tag validation, not addressed here).

## 1. Problem & Context

`wp_http_validate_url()` rejects uncommon TLDs like `.c`, blocking valid URLs and preventing widget saves (ED-25101). Switching to `filter_var()` with `FILTER_VALIDATE_URL` uses PHP's built-in validation, which is more permissive and aligns with the downstream `esc_url_raw()` sanitization.

## 2. What Changed (Where)

- `url-prop-type.php`: Replaced `wp_http_validate_url()` with `filter_var( $value, FILTER_VALIDATE_URL )`
- `test-image-prop-type.php`: Added test case validating uncommon TLD acceptance (`https://google.c`)

## 3. How It Works

Validation now delegates to PHP's native URL filter instead of WordPress's stricter curated TLD list. The `false !==` check ensures consistent boolean return. Sanitization via `esc_url_raw()` remains unchanged, maintaining output safety.

## 4. Risks

`filter_var()` is more permissive—malformed URLs like `ht!tp://example.com` may pass validation where `wp_http_validate_url()` would reject them. Mitigation: downstream `esc_url_raw()` still sanitizes output, and this aligns with user intent (allowing `.c` domains suggests accepting broader valid RFC 3986 URLs).

### DIFF
--- a/modules/atomic-widgets/module.php
+++ b/modules/atomic-widgets/module.php
@@ -16,6 +16,7 @@ use Elementor\Modules\AtomicWidgets\PlainResolvers\Resolvers\Dynamic_Plain_Resol
 use Elementor\Modules\AtomicWidgets\PlainResolvers\Resolvers\Html_V3_Plain_Resolver;
 use Elementor\Modules\AtomicWidgets\PlainResolvers\Resolvers\Passthrough_Plain_Resolver;
 use Elementor\Modules\AtomicWidgets\PlainResolvers\Resolvers\Number_Plain_Resolver;
+use Elementor\Modules\AtomicWidgets\PlainResolvers\Resolvers\Size_Plain_Resolver;
 use Elementor\Modules\AtomicWidgets\PlainResolvers\Resolvers\String_Plain_Resolver;
 use Elementor\Modules\AtomicWidgets\Elements\Atomic_Youtube\Atomic_Youtube;
 use Elementor\Modules\AtomicWidgets\Elements\Div_Block\Div_Block;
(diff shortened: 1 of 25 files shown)


Omerisra6:
Reworked after CI failure — previous approach (adding `->skip_validation()` on media src prop types) broke the MCP `Style_Applier` "URLs must be absolute" contract, causing `Test_Manage_Elements_Ability::test_update__unhandled_style_declaration_warns_with_declaration_and_url_hint` to fail.

New approach (minimal 2-line change): swap `wp_http_validate_url` for `filter_var($value, FILTER_VALIDATE_URL)` inside `Url_Prop_Type::validate_value()`. Verified against these cases:

| Input | Old (`wp_http_validate_url`) | New (`filter_var`) |
|---|---|---|
| `https://google.c` (user bug) | INVALID | **VALID** |
| `https://google.com` | VALID | VALID |
| `http://localhost:8888` | INVALID | VALID |
| `/wp-content/uploads/x.png` (MCP contract) | INVALID | INVALID (preserved) |
| `invalid-url` | INVALID | INVALID |
| `#anchor` | INVALID | INVALID |
| `javascript:alert(1)` | (blocked in scheme) | INVALID |

Ntnelbaba (approved):

PR #36711, opened 2026-07-29
Feature: Support breakpoints and pseudo-classes in mcp [ED-24912]

eavichay-elementor:
## 1. Problem & Context
Implement support for responsive breakpoints and pseudo-classes (hover, focus, active) in MCP global class management [ED-24912]. Transitions the API from simple property-value maps to raw CSS strings to allow complex selectors and media queries.

## 2. What Changed (Where)

| File | Change |
| :--- | :--- |
| `css-block-scanner-trait.php` | New trait for brace matching and escape character handling in CSS. |
| `css-converter.php` | Added `parse_nested` to extract `&` pseudo-class blocks. |
| `css-media-splitter.php` | New class to split CSS into breakpoint-specific blocks. |
| `manage-classes-ability.php` | Updated API to handle CSS strings, bulk operations, and update modes. |
| `style-variants-merger.php` | New logic to merge CSS variants and handle `null` property resets. |
| `mcp-proxy-rest-api.php` | Added HTTP status resolution for partial errors (207). |
| `manage-classes-tool.ts` | Updated tool schema to support raw CSS strings and `patch`/`replace` modes. |
| `tests/...` | Comprehensive unit tests for parsing, splitting, and merging logic. |

## 3. How It Works
`Manage_Classes_Ability` accepts CSS strings $\rightarrow$ `Css_Media_Splitter` isolates breakpoint blocks $\rightarrow$ `Css_Converter` extracts nested pseudo-classes $\rightarrow$ `Style_Variants_Merger` blends results with existing data based on the selected mode (`patch` or `replace`).

## 4. Risks
*   **Regex Complexity**: `NULL_DECLARATION_PATTERN` could misidentify `null` inside complex CSS values; mitigated by boundary checks and string-awareness in the scanner trait.
*   **Performance**: Bulk processing of raw CSS strings increases parsing overhead compared to maps; mitigated by limiting batch size to 50.

[ED-24912]: https://elementor.atlassian.net/browse/ED-24912?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/mcp/abilities/manage-classes-ability.php
```diff
@@ -31,6 +31,7 @@ class Manage_Classes_Ability extends Abstract_Ability {
 	const CLASS_TYPE = 'class';
```
eavichay-elementor, line 31: File-level: file is getting big - consider extraction of helpers

modules/mcp/abilities/manage-classes-ability.php
```diff
@@ -88,7 +89,21 @@ protected function get_definition(): Ability_Definition {
 								'label' => [ 'type' => 'string' ],
 								'css' => [
 									'type' => 'object',
-									'description' => 'Raw CSS declarations (property → value).',
+									'description' => 'Legacy: flat property → value map. Ignored when styles is present.',
```
eavichay-elementor: Do we need 'legacy' in the description?

modules/mcp/abilities/manage-classes-ability.php
```diff
@@ -258,17 +306,96 @@ private function translate_update( int $index, array $operation, array &$intents
...
 		}
 	}
 
+	protected function get_active_breakpoint_keys(): array {
+		return Plugin::$instance->breakpoints->get_active_devices_list();
+	}
+
+	private function parse_styles_field( int $index, string $action, array $operation, Bulk_Operations_Result $results ): ?array {
+		$raw_styles          = $operation['styles'];
+		$styles_tuples       = [];
```
eavichay-elementor: naming convention: 'tuples' describe the shape, not the reason

modules/mcp/abilities/manage-classes-ability.php
```diff
@@ -318,19 +456,146 @@ private function build_touched_items( array $intents, array $all_labels, Bulk_Op
...
+
 			$touched_items[ $intent['id'] ] = $class_item;
 		}
 
 		return $touched_items;
 	}
 
+	private function extract_null_props( string $css ): array {
+		$null_props = [];
+		preg_replace_callback(
```
eavichay-elementor: check for duplication of regexp

tests/phpunit/elementor/modules/atomic-widgets/css-converter/test-css-converter.php
```diff
@@ -181,6 +181,123 @@ public function test_convert__blocked_declarations_are_dropped_entirely() {
 		$this->assertSame( '', $result['customCss'] );
```
eavichay-elementor, line 181: File-level: test file getting big - extract parsed to separate file

tests/phpunit/elementor/modules/mcp/test-manage-classes-ability.php
```diff
@@ -2,6 +2,7 @@
```
eavichay-elementor, line 2: File-level: test file super big - break into several files - shared assertions as utility/base class?

tests/phpunit/unit-bootstrap.php
Omerisra6, line 1: Remove that I think it's only required locally

modules/mcp/abilities/utils/style-variants-merger.php
```diff
@@ -0,0 +1,206 @@
...
+
+use Elementor\Modules\AtomicWidgets\CssConverter\Css_Converter;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+class Style_Variants_Merger {
+
+	const PSEUDO_STATES = [ 'hover', 'focus', 'active' ];
```
Omerisra6, line 13: we have extra states for tabs, check if we can support element specific sates (if its not easy win ignore this comment)

modules/atomic-widgets/css-converter/css-converter.php
Omerisra6, line 1: lets support one shape to keep things simple, I would go with plain css and add support for media querys (if you want we can go with the style object)

modules/mcp/abilities/manage-classes-ability.php
```diff
@@ -88,7 +89,21 @@ protected function get_definition(): Ability_Definition {
...
+									'type' => 'object',
+									'description' => 'Map of breakpoint key to CSS string. Use "default" for desktop. Supports &:hover/&:focus/&:active nesting. Takes precedence over css. In patch mode: "prop: null" removes that prop; "all: null" wipes the variant (or deletes it when nothing remains); works per-block, e.g. "&:hover { all: null; }" deletes only hover.',
+									'additionalProperties' => [
+										'type' => [ 'string', 'null' ],
+										'description' => 'CSS for the breakpoint. null or empty string removes all variants for that breakpoint.',
+									],
+								],
+								'mode' => [
+									'type' => 'string',
+									'enum' => [ 'patch', 'replace' ],
```
Omerisra6, line 95: Let's keep the original behavior; we can keep the support for the ways you've added to delete props.

But I dont think an extra mode is needed.

eavichay-elementor: super required - saves multiple requests when needs to fast change css - keep this - it works - claude says it's functional, cursor agrees. really remove this?

modules/mcp/abilities/utils/style-variants-merger.php
eavichay-elementor: not an easy win

Omerisra6 (commented):
Review notes — see inline comments.

modules/atomic-widgets/css-converter/css-media-splitter.php
```diff
@@ -0,0 +1,188 @@
...
+						'error'       => 'Nested breakpoints are not allowed.',
+					];
+				}
+
+				$target                  = $result['breakpoint'];
+				$breakpoints[ $target ] = isset( $breakpoints[ $target ] )
+					? $breakpoints[ $target ] . ' ' . $block_content
+					: $block_content;
+			} else {
+				$custom_blocks[] = '@media' . $raw_selector . ' { ' . $block_content . ' }';
```
Omerisra6: `$raw_selector` is trimmed, so `@media print { ... }` gets stored as `@mediaprint { ... }`. Needs `'@media ' . $raw_selector`.

modules/atomic-widgets/css-converter/css-media-splitter.php
```diff
@@ -0,0 +1,188 @@
...
+			if ( '{' === $char ) {
+				++$depth;
+				++$i;
+				continue;
+			}
+
+			if ( '}' === $char ) {
+				--$depth;
+				if ( $depth < 0 ) {
+					return [
```
Omerisra6: Same error shape returned four times in `split()` — extract a private `error()` helper. Nit: `6` is hardcoded three times as `strlen('@media')`.

modules/atomic-widgets/css-converter/css-converter.php
```diff
@@ -34,6 +37,110 @@ public function __construct(
...
+				return [ 'error' => 'Unclosed brace detected in CSS input.' ];
+			}
+
+			if ( '&' !== $char ) {
+				++$i;
+				continue;
+			}
+
+			$j = $i + 1;
+			while ( $j < $len && '{' !== $css[ $j ] ) {
```
Omerisra6: Scans past `;` looking for `{`, so an `&` in a value corrupts the parse silently. `background: url(a&b.png); &:hover { color: red; }` splits mid-value with no error. Bail on `;` and `}`.

modules/atomic-widgets/css-converter/css-converter.php
```diff
@@ -34,6 +37,110 @@ public function __construct(
...
+
+			if ( '"' === $char || "'" === $char ) {
+				$in_string   = true;
+				$string_char = $char;
+				++$i;
+				continue;
+			}
+
+			if ( '{' === $char || '}' === $char ) {
+				return [ 'error' => 'Unclosed brace detected in CSS input.' ];
```
Omerisra6, line 98: Error says "unclosed brace" even when braces are balanced — any non-`&` nested selector hits this (e.g. `color: blue; span { color: green; }`). Message should say nested selectors other than `&:hover/&:focus/&:active` aren't supported.

modules/atomic-widgets/css-converter/css-block-scanner-trait.php
```diff
@@ -0,0 +1,44 @@
...
+
+namespace Elementor\Modules\AtomicWidgets\CssConverter;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+trait Css_Block_Scanner_Trait {
+
+	private function find_block_end( string $css, int $start, int $len ): ?int {
```
Omerisra6, line 26: The quote/escape scanner is duplicated in three places (here, `Css_Media_Splitter::split`, `scan_ampersand_blocks`). Move it fully into this trait as `skip_string()`. While at it: the single-char lookback treats `content: "a\\"` as an unterminated string — needs to count consecutive backslashes.

modules/mcp/abilities/utils/style-variants-merger.php
```diff
@@ -0,0 +1,206 @@
...
+
+	const NULL_DECLARATION_PATTERN = '/([a-zA-Z][a-zA-Z0-9-]*)\s*:\s*null\s*;?/';
```
Omerisra6: Regex isn't string/paren aware, unlike the careful quote tracking elsewhere in the PR. `content: "color: null;"` falsely captures `color` and strips text out of the string; `--my-var: null` captures `my-var` and leaves a stray `--`. Reuse the block scanner instead.

modules/mcp/abilities/utils/style-variants-merger.php
```diff
@@ -0,0 +1,206 @@
...
+				$css      = $block['css'];
+
+				if ( null === $selector ) {
+					$base_block_css = $css;
+					continue;
+				}
+
+				$state = ltrim( $selector, ':' );
+
+				if ( in_array( $state, self::PSEUDO_STATES, true ) ) {
```
Omerisra6: Two bugs on lines 35–37: `ltrim(':')` turns `& :hover` (a descendant selector in CSS) into the element's own hover state, and this `in_array` is case-sensitive so `&:HOVER` silently degrades to `custom_css`. Lowercase the state, and reject whitespace between `&` and the pseudo.

modules/mcp/abilities/utils/style-variants-merger.php
```diff
@@ -0,0 +1,206 @@
...
+		return array_keys(
+			array_filter( $css, fn( $v ) => null === $v || 'null' === $v )
+		);
+	}
+
+	public static function merge_custom_css( ?array $existing, ?array $incoming ): ?array {
+		$existing_raw = isset( $existing['raw'] ) ? base64_decode( $existing['raw'] ) : ''; // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.obfuscation_base64_decode -- Decoding stored base64 custom_css for merge.
+		$incoming_raw = isset( $incoming['raw'] ) ? base64_decode( $incoming['raw'] ) : ''; // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.obfuscation_base64_decode -- Decoding stored base64 custom_css for merge.
+
+		$merged = trim( $existing_raw . ' ' . $incoming_raw );
```
Omerisra6, line 142: Plain concat means `patch` isn't idempotent — same call three times stores the block four times, and it never shrinks. Dedupe or replace matching blocks.

modules/mcp/abilities/utils/style-variants-merger.php
```diff
@@ -0,0 +1,206 @@
...
+							: null,
+						'null_props' => $null_props,
+					];
+					continue;
+				}
+
+				$base_custom_css_parts[] = '&' . $selector . ' { ' . trim( $css ) . ' }';
+			}
+
+			$base_null_props = self::extract_null_props( $base_block_css );
```
Omerisra6: Base-block tail repeats the pseudo branch step for step (extract nulls, strip, convert, base64-wrap). Extract `make_variant($breakpoint, $state, $css)` and remove one of the two copies.

modules/mcp/abilities/utils/style-variants-merger.php
```diff
@@ -0,0 +1,206 @@
...
+
+			$result[ $match ]['props']      = $merged_props;
+			$result[ $match ]['custom_css'] = $merged_custom_css;
+			unset( $result[ $match ]['null_props'] );
+		}
+
+		return $result;
+	}
+
+	public static function extract_null_reset_props( array $css ): array {
```
Omerisra6: Dead: `extract_null_reset_props` has zero callers anywhere in the repo. Same for `apply_mode`'s 5th param `$props_to_remove` — every live call passes 4 args.

modules/mcp/abilities/manage-classes-ability.php
```diff
@@ -258,17 +291,102 @@ private function translate_update( int $index, array $operation, array &$intents
...
+			$intents['updates'][ $index ] = [
+				'id'                  => $id,
+				'label'               => $label,
+				'css'                 => [],
+				'breakpoint_blocks'   => $parsed['breakpoint_blocks'],
+				'removal_breakpoints' => $parsed['removal_breakpoints'],
+				'mode'                => $operation['mode'] ?? 'patch',
+				'existing_variants'   => $existing['variants'] ?? [],
+			];
+		} else {
```
Omerisra6, line 318: Label-only update has no working downstream path: intent carries no `breakpoint_blocks`, so `build_touched_items` falls into `build_class_item(..., [])` which always returns "CSS produced no valid style properties". Either support the rename properly or reject it here with an accurate message.

modules/mcp/abilities/manage-classes-ability.php
```diff
@@ -318,13 +447,45 @@ private function build_touched_items( array $intents, array $all_labels, Bulk_Op
...
 				$results->add_error( $index, 'update', $class_item->get_error_code(), $class_item->get_error_message() );
 				continue;
 			}
 
+			$class_item['variants'] = Style_Variants_Merger::apply_mode(
+				$intent['existing_variants'],
+				$class_item['variants'],
+				$intent['mode'],
+				[ self::DESKTOP_BREAKPOINT ],
+				$intent['null_reset_props']
```
Omerisra6: `$intent['null_reset_props']` is never written by `translate_update` in either branch. Production is only saved by the `is_wp_error` guard above; tests would fatal on `null` into an `array`-typed parameter.

modules/mcp/abilities/manage-classes-ability.php
```diff
@@ -307,6 +425,17 @@ private function build_touched_items( array $intents, array $all_labels, Bulk_Op
...
+				$touched_items[ $intent['id'] ] = [
+					'id'       => $intent['id'],
+					'label'    => $intent['label'],
+					'type'     => self::CLASS_TYPE,
+					'variants' => $variants,
+				];
+				continue;
+			}
+
 			$class_item = $this->build_class_item( $intent['id'], $intent['label'], $intent['css'] );
```
Omerisra6: Dead code: `build_class_item` (line 615) and `convert_css_to_variant` (line 640) are unreachable now that both intent branches always set `breakpoint_blocks` — only the broken label-only update path reaches them. Same story for `'css' => []` in every intent, and `as_map` at line 748.

modules/mcp/abilities/manage-classes-ability.php
```diff
@@ -258,17 +291,102 @@ private function translate_update( int $index, array $operation, array &$intents
...
+				'invalid_css',
+				$result['error'] . sprintf( ' Valid breakpoints: %s.', implode( ', ', $active_breakpoints ) )
+			);
+			return null;
+		}
+
+		return $this->parse_styles_field( $index, $action, $result['breakpoints'], $results );
+	}
+
+	private function parse_styles_field( int $index, string $action, array $raw_styles, Bulk_Operations_Result $results ): ?array {
```
Omerisra6, line 354: All three guards here are unreachable: `Css_Media_Splitter` already normalizes `default` → `desktop`, validates the breakpoint list, and always emits string values. Only the `parse_nested` loop is doing real work; the second `get_active_breakpoint_keys()` call can go.

modules/mcp/abilities/manage-classes-ability.php
```diff
@@ -218,32 +234,49 @@ private function translate_operations( array $operations, array $all_labels, arr
...
 
 		$intents['creates'][ $index ] = [
-			'id' => $class_id,
-			'label' => $label,
-			'css' => $css,
+			'id'                  => $class_id,
+			'label'               => $label,
+			'css'                 => [],
+			'breakpoint_blocks'   => $parsed['breakpoint_blocks'],
+			'removal_breakpoints' => $parsed['removal_breakpoints'],
```
Omerisra6: `removal_breakpoints` is written for creates but only read on the update path.

modules/mcp/abilities/manage-classes-ability.php
```diff
@@ -148,6 +156,14 @@ private function handle_bulk( array $operations ): array {
 
 		$new_order = $this->compute_new_order( $current_order, $added_ids, $deleted_ids );
 
+		// NOTE: null_props is an internal patch-merge signal; strip it from all variants before persisting.
+		foreach ( $touched_items as &$item ) {
+			foreach ( $item['variants'] as &$variant ) {
+				unset( $variant['null_props'] );
```
Omerisra6, line 160: `null_props` is stripped in four places — this one plus three inside `apply_mode`. This pass is enough (creates bypass `apply_mode`), so the three inside are redundant. Cleaner still: have `build_variants` return null_props as a sibling instead of embedding an internal signal in the persisted structure.

.github/scripts/prepare-slack-release-announcement.sh
Omerisra6, line 1: please remove all unrelated files

Omerisra6 (dismissed):

Omerisra6 (approved):

PR #36748, opened 2026-08-03
Internal: Background Video QA Fixes [ED-25117]

max-zu:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #

## 1. Problem & Context

Background Video element needs children dependency reconciliation at render time (matching editor canvas behavior) and flexible HTML tagging for semantic markup. Current implementation lacks server-side child validation and uses hardcoded `<div>`.

## 2. What Changed (Where)

- **Children dependency evaluation**: New `Children_Dependency_Evaluator` class (PHP) + `reconcileComponentInstanceElements` (TS/PHP) apply conditional child rules at render and component instantiation.
- **Background Video**: Added `tag` prop for semantic HTML, refactored padding strategy (10px root, 0 content), updated widget descriptions clarifying required children structure.
- **Play/Pause buttons**: Extract label paragraph generation into reusable methods, mark as required children, added `get_props_schema()` to hide display-conditions on locked sub-elements.
- **MCP validation**: New `Default_Children_Utils`, `collect_required_child_errors()` validates required children during XML composition.
- **Styling**: Fixed flex basis (auto/custom instead of 0/px), improved empty-state CSS scoping, content area gets `flex: 1 1 auto` with editor-specific min-height handling.

## 3. How It Works

When a component renders, `Reconcile_Component_Instance_Elements` traverses the element tree, evaluates `children_dependencies` rules against resolved settings, and inserts/removes child elements. Inserted children get stable IDs derived from parent ID + position (hashed identically in TS/PHP). Display-conditions and overridable props are evaluated server-side before ID derivation. Play/Pause buttons now generate from `build_default_element()` factories marked as required, ensuring controls always exist unless explicitly hidden via `show_controls` setting.

## 4. Risks

ID derivation must remain in sync between PHP (`Format_Element_Ids::hash_string`) and TS (`hashString`) or canvas/render diverge—mitigated by cross-platform test assertions. Locked sub-elements (`get_props_schema()` filter) hides display-conditions to prevent accidental user overrides; verify UI doesn't expose these controls. Padding flip (root 10px, content 0) reverses prior logic; confirm visual spacing matches design on nested content. Test coverage solid (215 TS + 310 PHP unit tests), but component-level e2e tests recommended for override application flow.


modules/components/transformers/component-instance-transformer.php
```diff
@@ -58,6 +59,7 @@ private function get_rendered_content( int $component_id, ?string $instance_elem
 		$data = apply_filters( 'elementor/frontend/builder_content_data', $data, $component_id );
 
 		$data = Format_Component_Elements_Id::format( $data, [ $instance_element_id ] );
+		$data = Reconcile_Component_Instance_Elements::apply( $data );
```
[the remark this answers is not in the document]
max-zu: Fixed in 28b8296 — reconcile now runs before `Format_Component_Elements_Id::format()`.

One addition beyond the suggested swap: `default_model` payloads are built by `Element_Builder`, which never emits an `id`, so swapping alone would have left `Format_Element_Ids::format()` hashing a missing key. Nested default children (Play/Pause under Controls) would then have hashed to the same value and collided. `Reconcile_Component_Instance_Elements::derive_ids()` now assigns each inserted node an id derived from the parent id and its position in the subtree, so the tree is fully formed before formatting and stays identical across renders.

packages/packages/core/editor-components/src/create-component-type.ts
```diff
@@ -220,12 +221,17 @@ function createComponentView( options: ComponentTypeOptions ): typeof TemplatedE
 				const instanceId = this.model.get( 'id' );
 				const elements = componentInstance.elements ?? [];
 				const formattedElements = formatComponentElementsId( elements, [ instanceId ] );
+				const reconciledElements = reconcileComponentInstanceElements(
+					formattedElements,
+					componentInstance.overrides ?? {}
+				);
```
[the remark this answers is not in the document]
max-zu: Fixed in 28b8296 — reconcile now runs on the raw component elements and `formatComponentElementsId` formats the complete tree.

The swap on its own would have kept the collision risk, just moved it: `reconcileInitialChildren` fills inserted models via `ensureModelId`, so the pre-format id was still a random `generateElementId()` value and the resulting hash differed on every render and from the PHP render path. `deriveIds()` now replaces the ids of the inserted subtree using the same seed scheme as `derive_ids()` in PHP, so both paths produce identical ids. `reconcile-component-instance-elements.test.ts` asserts the hashes verbatim against the PHP test to keep the two in sync.

modules/atomic-widgets/children-dependencies/children-dependency-evaluator.php
```diff
@@ -0,0 +1,93 @@
+<?php
+
+namespace Elementor\Modules\AtomicWidgets\ChildrenDependencies;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+class Children_Dependency_Evaluator {
```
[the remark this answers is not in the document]
max-zu: Added `tests/phpunit/elementor/modules/atomic-widgets/children-dependencies/test-children-dependency-evaluator.php` in 28b8296: all operators (`eq`/`ne`, `gt`/`gte`/`lt`/`lte` incl. non-numeric rejection, `in`/`nin`, `contains`/`ncontains`, `exists`/`not_exist`), the unknown-operator default, missing and non-traversable paths, `and`/`or` relations with the implicit `or` default, nested groups, and the real `show_controls ne false` rule.

Writing the tests surfaced a genuine parity gap: `contains`/`ncontains` were missing from the PHP `compare()` and fell through to `default: return true`, i.e. silently always-met. Added them mirroring `evaluateTerm`, including unwrapping transformable array items.

modules/components/utils/reconcile-component-instance-elements.php
```diff
@@ -0,0 +1,111 @@
...
+use Elementor\Modules\AtomicWidgets\PropsResolver\Render_Props_Resolver;
+use Elementor\Modules\AtomicWidgets\Utils\Utils;
+use Elementor\Plugin;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+class Reconcile_Component_Instance_Elements {
+	public static function apply( array $elements ): array {
```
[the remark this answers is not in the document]
max-zu: Both sides are covered in 28b8296:

- `tests/phpunit/elementor/modules/components/utils/test-reconcile-component-instance-elements.php`
- `packages/packages/core/editor-components/src/utils/__tests__/reconcile-component-instance-elements.test.ts`

They deliberately mirror each other: add-on-condition-met, add-on-schema-default, remove-on-condition-failed, existing child left untouched, elements without dependencies (and unknown types) untouched, nested reconciliation, `default_model` payload preserved, and id derivation (stable across runs, distinct per parent, distinct between sibling default children). The derived hashes are asserted verbatim in both suites, so a divergence between the PHP render path and the canvas fails the build. The TS suite additionally covers override application and unwrapping, which has no PHP counterpart.

While writing these I also removed a double recursion in `reconcile_element()` — children were being walked once before the rules ran and again after.

modules/mcp/abilities/build-composition/widget-type-resolver.php
```diff
@@ -138,4 +141,51 @@ private function collect_child_type_errors( \DOMElement $node, array $widget_con
 			$this->collect_child_type_errors( $child, $widget_configs, $errors );
 		}
 	}
+
+	private function collect_required_child_errors( \DOMElement $node, array $widget_configs, array &$errors ): void {
```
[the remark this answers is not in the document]
max-zu: Added `tests/phpunit/elementor/modules/mcp/abilities/build-composition/test-widget-type-resolver.php` in 28b8296 rather than extending `test-element-config-applier.php`, since `validate_child_types()` only needs `Xml_Parser` plus a `$widget_configs` array and reads better as its own suite.

It covers the `meta.required` / `elType` / `widgetType` resolution you flagged: required child present, missing, parent with no children at all, children without `meta.required` not being required, a nested-but-not-direct child failing to satisfy the requirement, required widgets resolved by `widgetType` instead of `elType`, recursive validation of nested parents, unknown tags ignored, and disallowed plus missing-required errors being reported together in one `WP_Error`.

assets/dev/scss/editor/preview/_container.scss
ManorHazaz, line 1: Verify that this change is intended, I think it's a high-risk file

modules/mcp/abilities/build-composition/widget-type-resolver.php
```diff
@@ -138,4 +141,51 @@ private function collect_child_type_errors( \DOMElement $node, array $widget_con
...
+				}
+			}
+		}
+
+		foreach ( $this->xml_parser->get_child_elements( $node ) as $child ) {
+			$this->collect_required_child_errors( $child, $widget_configs, $errors );
+		}
+	}
+
+	private function get_required_child_types( array $default_children ): array {
```
Omerisra6, line 174: Byte-for-byte duplicate of `Llm_Guidance_Builder::get_required_default_child_types()`. Extract to a shared helper so the guidance surfaced to the LLM and the server-side enforcement can't drift.

modules/atomic-widgets/children-dependencies/children-dependency-evaluator.php
```diff
@@ -0,0 +1,123 @@
...
+				return ( 'contains' === $operator ) === $contains;
+
+			case 'exists':
+			case 'not_exist':
+				$exists = (bool) $actual_value || 0 === $actual_value || false === $actual_value;
+
+				return ( 'exists' === $operator ) === $exists;
+
+			default:
+				return true;
```
Omerisra6, line 100: Unknown operator silently returns true — a typo'd operator would keep a hidden child visible. Fail-closed (`return false`) or `Utils::safe_throw`.

modules/atomic-widgets/children-dependencies/children-dependency-evaluator.php
```diff
@@ -0,0 +1,123 @@
...
+
+			$value = $value[ $segment ];
+		}
+
+		return $value;
+	}
+
+	private static function compare( string $operator, $actual_value, $expected_value ): bool {
+		switch ( $operator ) {
+			case 'eq':
```
Omerisra6, line 55: `PropDependencies\Manager::OPERATORS` already declares the canonical operator list. Reference it here so both sides can't drift.

packages/packages/core/editor-components/src/create-component-type.ts
```diff
@@ -219,13 +220,18 @@ function createComponentView( options: ComponentTypeOptions ): typeof TemplatedE
...
+				const formattedElements = formatComponentElementsId( reconciledElements, [ instanceId ] );
 
 				this.collection = legacyWindow.elementor.createBackboneElementsCollection( formattedElements );
 
 				this.collection.models.forEach( setInactiveRecursively );
 
 				settings.component_instance = '<template data-children-placeholder></template>';
+				settings.__componentOverridesFingerprint = JSON.stringify( componentInstance.overrides ?? {} );
```
Omerisra6, line 234: Where is `__componentOverridesFingerprint` read? If the consumer isn't in this PR, this is dead state on `settings` — drop the writer.

modules/components/utils/reconcile-component-instance-elements.php
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	/**
+	 * `default_model` payloads are static configuration and carry no ids, so every
+	 * inserted node gets one derived from the parent id and its position in the subtree.
+	 * Keep in sync with `deriveIds()` in the editor's
+	 * `reconcile-component-instance-elements.ts` so ids match between render and canvas.
+	 */
+	private static function derive_ids( array $element, string $seed ): array {
+		$element['id'] = Format_Element_Ids::hash_string( $seed, self::ELEMENT_ID_LENGTH );
```
Omerisra6, line 91: PHP `derive_ids` and TS `deriveIds` must stay identical. Add a parity test asserting `Format_Element_Ids::hash_string(seed, 7)` matches the TS `hashString` output for a fixed seed, so they can't silently drift.

Omerisra6 (commented):
Scope note: title says "Background Video QA Fixes" but this ships a runtime reconciliation subsystem (PHP+TS) used by Collection Loop as well. Consider updating the description. Also — the pinned CI run shows a PHPUnit failure; can we get a green run on the latest head before approving?

assets/dev/scss/editor/preview/_container.scss
max-zu: Yes, intentional — but the `:where()` approach from the first iteration was rolled back.

On `main`, `.e-atomic-element:has( .elementor-empty-view )` matches any descendant empty view. For background-video the empty view sits on `.e-background-video__content`, not the root, so the root picked up `min-height: 120px` and overrode Settings → Size on the widget.

The fix scopes the rule to a direct child only (`:has( > .elementor-empty-view )`). Atomic containers always mount empty view as a direct child (`reInitEmptyView` uses `:scope > .elementor-empty-view`), so empty div-block / flexbox / grid still get the 120px placeholder. Non-empty ancestors (including the video root) no longer inherit it.

Background-video content area keeps its own editor rules in `module.php` inline CSS. Added a short comment in the scss block so the intent is obvious on re-read.

modules/mcp/abilities/build-composition/widget-type-resolver.php
max-zu: Extracted to `Default_Children_Utils::get_required_child_types()` under `modules/mcp/abilities/utils/`. Both `Widget_Type_Resolver` and `Llm_Guidance_Builder` now call the shared helper.

modules/atomic-widgets/children-dependencies/children-dependency-evaluator.php
max-zu: Done — unknown operators now fail closed (`return false`), so a typo hides the conditional child instead of leaving it visible. Guard runs against `PropDependencies\Manager::OPERATORS` before the switch. Did not use `Utils::safe_throw` here: with `ELEMENTOR_DEBUG` that would throw during render/tests rather than just hiding the child.

Note: TS `evaluateTerm` in editor-props still defaults to `true` for panel conditional fields — left as-is since that path is separate; can align in a follow-up if we want strict parity there too.

modules/atomic-widgets/children-dependencies/children-dependency-evaluator.php
max-zu: Same change — `compare()` now rejects anything outside `Dependency_Manager::OPERATORS` before entering the switch.

packages/packages/core/editor-components/src/create-component-type.ts
max-zu: Removed — no reads anywhere in the repo. Looked like leftover debug/state from an earlier iteration.

modules/components/utils/reconcile-component-instance-elements.php
max-zu: Added shared seed→hash cases in `hash.test.ts` and `test-format-component-elements-id.php` (same inputs, same expected 7-char hashes). Reconcile integration tests on both sides still assert the derived subtree ids verbatim (`0f3lpg5` / `0p2wcv5` / `1ed1jog`) so PHP render and canvas stay locked together.

Omerisra6 (approved):

PR #36749, opened 2026-08-03
Internal: Add preview link tool to our mcp [ED-25184]

Omerisra6:
## Summary

Adds a new MCP ability `elementor/create-preview-link` that returns a signed, time-limited URL to render an Elementor post from a WordPress revision snapshot — without publishing the post and without requiring the visitor to be logged in.

Intended primarily as a self-verification tool for the agent: it can generate a preview URL, fetch/screenshot it, and validate the rendered output before showing anything to the user.

## What's included

- **`Create_Preview_Link_Ability`** — validates the target post (must exist, be built with Elementor, and the caller must have `edit_post` capability), snapshots the current state via `wp_save_post_revision`, and returns `{ url, edit_url, post_id, revision_id, expires_at, expires_at_unix }`. TTL is fixed (5 minutes) and controlled server-side.
- **`Preview_Token`** — stateless HMAC-SHA256 encode/verify. Secret derives from `wp_salt('auth')` + a namespace, so tokens can't be replayed across sites or reused after WP secret rotation.
- **`Public_Preview_Handler`** — hooks `parse_request`; on a valid token it:
  - validates the post (exists, is not itself a revision, is Elementor-built) and the revision (belongs to that post),
  - rewrites the main query to serve the parent post regardless of status,
  - primes an in-memory cache of the revision's Elementor postmeta and swaps it in via `get_post_metadata` for the parent post only,
  - emits `X-Robots-Tag: noindex, nofollow, noarchive`, `Cache-Control: private, no-store`, and `Referrer-Policy: no-referrer`.

Registered in `Modules\Mcp\Module` alongside the existing abilities.

## Security notes

- Tokens are opaque, signed, and time-limited; no state stored on the server.
- Any token failure (bad signature, expired, revision/post mismatch, missing post, non-Elementor post) results in a silent no-op — WP renders the request as it normally would.
- Meta override is scoped to the post ID from the token and a fixed allowlist of `_elementor_*` keys; other meta pass through untouched.
- The public URL is explicitly separated from the `edit_url` in the response so agents don't accidentally share the editor link with end users.

## Tests

- `tests/phpunit/elementor/modules/mcp/test-create-preview-link-ability.php` — happy path, 404/400/403 error cases, revision snapshot contents, token round-trip, fixed TTL.
- `tests/phpunit/elementor/modules/mcp/preview/test-preview-token.php` — encode/verify, tamper detection, expiry.
- `tests/phpunit/elementor/modules/mcp/preview/test-public-preview-handler.php` — no-token / invalid / expired / missing-post / non-Elementor / parent-mismatch flows, valid-token override behavior, unrelated meta pass-through, query rewrite.

## Jira

ED-XXXX

## 1. Problem & Context

The MCP tool needed a dedicated ability to generate time-limited preview links so agents can self-validate rendered output before sharing editor URLs to users. This decouples preview generation from mutation operations (build-composition, manage-elements), simplifying response schemas and enabling reusable preview logic.

## 2. What Changed (Where)

| File | Change |
|------|--------|
| `create-preview-link-ability.php` | New ability: encodes signed tokens, creates revisions, returns preview + editor URLs |
| `preview-token.php` | New crypto: HMAC-SHA256 token encoding/decoding with expiry validation |
| `public-preview-handler.php` | New middleware: hooks parse_request, decodes tokens, shadows post meta with revision snapshots, blocks indexing |
| `build-composition-ability.php` | Removed `Document_Mutation_Links` import/use; returns `edit_url` directly, dropped `preview_url`/`llm_instructions` |
| `manage-elements-ability.php` | Same refactor: `with_edit_url()` helper replaces `with_mutation_links()` |
| `module.php` | Registers `Public_Preview_Handler`, registers new ability in tools list |
| Tests | Added unit/integration tests for token crypto, preview handler, new ability |

## 3. How It Works

**Entry point:** `elementor/create-preview-link` ability accepts post_id → validates edit permission → snapshots revision via `wp_save_post_revision()` → encodes token with post_id, revision_id, 5min expiry using HMAC-SHA256 → returns anonymous preview URL + editor URL.

**Happy path:** Frontend visits preview URL with token query param → `Public_Preview_Handler::maybe_activate()` validates token signature/expiry → rewrites main query to fetch post → intercepts `get_post_metadata` to shadow post meta with revision's `_elementor_data` → renders draft/snapshot without publish.

**Edge cases:** Expired/tampered tokens silently fail (return null from decode). Revision parent mismatch blocked. Non-Elementor posts rejected. Privacy headers (`X-Robots-Tag: noindex`, `Cache-Control: private`) prevent indexing.

## 4. Risks

**Token secret derivation:** Uses `wp_salt('auth')` — if salts rotate mid-flight, tokens become unverifiable. Acceptable: TTL is short (5min), rotation is rare. **No explicit token revocation:** Once issued, token valid until expiry. Acceptable: short TTL + revision is immutable snapshot. **Cache bypass assumption:** `get_post_metadata` filter assumes no persistent cache layer between filter and DB; if caching layer exists upstream, shadow won't apply. Low risk for drafts in typical setups.


tests/phpunit/elementor/modules/mcp/test-create-preview-link-ability.php
```diff
@@ -0,0 +1,135 @@
...
+		$excessive_ttl_minutes = 999999;
+
+		$before = time();
+		$result = $this->ability->execute( [ 'post_id' => $post_id, 'ttl_minutes' => $excessive_ttl_minutes ] );
+		$after = time();
+
+		$this->assert_expires_after_fixed_ttl( $result, $before, $after );
+	}
+
+	private function assert_expires_after_fixed_ttl( array $result, int $before, int $after ): void {
```
ronros-elementor: it seems like the before and after are always the same, am I wrong?

if that's the case, I think a simpler assertion would simply be now() + Create_Preview_Link_Ability::TTL_MINUTES * MINUTE_IN_SECONDS

ronros-elementor (dismissed):

ronros-elementor (approved):

PR #36751, opened 2026-08-03
Internal: [V4] Remove HTML-v3 children support

ronenboxer:
## 1. Problem & Context

Migrate text content properties from `Html_V3_Prop_Type` (which supports nested children elements) to `Escaped_Html_Prop_Type` (plain string with HTML sanitization). Removes legacy children parsing complexity that's unnecessary for inline text fields. Ticket: ED-25027.

## 2. What Changed (Where)

| Component | Change |
|-----------|--------|
| **Backend Props** | New `Escaped_Html_Prop_Type` extending `Html_Prop_Type`; migration rules for html-v3→escaped-html |
| **Widgets** | Button, Heading, Paragraph, Tabs, Form elements: text/title/paragraph props switch from Html_V3 to Escaped_Html |
| **Transformers** | New `Escaped_Html_Transformer` registers sanitizer; phased out children parsing in inline editing |
| **Dynamic Tags** | Updated mapping to support both Escaped_Html and Html_V3 for TEXT_CATEGORY |
| **Frontend (TS)** | `InlineEditingControl` simplified—no debounced child parsing; `escapedHtmlPropTypeUtil` added to eligibility checks |
| **Migrations** | `html-v3-to-escaped-html.json` unwraps content.value, drops children, handles null/primitive cases; reversible via "down" |
| **Tests** | 100+ test updates: mocks, fixtures, assertions now expect `escaped-html` instead of `html-v3` structures |

## 3. How It Works

**Migration layer:** Content wrapped as `{content: {$type: 'string', value: 'text'}, children: []}` unwraps to plain `'text'` on upgrade; reverses on downgrade. 

**Sanitization:** `Html_Prop_Type::sanitize_allowed_html()` uses dynamic `allowedHTMLWrapperTags` filter; frontend uses DOMPurify with same allowlist. Strips `<script>`, event handlers, `javascript:` URIs.

**Inline editing (frontend):** Removed `parseHtmlChildren` debounce loop—now direct value passthrough for escaped-html; falls back to html-v3 extraction for unmigrated data.

**Edge case:** Empty strings persist (not replaced with default); null content→null value; backward-compatible union props (dynamic + escaped-html) check escaped-html first.

## 4. Risks

**Migration reversibility:** Down-migration wraps content but children lost—acceptable (children unsupported in new model anyway). Verify production docs show one-way intent.

**Sanitization coverage:** Frontend sanitization via DOMPurify + backend via wp_kses may diverge if allowlist configs desync. Monitor `elementor/allowed_html_wrapper_tags` filter registration in module init.

**Test snapshot churn:** Massive fixture updates across 25+ files—low risk functionally but high noise in review. Confirm snapshot regeneration was complete (no stale html-v3 leaking through).


modules/atomic-widgets/module.php
```diff
@@ -288,6 +290,27 @@ private function register_hooks() {
 		( new Atomic_Widgets_Database_Updater() )->register();
 		( new Css_Converter_REST_API() )->register_hooks();
 		( new Pro_Promotion_Data_Preservation() )->register_hooks();
+
+		add_filter( 'elementor/allowed_html_wrapper_tags', [ $this, 'add_inline_html_tags' ] );
+	}
+
+	public function add_inline_html_tags( array $tags ): array {
+		return array_merge( $tags, [
```
Ntnelbaba, line 298: Add span to this list

packages/packages/core/editor-canvas/src/utils/sanitize-escaped-html.ts
```diff
@@ -0,0 +1,18 @@
...
+export function sanitizeEscapedHtml( value: string | null ): string {
+	if ( ! value ) {
+		return '';
+	}
+
+	const allowedTags = [ ...getAllowedHtmlWrapperTags() ];
+
+	return DOMPurify.sanitize( value, {
+		ALLOWED_TAGS: allowedTags,
+		ALLOWED_ATTR: [ 'href', 'target' ],
```
Ntnelbaba: Expand this list to other non operationals attrs (such as class, ID, data-*, style and such, **no functional**)

tests/phpunit/elementor/modules/atomic-widgets/prop-type-migrations/fixtures/document-migrations/html-v3-to-escaped-html.json
Ntnelbaba, line 1: Ohh this is a big mock
Create a ticket to remove this mocks, it suppose to use the real infra

Ntnelbaba (dismissed):

Ntnelbaba (approved):

PR #36772, opened 2026-08-05
Fix: Batch component document config requests on editor load [ED-25172]

Ntnelbaba:
## Summary

V4 Components on editor open triggered one concurrent `admin-ajax.php` request per distinct component document via `documentManager.request()` (`immediately: true`). On pages with many or nested components this burst can hit host rate limits (HTTP 429) and leave some components unloaded.

`getComponentDocumentData()` now issues the same `get_document_config` request through `ajax.load()` **without** `immediately`, so concurrent component fetches merge into a single `elementor_ajax` multi-action request.

## Changes

- `component-document-data.ts`: build the request args locally and load them through the batching ajax path
- Unit tests covering the batched request args and the failure path

## Why not `apiClient.getComponentConfig()`

The obvious route was the existing `apiClient.getComponentConfig()`, but its `getParams()` uses a `document-config-<id>` `unique_id` that nothing ever fetched with, so `invalidateComponentConfigCache()` has always been a no-op. Routing component loading through it breaks in two ways:

- Component freshness: every invalidation path (`invalidateDocumentData` on component switch, `save.js` on document save) clears `document-<id>`, so an edited component would keep rendering stale content.
- Unrelated documents: `removeComponentStyles()` runs on every `attach-preview` for *any* document, so making that key real turns it into "clear the V1 documents manager cache on each preview attach". That regressed the site settings save/reload flow and failed `image-carousel.test.ts › Image Carousel Responsive Spacing` in CI.

So the request args are built in the component loader and `api.ts` is left untouched.

## Jira

https://elementor.atlassian.net/browse/ED-25172

## Testing

- `npm run test:packages -- --testPathPattern="editor-components"` — 217 passed
- `npm run lint` — ESLint and `tsc` both exit 0
- WordPress Playground, page with 5 distinct Components: 1 batched request on editor load carrying all 5 ids, all 5 components rendered, 0 requests for a cached read through the V1 documents manager, exactly 1 refetch after `elementor.documents.invalidateCache()`, and no failed `admin-ajax.php` responses

<div><a href="https://cursor.com/agents/bc-019fcff5-fccb-773a-a266-2458736b9875?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-019fcff5-fccb-773a-a266-2458736b9875&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

## 1. Problem & Context

Switching from sequential document requests to batched ajax calls to reduce waterfall requests and leverage the v1 documents manager's existing cache layer (ED-25172). Enables automatic cache invalidation across both request paths.

## 2. What Changed (Where)

- `component-document-data.ts`: Replaced `documentManager.request()` with `ajax.load()` using batched request params with `unique_id` matching v1 cache keys
- `component-document-data.test.ts`: Added comprehensive test coverage for batched request behavior and error handling

## 3. How It Works

`getComponentDocumentData(id)` now builds request params (`action`, `unique_id`, `data`) that match v1 manager's ajax cache entry signature, allowing batching middleware to coalesce concurrent requests into a single call. The `unique_id` ensures the ajax cache treats both component and manager requests identically, so `elementor.documents.invalidateCache()` invalidates both codepaths.

## 4. Risks

Minimal—mirrors existing v1 manager behavior but removes the `immediately` flag, intentionally delaying requests to enable batching. Verify no downstream code depends on synchronous request timing.


ronros-elementor (commented):
the single concern I have around this, is just the fact we're bypassing the documents manager, however upon changes to components' content - we're expected to see the publish button enabled on the root page, even if no changes occurred on that page directly, but only to components used in it

the only reason for concern here is depending on whether we use the documents manager's corresponding doc for changes tracking, and if so - would this bypass fail to populate the documents manager with relevant documents (I actually assume it's not a true issue as we use the `switch` command, which I expect triggers that documents manager fetching regardless)

ronros-elementor (approved):

PR #36773, opened 2026-08-05
Fix: Enforce MCP ability permissions on proxy path [ED-25102]

Omerisra6:
## Summary
- The in-editor `mcp-proxy` REST endpoint only checked `edit_posts` before calling ability `execute()`, bypassing each ability's `permission_callback`.
- Enforce per-ability permissions in the proxy (tools + resources) and in `read-resource` delegation so Editors cannot mutate global variables/classes or read admin-only resources.
- Add PHPUnit coverage for editor vs admin on proxy tools/resources and for `Abstract_Ability` permission helpers.

## Jira
[ED-25102](https://elementor.atlassian.net/browse/ED-25102)

## Test plan
- [ ] As Editor: `POST /elementor/v1/mcp-proxy` with `manage-global-variable` returns 403
- [ ] As Editor: `POST /elementor/v1/mcp-proxy` with `manage-classes` returns 403
- [ ] As Editor: `GET /elementor/v1/mcp-proxy?uri=elementor://variables/tools/manage-global-variable-guide` returns 403
- [ ] As Admin: same calls proceed past auth (validation/success as appropriate)
- [ ] As Editor: `read-resource` for the manage-global-variable guide returns 403
- [ ] Run: `vendor/bin/phpunit --filter 'Test_Mcp_Proxy_REST_API|Test_Abstract_Ability_Permissions'`

[ED-25102]: https://elementor.atlassian.net/browse/ED-25102?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## 1. Problem & Context

MCP proxy endpoints weren't enforcing ability-level permissions—they only called execute() directly. ED-25102 requires permission checks at the proxy layer to deny unauthorized access before execution.

## 2. What Changed (Where)

- **abstract-ability.php**: Added `check_permission()` method that invokes the ability definition's permission callback
- **read-resource-ability.php**: Refactored executor map to store ability instances instead of lambdas; added permission check before execute()
- **mcp-proxy-rest-api.php**: Restructured tools/resources maps to return ability factories; added permission checks in both `handle_tool()` and `handle_resource()` paths with shared `forbidden_error()` helper
- **test-abstract-ability-permissions.php** (new): Unit tests for permission callback invocation with editor/admin contexts
- **test-mcp-proxy-rest-api.php** (new): Integration tests verifying REST endpoint rejection/allowance by role across manage-variable, manage-classes, and resource endpoints

## 3. How It Works

Permission check flow: request → route handler instantiates ability via factory → `check_permission()` calls ability definition's callback → if denied, returns 403 before execute() runs. Both tool and resource paths follow identical pattern. Tests confirm editors get FORBIDDEN (403) while admins proceed.

## 4. Risks

Permission callback execution now happens twice for resources (once in proxy, once in Read_Resource_Ability internally)—minor redundancy but functionally safe since callbacks are idempotent guards. No breaking changes to existing ability contracts.

### DIFF
--- a/modules/mcp/abilities/abstract-ability.php
+++ b/modules/mcp/abilities/abstract-ability.php
@@ -14,6 +14,22 @@ abstract class Abstract_Ability {
 
 	abstract public function execute( $input = [] );
 
+	public function check_permission(): bool {
+		return (bool) call_user_func( $this->get_definition()->permission_callback );
+	}
+
+	public function permission_error(): ?\WP_Error {
+		if ( $this->check_permission() ) {
+			return null;
+		}
+
+		return new \WP_Error(
+			'rest_forbidden',
+			__( 'Sorry, you are not allowed to perform this action.', 'elementor' ),
+			[ 'status' => \WP_Http::FORBIDDEN ]
+		);
+	}
+
 	public function register(): void {
 		$definition = $this->get_definition()->to_array();
 		$definition['execute_callback'] = [ $this, 'execute' ];
--- a/modules/mcp/abilities/abstract-ability.php
+++ b/modules/mcp/abilities/abstract-ability.php
@@ -18,18 +18,6 @@ abstract class Abstract_Ability {
 		return (bool) call_user_func( $this->get_definition()->permission_callback );
 	}
 
-	public function permission_error(): ?\WP_Error {
-		if ( $this->check_permission() ) {
-			return null;
-		}
-
-		return new \WP_Error(
-			'rest_forbidden',
-			__( 'Sorry, you are not allowed to perform this action.', 'elementor' ),
-			[ 'status' => \WP_Http::FORBIDDEN ]
-		);
-	}
-
 	public function register(): void {
 		$definition = $this->get_definition()->to_array();
 		$definition['execute_callback'] = [ $this, 'execute' ];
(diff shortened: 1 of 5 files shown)


modules/mcp/abilities/abstract-ability.php
```diff
@@ -14,6 +14,22 @@ abstract protected function get_definition(): Ability_Definition;
 
 	abstract public function execute( $input = [] );
 
+	public function check_permission(): bool {
```
IshayMaya, line 17: what about pro?

modules/mcp/abilities/abstract-ability.php
```diff
@@ -14,6 +14,22 @@ abstract protected function get_definition(): Ability_Definition;
...
+		return (bool) call_user_func( $this->get_definition()->permission_callback );
+	}
+
+	public function permission_error(): ?\WP_Error {
```
IshayMaya: i think it should be the other way - check permission that returns the error if exist

Omerisra6:
Addressed review feedback:

1. Removed `permission_error()` — consumers (`mcp-proxy`, `read-resource`) now call `check_permission()` and build the `WP_Error` themselves.
2. Pro parity: https://github.com/elementor/elementor-pro/pull/7424

IshayMaya (approved):

PR #36776, opened 2026-08-05
Internal: Add MCP ability to reorder global classes [ED-25137]

ronros-elementor:
## Summary
- Adds `elementor/reorder-classes`, a new MCP ability that changes global CSS class priority via sequential relative moves (`before`/`after`/`start`/`end`) or an explicit order, so agents can fix cross-class style overrides without recreating classes.
- Restructures the `elementor://global-classes` MCP resource payload from an id→label object (order only implied by JSON key order) to an explicit `{ priority, classes: [{ id, label }] }` list, and documents that earlier entries override later ones.
- Registers the new ability in the MCP server tool list and the `mcp-proxy` REST route, with its own capability check (`elementor_global_classes_update_class`) enforced in the proxy.
- Updates MCP resource/ability docs (`docs/atomic-builder/mcp/*`) and the `build-composition` static prompt to reference the new ability and priority ordering.

## Test plan
- [x] `php -l` on all changed/added PHP files
- [ ] `tests/phpunit/elementor/modules/mcp/test-reorder-classes-ability.php` via `composer run test` (local `vendor/bin/phpunit` unavailable in this sandbox; needs CI or a `composer install` run)
- [ ] Manual smoke test: call `elementor/reorder-classes` via the MCP proxy with `moves` and `order` inputs against a kit with several global classes, confirm generated CSS priority updates

## 1. Problem & Context

Agents need to reorder global classes by priority to resolve CSS cascade conflicts. Currently they can only read classes or manage definitions, not adjust precedence—a critical gap for handling specificity issues.

## 2. What Changed (Where)

| File | Change |
|------|--------|
| `reorder-classes-ability.php` | New 292-line ability: sequential moves or explicit order input, validates IDs, persists via repository |
| `global-classes-resource-ability.php` | Refactored to return structured `{ priority_description, classes[] }` instead of flat map; added constructor injection for testing |
| `mcp/module.php` | Registered new ability and added to allowed tools list |
| `mcp-proxy-rest-api.php` | Added tool instantiation for reorder-classes |
| Docs & tests | Added ability docs, tests for both abilities, updated resource schema docs |

## 3. How It Works

Entry: `reorder-classes` accepts either `moves` (1–50 sequential repositions with before/after/start/end semantics) or `order` (complete/partial priority list). Validation rejects unknown IDs, self-moves, or both inputs together. Changes only persist if final order differs from current. Omitted IDs in partial `order` append automatically and report back in `appended_ids`. Repository writes via `apply_changes()` with order-only flag, triggering cache invalidation.

## 4. Risks

Max-moves cap (50) prevents DOS but may frustrate large reorderings—acceptable since sequential moves are rare at scale. Partial `order` auto-append is implicit; document clearly to avoid confusion. Test coverage is solid; constructor injection in `Global_Classes_Resource_Ability` enables mocking.


docs/atomic-builder/global-classes/api.md
```diff
@@ -70,6 +70,22 @@ Responses: `204` success; `400` `DUPLICATED_LABEL` or `global_classes_limit_exce
 
 Use `elementor://global-classes` for discovery; use **labels** in `build-composition`.
 
+### MCP `reorder-classes`
+
+The first class in the global order has the highest priority. When classes applied to the same element declare the same property, the earlier class overrides the later one.
+
+Use relative moves for targeted changes:
+
+```json
```
IshayMaya: do we need all this on top of `reorder-classes.md` ?

modules/mcp/abilities/global-classes-resource-ability.php
```diff
@@ -11,6 +11,7 @@
 
 class Global_Classes_Resource_Ability extends Abstract_Ability {
 	const URI = 'elementor://global-classes';
+	const PRIORITY_DESCRIPTION = 'Classes are ordered from highest to lowest priority. When classes on the same element set the same CSS property, the earlier class overrides the later one.';
```
IshayMaya: this is user facing no? should it be translated?

modules/mcp/abilities/global-classes-resource-ability.php
```diff
@@ -38,8 +39,17 @@ protected function get_definition(): Ability_Definition {
...
-		return wp_json_encode( (object) $classes_payload );
+		$classes = [];
+		foreach ( Global_Classes_Repository::make( $kit )->all_labels() as $id => $label ) {
+			$classes[] = [
+				'id' => $id,
+				'label' => $label,
+			];
+		}
+
+		return wp_json_encode( [
```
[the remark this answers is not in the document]
IshayMaya: @ronros-elementor  no consumers for this yet right?

modules/mcp/abilities/reorder-classes-ability.php
```diff
@@ -0,0 +1,291 @@
...
+						'items' => [ 'type' => 'string' ],
+					],
+					'appended_ids' => [
+						'type' => 'array',
+						'items' => [ 'type' => 'string' ],
+					],
+					'moves' => [
+						'type' => 'array',
+						'items' => [
+							'type' => 'object',
```
IshayMaya, line 50: maybe we should calrify?
```suggestion
							'type' => 'object',
							'description' => 'Applied moves in order. Each from/to is relative to the array state when that move ran, not the original input order.',
```

modules/mcp/abilities/global-classes-resource-ability.php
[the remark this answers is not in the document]
ronros-elementor: no mutual ability on the web mcp, so no need to adjust to any prior existing structure

IshayMaya (dismissed):

IshayMaya (approved):

PR #36778, opened 2026-08-05
Internal: Document-root validity + component-instance access [ED-24967] (2/4)

ronros-elementor:
## Summary
Enforce document-root validity (reject widgets as direct document children), allow composition against a null document during component create, and centralize insufficient-permissions errors for component instance placement.

## Stack
- **Base:** `feat/ED-24967-1-composition-compiler` (1/4)
- **This PR:** 2/4 in the ED-24967 stack
- **Follow-ups:** manage-component CRUD (3/4), overridable props (4/4)

## Jira
[ED-24967](https://elementor.atlassian.net/browse/ED-24967)

## Test plan
- [ ] PHPUnit: component-instance / element-config / build-composition / manage-elements coverage for the new gates
- [ ] Confirm widget-as-document-root is rejected; component create with null document still works

## 1. Problem & Context
ED-24967: Prevents invalid document trees by forbidding widgets as direct root children and enforces license-based access control for component instance placement.

## 2. What Changed (Where)

| File | Change |
| :--- | :--- |
| `document-mutator.php` | Added root-level child validity checks; restricted widgets at document root. |
| `component-instance-applier.php` | Integrated `Components_Access_Controller` for create/update/add permissions. |
| `element-config-applier.php` | Removed mandatory document context requirement for component entries. |
| `composition-compiler.php` | Implemented auto-wrapping of root widgets into `e-div-block` containers. |
| `insufficient-permissions-error.php` | New utility for standardized permission error responses. |
| `tests/...` | Added coverage for root wrapping, permission tiers, and invalid reparenting. |
| `build-composition.md` | Updated LLM instructions for component usage and capability checks. |

## 3. How It Works
`Composition_Compiler` now intercepts the DOM; if widgets are found at the root, it wraps them in an `e-div-block` to maintain editor compatibility. `Component_Instance_Applier` calls `get_access_error` to verify the user's license tier against the action (create/update/add) before placing components. `Document_Mutator` acts as the final guard, returning a `WP_Error` if any operation attempts to place a widget directly under the document root.

### DIFF
--- a/core/utils/document/document-mutator.php
+++ b/core/utils/document/document-mutator.php
@@ -14,6 +14,8 @@ class Document_Mutator {
 
 	const DOCUMENT_ROOT = 'document';
 
+	const WIDGET_EL_TYPE = 'widget';
+
 	/** @var \Elementor\Elements_Manager */
 	private $element_manager;
 
@@ -91,7 +93,13 @@ class Document_Mutator {
 			$element['id'] = $this->generate_id();
 		}
 
-		if ( 'document' === $parent_id ) {
+		if ( self::DOCUMENT_ROOT === $parent_id ) {
+			$root_child_error = $this->check_document_root_child_allowed( $element );
+
+			if ( is_wp_error( $root_child_error ) ) {
+				return $root_child_error;
+			}
+
 			return $this->splice_into( $tree, $index, $element );
 		}
 
@@ -283,7 +291,7 @@ class Document_Mutator {
 	private function insert_into_tree( array $tree, string $parent_id, ?int $index, array $element ) {
 		foreach ( $tree as $i => $node ) {
 			if ( isset( $node['id'] ) && $node['id'] === $parent_id ) {
-				if ( isset( $node['elType'] ) && 'widget' === $node['elType'] ) {
+				if ( isset( $node['elType'] ) && self::WIDGET_EL_TYPE === $node['elType'] ) {
 					return new \WP_Error(
 						'elementor_invalid_parent',
 						__( 'Cannot insert into a widget element.', 'elementor' ),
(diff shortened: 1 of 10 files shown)


core/utils/document/document-mutator.php
```diff
@@ -341,6 +349,28 @@ private function remove_from_tree( array $tree, string $id ) {
 		return false;
 	}
 
+	/**
+	 * Mirrors the editor's `Document.isValidChild`: only element types may sit at the document
+	 * root, so widgets inserted here would produce a tree the editor itself refuses to create.
+	 *
+	 * @return null|\WP_Error
+	 */
+	private function check_document_root_child_allowed( array $element ): ?\WP_Error {
```
Omerisra6, line 358: Maybe instead we could auto-wrap it with a div-block? (just because a build composition tool call is a long operation for the agent and its better to allow with a warning)

modules/mcp/abilities/appliers/component-instance-applier.php
```diff
@@ -107,30 +114,11 @@ private function build_envelope( string $config_id, array $shorthand, Document $
 			return null;
 		}
 
-		$component = $this->repository->get( $component_id, false );
-
+		$component = $this->load_valid_component( $config_id, $component_id, $document, $errors );
 		if ( ! $component ) {
-			$errors[] = sprintf( '[%s] Component %d not found.', $config_id, $component_id );
```
Omerisra6, line 113: why don't we need these checks anymore?

ronros-elementor: they are already being added in the load_valid_component function itself, that was actually a logic duplication

core/utils/document/document-mutator.php
ronros-elementor: added now the auto wrap + warnings to the composition-compiler, but will leave it here just as a safe guard if for any reason the execution got here with invalid direct children as a last line of defense

PR #36780, opened 2026-08-05
Internal: manage-component overridable props [ED-24967] (4/4)

ronros-elementor:
## Summary
Wire `Overridable_Props_Builder` into manage-component create/update, document expose-further behavior, and cover it in a dedicated PHPUnit file.

## Stack
- **Base:** `feat/ED-24967-3-manage-component-crud` (3/4)
- **This PR:** 4/4 (top of the ED-24967 stack)
- **Stack base:** `feature/ED-24967-stack` → merge upward through 1–4

## Jira
[ED-24967](https://elementor.atlassian.net/browse/ED-24967)

## Test plan
- [ ] PHPUnit: `test-manage-component-overridable-props.php`
- [ ] Create/update a component with overridable props and verify expose-further documentation path

## 1. Problem & Context

Adds overridable props support to the manage-component ability, allowing callers to expose configurable settings on component instances. Part 4/4 of ED-24967 — completes the "expose-further" feature that chains nested component overrides through arbitrary depth.

## 2. What Changed (Where)

- **manage-component-ability.php**: Integrated `apply_overridable_props()` into create/update flows; split update into two modes (with/without xml_structure).
- **composition-compiler.php**: Refined widget detection to exclude component instances; added `COMPONENT_PARENT_ID` constant.
- **overridable-props-builder.php** (new): 568-line builder validates definitions, maps to element tree, handles nested expose-further and group generation.
- **test-manage-component-overridable-props.php** (new): 777-line test suite covering raw widgets, nested instances, multi-hop chaining, error cases.
- **manage-component.md**: Documentation for `overridable_props` parameter and two update modes with examples.

## 3. How It Works

Entry point: `apply_overridable_props()` called during create or update. Builder iterates prop definitions, finds target elements by id or configuration-id, wraps settings in `Overridable_Prop_Type` envelopes, and stamps `origin_value`. For nested e-components, extracts inner component id, resolves exposed override keys, threads `origin_prop_fields` through multi-hop chains, and upserts `component_instance.overrides`. Groups auto-generated with slug-safe ids (never "default" to avoid Sortable collisions). Returns persisted metadata keyed by override key; element tree mutations only applied on success (atomic).

## 4. Risks

- **Multi-hop resolution complexity**: Three levels of indirection (wrapper → inner component → raw widget). Mitigated by comprehensive test coverage including multi-hop threading (test_update__expose_further_threads_origin_prop_fields_through_multi_hop).
- **Element tree mutation on partial failure**: Builder doesn't revert on second/third prop error. Mitigated by collecting all errors upfront and rejecting entire batch if any invalid.
- **Group ID collision logic**: Relies on magic prefix + timestamp + random suffix; clock skew or concurrent requests could theoretically collide. Low risk given collision detection tests pass.


modules/mcp/abilities/utils/overridable-props-builder.php
```diff
@@ -0,0 +1,566 @@
...
+use Elementor\Modules\Components\PropTypes\Overrides_Prop_Type;
+use Elementor\Modules\Components\Utils\Parsing_Utils;
+use Elementor\Modules\Components\Widgets\Component_Instance;
+use WP_Http;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
```
Omerisra6, line 19: I think you can reduce the comments here

modules/mcp/abilities/utils/overridable-props-builder.php
```diff
@@ -0,0 +1,566 @@
...
+	public static function make( ?Components_Repository $repository = null ): self {
+		return new self( $repository );
+	}
+
+	/**
+	 * @param array $elements       Elements tree containing the referenced target ids. Mutated in place:
+	 *                              each referenced element's setting is rewritten into an `overridable` envelope
+	 *                              (raw widget target) or spliced into `component_instance.value.overrides`
+	 *                              (`<e-component>` target).
+	 * @param array $friendly_props Friendly overridable-props input, keyed by override key.
```
Omerisra6: lets find a better name for that

modules/mcp/abilities/manage-component-ability.php
```diff
@@ -567,6 +615,10 @@ private function get_input_schema(): array {
 					'default' => (object) [],
 					'description' => 'Same shape as elementor/build-composition interactions. Only used with xml_structure.',
 				],
+				'overridable_props' => [
```
Omerisra6, line 650: The friendly input is keyed by override_key from the caller (and sanitize_key( $override_key ) !== $override_key is a hard failure). Only groupId and origin_value are actually server-generated. The web-MCP tool, in contrast, does generate the override key itself (generateUniqueId('prop')). Either fix the description or make the builder generate override keys and treat the friendly map's keys as caller-hints only. The current wording will mislead LLMs.

check if valid?

modules/mcp/abilities/utils/overridable-props-builder.php
```diff
@@ -0,0 +1,566 @@
...
+			'prop_key' => $prop_key,
+			'label' => $label,
+			'group_label' => $group_label,
+		];
+	}
+
+	/**
+	 * @return array|\WP_Error
+	 */
+	private function build_raw_widget_prop(
```
Omerisra6, line 202: Duplicate (target, prop_key) is silently overwritten. In build_raw_widget_prop the last entry wins on $element['settings'][$prop_key] = ..., but both entries stay in props and in their groups. The editor action createComponentOverridableProp explicitly de-duplicates via removePropsFromState. The web-MCP's enrichOverridableProps doesn't dedupe either, so this matches the buggier of the two — but if the intent is "same as editor", editor semantics is dedupe. Consider matching that.

Check if valid?

modules/mcp/abilities/utils/overridable-props-builder.php
```diff
@@ -0,0 +1,566 @@
...
+				'invalid_overridable_prop_definition',
+				sprintf(
+					'[%s] target "%s" is an <e-component> instance but has no valid component_instance settings. Set component_id on the instance (via element_config) before exposing an override through it.',
+					$override_key,
+					(string) ( $element['id'] ?? '' )
+				)
+			);
+		}
+
+		$inner_component = $this->get_repository()->get( $inner_component_id, false );
```
Omerisra6, line 275: Nested-component lookup by unpublished state. $this->get_repository()->get( $inner_component_id, false ) — the second arg matters. If it filters to published-only, exposing further on an unpublished nested component will fail; the editor's expose-further UI works in-memory via componentsSelectors, which sees both. Worth double-checking that this doesn't drop the unpublished case.

Check if valid?

ronros-elementor: Checked this — `false` here doesn't filter to published-only, it only disables autosave inclusion:

```php
public function get( $id, bool $include_autosave = true ) {
    $doc = $include_autosave
        ? Plugin::$instance->documents->get_doc_or_auto_save( $id, get_current_user_id() )
        : Plugin::$instance->documents->get( $id );
    ...
}
```

`Plugin::$instance->documents->get( $id )` resolves the main document regardless of its post status, so a `draft` inner component is still found. The intentional choice is to read the inner component's committed/published props snapshot rather than another user's in-flight autosave — that matches "expose-further" being a server-side, canonical operation rather than an editor in-memory preview. So the unpublished-nested-component case works; leaving this as-is.

Omerisra6 (approved):

PR #36782, opened 2026-08-05
Internal: Align MCP read tools with raw-CSS style format [ED-25195]

Omerisra6:
## Summary

Close the MCP style round-trip gap for **local element styles**: `get-page-structure` now emits the same raw-CSS shape the write tools already accept (after #36789).

Jira: [ED-25195](https://elementor.atlassian.net/browse/ED-25195)

## Read path

- **`Style_Variants_To_Css`** — serializes atomic variants to a raw CSS string (`&:hover/&:focus/&:active` + `@media(--<bp>)`).
- **`Local_Style_Serializer`** — emits `{ __style_id, css }` for `elementor/get-page-structure` when `include_content=true`.
- Ability descriptions updated for the new shape.

## Breaking API changes

- `get-page-structure` `styles`: flat `{ __style_id, __variants, __custom_css, <prop>: <val> }` → `{ __style_id, css }`.

No external stability guarantees on the MCP schema yet — these are internal LLM tools.

## Out of scope

- Write-path changes already landed in #36789.
- `elementor://global-classes` stays label-only — attaching CSS would require `Global_Classes_Repository::all()`, which is too expensive for this resource.
- Other MCP tools (variables, widget schemas, list-*) do not emit style payloads.

## Tests

- New: `Style_Variants_To_Css` unit tests.
- Migrated `get-structure` `include_content` assertions to the `{ __style_id, css }` shape.

## Test plan

- [ ] `composer run test -- --filter 'Style_Variants_To_Css|Get_Structure_Ability|Manage_Elements_Ability'`
- [ ] Manual: `elementor/get-page-structure` with `include_content=true` returns `styles.css` with desktop + `&:hover` + `@media(--mobile)` when present
- [ ] Manual: paste `styles.css` into `manage-elements.update` with `replace` mode = no-op round-trip

[ED-25195]: https://elementor.atlassian.net/browse/ED-25195?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## 1. Problem & Context

MCP read tools (get-page-structure) were returning styles as flattened CSS properties; this changes them to raw CSS format (`{ __style_id, css }`) matching the shape that write-side tools (manage-elements, build-composition) accept, enabling true round-trip serialization without consumer transformation.

## 2. What Changed (Where)

- `local-style-serializer.php`: Simplified from 121→38 lines; now delegates CSS rendering to domain objects instead of flattening props
- `local-style.php` (new): Renders variants into `@media(--breakpoint)` + `&:state` format; groups by breakpoint
- `local-style-variant.php` (new): Handles individual variant→CSS fragment conversion with pseudo-state wrapping
- `get-structure-ability.php`: Updated ability descriptions to document raw CSS format and `@media(--breakpoint)` syntax
- Tests: New `test-local-style.php` + updated `test-get-structure-ability.php` assertions to verify CSS string output instead of flat properties

## 3. How It Works

Serializer → `Local_Style::from_styles_map()` → groups variants by breakpoint → renders desktop top-level, wraps non-desktop in `@media(--breakpoint)`, wraps pseudo-states in `&:selector { }`, concatenates all fragments into single CSS string. Variant rendering handles both structured props (via `Style_Props_To_Css`) and decoded custom CSS inline.

## 4. Risks

Minor: consumers expecting flat property structure will break (intentional API change). CSS fragment ordering assumptions if variant iteration order changes (mitigated by explicit grouping logic).


modules/mcp/abilities/global-classes-resource-ability.php
```diff
@@ -38,8 +39,25 @@ protected function get_definition(): Ability_Definition {
 	public function execute( $input = [] ) {
 		$kit = Plugin::$instance->kits_manager->get_active_kit();
 
-		$classes_payload = Global_Classes_Repository::make( $kit )->all_labels();
+		$classes = Global_Classes_Repository::make( $kit )->all();
+		$items = $classes->get_items()->all();
+		$order = $classes->get_order()->all();
```
Omerisra6: all shouldn't be used, its a very expensive operation. for now return only labels as before

modules/atomic-widgets/styles/style-variants-to-css.php
Omerisra6, line 1: cant we reuse the `modules/mcp/abilities/utils/style-variants-merger.php`?

eavichay-elementor (changes requested):
The direction (raw CSS) is better than the previous flat-prop-map, but it's inconsistent with the existing LLM/json dialect pipeline. The right fix is to apply similar to Plain_Llm_Schema_Converter, but for style variant props — same as settings — and emit a structured variants array instead of a
  CSS string.

We already have a solid pipeline for converting PropTypes outside to consumers - why not reuse the same thing for the StylePropType?

eavichay-elementor (changes requested):
inline comments

eavichay-elementor (approved):
oh well

PR #36804, opened 2026-08-06
Fix: Batch component document requests without dropping a caller [ED-25172]

Ntnelbaba:
## PR Checklist

## PR Type

- [x] Bug fix

## Summary

* Fix: Component documents are loaded in a single batched request on editor open, without dropping any of the concurrent callers

## Description

Re-lands the batching from #36772 (reverted in #36796) together with the fix for the regression it caused, per the analysis on [ED-25172](https://elementor.atlassian.net/browse/ED-25172?focusedCommentId=356640).

**Why #36772 regressed.** `getComponentDocumentData` issued its request without `immediately`, so it went through `addRequest()`'s batch queue, which stores one request per `unique_id`. Two callers request the same component document inside the 500ms debounce window on editor open — `loadComponentsAssets` and `componentInstanceTransformer` — so the second overwrote the first and the first `jQuery.Deferred` was never resolved and never rejected. `ajax.load` in `@elementor/editor-v1-adapters` wraps that deferred in a promise, so the losing caller hung forever and its `catch` never ran. `loadComponentsAssets` lost the race, so `embeddedDocumentsManager.setDocument()` and `loadComponentsOverridableProps()` never ran: components rendered without their local styles and the instance editing panel showed no override controls. Both call sites are `void`-ed, so nothing was logged.

**Two layers, both keeping the batching:**

1. `core/common/modules/ajax/assets/js/ajax.js` — `addRequest()` no longer replaces a pending non-immediate request silently. An identical request (same `unique_id` **and** same data) chains onto the pending deferred, so both callers settle from one HTTP request. A request that cannot share the batch entry rejects the deferred it displaces with `Request replaced` instead of leaving it pending forever.
2. `packages/packages/core/editor-components/src/utils/component-document-data.ts` — restores the batched `get_document_config` request and dedupes concurrent callers per component id, so the component path no longer depends on ajax-layer internals. The `document-${ id }` cache key is kept, which is what keeps `elementor.documents.invalidateCache()` and the save flow invalidating component data. The map holds only in-flight promises and is cleared on settle.

Caller audit for layer 1: nothing depends on the old clobbering. `load-templates.ts` and the embedded-documents manager use per-id keys; `documents/component.js` and font enqueueing pass `immediately: true`; the finder and template-search paths reuse a `unique_id` with different data, and those previously left the displaced deferred pending anyway (the finder passes no `error` handler, so a rejection is a no-op there).

Out of scope, tracked on the ticket as follow-ups: 429 retry with backoff honouring `Retry-After` in `ajax.js`, a purpose-built `get_documents_config` batch endpoint taking `ids[]`, and restoring `$post` / current document / `editor.post_id` after `ajax_get_document_config` in `core/documents-manager.php`.

## Test instructions

* On a page with several Component instances (nested included), open the editor with the Network tab filtered to `admin-ajax.php`: the `get_document_config` actions arrive in one batched `elementor_ajax` request per nesting level instead of one HTTP request per document.
* The components render with their local styles, and selecting an instance shows its exposed-prop controls in the instance editing panel — the two things that broke in #36772.

## Quality assurance

### Unit tests

* `tests/jest/unit/core/common/modules/ajax/assets/js/ajax.test.js` — runs against the real `ajax.js` with a focused local `jQuery.Deferred` test double (no npm `jquery` dependency). Three of its four cases fail on `main`: both callers of an identical batched request settle, both reject when the shared entry fails, and a displaced request with different data rejects. The fourth (repeated request served from cache) is a control that passes either way.
* `packages/.../editor-components/src/utils/__tests__/component-document-data.test.ts` — request shape on the v1 cache key, `null` on failure, one `ajax.load` for concurrent callers of the same id, and a fresh request once the previous one settled.
* Full suites green: `test:jest` 837 passed (108 suites), `test:packages` 3669 passed, `tsc` and ESLint (root + packages) clean.

### Pro E2E against this Core

Ran Elementor Pro's component Playwright tests against this branch's Core build in WP Playground (Core + Pro mounted, `e_atomic_elements` active, advanced license seeded), rebuilding Core between configurations:

| Core build | `atomic-components.test.ts` › should load local styles used within a component on page |
| --- | --- |
| This PR | **passed** (reproduced twice) |
| #36772 as merged (batching only) | **failed** — canvas `h2` computed `rgba(0, 0, 0, 0)` instead of `rgb(255, 0, 0)`, the exact symptom reported on the ticket |

Whole `atomic-components.test.ts` file on this branch: 6 passed, 2 failed, 2 not run (the Pro-gating cases, which need license switching through wp-env and cannot run in Playground). Both failures — `should allow exposing props and overriding at instance and nested component level` and `should show correct placeholder in nested exposed props` — fail identically on unpatched `main` in this environment, each verified with its own baseline rebuild: the instance panel stays on its "No properties yet" empty state even with a 45s expect timeout. They are environment limitations here, not signal about this change, and they exercise the same `loadComponentsAssets` path as the test that does discriminate.

[ED-25172]: https://elementor.atlassian.net/browse/ED-25172?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## 1. Problem & Context

Concurrent batched AJAX requests with identical `unique_id` but different parameters were causing callers to hang indefinitely. ED-25172 fixes this by properly settling displaced requests and deduping component document fetches at the application layer.

## 2. What Changed (Where)

- **ajax.js**: Added pending request detection; when same `unique_id` reused with different cache key, reject displaced deferred, otherwise merge both callers into single response.
- **component-document-data.ts**: Replaced v1 manager dependency with direct ajax.load batching; maintains request Map to dedupe concurrent calls for same document.
- **Test coverage**: Added jquery.js mock with deferred implementation; full test suites for ajax batch collision handling and component document deduping logic.

## 3. How It Works

When `ajax.load()` is called with a `unique_id` already pending:
1. If cache keys match (identical request), attach new caller's deferred to existing request—both resolve/reject together.
2. If cache keys differ (different params), reject the old deferred with "Request replaced" and queue the new one.
3. Component layer adds its own deduping: concurrent calls for same document ID share one Promise via `pendingRequests` Map, cleared on settlement to allow fresh requests.

## 4. Risks

Low. Deferred attachment pattern is standard; Map cleanup via finally() is reliable. Only real edge case is rapid parameter changes on same unique_id—correctly handled by rejection. Test coverage validates both collision scenarios and cache hit behavior.

<div><a href="https://cursor.com/agents/bc-6a97b81f-0ee1-4694-829d-5ce4912305a4?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-6a97b81f-0ee1-4694-829d-5ce4912305a4&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

[ED-25172]: https://elementor.atlassian.net/browse/ED-25172?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## 1. Problem & Context

Concurrent batched AJAX requests with identical `unique_id` but different parameters were causing callers to hang indefinitely. ED-25172 fixes this by properly settling displaced requests and deduping component document fetches at the application layer.

## 2. What Changed (Where)

- **ajax.js**: Added pending request detection; when same `unique_id` reused with different cache key, reject displaced deferred, otherwise merge both callers into single response.
- **component-document-data.ts**: Replaced v1 manager dependency with direct ajax.load batching; maintains request Map to dedupe concurrent calls for same document.
- **Test coverage**: Added jquery.js mock with deferred implementation; full test suites for ajax batch collision handling and component document deduping logic.

## 3. How It Works

When `ajax.load()` is called with a `unique_id` already pending:
1. If cache keys match (identical request), attach new caller's deferred to existing request—both resolve/reject together.
2. If cache keys differ (different params), reject the old deferred with "Request replaced" and queue the new one.
3. Component layer adds its own deduping: concurrent calls for same document ID share one Promise via `pendingRequests` Map, cleared on settlement to allow fresh requests.

## 4. Risks

Low. Deferred attachment pattern is standard; Map cleanup via finally() is reliable. Only real edge case is rapid parameter changes on same unique_id—correctly handled by rejection. Test coverage validates both collision scenarios and cache hit behavior.


package.json
```diff
@@ -134,6 +134,7 @@
     "husky": "^9.1.7",
     "jest": "^29.7.0",
     "jest-environment-jsdom": "^29.7.0",
+    "jquery": "^3.7.1",
```
ronros-elementor: 😢

ronros-elementor (dismissed):

core/common/modules/ajax/assets/js/ajax.js
```diff
@@ -130,6 +130,22 @@ export default class extends elementorModules.Module {
...
+				// deferred its caller would wait forever.
+				if ( this.getCacheKey( pendingRequest.options ) === this.getCacheKey( options ) ) {
+					pendingRequest.options.deferred
+						.done( ( data ) => options.deferred.resolve( data ) )
+						.fail( ( data ) => options.deferred.reject( data ) );
+
+					return options.deferred;
+				}
+
+				pendingRequest.options.deferred.reject( 'Request replaced' );
```
ronros-elementor, line 146: maybe just wrap with the translate util? (e.g. if some consumer of this API shows the error in some modal)

ronros-elementor (approved):

PR #36829, opened 2026-08-09
Internal: Add V3 widget allowlist to MCP schema abilities [ED-25071]

Omerisra6:
## Summary

Layer 1 of the [ED-25071 stacked series](https://github.com/elementor/elementor/pull/36801). Opens a closed V3 allowlist (`nav-menu`, `theme-post-content`, `theme-post-title`, `theme-post-featured-image`, `theme-post-excerpt`, `theme-archive-title`) to `elementor/list-widget-schemas` and `elementor/get-widget-schema` (V3 fallback shape), and permits allowlisted V3 targets in `build-composition` and `manage-elements`.

Same LLM contract as V4; internal V3 shape bridge:
- `classes`: labels → `settings._css_classes` (space-separated string).
- `style`: CSS wrapped as `selector { ... }` into `settings.custom_css` (requires Elementor Pro; otherwise a warning is emitted and the style is skipped).
- `settings`: raw merge (no schema validation, matching V3 fallback behavior in `Element_Config_Applier`).

Also forces V3 widget stack initialization in `Widget_Context_Helper` so `get_config()` surfaces controls for the six allowlisted types via the V3 fallback shape.

## Depends on

- Base: `main`
- Top of stack: [PR4 registry refactor](../pull/) (see other PRs in the ED-25071 chain)

## Test plan

- [ ] PHPUnit: allowlist behavior in `test-list-widget-schemas-ability.php`, `test-get-widget-schema-ability.php`, `test-build-composition-ability.php`, `test-manage-elements-ability.php`
- [ ] Manual: call `elementor/list-widget-schemas` and confirm the six V3 types are present

Made with [Cursor](https://cursor.com)

## 1. Problem & Context
ED-25071: Extends MCP schema abilities to support a closed allowlist of V3 widgets (nav-menu, theme-post-*) with the same V4 contract, enabling LLM interactions with legacy widgets without editor modification. Bridges V4-shape inputs (global class labels, CSS strings) onto V3 settings format.

## 2. What Changed (Where)
- **Class_Applier**: Refactored label validation into `resolve_labels()` method; added V3_Node_Bridge detection for allowlisted widgets to write to `_css_classes`.
- **Style_Applier**: Moved `$node` assignment conditionally; added V3_Node_Bridge detection to write CSS to `custom_css` with Pro check; tracks warnings separately.
[...]

PR #36801, opened 2026-08-06
Internal: Add V3 widget allowlist to MCP schema abilities [ED-25071]

Omerisra6:
## Summary

Opens a closed V3 allowlist to the MCP schema abilities so a curated set of legacy widgets is discoverable, fetchable, and emit-able alongside V4 atomic widgets — with no V3→V4 conversion. Same LLM contract as V4; the V3 shape is bridged internally.

Allowlist (six widgets):
- `nav-menu`
- `theme-post-content`, `theme-post-title`, `theme-post-featured-image`, `theme-post-excerpt`, `theme-archive-title`

### Ability changes
- `elementor/list-widget-schemas` — includes allowlisted V3 types (V3 fallback shape); non-allowlisted V3 widgets remain excluded.
- `elementor/get-widget-schema` — returns V3 fallback (`widget_version: 'v3'`, message, `properties` built from legacy controls) for allowlisted types instead of `elementor_v3_not_supported`.
- `elementor/build-composition` and `elementor/manage-elements` — accept allowlisted V3 nodes as operation targets.

### Internal V3 bridge (same LLM contract as V4)
Introduces `Appliers\V3_Node_Bridge` used by `Class_Applier` and `Style_Applier`:
- `element_config` → raw V3 settings merge (no schema validation, matches existing V3 fallback semantics).
- `classes` (global class labels) → V3 `settings._css_classes` (space-separated string).
- `style` (plain CSS string) → V3 `settings.custom_css` wrapped as `selector { ... }`. Requires Elementor Pro (Custom CSS module); without Pro the style is skipped and a warning is emitted.
- Interactions and V4 nested `&:hover` / `@media (--mobile)` in `style` are NOT translated for V3 in this iteration — V3 style strings should stay flat.

### Registration fix
`Widget_Context_Helper` now triggers `get_stack()` for allowlisted V3 types before reading `get_config()`. V3 widgets expose their controls only after their control stack is initialized; without this, `get_config()` returned no controls in a REST context and the allowlisted widgets were silently filtered out of the LLM-eligible set. Only the six allowlisted types are touched — no effect on other widgets.

### Docs
Updated `list-widget-schemas.md`, `get-widget-schema.md`, and `build-composition.md` to describe the allowlist and V3 mapping.

## Test plan

PHPUnit (added):
- `test-list-widget-schemas-ability.php` — allowlisted V3 included, non-allowlisted V3 excluded, in both full and summary modes.
[...]

PR #36830, opened 2026-08-09
Internal: Map V3 allowlisted widget styles via CSS bridge [ED-25071]

Omerisra6:
## Summary

Layer 2 of the [ED-25071 stacked series](https://github.com/elementor/elementor/pull/36801). Adds the V3 style read/write path so the six allowlisted V3 widgets accept plain-CSS `style` on `build-composition`/`manage-elements` and surface their current styles back through `get-page-structure`.

- `V3_Style_Mapper`: parses LLM CSS declarations, maps them onto legacy V3 style controls via a widget bridge registry, and wraps anything unmapped in `selector { ... }` for the Custom CSS module.
- `V3_Style_Serializer`: inverse path — reconstructs a CSS string from current V3 settings for `get-page-structure` output.
- `V3_Widget_Bridge_Registry` / `V3_Style_Settings_Index` / `V3_Value_Resolvers`: per-widget mapping tables (base + responsive + pseudo-state) with value resolvers for typography, borders, shadows, dimensions, etc.
- `V3_Json_Schema_Builder`: exposes the V3 shape via list/get-widget-schemas.
- `Style_Applier` / `Element_Config_Applier`: dispatch V3 style writes through the mapper; V4 atomic path unchanged.
- `get-structure-ability`: emits a `style` field for allowlisted V3 nodes.

## Depends on

- Base: #36829 (PR1 — V3 allowlist)

## Test plan

- [ ] PHPUnit: `test-v3-style-mapper.php`, `test-v3-style-serializer.php`, `test-v3-value-resolvers.php`, `test-v3-non-style-allowlist.php`, `test-v3-controls-metadata.php`, `test-v3-json-schema-builder.php`, `test-style-applier.php`, `test-element-config-applier.php`
- [ ] Manual: call `elementor/manage-elements` with a plain-CSS `style` payload on `theme-post-title` and observe the mapped V3 settings; verify `get-page-structure` round-trips the same CSS.

## 1. Problem & Context

V3 allowlisted widgets (nav-menu, theme-post-*) now accept CSS via the same `style` input as V4 widgets, mapping declarative CSS to legacy settings (typography, border, color, etc.) with fallback to custom_css. Closes ED-25071.

## 2. What Changed (Where)

| File | Change |
|------|--------|
| `v3/v3-widget-bridge-registry.php` | New: per-widget CSS→setting override registry (6 allowlisted types, 335 LOC) |
| `v3/v3-style-mapper.php` | New: CSS parser mapping rules to V3 settings, handles responsive breakpoints (338 LOC) |
| `v3/v3-style-serializer.php` | New: reverse mapping—settings back to CSS for `get-structure` (384 LOC) |
| `v3/v3-value-resolvers.php` | New: pure functions resolving CSS values to V3 control shapes (355 LOC) |
| `v3/v3-style-settings-index.php` | New: generic CSS→setting reverse lookup from control selectors (150 LOC) |
| `v3/v3-non-style-allowlist.php` | New: per-widget non-style settings allowlist validator (66 LOC) |
| `appliers/v3-node-bridge.php` | Extended: `clear_style_settings()` wipes mapped settings on replace mode; introspection helpers for collecting/expanding override keys |
| `appliers/style-applier.php` | `apply()` signature: +`$style_apply_mode`, +`$widget_configs` params; new `apply_v3_style()` delegates to V3_Style_Mapper |
| `appliers/element-config-applier.php` | V3 nodes now filter settings through V3_Non_Style_Allowlist before merge |
| `get-structure-ability.php` | New `populate_v3_content()` serializes V3 styles + filters settings when `include_content=true`; dropped `version` field |
| `build-composition-ability.php`, `manage-elements-ability.php` | Pass `$widget_configs` to `style_applier->apply()` |
| Tests | 10 new test files covering mappers, serializers, resolvers, allowlist filtering; updated build/manage/get-structure tests |
| Docs | Updated `.md` guides—removed v3-specific caveats, unified contract language |

## 3. How It Works

**Entry:** V3 node + CSS string → Style_Applier detects `V3_Node_Bridge::is_v3_node()` → delegates to `apply_v3_style()`.

**Happy path:**
1. V3_Style_Mapper parses CSS, splits by breakpoint, matches selectors to pseudo-states
2. Per-widget registry overrides (typography_prefix, border_prefix, box_shadow_prefix, setting+resolver) apply first
3. Fallback to generic V3_Style_Settings_Index (reverse lookup from control selectors)
4. V3_Value_Resolvers converts CSS values to V3 shape (e.g., `2rem` → `{unit: 'rem', size: 2.0}`)
5. Responsive suffixes (`_tablet`, `_mobile`) written if control exists
6. Unmapped CSS accumulates, merged into custom_css (Pro only) or warning

**Edge cases:**
- Replace mode clears all known style settings + custom_css before applying new styles
- Empty CSS in replace mode wipes styles cleanly
- Unmapped CSS truncated (200 chars) in warnings
- Generic index excludes overridden keys; only unique 1:1 selector→setting mappings included

**Serialization (reverse):** V3 settings → CSS via V3_Style_Serializer for `get-structure` with `include_content=true`. Re-wraps custom_css if needed, groups by breakpoint/pseudo-state.

## 4. Risks

**Schema stability:** Registry is hand-curated per widget. Missing or incorrect CSS→setting mappings silently drop to unmapped. *Mitigation:* registry tests cover all 6 types; mapper tests verify round-trip idempotence.

**Responsive control existence:** Mapper checks control existence before applying breakpoint suffixes. Silently skips if control missing. *Mitigation:* intentional—avoids orphaned settings; test covers both cases.

**Pro dependency:** Unmapped CSS requires Pro custom_css module. Non-Pro warns but doesn't fail. *Mitigation:* explicit warning; tests mock Utils::has_pro() to verify both paths.

**Backward compat:** V3 widget get-structure no longer returns `version` field. *Mitigation:* docs updated; clients already ignore unknown fields; test corrected.

### DIFF
--- a/modules/mcp/abilities/appliers/style-applier.php
+++ b/modules/mcp/abilities/appliers/style-applier.php
@@ -60,7 +60,7 @@ class Style_Applier {
 			$node = &$config_id_index[ $config_id ];
 
 			if ( V3_Node_Bridge::is_v3_node( $node ) ) {
-				$v3_warnings = $this->apply_v3_style( $node, $css_string, $widget_configs );
+				$v3_warnings = $this->apply_v3_style( $node, $css_string, $style_apply_mode, $widget_configs );
 				foreach ( $v3_warnings as $warning ) {
 					$warnings[] = sprintf( '[%s] %s', $config_id, $warning );
 				}
@@ -132,7 +132,7 @@ class Style_Applier {
 	 * @param array<string, array> $widget_configs
 	 * @return string[] Warnings (without config-id prefix).
 	 */
-	private function apply_v3_style( array &$node, string $css_string, array $widget_configs = [] ): array {
+	private function apply_v3_style( array &$node, string $css_string, string $style_apply_mode = 'patch', array $widget_configs = [] ): array {
 		$warnings = [];
 		$widget_type = $node['widgetType'] ?? '';
 		$widget_config = [];
@@ -141,6 +141,17 @@ class Style_Applier {
 			$widget_config = $widget_configs[ $widget_type ]
 				?? Widget_Context_Helper::get_widget_config( $widget_type )
 				?? [];
+		}
+
+		$is_empty_css = '' === trim( $css_string );
+		$is_replace = 'replace' === $style_apply_mode;
+
+		if ( $is_replace ) {
+			V3_Node_Bridge::clear_style_settings( $node, (string) $widget_type, $widget_config );
+		}
+
+		if ( $is_empty_css ) {
+			return $warnings;
 		}
 
 		$mapper = new V3_Style_Mapper( $this->css_converter, $this->get_active_breakpoints() );
@@ -155,14 +166,36 @@ class Style_Applier {
 		}
 
 		$unmapped = $result['unmapped_css'] ?? '';
-		$warning = V3_Node_Bridge::apply_custom_css( $node, $unmapped );
-		if ( null !== $warning ) {
-			$warnings[] = $warning;
-		} elseif ( '' !== trim( $unmapped ) ) {
-			$warnings[] = __( 'Some CSS could not be mapped to V3 settings and was written to custom_css.', 'elementor' );
+		$pro_warning = V3_Node_Bridge::apply_custom_css( $node, $unmapped );
+		if ( null !== $pro_warning ) {
+			$warnings[] = $pro_warning;
+		}
+
+		if ( '' !== trim( $unmapped ) ) {
+			$snippet = self::truncate_css_snippet( $unmapped );
+			$warnings[] = null !== $pro_warning
+				? sprintf(
+					/* translators: %s: CSS snippet that could not be mapped */
+					__( 'Some CSS could not be mapped to V3 settings and was dropped: %s', 'elementor' ),
+					$snippet
+				)
+				: sprintf(
+					/* translators: %s: CSS snippet that could not be mapped */
+					__( 'Some CSS could not be mapped to V3 settings and was written to custom_css: %s', 'elementor' ),
+					$snippet
+				);
 		}
 
 		return $warnings;
+	}
+
+	private static function truncate_css_snippet( string $css, int $max_length = 200 ): string {
+		$css = trim( preg_replace( '/\s+/', ' ', $css ) ?? $css );
+		if ( strlen( $css ) <= $max_length ) {
+			return $css;
+		}
+
+		return substr( $css, 0, $max_length - 3 ) . '...';
 	}
 
 	private function get_active_breakpoints(): array {
--- a/modules/mcp/abilities/appliers/style-applier.php
+++ b/modules/mcp/abilities/appliers/style-applier.php
@@ -3,8 +3,10 @@
 namespace Elementor\Modules\Mcp\Abilities\Appliers;
 
 use Elementor\Modules\AtomicWidgets\CssConverter\Css_Converter;
+use Elementor\Modules\Mcp\Abilities\Appliers\V3\V3_Style_Mapper;
 use Elementor\Modules\Mcp\Abilities\Utils\Bulk_Operations_Result;
 use Elementor\Modules\Mcp\Abilities\Utils\Style_Variants_Merger;
+use Elementor\Modules\Mcp\Abilities\Utils\Widget_Context_Helper;
 use Elementor\Plugin;
 
 if ( ! defined( 'ABSPATH' ) ) {
@@ -29,9 +31,11 @@ class Style_Applier {
 	/**
 	 * @param array<string, array&> $config_id_index Index of subtree refs.
 	 * @param array<string, string> $styles          Per-config-id CSS strings.
+	 * @param string                $style_apply_mode `patch` or `replace`.
+	 * @param array<string, array>  $widget_configs  Optional widget_type => config map (used for V3 mapping).
 	 * @return array{error: \WP_Error|null, warnings: string[]}
 	 */
-	public function apply( array $config_id_index, array $styles, string $style_apply_mode = 'patch' ): array {
+	public function apply( array $config_id_index, array $styles, string $style_apply_mode = 'patch', array $widget_configs = [] ): array {
 		if ( empty( $styles ) ) {
 			return [
 				'error'    => null,
@@ -56,8 +60,8 @@ class Style_Applier {
 			$node = &$config_id_index[ $config_id ];
 
 			if ( V3_Node_Bridge::is_v3_node( $node ) ) {
-				$warning = V3_Node_Bridge::apply_custom_css( $node, $css_string );
-				if ( null !== $warning ) {
+				$v3_warnings = $this->apply_v3_style( $node, $css_string, $widget_configs );
+				foreach ( $v3_warnings as $warning ) {
 					$warnings[] = sprintf( '[%s] %s', $config_id, $warning );
 				}
 				unset( $node );
@@ -122,6 +126,43 @@ class Style_Applier {
 			) : null,
 			'warnings' => $warnings,
 		];
+	}
+
+	/**
+	 * @param array<string, array> $widget_configs
+	 * @return string[] Warnings (without config-id prefix).
+	 */
+	private function apply_v3_style( array &$node, string $css_string, array $widget_configs = [] ): array {
+		$warnings = [];
+		$widget_type = $node['widgetType'] ?? '';
+		$widget_config = [];
+
+		if ( is_string( $widget_type ) && '' !== $widget_type ) {
+			$widget_config = $widget_configs[ $widget_type ]
+				?? Widget_Context_Helper::get_widget_config( $widget_type )
+				?? [];
+		}
+
+		$mapper = new V3_Style_Mapper( $this->css_converter, $this->get_active_breakpoints() );
+		$result = $mapper->apply( $css_string, (string) $widget_type, $widget_config );
+
+		foreach ( $result['warnings'] as $warning ) {
+			$warnings[] = $warning;
+		}
+
+		if ( ! empty( $result['settings_patch'] ) ) {
+			$node['settings'] = array_merge( $node['settings'] ?? [], $result['settings_patch'] );
+		}
+
+		$unmapped = $result['unmapped_css'] ?? '';
+		$warning = V3_Node_Bridge::apply_custom_css( $node, $unmapped );
+		if ( null !== $warning ) {
+			$warnings[] = $warning;
+		} elseif ( '' !== trim( $unmapped ) ) {
+			$warnings[] = __( 'Some CSS could not be mapped to V3 settings and was written to custom_css.', 'elementor' );
+		}
+
+		return $warnings;
 	}
 
 	private function get_active_breakpoints(): array {
--- a/modules/mcp/abilities/appliers/style-applier.php
+++ b/modules/mcp/abilities/appliers/style-applier.php
@@ -129,6 +129,9 @@ class Style_Applier {
 	}
 
 	/**
+	 * @param array                $node
+	 * @param string               $css_string
+	 * @param string               $style_apply_mode
 	 * @param array<string, array> $widget_configs
 	 * @return string[] Warnings (without config-id prefix).
 	 */
--- a/modules/mcp/abilities/appliers/v3/v3-style-mapper.php
+++ b/modules/mcp/abilities/appliers/v3/v3-style-mapper.php
@@ -188,13 +188,10 @@ class V3_Style_Mapper {
 		$key = $setting;
 		if ( ! empty( $override['responsive'] ) && self::DESKTOP_BREAKPOINT !== $breakpoint ) {
 			$suffixed = $setting . '_' . $breakpoint;
-			if ( $this->control_exists( $widget_config, $suffixed ) ) {
-				$key = $suffixed;
-			} elseif ( $this->control_exists( $widget_config, $setting ) ) {
+			if ( ! $this->control_exists( $widget_config, $suffixed ) ) {
 				return null;
-			} else {
-				$key = $suffixed;
 			}
+			$key = $suffixed;
 		}
 
 		return [ $key => $resolved ];
@@ -213,13 +210,10 @@ class V3_Style_Mapper {
 		$key = $rule['setting'];
 		if ( self::DESKTOP_BREAKPOINT !== $breakpoint ) {
 			$suffixed = $key . '_' . $breakpoint;
-			if ( $this->control_exists( $widget_config, $suffixed ) ) {
-				$key = $suffixed;
-			} elseif ( empty( $rule['responsive'] ) ) {
-				return null;
-			} else {
+			if ( ! $this->control_exists( $widget_config, $suffixed ) ) {
 				return null;
 			}
+			$key = $suffixed;
 		}
 
 		return [ $key => $resolved ];
--- a/modules/mcp/abilities/appliers/v3/v3-style-mapper.php
+++ b/modules/mcp/abilities/appliers/v3/v3-style-mapper.php
@@ -1,0 +1,344 @@
+<?php
+
+namespace Elementor\Modules\Mcp\Abilities\Appliers\V3;
+
+use Elementor\Modules\AtomicWidgets\CssConverter\Css_Converter;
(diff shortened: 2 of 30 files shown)


modules/mcp/abilities/appliers/v3/v3-style-mapper.php
Omerisra6, line 1: Maybe we can create some kind of a readable schema, not sure that it worth it

modules/mcp/abilities/appliers/v3/v3-style-mapper.php
```diff
@@ -0,0 +1,344 @@
...
+		if ( null === $resolved ) {
+			return null;
+		}
+
+		$key = $rule['setting'];
+		if ( self::DESKTOP_BREAKPOINT !== $breakpoint ) {
+			$suffixed = $key . '_' . $breakpoint;
+			if ( $this->control_exists( $widget_config, $suffixed ) ) {
+				$key = $suffixed;
+			} elseif ( empty( $rule['responsive'] ) ) {
```
IshayMaya, line 218: do we need to do something special in this case? check if it somehow missed, otherwise there's no meaning to this elseif

modules/mcp/abilities/appliers/v3/v3-style-mapper.php
```diff
@@ -0,0 +1,344 @@
...
+
+		$key = $setting;
+		if ( ! empty( $override['responsive'] ) && self::DESKTOP_BREAKPOINT !== $breakpoint ) {
+			$suffixed = $setting . '_' . $breakpoint;
+			if ( $this->control_exists( $widget_config, $suffixed ) ) {
+				$key = $suffixed;
+			} elseif ( $this->control_exists( $widget_config, $setting ) ) {
+				return null;
+			} else {
+				$key = $suffixed;
```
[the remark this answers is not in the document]
IshayMaya: @Omerisra6 please check

modules/mcp/abilities/appliers/v3/v3-style-settings-index.php
```diff
@@ -0,0 +1,150 @@
...
+					'setting' => $setting_key,
+					'resolver' => self::infer_resolver( $control, $property ),
+					'responsive' => isset( $controls[ $setting_key . '_tablet' ] ) || isset( $controls[ $setting_key . '_mobile' ] ),
+				];
+			}
+		}
+
+		$index = [];
+		foreach ( $candidates as $match_key => $matches ) {
+			if ( 1 !== count( $matches ) ) {
```
IshayMaya, line 86: is that mean we don't handle variants? maybe we should at least pass a warning to the llm

modules/mcp/abilities/get-structure-ability.php
```diff
@@ -138,43 +137,59 @@ private function prune_elements( array $elements, bool $include_content = false
...
+			$skeleton['settings'] = (object) [];
+			$skeleton['style'] = '';
+			return;
+		}
+
+		$filter = V3_Non_Style_Allowlist::filter( $widget_type, $raw_settings );
+		$allowed_settings = $filter['allowed'];
+
+		$widget_config = Widget_Context_Helper::get_widget_config( $widget_type );
+		$style = ( new V3_Style_Serializer() )->serialize( $raw_settings, $widget_type, $widget_config ?? [] );
```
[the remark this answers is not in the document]
IshayMaya: @Omerisra6 please check

modules/mcp/abilities/appliers/style-applier.php
```diff
@@ -56,8 +60,8 @@ public function apply( array $config_id_index, array $styles, string $style_appl
 			$node = &$config_id_index[ $config_id ];
 
 			if ( V3_Node_Bridge::is_v3_node( $node ) ) {
-				$warning = V3_Node_Bridge::apply_custom_css( $node, $css_string );
-				if ( null !== $warning ) {
+				$v3_warnings = $this->apply_v3_style( $node, $css_string, $widget_configs );
```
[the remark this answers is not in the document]
IshayMaya: @Omerisra6 please check

modules/mcp/abilities/appliers/v3/v3-value-resolvers.php
```diff
@@ -0,0 +1,322 @@
...
+	/**
+	 * Spreads typography-related CSS properties into legacy typography_* keys and
+	 * forces the group toggle to `custom`.
+	 *
+	 * @param array<string, string> $declarations property => css value
+	 * @param string                $prefix       Control group prefix, e.g. `typography` or `menu_typography`.
+	 * @return array<string, mixed>
+	 */
+	public static function resolve_typography_group( array $declarations, string $prefix = 'typography' ): array {
+		$patch = [
```
IshayMaya, line 164: the toggle is flipped before we know if any declaration resolves. If the LLM sends font-size: potato; line-height: banana;, both resolve to null, but we still write typography_typography = 'custom'
this will the user's preset without applying anything. maybe defer this line until after the loop and only setting it if $patch gained at least one real value.

modules/mcp/abilities/appliers/v3/v3-style-serializer.php
```diff
@@ -0,0 +1,347 @@
...
+		}
+
+		$this->push( $blocks, self::DEFAULT_BREAKPOINT, $state, 'border', implode( ' ', $parts ) );
+
+		if ( ! $responsive ) {
+			return;
+		}
+
+		foreach ( self::RESPONSIVE_SUFFIXES as $suffix => $breakpoint ) {
+			$responsive_width = $this->format_value( 'sides', $settings[ $prefix . '_width' . $suffix ] ?? null );
```
IshayMaya, line 152: we emit border-width here but the mapper only handles the border shorthand, so this can't be read back on the next round-trip and will fall into custom_css. 
either emit full shorthand, or add a border-width override to the mapper.

modules/mcp/abilities/appliers/v3/v3-value-resolvers.php
```diff
@@ -0,0 +1,322 @@
...
+
+			if ( in_array( $property, [ 'font-size', 'line-height', 'letter-spacing', 'word-spacing' ], true ) ) {
+				$dimension = self::resolve_dimension( $value );
+				if ( null !== $dimension ) {
+					$patch[ $key ] = $dimension;
+				}
+				continue;
+			}
+
+			$patch[ $key ] = trim( $value );
```
IshayMaya, line 194: V3's typography control expects a single named font, so the editor won't recognize a comma-separated stack. renders fine, but not editable maybe mention it in the docs

modules/mcp/abilities/appliers/v3/v3-style-mapper.php
```diff
@@ -0,0 +1,344 @@
...
+		return in_array( $state, Style_Variants_Merger::PSEUDO_STATES, true ) ? $state : null;
+	}
+
+	/**
+	 * @return array<int, array{property: string, value: string}>
+	 */
+	private function parse_declarations( string $css ): array {
+		$rules = [];
+
+		foreach ( explode( ';', $css ) as $declaration ) {
```
IshayMaya, line 275: this breaks on ; inside url(...), quoted strings, etc. Css_Converter already tokenizes declarations correctly, let's reuse that instead of a naive split.

modules/mcp/abilities/appliers/style-applier.php
```diff
@@ -124,6 +128,43 @@ public function apply( array $config_id_index, array $styles, string $style_appl
...
+		if ( ! empty( $result['settings_patch'] ) ) {
+			$node['settings'] = array_merge( $node['settings'] ?? [], $result['settings_patch'] );
+		}
+
+		$unmapped = $result['unmapped_css'] ?? '';
+		$warning = V3_Node_Bridge::apply_custom_css( $node, $unmapped );
+		if ( null !== $warning ) {
+			$warnings[] = $warning;
+		} elseif ( '' !== trim( $unmapped ) ) {
+			$warnings[] = __( 'Some CSS could not be mapped to V3 settings and was written to custom_css.', 'elementor' );
```
IshayMaya, line 162: when Pro is missing and there's unmapped CSS, only the Pro warning fires and the LLM never learns anything was dropped
maybe emit both and include a snippet of the unmapped CSS so the model can retry.

modules/mcp/abilities/appliers/style-applier.php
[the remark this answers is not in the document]
Omerisra6: Fixed in 38ea880927 — `apply_v3_style` now honors `style_apply_mode`. On `replace`, `V3_Node_Bridge::clear_style_settings()` wipes all known mapped style keys (registry overrides + generic index + responsive suffixes) and `custom_css` before applying, matching V4 semantics. Empty CSS + replace clears without re-applying.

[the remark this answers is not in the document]
Omerisra6: Fixed — see above. V3 replace now mirrors V4: full wipe of mapped style keys + `custom_css`, then apply.

modules/mcp/abilities/get-structure-ability.php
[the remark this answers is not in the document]
Omerisra6: Fixed in 38ea880927 — `V3_Style_Serializer` now unwraps `settings.custom_css` (`selector { ... }` → declarations) and appends it to the serialized `style` string so get-page-structure round-trips unmapped CSS.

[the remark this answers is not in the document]
Omerisra6: Fixed — serializer now includes unwrapped `custom_css` in the `style` field.

modules/mcp/abilities/get-structure-ability.php
```diff
@@ -122,11 +126,6 @@ private function prune_elements( array $elements, bool $include_content = false
 				$skeleton['widgetType'] = $node['widgetType'];
 			}
 
-			$version = $this->resolve_element_version( $node );
-			if ( null !== $version ) {
-				$skeleton['version'] = $version;
-			}
-
 			$title = Element_Structure_Title::resolve( $node );
```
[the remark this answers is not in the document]
Omerisra6: Fixed in 38ea880927 — dropped `version` assertions, renamed the tagging test to assert version is omitted, and added `test_execute__serializes_allowlisted_v3_style_when_include_content_true` covering the `style` field + `custom_css` round-trip.

modules/mcp/abilities/appliers/v3/v3-style-mapper.php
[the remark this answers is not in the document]
Omerisra6: Fixed in 38ea880927 — if the suffixed responsive control does not exist, we `return null` and fall through to unmapped CSS instead of writing an orphan key.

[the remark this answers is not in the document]
Omerisra6: Fixed — orphan responsive writes removed; missing suffixed control now falls back to unmapped CSS.

modules/mcp/abilities/appliers/v3/v3-style-mapper.php
Omerisra6: Agreed — both branches were `return null`. Collapsed to a single `return null` when the suffixed control is missing. No special handling needed.

modules/mcp/abilities/appliers/v3/v3-value-resolvers.php
Omerisra6: Fixed in 38ea880927 — `typography_*_typography = custom` is only set after the loop, and only if at least one declaration actually resolved. Invalid values now return an empty patch and leave the preset alone.

modules/mcp/abilities/appliers/v3/v3-style-serializer.php
Omerisra6: Fixed in 38ea880927 — responsive path now emits the full `border: <width> <style> <color>` shorthand (same as desktop) so the mapper can round-trip it.

modules/mcp/abilities/appliers/style-applier.php
Omerisra6: Fixed in 38ea880927 — both warnings are emitted independently, and the unmapped warning includes a truncated CSS snippet. When Pro is missing the message says "dropped" instead of "written to custom_css".

modules/mcp/abilities/appliers/v3/v3-style-settings-index.php
Omerisra6: Yes — ambiguous selector matches (multiple controls for the same property/state) are intentionally skipped so we don't guess the wrong setting. Unmapped CSS still lands in `custom_css` with a warning. Surfacing a dedicated "ambiguous match" warning is a reasonable follow-up; leaving it out of this PR to keep scope tight.

modules/mcp/abilities/appliers/v3/v3-style-mapper.php
Omerisra6: Fair point. Current input is already a declaration block from `Css_Converter::parse_nested`, so the practical risk is `;` inside `url()`/quotes. Prefer a follow-up to reuse the Css_Converter tokenizer rather than expand this PR — unless you want it in-scope now.

modules/mcp/abilities/appliers/v3/v3-style-settings-index.php
```diff
@@ -0,0 +1,150 @@
...
+		'dropdown_divider_',
+	];
+
+	/**
+	 * @param array                $controls Widget controls from get_config()['controls'].
+	 * @param array<string, array> $style_overrides Registry overrides (excluded from generic index).
+	 * @return array<string, array{setting: string, resolver: string, responsive: bool}>
+	 *         Keyed by "property" or "property@state". Only unique matches are kept.
+	 */
+	public static function build( array $controls, array $style_overrides = [] ): array {
```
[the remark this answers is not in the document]
Omerisra6: Deferring dedicated `V3_Style_Settings_Index` unit tests to a follow-up. Mapper/serializer coverage already exercises the unique-match path; ambiguous/group-control cases are worth a small focused test file next.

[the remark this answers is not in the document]
Omerisra6: Leaving dedicated `V3_Style_Settings_Index` tests out of this PR. Mapper/serializer coverage already exercises the unique-match path; ambiguous/group-control cases can land in a follow-up if we see regressions.

modules/mcp/abilities/appliers/v3/v3-style-mapper.php
Omerisra6: Not worth it for this PR. The registry + allowlist already define the contract; a separate “readable schema” would duplicate that surface without changing LLM behavior. Closing this idea.

modules/mcp/abilities/appliers/v3/v3-style-settings-index.php
Omerisra6: Keeping the silent skip. Ambiguous property→control matches stay out of the index so we don’t guess the wrong setting; those declarations still fall through to `custom_css` with the existing unmapped warning. A dedicated “ambiguous match” warning isn’t needed for this PR.

modules/mcp/abilities/appliers/v3/v3-style-mapper.php
Omerisra6: Explaining rather than changing in this PR:

`Css_Converter::parse()` is also a naive `explode(';')` — same limitation (documented as acceptable for clean LLM input). There isn’t a declaration-level tokenizer to reuse today; `Css_Token_Splitter` only splits values by comma/whitespace inside a single value.

Practical risk for V3 bridge: `;` inside `url("...")` / quoted strings in a flat declaration list. That input is uncommon for our allowlisted widgets (mostly color/typography/border), and unmapped leftovers already go to `custom_css`.

If we harden this later, the right move is a shared paren/string-aware splitter used by both `Css_Converter` and `V3_Style_Mapper` — not wiring V3 to the current naive parse.

modules/mcp/abilities/appliers/v3/v3-value-resolvers.php
Omerisra6: Taking the first family from the stack now (strip quotes) so the V3 typography control stays a single named font.

IshayMaya (dismissed):

IshayMaya (approved):

PR #36853, opened 2026-08-11
Internal: Editor<->MCP Locking signal [ED-24605]

eavichay-elementor:
- **session-guard init**
- **applying mutation signals**
- **heartbeat watch on change**
- **mutation marking**

## 1. Problem & Context

Prevents MCP mutations from conflicting with unsaved editor changes by adding bidirectional locking: editor signals unsaved state to heartbeat; heartbeat signals mutations back to editor. Blocks destructive MCP operations when editor holds the lock with pending changes.

## 2. What Changed (Where)

| Component | Change |
|-----------|--------|
| `Editor_Sync_State` (new) | Transient-based state manager: tracks editor unsaved/MCP mutation signals, guards mutations |
| `heartbeat.php` | Fires `elementor/heartbeat/unsaved_signal` action; injects `elementor_mcp_mutation` marker + server time in response |
| `Abstract_Ability` | Added `execute_guarded()` wrapper that checks mutation guard before execute, sets mutation flag after success |
| MCP Abilities | Updated `destructive` flag accuracy: `true` for Build/Publish/UpdateSettings, `false` for ManageClasses/ManageVariable |
| `heartbeat.js` | Detects MCP mutations via response marker; shows modal warning and forces reload/save based on dirty state |
| Tests | New unit tests for guard logic, transient lifecycle, signal handling; QUnit for heartbeat modal UX |

## 3. How It Works

**Flow on destructive MCP call:**
1. `execute_guarded()` runs `pre_execute_guard` filter → checks lock + unsaved transient → returns 409 error if both exist
2. On success, `Editor_Sync_State::set_mcp_mutation()` stores mutation timestamp (3600s TTL)
3. Next heartbeat sends `elementor_mcp_mutation` with mutated_at timestamp
4. Editor JS detects mutation newer than last sync → shows modal with Force Save/Reload options
5. Document save clears both unsaved + mutation transients (cleanup on `document/after_save`)

Transient ownership prevents cross-user cleanup; mutation guard only fires when editor lock owned by **different** user than MCP caller.

## 4. Risks

**Stale transients:** 5min unsaved TTL could expire before heartbeat fires, allowing MCP to proceed while editor still editing. Mitigated by guard checking both lock + transient (dual condition).

**Modal noise:** Multiple rapid mutations could fire multiple modals. No deduplication per sync cycle—rely on heartbeat interval (typically 15s) to batch signals naturally.


assets/dev/js/editor/utils/heartbeat.js
```diff
@@ -69,6 +78,12 @@ export default class Heartbeat {
 			this.getModal().hide();
 		}
 
+		const mutatedAt = response.elementor_mcp_mutation?.mutated_at;
+
+		if ( mutatedAt && mutatedAt > this.lastSyncedAt ) {
+			this.showExternalChangeModal();
```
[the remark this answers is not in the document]
eavichay-elementor: fixed

assets/dev/js/editor/utils/heartbeat.js
```diff
@@ -86,12 +101,87 @@ export default class Heartbeat {
...
+			}
+			: {
+				headerMessage: __( 'Page Updated by AI', 'elementor' ),
+				message: __( 'This page was changed externally. Reload to get the latest version.', 'elementor' ),
+				strings: {
+					confirm: __( 'Reload', 'elementor' ),
+				},
+				defaultOption: 'confirm',
+				onConfirm: () => this.reloadDocument(),
+			};
```
eavichay-elementor, line 157: Consider extracting the common (headerMessage, cancelMessage, defaultOption, onConfirm) outside the object and then spread+append if `isDirty`

assets/dev/js/editor/utils/heartbeat.js
```diff
@@ -86,12 +101,87 @@ export default class Heartbeat {
...
+
+		this.externalChangeModal = elementorCommon.dialogsManager.createWidget( isDirty ? 'confirm' : 'alert', {
+			...config,
+			closeButton: false,
+			closeButtonOptions: { iconClass: '' },
```
eavichay-elementor: necessary if no closeButton?

assets/dev/js/editor/utils/heartbeat.js
```diff
@@ -86,12 +101,87 @@ export default class Heartbeat {
...
+				onEscKeyPress: false,
+				onBackgroundClick: false,
+			},
+		} );
+
+		this.externalChangeModal.show();
+	}
+
+	onDocumentChanged() {
+		const newChangeOfDocumentState = this.document.editor.isChanged;
```
eavichay-elementor, line 173: newStateOfDocumentChanged or newDocumentChangedState (just naming)

modules/mcp/module.php
```diff
@@ -35,6 +36,11 @@ public function __construct() {
 		( new Mcp_Proxy_REST_API( $this->registry ) )->register_hooks();
 		( new Public_Preview_Handler() )->register();
 
+		add_filter( 'elementor/mcp/pre_execute_guard', [ $this, 'check_mutation_guard' ], 10, 2 );
+		add_action( 'elementor/heartbeat/unsaved_signal', [ $this, 'handle_unsaved_signal' ], 10, 2 );
+		add_filter( 'elementor/heartbeat/mutation_marker', [ $this, 'build_mutation_marker' ], 10, 2 );
+		add_action( 'elementor/document/after_save', [ $this, 'clear_unsaved_signal_on_save' ] );
```
eavichay-elementor: these are functions calling the session utility, perhaps these events should be registered and handled there, and the module only constructs and ensures it registers?

modules/mcp/module.php
```diff
@@ -101,6 +107,55 @@ public function register_server( $adapter ) {
 		}
 	}
 
+	public function clear_unsaved_signal_on_save( $document ): void {
```
eavichay-elementor: moving this code into the session 'guard' keeps the module clean, and bettere SOC

modules/mcp/utils/editor-session-guard.php
```diff
@@ -0,0 +1,66 @@
+<?php
```
eavichay-elementor: File-level: SessionGuard sounds like a security/connectivity related functionality - it only deals with sync. Perhaps editor_session_sync_util?

modules/mcp/utils/editor-sync-state.php
```diff
@@ -0,0 +1,122 @@
...
+	public function register_hooks(): void {
+		add_filter( 'elementor/mcp/pre_execute_guard', [ $this, 'check_mutation_guard' ], 10, 2 );
+		add_action( 'elementor/heartbeat/unsaved_signal', [ $this, 'handle_unsaved_signal' ], 10, 2 );
+		add_filter( 'elementor/heartbeat/mutation_marker', [ $this, 'build_mutation_marker' ], 10, 2 );
+		add_action( 'elementor/document/after_save', [ $this, 'on_document_saved' ] );
+	}
+
+	public function handle_unsaved_signal( int $post_id, $signal_value ): void {
+		if ( $signal_value ) {
+			self::set_editor_unsaved( (int) $signal_value );
```
Omerisra6: Any logged-in user with heartbeat access can set `elementor_has_unsaved: <ANY_POST_ID>` in their heartbeat payload and mark arbitrary posts as unsaved, blocking admins' MCP mutations on those posts with 409. Use `$post_id` (from `elementor_post_lock.post_ID`) and add a `current_user_can( 'edit_post', $post_id )` check before setting the transient.

```php
self::set_editor_unsaved( (int) $signal_value );
// ->
if ( current_user_can( 'edit_post', $post_id ) ) {
    self::set_editor_unsaved( $post_id );
}
```

modules/mcp/utils/editor-sync-state.php
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	public function check_mutation_guard( $error, $input ): ?\WP_Error {
+		$post_id = isset( $input['post_id'] ) ? (int) $input['post_id'] : 0;
+
+		if ( $post_id <= 0 ) {
+			return null;
+		}
+
+		$has_lock    = ! empty( get_post_meta( $post_id, '_edit_lock', true ) );
```
Omerisra6: `_edit_lock` stores `TIMESTAMP:USER_ID` and expires after ~150s in WordPress, but `! empty(get_post_meta(...))` treats a stale value as still locked. Prefer `wp_check_post_lock( $post_id )` which handles the timestamp comparison.

modules/mcp/abilities/manage-classes-ability.php
```diff
@@ -64,7 +64,7 @@ protected function get_definition(): Ability_Definition {
 				'annotations' => [
 					'readonly' => false,
 					'idempotent' => false,
-					'destructive' => true,
+					'destructive' => false,
```
Omerisra6, line 67: These abilities mutate global state (classes / variables) that affects every document, and they now skip the guard entirely — so an AI-triggered change can silently rewrite globals while a user has unsaved changes elsewhere. Was this intentional, and if so how should conflicts against an in-progress edit be surfaced?

modules/mcp/abilities/abstract-ability.php
```diff
@@ -20,13 +22,35 @@ abstract protected function get_definition(): Ability_Definition;
...
+
+		return $result;
+	}
+
 	public function check_permission(): bool {
 		return (bool) call_user_func( $this->definition()->permission_callback );
 	}
 
-	public function register(): void {
+	final public function register(): void {
```
Omerisra6: `register()` was previously overridable; marking it `final` breaks any Pro or third-party subclass that customized registration. Worth confirming no external subclass overrides this before landing.

assets/dev/js/editor/utils/heartbeat.js
```diff
@@ -54,6 +61,10 @@ export default class Heartbeat {
 		data.elementor_post_lock = {
 			post_ID: this.document.id,
 		};
+
+		if ( this.document.editor.isChanged ) {
+			data.elementor_has_unsaved = this.document.id;
+		}
```
Omerisra6: `onSend` only writes `elementor_has_unsaved` when dirty, so after a dirty→clean transition (undo, external revert) the server never clears the transient and MCP calls stay blocked until the 300s TTL or a save. The new QUnit test also asserts `elementor_has_unsaved === null` for the clean case, which will fail against the current code (`undefined !== null`) — add an `else` branch that sends `null` (or the doc id with a "clean" marker) so `handle_unsaved_signal`'s falsy path is actually reachable.

```js
if ( this.document.editor.isChanged ) {
    data.elementor_has_unsaved = this.document.id;
} else {
    data.elementor_has_unsaved = null;
}
```

modules/mcp/utils/editor-sync-state.php
```diff
@@ -0,0 +1,122 @@
...
+
+namespace Elementor\Modules\Mcp\Utils;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+class Editor_Sync_State {
+
+	const EDITOR_UNSAVED_TTL = 300;
```
Omerisra6, line 11: Heartbeat intervals can stretch to minutes when the editor tab is backgrounded/throttled, but `EDITOR_UNSAVED_TTL = 300` will expire silently and let MCP mutations through even though the editor still has unsaved changes. Was this bypass window considered, or should the client refresh the signal on every tick while dirty (not only on state transitions)?

assets/dev/js/editor/utils/heartbeat.js
```diff
@@ -86,12 +103,90 @@ export default class Heartbeat {
 		}
 	}
 
+	onDocumentLoaded() {
+		this.lastSyncedAt = Math.floor( Date.now() / 1000 );
```
Omerisra6: `Date.now()` is client-local and drifts from server `time()` — a laptop with a fast clock will initialize `lastSyncedAt` ahead of any server-recorded `mutated_at`, and the "Page Updated by AI" modal will never fire. Seed `lastSyncedAt` from the first heartbeat response (server-authoritative) instead.

assets/dev/js/editor/utils/heartbeat.js
```diff
@@ -86,12 +103,90 @@ export default class Heartbeat {
...
+		$e.internal( 'document/save/set-is-modified', { status: false } );
+		this._doReload();
+	}
+
+	_doReload() {
+		window.location.reload();
+	}
+
+	forceSave() {
+		this.lastSyncedAt = Math.floor( Date.now() / 1000 );
```
Omerisra6: `lastSyncedAt` is set to "now" synchronously, before `$e.run('document/save/save')` actually resolves — if the save fails, any mutation marker that arrives afterwards is silently suppressed and the user never sees the alert. Move the update into a save-success handler.

Omerisra6 (commented):
Security and edge-case review focused on the editor↔MCP locking signal.

modules/atomic-widgets/prop-types/utils/plain-llm-schema-converter.php
```diff
@@ -34,6 +34,9 @@ private function walk( array $schema ): array {
 
 		if ( isset( $schema['anyOf'] ) && is_array( $schema['anyOf'] ) ) {
 			$schema['anyOf'] = $this->walk_any_of( $schema['anyOf'] );
+			if ( count( $schema['anyOf'] ) === 1 ) {
+				return $schema['anyOf'][0];
```
Omerisra6: Returning `$schema['anyOf'][0]` when the union collapses to one entry drops every sibling key on `$schema` (`description`, `title`, `default`, `nullable`, etc.), silently stripping constraints from the LLM-facing schema. This also looks unrelated to the locking-signal feature — please split it into its own PR, or at minimum merge the sibling keys into the returned schema.

modules/mcp/abilities/manage-classes-ability.php
eavichay-elementor: they do not have page_id - therefore cannot be used as an AI locking mechanism, same behavior as if a different user updates variables/classes

modules/mcp/utils/editor-sync-state.php
eavichay-elementor: The TTL bypass is handled by the updated `onSend` — it now always sends `elementor_has_unsaved` (document ID when dirty, null when clean) on every heartbeat tick, which resets the 300s transient TTL. With a 15–60s heartbeat interval, the transient stays fresh as long as the editor is dirty.

modules/mcp/abilities/manage-classes-ability.php
eavichay-elementor: Intentional. manage-classes and manage-variable operate on global design tokens, not per-document content, so there is no per-document unsaved state to conflict with. The guard is scoped to document mutations (destructive tools with a post_id). Global token changes are non-destructive from a document perspective and are reversible via their own undo history.

Omerisra6 (approved):

PR #36891, opened 2026-08-12
Internal: Atomic e-list - markers [ED-25348]

mike-elementor:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #

## 1. Problem & Context

Markers in atomic list items are now conditionally rendered based on a parent list setting. Previously hard-coded as always present, this enables toggling marker visibility at the list level while preserving marker customizations via stashing.

## 2. What Changed (Where)

| Component | Change |
|-----------|--------|
| `atomic-list-item-marker.php` | Added `build_template_context()` to provide base context. |
| `atomic-list-item.php` | Removed marker from default children; added `show_markers` prop, children dependency logic, and render context syncing. |
| `atomic-list.php` | Added `show_markers` boolean prop, Switch_Control UI, and render context propagation to items. |
| `legacy/list-type.ts` | New file: custom element type extending base with `_syncShowMarkersToChildren()` and render context getters. |
| `editor-canvas init.tsx` | Reordered initialization to call `initListType()` before `initLegacyViews()`. |
| `list-actions.ts` & test | New items now include `show_markers: true` in settings. |
| `legacy/types.ts` | Added `initialize()` and `listenTo()` declarations to ElementView. |

## 3. How It Works

List-level `show_markers` setting flows down via render context to each item's children dependencies. On list init/render, legacy view syncs the prop to all child items' settings, triggering dependencies to conditionally attach/detach markers with stash preservation. Avoids marker as default child; declarative control via prop instead.

## 4. Risks

Stashing logic assumes marker customizations are recoverable—verify marker removal/re-addition preserves all styling/settings. Legacy view sync occurs on every render; validate performance with large lists. Type assertions on prop values (lines 72, 86, 105-106) could mask schema mismatches—consider stricter typing.


modules/atomic-widgets/elements/atomic-list/atomic-list-item/atomic-list-item.php
```diff
@@ -89,17 +106,66 @@ protected function define_base_styles(): array {
...
+	 * This propagates the list-level setting to each item, where children
+	 * dependencies use it to determine marker visibility.
+	 */
+	protected function define_render_context(): array {
+		$list_context = Render_Context::get( Atomic_List::class );
+
+		return [
+			[
+				'context' => [
+					'show_markers' => $list_context['show_markers'] ?? true,
```
IshayMaya, line 163: 🔥

modules/atomic-widgets/module.php
```diff
@@ -600,10 +600,10 @@ private function add_inline_styles() {
 			'.e-background-video:not(.e-background-video--playing):not(.e-background-video--paused) .e-background-video__play,',
 			'.e-background-video:not(.e-background-video--playing):not(.e-background-video--paused) .e-background-video__pause { display: none; }',
 			'.e-list { margin: 0; padding: 0; }',
-			'.e-list-item { list-style: none; }',
-			'.e-list-item__marker { align-self: flex-start; }',
+			'.e-list-item { list-style: none; display: flex; flex-direction: row; align-items: flex-start; gap: 8px; }',
+			'.e-list-item__marker { align-self: flex-start; flex: 0 0 auto; width: auto; }',
 			'.e-list-item__marker .e-svg-base { width: 1rem; height: 1rem; }',
-			'.e-list-item__content { min-width: 0; }',
+			'.e-list-item__content { min-width: 0; flex: 1 1 0%; }',
```
IshayMaya: please check why we need all that, we want to stick with the atomic styles

packages/packages/core/editor-canvas/src/legacy/list-type.ts
```diff
@@ -0,0 +1,125 @@
...
+			this._syncShowMarkersToChildren();
+		},
+
+		/**
+		 * Sync show_markers from list to all list items.
+		 *
+		 * This propagates the list-level setting to each item, where children dependencies
+		 * will conditionally add/remove markers based on the value.
+		 */
+		_syncShowMarkersToChildren() {
```
IshayMaya, line 69: why do we need this custom action here? the existing reconcile mechanism should handle it

modules/atomic-widgets/elements/atomic-list/atomic-list-item-marker/atomic-list-item-marker.php
```diff
@@ -91,6 +91,10 @@ protected function define_default_children() {
 		];
 	}
 
+	protected function build_template_context(): array {
+		return $this->build_base_template_context();
+	}
+
```
mike-elementor, line 97: will be removed in next PR's

packages/packages/core/editor-canvas/src/legacy/list-type.ts
mike-elementor: for some reason, it didn't work for me "live"
only with the custom... (need to investigate some more)..

mike-elementor: @IshayMaya can I address/investigate this in a sep. task later?
( I know current impl. works, but not so sure about the other )

packages/packages/core/editor-canvas/src/legacy/list-type.ts
```diff
@@ -0,0 +1,125 @@
...
+function createListView( options: CreateNestedTemplatedElementTypeOptions ) {
+	const BaseView = createNestedTemplatedElementView( options );
+
+	return BaseView.extend( {
+		initialize( ...args: unknown[] ) {
+			BaseView.prototype.initialize?.apply( this, args );
+
+			const settings = this.model.get( 'settings' );
+			if ( settings ) {
+				this.listenTo( settings, 'change:show_markers', () => {
```
Omerisra6, line 52: check how loop solve this, (pagination show/hide) maybe there is a better approach

packages/packages/core/editor-canvas/src/init.tsx
```diff
@@ -27,6 +28,9 @@ export function init() {
 
 	initViewReplacements();
 
+	initTabsModelExtensions();
```
Omerisra6, line 31: Lets keep the old call order (put last)

PR #36901, opened 2026-08-13
Internal: Validate V3 element_config settings before merge [ED-25241]

Omerisra6:
## Summary

Adds a value-level gate for V3 (legacy) widget settings in the MCP `build-composition` applier. Before this change the applier only checked key names via `V3_Non_Style_Allowlist`; a well-shaped-looking value could still smuggle e.g. an array into a scalar `link` slot, and `esc_url()` would blow up at render time.

`V3_Settings_Validator` is a thin orchestrator:

- `V3_Non_Style_Allowlist` — key gate (which keys the widget accepts at all).
- `V3_Json_Schema_Builder::check_settings_shape()` — shallow shape guard co-located with schema emission (`type`, `enum`, one-level nested `properties.type`). This is **not** full JSON Schema validation — no `required`, `additionalProperties`, or `items` enforcement.

`Element_Config_Applier`'s V3 branch is wired through the validator and surfaces the aggregate `WP_Error` under the `config_id`, matching the existing V4 error-reporting pattern.

## Scope

- No dynamic-tag coercion yet — that lands in the stacked follow-up PR (LLM `{name, settings}` → V3 `__dynamic__[key]` shortcode).
- No JSON-Schema `anyOf` wrapping yet — also follow-up (needed only to advertise both primitive and dynamic-shape branches to the LLM).

## Jira

ED-TBD — replace with the real key before merge.

## Test plan

- [x] `tests/phpunit/run-unit.sh tests/phpunit/elementor/modules/mcp/abilities/utils/test-v3-json-schema-builder.php` — shape-check tests on `check_value_shape` / `check_settings_shape`, pass.
- [x] `tests/phpunit/run-unit.sh tests/phpunit/elementor/modules/mcp/abilities/appliers/v3/test-v3-settings-validator.php` — orchestrator + parity tests, pass.
- [x] `tests/phpunit/run-unit.sh tests/phpunit/elementor/modules/mcp/abilities/appliers/test-element-config-applier.php --filter v3_rejects` — applier integration, pass.
- [x] `vendor/bin/phpcs --standard=ruleset.xml` on touched files — clean.
- [ ] Full CI PHPUnit suite (WP-booted) green.

## 1. Problem & Context

V3 element_config settings were merged directly into nodes without shape validation, risking array-to-scalar field collisions and invalid enum values. This adds a shallow schema validator (type, enum, nested properties only) to catch mismatches before merge.

## 2. What Changed (Where)

- `element-config-applier.php`: Import `V3_Settings_Validator`; refactor controls extraction; validate primitives post-hoisting before merge; collect shape errors separately
- `v3-settings-validator.php` (new): Wraps `V3_Json_Schema_Builder::check_settings_shape()` with human-readable error messages
- `v3-json-schema-builder.php`: Add `check_value_shape()`, `check_settings_shape()`, type/enum/nested-property validators, loose equality helpers
- Test files: Extend coverage for validation paths, fixture support for control resolution

## 3. How It Works

Entry: Element settings flow through allowlist → dynamic hoister → **validator** → merge. Validator builds schema from controls, compares primitives against type/enum/nested-property constraints. Valid settings merge; invalid ones reported as WP_Error, blocking partial updates.

## 4. Risks

Shallow validation (no deep recursion, loose type coercion for numerics/enums) matches V3's runtime-free design but may miss complex object mismatches. Mitigation: two-level property check covers common cases; full JSON-Schema unnecessary for V3 use case.


modules/mcp/abilities/appliers/v3/v3-settings-validator.php
```diff
@@ -0,0 +1,129 @@
...
+				[ 'status' => \WP_Http::BAD_REQUEST ]
+			),
+		];
+	}
+
+	private static function format_error( string $widget_type, string $key, string $reason ): string {
+		return sprintf( 'V3 widget "%s" property "%s": %s', $widget_type, $key, $reason );
+	}
+
+	private static function validate_primitive( $value, ?array $entry_schema ): ?string {
```
Omerisra6: Can we validate in a more elegant way? how V3 settings are validated?

modules/mcp/abilities/appliers/v3/v3-settings-validator.php
```diff
@@ -0,0 +1,56 @@
...
+	 *     allowed: array<string, mixed>,
+	 *     error: \WP_Error|null,
+	 * }
+	 */
+	public static function validate( string $widget_type, array $settings, array $widget_config ): array {
+		$key_filter = V3_Non_Style_Allowlist::filter( $widget_type, $settings );
+
+		$controls = is_array( $widget_config['controls'] ?? null ) ? $widget_config['controls'] : [];
+		$schema = V3_Json_Schema_Builder::build( $controls, array_keys( $key_filter['allowed'] ) );
+		$shape = V3_Json_Schema_Builder::check_settings_shape( $key_filter['allowed'], $schema );
```
IshayMaya: if the widget_config['controls'] is empty or missing this key, the shape check does nothing and any value is accepted, can we fail (or at least skip merge) when there is no schema for an allowlisted key?

modules/mcp/abilities/utils/v3-json-schema-builder.php
```diff
@@ -76,6 +76,109 @@ public static function build( $controls, ?array $allowed_keys = null ): array {
...
+	 * This is not full JSON Schema validation — only `type`, `enum`, and one-level nested
+	 * `properties.type` are enforced. V3 has no runtime `Props_Parser`; this catches the
+	 * array-vs-scalar and unknown-enum classes without pulling in a general JSON-Schema validator.
+	 *
+	 * @param mixed      $value         Setting value to check.
+	 * @param array|null $entry_schema  Schema entry from `build()['properties'][$key]`.
+	 * @return string|null Human-readable reason when shape mismatches; null when acceptable.
+	 */
+	public static function check_value_shape( $value, ?array $entry_schema ): ?string {
+		if ( ! is_array( $entry_schema ) ) {
```
IshayMaya, line 93: that means a key on the allowlist with no control entry is never shape checked, is that intended?

modules/mcp/abilities/utils/v3-json-schema-builder.php
```diff
@@ -76,6 +76,109 @@ public static function build( $controls, ?array $allowed_keys = null ): array {
...
+		}
+
+		if ( 'object' === $expected_type && is_array( $value ) && isset( $entry_schema['properties'] ) && is_array( $entry_schema['properties'] ) ) {
+			foreach ( $entry_schema['properties'] as $prop_key => $prop_schema ) {
+				if ( ! is_string( $prop_key ) || ! array_key_exists( $prop_key, $value ) ) {
+					continue;
+				}
+
+				$sub_expected = $prop_schema['type'] ?? null;
+				$sub_actual = self::json_type_of( $value[ $prop_key ] );
```
IshayMaya: empty [] is treated as object, a valid empty repeater (array) will fail no?

modules/mcp/abilities/utils/v3-json-schema-builder.php
```diff
@@ -76,6 +76,109 @@ public static function build( $controls, ?array $allowed_keys = null ): array {
...
+		}
+
+		$expected_type = $entry_schema['type'] ?? null;
+		$actual_type = self::json_type_of( $value );
+
+		if ( $expected_type && $expected_type !== $actual_type ) {
+			return sprintf( 'invalid shape (expected %s, got %s).', $expected_type, $actual_type );
+		}
+
+		if ( isset( $entry_schema['enum'] ) && is_array( $entry_schema['enum'] ) && ! in_array( $value, $entry_schema['enum'], true ) ) {
```
IshayMaya: strict types will reject common legacy values, numeric option keys as strings ("1" vs 1), and "123" for id/size this might block real compositions.

modules/mcp/abilities/appliers/element-config-applier.php
```diff
@@ -57,13 +57,17 @@ public function apply( array &$config_id_index, array $element_config, array $wi
...
-				$filter = V3_Non_Style_Allowlist::filter( (string) $tag, $settings );
-				if ( $filter['error'] ) {
-					$errors[] = sprintf( '[%s] %s', $config_id, $filter['error']->get_error_message() );
+				$widget_type = (string) $tag;
+				$widget_config = is_array( $widget_configs[ $widget_type ] ?? null ) ? $widget_configs[ $widget_type ] : [];
+
+				$validated = V3_Settings_Validator::validate( $widget_type, $settings, $widget_config );
+
+				if ( $validated['error'] ) {
+					$errors[] = sprintf( '[%s] %s', $config_id, $validated['error']->get_error_message() );
```
IshayMaya: one bad key drops the whole element_config entry. the validator still returns a partial allowed set that we never use. maybe merge the valid keys or dont return a partial set.

modules/mcp/abilities/appliers/v3/v3-settings-validator.php
Omerisra6: Fixed: check_settings_shape() now emits a per-key no schema for allowlisted key error when an allowlisted key has no matching control/schema entry.

modules/mcp/abilities/appliers/v3/v3-settings-validator.php
Omerisra6: Addressed in 1570304 — shape checking is co-located in V3_Json_Schema_Builder::check_value_shape() / check_settings_shape(), and V3_Settings_Validator is a thin orchestrator. Latest commit 3e991f1 tightens the remaining review gaps.

modules/mcp/abilities/utils/v3-json-schema-builder.php
Omerisra6: Not intended — missing schema entries are now rejected per-key (skipped from valid, surfaced in aggregate error) rather than passing through check_value_shape(null schema) → null.

modules/mcp/abilities/utils/v3-json-schema-builder.php
Omerisra6: Fixed: json_type_of([]) now returns array, and value_matches_type explicitly accepts [] for both array and object expected types so empty repeaters pass.

modules/mcp/abilities/utils/v3-json-schema-builder.php
Omerisra6: Fixed: value_matches_type accepts is_numeric() strings for number, and enum matching runs first with loose scalar equality (1 vs "1") before strict type comparison.

modules/mcp/abilities/appliers/element-config-applier.php
Omerisra6: Fixed: V3 branch now best-effort merges validated['allowed'] before surfacing the aggregate error, so valid keys land while invalid ones are reported (test_apply__v3_merges_valid_keys_and_reports_invalid_keys).

IshayMaya (approved):

PR #36903, opened 2026-08-13
Internal: Coerce LLM dynamic-tag shapes on V3 element_config [ED-25241]

Omerisra6:
## Summary

Teaches the MCP `build-composition` applier to accept the LLM-facing `{ name, settings }` (or the nested URL/media `{ url: { name, settings } }`) dynamic-tag shape on V3 primitive slots and rewrite it into V3's `__dynamic__[key]` shortcode format at the node level, matching how legacy widgets render dynamic content.

Standalone off `main` — independent of the settings-validator work in [#36901](https://github.com/elementor/elementor/pull/36901).

## Pieces

- `V3_Dynamic_Resolver` — pure, static coercion. Strict `is_dynamic_capable` (widget-level `dynamic.active` or non-empty `dynamic.default` only — never merges control-type defaults, which otherwise made every text control look dynamic and broke the schema for plain scalar slots). Test seams (`set_tag_info_resolver`, `set_shortcode_builder`) let DB-less unit tests run without a WP-booted registry; production paths fall back to `Plugin::$instance->dynamic_tags`.
- `V3_Json_Schema_Builder` — wraps dynamic-capable controls in `anyOf` over the primitive branch and a `{ name, settings }` branch, so the schema advertises both shapes to the LLM.
- `Element_Config_Applier` — routes each V3 slot through the resolver and merges the resulting shortcodes into `settings['__dynamic__']`, preserving already-populated entries. Keeps the existing `V3_Non_Style_Allowlist` key gate.
- `build-composition.md` — documents the accepted shapes for V3 slots.

## Origin bug

LLM sent `{ url: { name: "post-url", settings: [] } }` on the V3 `link` control of `theme-post-title`. The old path merged the array into the primitive slot, and `esc_url()` crashed at render on the array.

## Jira

ED-TBD — replace with the real key before merge.

## Test plan

- [x] `tests/phpunit/run-unit.sh` on resolver + schema-builder tests — 18 tests, 59 assertions, pass.
- [x] Applier regression test (`v3_dynamic_tag_on_url_control_is_hoisted_into__dynamic__`) — pass.
- [x] `vendor/bin/phpcs --standard=ruleset.xml` on touched files — clean.
- [ ] Full CI PHPUnit suite (WP-booted) green — the previous failure was `is_dynamic_capable` walking into `Plugin::$instance->controls_manager` and merging type-level defaults; strict widget-level check removes that path.

## 1. Problem & Context

V3 widgets need runtime coercion of LLM dynamic-tag shapes from the allowlist layer into hoisted shortcodes and primitives, matching V2 behavior with category validation. This enables LLM to bind dynamic tags to V3 control properties while respecting control-declared category constraints.

## 2. What Changed (Where)

| File | Change |
|------|--------|
| `element-config-applier.php` | Inject `V3_Dynamic_Hoister`, call it for V3 nodes to hoist dynamic tags into `__dynamic__` |
| `v3-dynamic-hoister.php` | New: validates tag categories, serializes to shortcodes, preserves non-dynamic sibling fields |
| `v3-dynamic-resolver.php` | New: detects dynamic-capable controls, extracts top-level and nested dynamic shapes |
| `v3-json-schema-builder.php` | Wrap dynamic-capable controls in `anyOf` union (primitive ∣ `{name, settings}`) with category descriptions |
| Tests | Unit coverage for hoisting, resolver, schema builder; stub dynamic tags manager |
| Documentation | Note category intersection requirement for dynamic tag selection |

## 3. How It Works

**Entry:** V3 node settings flow through `Element_Config_Applier.apply()`. For V3 nodes, after allowlist filtering, `V3_Dynamic_Hoister.hoist()` processes each allowed setting.

**Hoisting logic:** For each control, `V3_Dynamic_Resolver.is_dynamic_capable()` checks if control has `dynamic.active=true` or non-empty `dynamic.default`. If capable, extracts dynamic shape from value (top-level `{name, settings}` or nested at `dynamic.property`). Validates tag exists and category intersection. Serializes to shortcode → `__dynamic__[key]`. Remainder fields (e.g., `is_external` on URL control) merge back into primitives.

**Schema:** `V3_Json_Schema_Builder` now exposes dynamic capability via `anyOf` with category hints for LLM to make informed tag selection.

## 4. Risks

**Category mismatch silently skips field:** If LLM picks incompatible tag, the control won't merge—no fallback. Mitigation: schema descriptions list allowed categories; robust LLM prompt needed upstream.

**Nested dynamic only on `property` key:** If control declares `dynamic.property` but LLM sends top-level shape, extraction fails silently. Mitigation: schema must be clear; test coverage exists.


modules/mcp/abilities/appliers/v3/v3-dynamic-resolver.php
```diff
@@ -0,0 +1,259 @@
...
+
+namespace Elementor\Modules\Mcp\Abilities\Appliers\V3;
+
+use Elementor\Plugin;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
```
Omerisra6, line 9: reduce doc length

modules/mcp/abilities/appliers/v3/v3-dynamic-resolver.php
```diff
@@ -0,0 +1,259 @@
...
+	 * @return array{
+	 *     matched: bool,
+	 *     shortcode?: string,
+	 *     primitive?: mixed,
+	 *     error?: \WP_Error,
+	 * }
+	 */
+	public static function try_resolve( string $key, $value, array $control ): array {
+		if ( ! self::is_dynamic_capable( $control ) ) {
+			return [ 'matched' => false ];
```
Omerisra6: move the check to the consumer

modules/mcp/abilities/appliers/v3/v3-dynamic-resolver.php
```diff
@@ -0,0 +1,259 @@
...
+
+		$category_error = self::validate_categories( $key, $dynamic_input['name'], $tag_info, $control['dynamic']['categories'] ?? [] );
+		if ( $category_error ) {
+			return [
+				'matched' => true,
+				'error' => $category_error,
+			];
+		}
+
+		$shortcode = self::build_shortcode( $dynamic_input['name'], $dynamic_input['settings'] );
```
Omerisra6: Do all tags resolve to a shortcode? Why can't we use the tag render method?

modules/mcp/abilities/appliers/v3/v3-dynamic-resolver.php
```diff
@@ -0,0 +1,259 @@
...
+			return null;
+		}
+
+		return self::pluck_dynamic_shape( $nested );
+	}
+
+	/**
+	 * @return array{name: string, settings: array<string, mixed>}|null
+	 */
+	private static function pluck_dynamic_shape( array $value ): ?array {
```
Omerisra6: do we need that?

modules/mcp/abilities/appliers/v3/v3-dynamic-resolver.php
```diff
@@ -0,0 +1,259 @@
...
+			return '';
+		}
+
+		$built = Plugin::$instance->dynamic_tags->tag_data_to_tag_text( $id, $name, $settings );
+		return is_string( $built ) ? $built : '';
+	}
+
+	private static function generate_tag_id(): string {
+		return substr( md5( uniqid( 'mcp-v3-dyn', true ) ), 0, 7 );
+	}
```
Omerisra6: Lets check if we can just use V3 methods directly, expose if we need to

modules/mcp/abilities/appliers/v3/v3-dynamic-resolver.php
```diff
@@ -0,0 +1,259 @@
...
+
+	private static function empty_primitive_for_control( array $control ): mixed {
```
Omerisra6: why do we need that?

modules/mcp/abilities/appliers/element-config-applier.php
```diff
@@ -57,13 +58,45 @@ public function apply( array &$config_id_index, array $element_config, array $wi
...
-				$filter = V3_Non_Style_Allowlist::filter( (string) $tag, $settings );
+				$widget_type = (string) $tag;
+				$filter = V3_Non_Style_Allowlist::filter( $widget_type, $settings );
 				if ( $filter['error'] ) {
 					$errors[] = sprintf( '[%s] %s', $config_id, $filter['error']->get_error_message() );
 					continue;
 				}
 
-				$node['settings'] = $this->merge_with_clears( $node['settings'] ?? [], $filter['allowed'] );
+				$controls = is_array( $widget_configs[ $widget_type ]['controls'] ?? null )
```
Omerisra6, line 71: lets see how can we make it extendable and not inline in the applier

modules/mcp/abilities/utils/v3-json-schema-builder.php
```diff
@@ -83,11 +83,69 @@ private static function build_entry( array $control, ?string $control_type ): ?a
 			$entry['default'] = $control['default'];
 		}
 
-		if ( isset( $control['description'] ) && is_string( $control['description'] ) ) {
-			$entry['description'] = trim( strip_tags( $control['description'] ) );
+		$description = isset( $control['description'] ) && is_string( $control['description'] )
```
Omerisra6, line 88: seems redundant

modules/mcp/static-resources/abilities/build-composition.md
```diff
@@ -115,10 +115,12 @@ Read [elementor://global-classes] before composing. Create or update via `elemen
 - Global classes are prepended before any local styles from `style`; local styles still win on conflicts
 
 # DYNAMIC TAGS
-- A value can be made dynamic wherever the widget schema allows a dynamic variant (often a union on the prop or a nested field such as an image's `src`).
+- A value can be made dynamic wherever the widget schema allows a dynamic variant (often an `anyOf` on the prop or a nested field such as an image's `src`).
 - Put the plain dynamic object at that node, in place of the static variant. Read [elementor://dynamic-tags] for allowed tag names and each tag's settings schema.
 - Plain dynamic shape: `{ "name": "<allowed tag>", "settings": { ... } }`
 - Example (image `src`): `"image": { "src": { "name": "<image tag>", "settings": { ... } }, "size": "full" }`
+- V3 widgets (e.g. `theme-post-title`, `theme-post-featured-image`) accept the SAME dynamic shape at the field's slot. On URL controls the dynamic goes on the inner `url` slot: `"link": { "url": { "name": "post-url", "settings": {} } }`. The server rewrites it into V3's `__dynamic__` map — do NOT hand-craft `__dynamic__` shortcodes yourself.
```
Omerisra6: Why do we need that? The agent doesnt knows that its a V3 widget, and we explain that at the schema, remove

modules/mcp/abilities/appliers/v3/v3-dynamic-hoister.php
```diff
@@ -0,0 +1,95 @@
...
+		$errors = [];
+
+		foreach ( $allowed as $key => $value ) {
+			$control = is_array( $controls[ $key ] ?? null ) ? $controls[ $key ] : [];
+
+			if ( ! V3_Dynamic_Resolver::is_dynamic_capable( $control ) ) {
+				$primitives[ $key ] = $value;
+				continue;
+			}
+
```
IshayMaya: when the LLM sends a nested dynamic like { url: {name, settings}, is_external: "on" } we drop the whole link slot and only push the shortcode. the is_external / nofollow / custom_attributes siblings never reach settings['link']. 
the legacy behavior keeps those siblings, please also add the non-dynamic siblings back to primitives[$key].

modules/mcp/abilities/appliers/v3/v3-dynamic-hoister.php
```diff
@@ -0,0 +1,95 @@
...
+			$input = V3_Dynamic_Resolver::extract_input( $value, $control['dynamic']['property'] ?? null );
```
IshayMaya: If the LLM sends a nested dynamic plus extra fields (like `is_external`), we only keep the shortcode and drop the extra fields. Please keep those extra fields in the primitive settings.

modules/mcp/abilities/utils/v3-json-schema-builder.php
```diff
@@ -83,13 +83,71 @@ private static function build_entry( array $control, ?string $control_type ): ?a
...
+		}
+
+		if ( null !== $description ) {
+			$entry['description'] = $description;
 		}
 
 		return $entry;
 	}
 
+	private static function is_dynamic_capable( array $control ): bool {
```
IshayMaya: These two `is_dynamic_capable` checks don’t match. The schema only looks at `dynamic.active`, but the resolver also accepts `dynamic.default`. Please use the same rule in both places.

modules/mcp/abilities/appliers/element-config-applier.php
```diff
@@ -57,13 +61,31 @@ public function apply( array &$config_id_index, array $element_config, array $wi
...
+				foreach ( $hoist_outcome['errors'] as $error_message ) {
+					$errors[] = sprintf( '[%s] %s', $config_id, $error_message );
+				}
+
+				$node['settings'] = $this->merge_with_clears( $node['settings'] ?? [], $hoist_outcome['primitives'] );
+
+				if ( ! empty( $hoist_outcome['shortcodes'] ) ) {
+					$existing = is_array( $node['settings']['__dynamic__'] ?? null ) ? $node['settings']['__dynamic__'] : [];
+					$node['settings']['__dynamic__'] = array_merge( $existing, $hoist_outcome['shortcodes'] );
+				}
```
IshayMaya, line 86: This merge overwrites existing `__dynamic__` values. The PR says we keep the old ones. Please either keep the old values, or change the wording.

modules/mcp/abilities/appliers/element-config-applier.php
```diff
@@ -57,13 +61,31 @@ public function apply( array &$config_id_index, array $element_config, array $wi
...
-				$filter = V3_Non_Style_Allowlist::filter( (string) $tag, $settings );
+				$widget_type = (string) $tag;
+				$filter = V3_Non_Style_Allowlist::filter( $widget_type, $settings );
 				if ( $filter['error'] ) {
 					$errors[] = sprintf( '[%s] %s', $config_id, $filter['error']->get_error_message() );
 					continue;
 				}
 
-				$node['settings'] = $this->merge_with_clears( $node['settings'] ?? [], $filter['allowed'] );
+				$controls = is_array( $widget_configs[ $widget_type ]['controls'] ?? null )
```
IshayMaya, line 71: We create a new hoister inside the loop, and the two real call sites don’t pass one in. Please inject it once, or at least create it once outside the loop.

modules/mcp/abilities/appliers/v3/v3-dynamic-hoister.php
```diff
@@ -0,0 +1,95 @@
...
+
+		return [
+			'primitives' => $primitives,
+			'shortcodes' => $shortcodes,
+			'errors' => $errors,
+		];
+	}
+
+	private function generate_tag_id(): string {
+		return substr( md5( uniqid( 'mcp-v3-dyn', true ) ), 0, 7 );
```
IshayMaya: The tag id changes on every run, so the same apply looks different each time. Please use a stable id.

modules/mcp/abilities/appliers/v3/v3-dynamic-resolver.php
```diff
@@ -0,0 +1,69 @@
...
+	 * @return array{name: string, settings: array<string, mixed>}|null
+	 */
+	public static function extract_input( $value, ?string $property ): ?array {
+		if ( ! is_array( $value ) ) {
+			return null;
+		}
+
+		$name = $value['name'] ?? null;
+		if ( is_string( $name ) && '' !== $name ) {
+			$settings = $value['settings'] ?? [];
```
IshayMaya: These two blocks do the same thing. Please extract one small helper and reuse it.

modules/mcp/abilities/utils/v3-json-schema-builder.php
```diff
@@ -83,13 +83,71 @@ private static function build_entry( array $control, ?string $control_type ): ?a
...
+	 * Wraps a primitive schema entry in an `anyOf` union with the plain dynamic-tag shape,
+	 * mirroring how V4 atomic prop types advertise dynamic bindings. Kept aligned with the
+	 * static resource `elementor://dynamic-tags`: the LLM picks a tag from there whose
+	 * categories intersect the ones declared on this control.
+	 *
+	 * @param array<string, mixed> $primitive_entry Primitive schema entry.
+	 * @param array<int, string>   $categories      Dynamic-tag categories accepted by this control.
+	 * @param string|null          $description     Optional shared description hoisted to the wrapper.
+	 */
+	private static function wrap_with_dynamic_branch( array $primitive_entry, array $categories, ?string $description ): array {
```
IshayMaya, line 103: The schema still allows extra keys like `group`. Please set `additionalProperties: false` so the LLM can’t invent them.

tests/phpunit/elementor/modules/mcp/abilities/appliers/v3/test-v3-dynamic-hoister.php
```diff
@@ -0,0 +1,150 @@
...
+		);
+
+		// Assert.
+		$this->assertSame( [], $result['primitives'] );
+		$this->assertArrayHasKey( 'title', $result['shortcodes'] );
+		$this->assertStringContainsString( 'name="post-title"', $result['shortcodes']['title'] );
+		$this->assertSame( [], $result['errors'] );
+	}
+
+	public function test_hoist__accepts_nested_dynamic_on_url_control_property() {
```
IshayMaya, line 55: Please add two tests: one for nested URL + extra fields, and one for a control that has only `dynamic.default`.

tests/phpunit/elementor/modules/mcp/abilities/appliers/v3/stub-dynamic-tags-manager.php
```diff
@@ -0,0 +1,59 @@
...
+
+namespace Elementor\Testing\Modules\Mcp\Abilities\Appliers\V3;
+
+use Elementor\Core\DynamicTags\Manager;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+class Stub_Dynamic_Tags_Manager extends Manager {
```
IshayMaya, line 11: Extending `Manager` with an empty constructor is a bit fragile. A small stub interface would be safer. Not blocking.

IshayMaya (commented):
Review notes on V3 dynamic-tag coercion. Main concerns: dropped URL sibling fields, mismatched `is_dynamic_capable` checks, and creating the hoister inside the loop.

modules/mcp/static-resources/abilities/build-composition.md
```diff
@@ -119,6 +119,7 @@ Read [elementor://global-classes] before composing. Create or update via `elemen
 - Put the plain dynamic object at that node, in place of the static variant. Read [elementor://dynamic-tags] for allowed tag names and each tag's settings schema.
 - Plain dynamic shape: `{ "name": "<allowed tag>", "settings": { ... } }`
 - Example (image `src`): `"image": { "src": { "name": "<image tag>", "settings": { ... } }, "size": "full" }`
+- The tag's categories must intersect the categories declared by the field (visible in the widget schema's dynamic branch). Pick a tag from [elementor://dynamic-tags] whose category list overlaps.
```
IshayMaya: Nice note. Please also say that a category mismatch returns an error and skips that field.

modules/mcp/abilities/appliers/v3/v3-dynamic-hoister.php
Omerisra6: Fixed: after hoisting nested dynamic on a property slot, sibling fields (e.g. is_external) are kept in primitives via extract_primitive_remainder().

modules/mcp/abilities/utils/v3-json-schema-builder.php
Omerisra6: Fixed: V3_Json_Schema_Builder now delegates to V3_Dynamic_Resolver::is_dynamic_capable(), so dynamic.default-only controls get the anyOf branch too.

modules/mcp/abilities/appliers/element-config-applier.php
Omerisra6: Intentional: array_merge(existing, new) so hoisted shortcodes from element_config win on key conflict (LLM overrides). PR wording updated to match.

modules/mcp/abilities/appliers/element-config-applier.php
Omerisra6: Fixed: hoister is lazy-initialized once via get_v3_dynamic_hoister(); optional constructor injection for tests unchanged.

modules/mcp/abilities/appliers/v3/v3-dynamic-hoister.php
Omerisra6: Fixed: tag ID is now deterministic (md5 of widget_type + key + tag name + settings JSON).

modules/mcp/abilities/appliers/v3/v3-dynamic-resolver.php
Omerisra6: Fixed: extracted normalize_dynamic_input() and reused it for top-level and nested paths.

modules/mcp/abilities/utils/v3-json-schema-builder.php
Omerisra6: Fixed: dynamic branch sets additionalProperties: false on the tag object.

tests/phpunit/elementor/modules/mcp/abilities/appliers/v3/test-v3-dynamic-hoister.php
Omerisra6: Added tests for nested URL + sibling fields and dynamic.default-only controls.

modules/mcp/static-resources/abilities/build-composition.md
Omerisra6: Added: category mismatch documented as returning an error and skipping that field.

IshayMaya (approved):

PR #36923, opened 2026-08-17
Fix: Improve shadow handling from LLM [ED-25055]

Omerisra6:
## Summary
- Widen `Box_Shadow_Property_Converter` to treat CSS-wide clearing keywords (`initial`, `inherit`, `unset`, `revert`, `revert-layer`) and unitless `0` as an empty `box-shadow` prop, avoiding a `custom_css` fallback that Pro 3.35+ strips.
- Make multi-layer parsing resilient: keep valid shadow layers and drop invalid ones instead of declining the entire declaration.
- Add shadow authoring guidance to `build-composition.md` so LLMs write convertible `box-shadow` values and avoid unsupported `var(...)` wrappers.

## Jira
https://elementor.atlassian.net/browse/ED-25055

## Test plan
- [x] `tests/phpunit/run-unit.sh tests/phpunit/elementor/modules/atomic-widgets/css-converter/converters/test-box-shadow-property-converter.php` (25 tests)
- [ ] Build a composition via MCP with `box-shadow: 0 20px 60px rgba(0,0,0,0.15)` and confirm the shadow renders in the editor
- [ ] Confirm `box-shadow: inherit` / `unset` clears the shadow instead of falling back to `custom_css`
- [ ] Confirm a mixed multi-layer value keeps the valid layers

## 1. Problem & Context

LLM responses were generating invalid `box-shadow` values with `var(...)` wrappers, which Elementor doesn't support. Documentation clarifies the constraint to prevent composition failures (ED-25055).

## 2. What Changed (Where)

- `build-composition.md`: Added constraint note that `box-shadow` only accepts literal values, no CSS variables.

## 3. How It Works

Guidance update prevents LLM from wrapping shadow values in `var(...)` during composition generation, ensuring Elementor accepts the output without fallback to stripped `custom_css`.

## 4. Risks

None — documentation-only change. Existing compositions unaffected; future LLM calls gain proper constraints.

### DIFF
--- a/modules/mcp/static-resources/abilities/build-composition.md
+++ b/modules/mcp/static-resources/abilities/build-composition.md
@@ -81,6 +81,7 @@ Some elements have internal tree structures (nesting). When using these elements
 - style is a plain CSS string (e.g. `color: red; padding-top: 1rem;`); supports `&:hover`/`&:focus`/`&:active` nesting and `@media (--breakpoint)` blocks (e.g. `@media (--mobile) { font-size: 2rem; }`); the server converts it to native styles. **Use Elementor breakpoint names only** (`--mobile`, `--tablet`, `--laptop`, etc.) — raw pixel queries like `@media (max-width: 768px)` are NOT converted to variants and fall back to `custom_css`, which is stripped by Pro 3.35+.
 - classes is configuration-id → array of existing global class **labels** from [elementor://global-classes]
 - **CSS shorthand properties may fall back to custom_css which is stripped by Pro 3.35+; prefer longhand properties (e.g., `padding-top`, `padding-right` instead of `padding`)**
+- **box-shadow**: literal values only — `var(...)` wrappers are not supported.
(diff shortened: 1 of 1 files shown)


modules/atomic-widgets/css-converter/converters/box-shadow-property-converter.php
```diff
@@ -38,12 +38,30 @@
...
+ *     every layer is invalid, so a single malformed layer no longer wipes an otherwise usable
+ *     shadow stack.
  */
 class Box_Shadow_Property_Converter extends Property_Converter_Base {
-	const INSET_KEYWORD = 'inset';
-	const DEFAULT_COLOR = 'currentColor';
-	const ZERO_SIZE     = [
+	const INSET_KEYWORD    = 'inset';
+	const DEFAULT_COLOR    = 'currentColor';
+	const CLEARING_KEYWORDS = [
```
ronros-elementor: if we'll remove these, we won't hve any way to prevent an inherited style field value of affecting the final widget style

e.g. we use a global class with many style fields, and we wish to revert only the box-shadow on a specific widget.
we would do that by setting on *local* the corresponding style field with `"unset"`

by setting it null, it means we're omitting this box shadow field entirely from local

IshayMaya (approved):

PR #36955, opened 2026-08-18
Internal: Expose default styles in MCP [ED-25294]

Omerisra6:
## PR Type
- [x] Feature

## Summary

Surface the active kit's site-wide default styles inside `elementor/get-page-structure` (per-element via `Styles_Renderer` layers), and add a new `elementor/manage-default-styles` MCP tool so agents can bulk update/delete the site-wide default style for any allowed HTML wrapper tag.

## Description

**`elementor/get-page-structure`** — when `include_content=true`, each V4 node now also returns:
- `tag`: the resolved rendered HTML wrapper tag (from `settings.tag` or the widget prop-schema default), filtered to `Utils::ALLOWED_HTML_WRAPPER_TAGS`.
- `default_styles`: a **raw CSS string** containing the widget's base layer followed by the kit's site-wide default layer for that tag, in browser cascade order. Each layer is rendered via the production `Styles_Renderer` (same pipeline as frontend enqueue), so the output includes selectors (e.g. `.e-heading-base`, `.e-default-h1`), `@media(--breakpoint)` blocks, pseudo-states, and custom CSS — not a flattened prop map.

New utilities:
- `Element_Tag_Resolver` — resolves the rendered tag from settings or schema default.
- `Element_Default_Styles_Builder::render()` — renders widget `base_styles` and kit default for the tag through `Styles_Renderer`, then concatenates in cascade order.

**`elementor/manage-default-styles`** (new tool) — bulk ops mirroring `elementor/manage-classes` but keyed by HTML tag:
- Actions: `update` (upsert per tag with raw CSS, patch/replace modes) and `delete`.
- Reuses `Style_Variants_Merger::parse_css_string` + `apply_mode` and the same CSS-converter pipeline; supports `@media(--breakpoint)` blocks and `&:hover/&:focus/&:active` states.
- Backed by `Default_Styles_Repository::put` / `delete`; `manage_options` capability; MAX 20 ops per call.

## Test instructions

1. `bash tests/phpunit/run-unit.sh tests/phpunit/elementor/modules/mcp/abilities/utils/test-element-tag-resolver.php tests/phpunit/elementor/modules/mcp/abilities/utils/test-element-default-styles-builder.php tests/phpunit/elementor/modules/mcp/abilities/test-manage-default-styles-ability.php --filter test_` — 23 tests, all pass.
2. In the editor, save a kit default style for `h1`, then call `elementor/get-page-structure` with `include_content=true` on a page containing an `e-heading`. Verify each heading node has `tag` and `default_styles` as a CSS string that includes both base selectors and `.e-default-h1` rules.
3. Call `elementor/manage-default-styles` with a mixed batch and confirm per-op results plus kit changes on the frontend.

## Quality assurance

Fixes ED-25294

## 1. Problem & Context

Expose computed HTML tags and default styles in MCP (ED-25294). Currently, atomic elements don't expose their dynamically computed tags or kit-level default styles through the MCP API, blocking external tools from inspecting element cascade and render structure.

## 2. What Changed (Where)

| Layer | Change |
|-------|--------|
| **Atomic Elements** | Added `get_computed_html_tag()` static method to all 30+ concrete elements; replaced instance `define_default_html_tag()` with class contract. |
| **Base Classes** | `Atomic_Element_Base` and `Atomic_Widget_Base` now populate `default_html_tag` and `html_tag_follows_link` config keys; `Has_Atomic_Base` trait provides defaults. |
| **MCP Abilities** | New `Get_Default_Styles_Ability` and `Manage_Default_Styles_Ability` for read/write; enhanced `Get_Structure_Ability` to include tag and default_styles in responses. |
| **Rendering** | New `Html_Tag_Computer` utility centralizes tag resolution (settings.tag → link.tag → default, with link-follow opt-out). Applied to templates and TS renderer. |
| **Templates** | Removed inline tag logic from Twig; now passed as computed `tag` variable from context. |
| **Tests** | Contract tests ensure all concrete elements declare `get_computed_html_tag()`; fixtures shared PHP/TS. |

## 3. How It Works

1. **Tag Resolution**: `Html_Tag_Computer::compute()` evaluates settings in precedence order—link href (if active + followLink=true) → settings.tag → default. Same logic mirrored in TS.
2. **Config Exposure**: Element classes declare static `get_computed_html_tag(array $settings)` returning computed tag; base classes call it at config build time, storing `default_html_tag` + `html_tag_follows_link` in element config.
3. **MCP API**: `Get_Structure_Ability` now resolves each V4 node's tag via `Element_Tag_Resolver` and includes raw CSS (`default_styles` = widget base layer + kit site-wide default for that tag) in responses.
4. **Template Consumption**: Twig/TS renderers receive pre-computed `tag` in context; no longer calculate inline, simplifying templates and keeping render logic centralized.

## 4. Risks

**Behavioral**: Some elements (Heading, Paragraph, Image) opt out of link-following (`html_tag_follows_link=false`) to maintain semantic tags—ensure editor/frontend agree on precedence. **Breakage**: Removed `define_default_html_tag()` instance method; any custom elements overriding it will fail at runtime—requires audit of plugins. **Performance**: MCP `Get_Structure_Ability` now resolves tags + renders default styles CSS per node; pagination/depth limits may be needed for large documents.


modules/atomic-widgets/elements/atomic-button/atomic-button.php
```diff
@@ -44,6 +44,10 @@ public function get_icon() {
 		return 'eicon-e-button';
 	}
 
+	protected static function define_default_html_tag() {
+		return 'button';
```
IshayMaya: maybe we should align the twigs to use those (and enforce it) ? to be on the safer side

modules/atomic-widgets/elements/atomic-heading/atomic-heading.php
```diff
@@ -45,6 +45,16 @@ public function get_icon() {
 		return 'eicon-e-heading';
 	}
 
+	protected static function define_default_html_tag() {
+		return 'h2';
+	}
+
+	public static function get_computed_html_tag( array $settings ): string {
```
IshayMaya, line 53: why do we need to compute on this level?

modules/atomic-widgets/elements/atomic-image/atomic-image.php
```diff
@@ -44,6 +44,14 @@ public function get_icon() {
 		return 'eicon-e-image';
 	}
 
+	protected static function define_default_html_tag() {
+		return 'img';
+	}
+
+	public static function get_computed_html_tag( array $settings ): string {
+		return static::define_default_html_tag();
```
IshayMaya: why?

modules/atomic-widgets/elements/atomic-paragraph/atomic-paragraph.php
```diff
@@ -46,6 +46,16 @@ public function get_icon() {
 		return 'eicon-paragraph';
 	}
 
+	protected static function define_default_html_tag() {
+		return 'p';
+	}
+
+	public static function get_computed_html_tag( array $settings ): string {
+		$tag = static::extract_html_tag_value( $settings['tag'] ?? null );
```
IshayMaya: seems like we are repeating the same behavior with this function, i think it can just accept the settings and return the resolved value

modules/atomic-widgets/elements/base/has-html-tag.php
```diff
@@ -0,0 +1,74 @@
...
+	exit;
+}
+
+trait Has_Html_Tag {
+
+	protected static function define_default_html_tag() {
+		return 'div';
+	}
+
+	public static function get_computed_html_tag( array $settings ): string {
```
IshayMaya: i'm not sure if its the naming that is confusing or the split in the utils but we have 2 extract and 1 compute for similar actions (some link related, some not) try to make it more clear

modules/atomic-widgets/elements/base/atomic-element-base.php
```diff
@@ -97,7 +97,7 @@ protected function get_initial_config() {
 		$config['children_dependencies'] = $this->get_children_dependencies_config();
 		$config['initial_attributes'] = $this->define_initial_attributes();
 		$config['include_in_widgets_config'] = true;
-		$config['default_html_tag'] = $this->define_default_html_tag();
+		$config['default_html_tag'] = static::define_default_html_tag();
```
IshayMaya: can we enforce all elements that extend it to pass the tag

modules/mcp/abilities/utils/element-default-styles-builder.php
```diff
@@ -0,0 +1,59 @@
...
+	): string {
+		$renderer = $renderer ?? Styles_Renderer::make( Plugin::$instance->breakpoints->get_breakpoints_config() );
+
+		$base_css = ! empty( $widget_base_style_defs )
+			? $renderer->render( array_values( $widget_base_style_defs ) )
+			: '';
+
+		$default_css = self::render_kit_default( $tag, $repository, $renderer );
+
+		return trim( $base_css . "\n" . $default_css );
```
IshayMaya, line 37: do we mirror some other styles behavior? maybe we can extract some logic to mutual utils

modules/atomic-widgets/elements/atomic-button/atomic-button.php
Omerisra6: Addressed in a0b65e6: `Has_Template` / `Has_Element_Template` now inject a precomputed `tag` into the Twig context via `static::get_computed_html_tag( $settings )`, and the tag-aware wrapper twigs (button, heading, paragraph, svg, div-block, flexbox, grid) use `{{ tag | e('html_tag') }}` instead of `settings.tag`.

Elements with fixed semantic wrappers (`hr`, `img`, tab `button`, etc.) still hardcode the tag in Twig, but it matches their `get_computed_html_tag()` default. Happy to migrate the rest and add a Twig contract test if you want full enforcement.

modules/atomic-widgets/elements/atomic-heading/atomic-heading.php
Omerisra6: This isn't custom logic — it delegates to `Html_Tag_Computer::compute()`. We declare it on heading/paragraph/image only to pass two things the trait default doesn't know:

1. The element-specific default (`h2` / `p` / `img`)
2. `FOLLOW_LINK_OPTION => false` — on these widgets the link wraps *inner* content, so the outer wrapper tag must stay the heading/paragraph tag (or `img`), not become the link tag.

MCP and Twig both call the same static method, so there's a single source of truth.

modules/atomic-widgets/elements/atomic-image/atomic-image.php
Omerisra6: Same pattern as heading: the wrapper is always `img`; when a link is present it wraps the `` in a separate `<a>`, so we disable link-following and pass `img` as the default. One-liner over `Html_Tag_Computer::compute()` — no element-specific resolution logic.

modules/atomic-widgets/elements/atomic-paragraph/atomic-paragraph.php
Omerisra6: Agreed — consolidated in a0b65e6. The paragraph override is now just `Html_Tag_Computer::compute( $settings, 'p', [ FOLLOW_LINK_OPTION => false ] )`; all settings/link/default resolution lives in `Html_Tag_Computer`.

modules/atomic-widgets/elements/base/has-html-tag.php
Omerisra6: Renamed and split in a0b65e6: `Has_Html_Tag` → `Html_Tag_Computer` with clearer responsibilities:

- `compute()` — main entry (link → settings.tag → default)
- `extract_html_tag_value()` — unwrap string / transformable envelope
- `extract_link_html_tag()` — resolve inner link tag
- `settings_have_active_link()` — link detection (href or rendered attributes)

Tests moved to `test-html-tag-computer.php`.

modules/atomic-widgets/elements/base/atomic-element-base.php
Omerisra6: Done in a0b65e6: every concrete atomic element now declares `public static function get_computed_html_tag()`, config uses `static::get_computed_html_tag( [] )`, and `test-computed-html-tag-contract.php` scans the elements tree and fails CI if any concrete element is missing the method. Chose test enforcement over PHP `abstract` to avoid breaking mid-refactor.

modules/mcp/abilities/utils/element-default-styles-builder.php
Omerisra6: It reuses the production `Styles_Renderer` (same as `Atomic_Styles_Manager::render_css()`), but the layer-concat + kit-default lookup is MCP-specific — agents need widget base + site-wide default as one raw CSS string in cascade order, which frontend enqueue doesn't do.

Kept it as a thin MCP util rather than extracting shared code, since the only overlap is `$renderer->render( $defs )`. Happy to extract a small shared helper if you feel strongly — otherwise I'd keep it local to avoid coupling MCP to the enqueue pipeline.

modules/atomic-widgets/elements/atomic-background-video/atomic-background-video-controls/atomic-background-video-controls.php
```diff
@@ -58,6 +59,10 @@ public function should_show_in_panel() {
 		return false;
 	}
 
+	public static function get_computed_html_tag( array $settings ): string {
+		return Html_Tag_Computer::compute( $settings, 'div' );
```
IshayMaya: did we not have these in the previous approach? (on this pr)

IshayMaya (dismissed):

IshayMaya (dismissed):

IshayMaya (approved):

PR #36968, opened 2026-08-19
Internal: Remove 'New' text from V4 Elements panel category chip [ED-25299]

avivu:
## Summary
- Removes the "New" text label from the V4 Elements panel section header chip
- Keeps the ⓘ (`eicon-info`) icon and the React promotion wrapper
- Removes the now-unused `margin-inline-start` from the chip's icon style

## Jira
[ED-25299](https://elementor.atlassian.net/browse/ED-25299)

## Test plan
- [ ] Open the editor
- [ ] Check the "V4 Elements" section header in the widget panel — should show only the ⓘ icon (no "New" text)
- [ ] Verify clicking the ⓘ icon still works as expected

[ED-25299]: https://elementor.atlassian.net/browse/ED-25299?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## 1. Problem & Context

Removes "New" text label from V4 Elements panel category chip (ED-25299), keeping only the icon. Reduces visual clutter while maintaining accessibility via aria-label.

## 2. What Changed (Where)

- `module.php`: Moved "New" text from visible span to aria-label attribute; restructured chip markup with icon wrapper.
- `_controls.scss`: Removed 4px margin rule for icon (no longer needed in new layout).

## 3. How It Works

The chip now displays only the info icon visually, with the "New" label hidden but accessible to screen readers via aria-label. Icon spacing rules were removed since the new structure doesn't require them.

## 4. Risks

None identified. Icon-only chips are common patterns; aria-label ensures accessibility compliance. Styling cleanup is safe.


modules/atomic-widgets/module.php
```diff
@@ -553,8 +553,8 @@ private function enqueue_scripts() {
 
 	private function render_panel_category_chip() {
 		?><# if ( 'v4-elements' === name )  { #>
-		<span class="elementor-panel-heading-category-chip">
-				<?php echo esc_html__( 'New', 'elementor' ); ?><i class="eicon-info"></i>
+		<span class="elementor-panel-heading-category-chip" aria-label="<?php echo esc_attr__( 'New', 'elementor' ); ?>">
```
Ntnelbaba, line 587: I'm sure that this is not the only needed deletion
If the new not exists now there is a lot of logic out there that needs to be deleted and a lot of screenshot tests to fix

avivu: Checked this - the chip element itself stays, so the surrounding JS, CSS, and Playwright tests all remain valid. The only thing removed is the 'New' text inside it. No orphaned logic, no screenshot tests to fix.

Ntnelbaba (approved):

avivu:
Superseded by #37047 which combines this with ED-25328 (New badge for Accordion + List). Snapshot updated there.

PR #36984, opened 2026-08-19
Internal: Persist dependency cache key through save step [ED-24997]

Ntnelbaba:
## Summary
- Save the dependency cache with `cache-primary-key` from the restore step so install and downstream jobs use the same key (lockfiles change during `npm ci`, so re-hashing at save time produced a different key).
- Remove `fail-on-cache-miss` from downstream PR and Playwright suite jobs that only restore the install cache.

## Jira
ED-24997

## Test plan
- [ ] Open a PR and confirm `install` saves cache under the same key it restored
- [ ] Confirm `build`, lint, jest, and qunit jobs restore that cache successfully
- [ ] Confirm Playwright suite jobs restore the install cache without hard-failing on miss

## 1. Problem & Context

Significant refactor addressing ED-24997: Persist dependency cache key through save step. Major architectural changes to onboarding workflow, CSS conversion pipeline (null-value handling), atomic widgets, global classes kit management, and security/sanitization flow.

## 2. What Changed (Where)

| Category | Changes |
|----------|---------|
| **Onboarding** | Removed theme-selection-conditional logic; always show theme step. Replaced `hello-biz` with `hello-elementor` only. Flattened 5-step flow to 3-step. Added `isHelloThemeActive` flag. |
| **CSS Converter** | Added null-value reset semantics throughout: dedupe rules, expand null to all longhands, cleanup collapsed nulls. New `Outline_Shorthand_Expander`. Changed border/radius keys from `border-width` → `border-width-v2`, etc. |
| **Atomic Widgets** | Refactored property converters: split `public convert()` into protected `do_convert()` + new `convert_null()` override pattern. Added `Preserves_Children_Subtree` trait + `Preserved_Element` for promotion data durability. Added webmcp form context. |
| **Global Classes** | New `Migrate_All_Kits_Post_IDs` migration (v4): ensures exclusive kit ownership of post IDs. Removed lazy backfill, added fallback-path support in `Migrations_Loader`. Added kit cloning on new-kit-create. |
| **Security/Sanitization** | Removed `Plugin::sanitize_post_data()` REST filter. Centralized in `Document::save()` + `Elementor_Post_Meta` callbacks using new `Utils::kses_post_deep()`. |
| **Misc** | Icon updates (loop: `eicon-loop-builder` → `eicon-loop-widget`). "Twitter" → "X" social share. Pro detection now uses `Utils::is_license_active()`. Menu item refactor (flyout exclusions, event IDs). |

## 3. How It Works

**CSS Null Semantics** (core complexity):
- Parser converts `'null'` string → PHP `null` before expanders run
- Shorthands (`border: null`) expand to per-longhand nulls via `expand_null()` override
- Converters call base `convert()`, which dispatches to `convert_null()` (custom reset) or `do_convert()` (normal path)
- After conversion, `cleanup_props()` recursively collapses objects where all leaves are null → single `null` at prop level
- Last-wins deduplication at CSS parse: identical properties keep only final value

**Kit Migration Strategy**:
- Pass A: Run `Migrate_To_Posts` on all kits (idempotent via order check)
- Pass B: Fill missing class_id → post_id mappings, conflict-aware (never reuse claimed post_id)
- Pass C: Break shared post_ids by cloning posts; lowest-ID kit owns, others get copies

**Atomic Form WebMCP** (optional feature):
- Only enabled when `!is_edit_mode()` and `webmcp-autosubmit` prop set
- Builds tool name/description from form-name; emits as HTML attributes

## 4. Risks

| Risk | Mitigation |
|------|-----------|
| **Null cleanup propagation bugs** | Recursive collapse logic can silently drop partial data if any nested leaf is null. Extensive test coverage (21 new tests) validates permutations. |
| **Kit migration data loss** | Pass A idempotency check could miss re-migrated kits. Conflict resolver re-queries all claimed post_ids each pass; verified by Pass C dedup check. |
| **Icon/text regressions** | String replacements (`eicon-loop-builder`, Twitter→X) may break CSS or frontend matching. No schema version bump; assumes consumers sync. |
| **CSS converter interface breakage** | `public convert()` → protected `do_convert()` forces all custom converters to update. Pre-existing pattern, but migration surface is 30+ converter classes. |


Ntnelbaba:
Closing: opened from cp/36965-to-4.02 by mistake. Recreating from fix/ED-24997-deps-cache-save-key.

PR #36999, opened 2026-08-20
Internal: [V4] add to SVG support for icons [ED-25018]

ronenboxer:
## 1. Problem & Context

Add icon library support to SVG control, enabling font icon selection alongside uploaded SVGs. Ticket: ED-25018.

## 2. What Changed (Where)

| Component | Change |
|-----------|--------|
| **SVG Control (PHP)** | Added `show_icon_library` flag; `Svg_Control::set_show_icon_library()` method |
| **Icon Prop Type** | New `Icon_Prop_Type` class; validates icon value + library strings |
| **Icon Transformer** | PHP transformer converts icon metadata to inline SVG via Font Awesome data |
| **Atomic Module** | Registered `EXPERIMENT_ICON_LIBRARY` experiment; `is_svg_library_active()` helper |
| **SVG Media Control (TS/React)** | Extracted `processSvgContent()` utility; added icon library button + preview |
| **Icon Transformer (TS)** | Fetches Font Awesome JSON, builds SVG, sanitizes output |
| **Migrations** | Added `svg-src-to-icon` bidirectional migration |
| **Tests** | 7 new test files covering prop types, transformers, controls |

## 3. How It Works

User selects SVG control → if library active, "Icon library" button appears → clicking opens icon manager → selecting font icon calls `Icon_Transformer` → fetches Font Awesome JSON, builds inline SVG, returns HTML. SVG uploads still work unchanged. Migration handles backward compat (icon → svg-src on downgrade).

## 4. Risks

Icon library fetches external Font Awesome JSON files—network failures degrade silently to null HTML. Sanitization via DOMPurify mitigates XSS. SVG + icon union prop type could confuse users; UI makes distinction clear with separate buttons.


packages/packages/core/editor-canvas/src/transformers/shared/__tests__/icon-transformer.test.ts
```diff
@@ -0,0 +1,74 @@
...
+		expect( result ).toEqual( { html: null, url: null } );
+	} );
+
+	it( 'returns null html when value or library is missing', async () => {
+		// Act.
+		const result = await iconTransformer( { value: 'fas fa-star' }, { key: 'svg' } );
+
+		// Assert.
+		expect( result ).toEqual( { html: null, url: null } );
+	} );
```
davseve, line 73: Can we add a test for when the next tag is [not SVG](https://github.com/elementor/elementor/pull/36999/changes#diff-420547d4ce526f82efd65f2ca5198f9de78042255ceeff419c60cac885c35022R62)?

tests/phpunit/elementor/modules/atomic-widgets/elements/test-atomic-svg-icon-library-experiment.php
davseve, line 1: Do we need this file? tests for the experiment?
I think that the experiments are already covered.

davseve (dismissed):

packages/packages/core/editor-canvas/src/transformers/shared/__tests__/icon-transformer.test.ts
ronenboxer: AI's response
No. Not in icon-transformer.test.ts.

iconTransformer always builds markup with buildFontAwesomeSvg(), which is hard-coded as <svg>…</svg>. After a successful FA lookup, processSvgContent always sees an svg tag. You cannot hit “next tag is not svg” without mocking processSvgContent or lying about FA JSON. That would not test the PHP next_tag( 'svg' ) branch.

That branch lives in processSvgContent, and it is already covered through svgSrcTransformer:

returns null html when content is not SVG
returns null html when response has no svg element
Same helper, same querySelector('svg') miss → null.

PHP Icon_Transformer::process_svg() is the same: build_svg() only emits <svg> when path is non-empty; empty path already returns '' before the processor. A dedicated “no svg tag” PHPUnit would only fire if sanitizer stripped the tag — not worth a fake fixture.

Do not add this case to the icon transformer tests. If you want it more visible, a small processSvgContent test is the right file; it is not required for this PR.

tests/phpunit/elementor/modules/atomic-widgets/elements/test-atomic-svg-icon-library-experiment.php
ronenboxer: removed

davseve (approved):

PR #36953, opened 2026-08-18
Internal: Add component variants data model [ED-25267]

IshayMaya:
## PR Checklist
- [x] The commit message follows our guidelines:  https://github.com/elementor/elementor/blob/master/.github/CONTRIBUTING.md

## PR Type
What kind of change does this PR introduce?
- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [x] Other... Please describe: Hidden experiment + PHP data model for Component Variants (Phase 0).

## Summary

This PR can be summarized in the following changelog entry:

* Internal: Add Component Variants data model behind a hidden experiment [ED-25267]

## Description
An explanation of what is done in this PR

* Adds hidden `e_component_variants` experiment (dev, inactive by default) in the components module.
* Stores named variants on the Component document as `_elementor_component_variants`.
* Adds `VARIANTS_META_KEY` to `COMPONENT_CUSTOM_META_KEYS` so autosave variants are promoted on publish (same path as overridable props).
* Adds parsers/DTOs under `modules/components/variants/` with additive-only `classes.add` semantics.
* Rejects nested `settings.variant`; nested pinning lives as a sibling `variant` field.
* Wires `elementor/document/after_save` → `set_component_variants` when the experiment is active, gated by `can_edit()`.
* Variant ids are format-checked (`v_` + 8 base36 chars); generation stays client-side, matching overridables.

## Test instructions
This PR can be tested by following these steps:

* Enable `e_atomic_elements` and `e_component_variants` experiments.
* Run PHPUnit: `vendor/bin/phpunit --filter Test_Component_Variants_Parser` and `--filter Test_Component_Variants_Meta`.
* Confirm save with `settings.variants` persists `_elementor_component_variants`.
* Confirm invalid payloads (`settings.variant`, duplicate ids, invalid class ids) are rejected and do not write meta.
* Confirm `VARIANTS_META_KEY` is listed in `COMPONENT_CUSTOM_META_KEYS` (publish/autosave promotion).

## Quality assurance

- [x] I have tested this code to the best of my abilities
- [x] I have added unittests to verify the code works as intended
- [ ] Docs have been added / updated (for bug fixes / features)

Fixes ED-25267

Made with [Cursor](https://cursor.com)

[...]

PR #37000, opened 2026-08-20
Internal: Add component variants class-collection filter [ED-25268]

IshayMaya:
## PR Checklist

- The commit message follows our guidelines: https://github.com/elementor/elementor/blob/master/.github/CONTRIBUTING.md

## PR Type

- Other: Hidden experiment infrastructure for Component Variants (Phase 0).

## Summary

* Internal: Wire component variant class ids into the global-classes usage/CSS pipeline behind the hidden `e_component_variants` experiment [ED-25268](https://elementor.atlassian.net/browse/ED-25268)

## Description

Follow-up to #36953 (ED-25267, merged). This PR targets `main` directly.

Today `Global_Classes_Relations::extract_class_ids_from_post()` walks the document's elementor data and only picks up class ids from atomic element `classes` settings. Variant class ids live in `_elementor_component_variants` meta at `variants[].widgets[].settings.classes.add[]` — the collector never sees them, so variant classes are missing from the frontend/preview CSS bundle produced by `Atomic_Global_Styles`.

This PR:
- Adds a new filter `elementor/global_classes/extract_class_ids_from_post` inside `Global_Classes_Relations::extract_class_ids_from_post()` so external modules can contribute additional class ids for a post before dedup.
- Adds a small `Component_Variant_Class_Collector` helper that flattens variant class ids out of a `Component_Variants` DTO.
- Subscribes to the new filter from `Components\Module` (inside the existing variants experiment gate) — contributes variant class ids only for `Component_Document` posts.
- Bumps `set_component_variants` from priority 10 → **9** on `elementor/document/after_save` so variant meta is persisted before `Global_Classes_Relations::on_document_save` (priority 10) reads it via the new filter.

Nested-component CSS aggregation is unchanged — parents still compose child CSS at render time via `elementor/document/related_posts`. Each post's own `_elementor_used_global_class` is now complete after a single save (including variant classes), so ancestor aggregation naturally picks them up.

## Test instructions

- Enable `e_atomic_elements` and `e_component_variants` experiments.
- Run PHPUnit:
  - `vendor/bin/phpunit --filter Test_Component_Variant_Class_Collector`
  - `vendor/bin/phpunit --filter Test_Component_Variants_Meta`
- Save a Component with a variant declaring `settings.classes.add: ["<some-global-class>"]` and confirm `_elementor_used_global_class` for the component post includes the variant class ids.
- Confirm non-component posts are unaffected: filter returns `$ids` unchanged.

## Quality assurance

Fixes ED-25268

[ED-25268]: https://elementor.atlassian.net/browse/ED-25268?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## 1. Problem & Context

Component variants now need to contribute class IDs to global class tracking. Without this, variant-added classes won't be indexed in post metadata, breaking class deduplication and management.

## 2. What Changed (Where)

- **module.php**: Added `Component_Variant_Class_Collector` import; added `SAVE_VARIANTS_PRIORITY=9` constant; wired `set_component_variants()` action and `add_variant_class_ids()` filter at priority 9 (before Global_Classes_Relations @10).
- **component-variant-class-collector.php**: New collector class extracts class IDs from variant widget settings.
- **global-classes-relations.php**: Refactored `collect_class_ids_from_post()` to apply `extract_class_ids_from_post` filter *after* element iteration, enabling hook-based ID injection.
- **Tests**: Added unit tests for collector and three integration tests validating end-to-end variant class indexing.

## 3. How It Works

On document save, `set_component_variants()` persists variant metadata (priority 9). `Global_Classes_Relations::collect_class_ids_from_post()` (priority 10) then iterates elements, applies the filter hook, which calls `add_variant_class_ids()` to collect variant class IDs via `Component_Variant_Class_Collector`. This merges variant classes into the post's used-classes metadata before Global_Classes_Relations stores it.

## 4. Risks

Priority ordering is critical—if `set_component_variants()` runs *after* `Global_Classes_Relations`, variants won't be persisted when the filter fires. Mitigated by explicit constant and test coverage validating hook execution order. Filter on non-component posts is safe (early return on type check).


modules/components/variants/component-variant-class-collector.php
Omerisra6, line 1: Maybe instead of applying the styles on the inner widgets lets create a new layer of context (before render like we did in the loop) and then the widgets will read the active variant and apply the classes accordingly

modules/global-classes/global-classes-relations.php
```diff
@@ -278,16 +278,20 @@ private function extract_class_ids_from_post( int $post_id ): array {
...
 		$elements_data = $document->get_elements_data();
 
-		if ( empty( $elements_data ) ) {
-			return [];
+		if ( ! empty( $elements_data ) ) {
+			Plugin::$instance->db->iterate_data( $elements_data, function ( $element_data ) use ( &$used_class_ids ) {
+				$used_class_ids = array_merge(
+					$used_class_ids,
+					Atomic_Elements_Utils::collect_class_ids_from_element_data( $element_data )
+				);
```
Omerisra6, line 286: maybe we can collect the classes here when we already iterating instead of iterating twice

IshayMaya: they are not the same iteration, `iterate_data` walks `$elements_data` (the element tree), while Component_Variant_Class_Collector walks _elementor_component_variants post meta, which isn't in scope inside the tree visitor.

modules/components/variants/component-variant-class-collector.php
IshayMaya: this is for save-time indexing, not render
 a pre-render context is the right place to decide which variant is active for this instance and merge classes onto the DOM (Phase 1 / ED-25274). It's the wrong place to declare which classes the component post can possibly use, that has to be known at save so the bundle and the reverse index are correct.

Omerisra6 (approved):

PR #37017, opened 2026-08-23
Fix: Show section indicator dots in default styles panel [ED-25259]

Omerisra6:
## Summary

- Include the active style ID from `useStyle()` in `useAppliedStyles()` so the inheritance snapshot covers the tag being edited in the default styles panel.
- Section dots (gray indicator on collapsed sections with defined values) now appear in the Design System → Defaults panel, matching behavior in the element Style tab.

## Root cause

The default styles panel reuses `StyleSections` and `StyleInheritanceProvider`, but `useAppliedStyles()` only gathered style IDs from element data (base styles, applied classes, preview tag). The edited tag ID lives in `StyleProvider.id` and was never included in the snapshot, so `StylesInheritanceSectionIndicators` had nothing to render.

## Test plan

- [ ] Open Design System → Defaults, select H1, set a Typography color, collapse Typography → gray dot appears on the section header
- [ ] Clear the value → dot disappears
- [ ] Regular element Style tab: dots still appear only for sections with values on the active class (no regression)

## Jira

ED-25259

## 1. Problem & Context

Section indicator dots weren't showing in the default styles panel because the active style wasn't included in the applicable styles filter. ED-25259 fixes this by explicitly tracking the currently active style.

## 2. What Changed (Where)

- `styles-inheritance-context.tsx`: Added `activeStyleId` to the `applicableIds` array so active styles render their section indicators
- `use-styles-inheritance-snapshot.test.ts`: Converted global mock setup to per-test setup; added test case validating active style inclusion without applied classes

## 3. How It Works

The `useAppliedStyles` hook now calls `useStyle()` to get the active style ID and includes it in the filter criteria alongside base styles, applied styles, and default tag styles. This ensures the repository filter returns the active style even if it's not in the applied classes, allowing section indicators to display. The test refactor parameterizes `activeStyleId` per test case rather than globally, enabling targeted validation of the active style behavior.

## 4. Risks

None identified. Change is additive (includes more styles in filter) with backward-compatible test structure.


packages/packages/core/editor-editing-panel/src/contexts/styles-inheritance-context.tsx
```diff
@@ -83,14 +83,20 @@ const useAppliedStyles = () => {
...
 
 	const classesProp = usePanelElementSetting< ClassesPropValue >( currentClassesProp );
 
 	const appliedStyles = classesPropTypeUtil.extract( classesProp ) ?? [];
 
-	const applicableIds = [ ...baseStyles, ...appliedStyles, ...( defaultTagStyleId ? [ defaultTagStyleId ] : [] ) ];
+	const applicableIds = [
+		...baseStyles,
+		...appliedStyles,
+		...( defaultTagStyleId ? [ defaultTagStyleId ] : [] ),
```
IshayMaya, line 97: maybe we can split the usage here since in our case - those are always empty

Omerisra6: I dont think it worth it for now, lets talk about it

IshayMaya (approved):

PR #37049, opened 2026-08-25
Fix: Portal editor notifications to body and offset by app bar height [ED-25336]

IshayMaya:
## Summary
- Portal notistack snackbars to `document.body` so they are no longer clipped by the v4 editor shell containing block.
- Offset snackbar content upward by the v4 app bar height so alerts render fully visible at the bottom of the viewport.

## Test plan
- [ ] Open the v4 editor and trigger any notification (`error`, `success`, `info`, `promotion`, or `default`).
- [ ] Confirm the notistack container is a direct child of `<body>` and the alert is fully visible above the bottom edge.
- [ ] Right-click a v3 widget (e.g. Button) → Save as component → confirm the "Components require atomic elements only…" message appears and is readable.
- [ ] `npx jest --config packages/jest.config.js packages/packages/core/editor-notifications`

Fixes ED-25336

## 1. Problem & Context

Editor notifications were hidden behind the app bar; this fix repositions them to the document body with negative top margin to account for app bar height, ensuring visibility. Addresses ED-25336.

## 2. What Changed (Where)

- **notifications.tsx**: Import `getAppBarHeight()`, apply negative top margin offset in both `DefaultCustomSnackbar` and `AlertSnackbar` components, move SnackbarProvider to document.body via `domRoot` prop
- **get-app-bar-height.ts** (new): Utility function queries editor wrapper element height; returns 0 if missing

## 3. How It Works

Notifications now render at document.body level instead of being constrained by ancestor positioning contexts. The `mt: `-${appBarHeight}px`` negative margin shifts them up by the app bar height, preventing overlap while keeping them viewport-visible. Graceful degradation: missing app bar element returns height of 0.

## 4. Risks

Minor: DOM query on every render could be optimized with memoization if notifications fire frequently. The hardcoded element ID `elementor-editor-wrapper-v2` creates brittle coupling—if that ID changes, notifications won't offset correctly.


IshayMaya:
Closing in favor of a PR opened from a branch on the main repository.

PR #37080, opened 2026-08-27
Fix: Support null overridable origin value across components render path [ED-25333]

max-zu:
ED-22351 (#34175) widened the `overridable` prop contract so that an exposed prop with no value persists `origin_value: null`, but the consumers were never updated:

- `Overridable_Transformer::is_origin_value_override()` still declared `array`, so a null origin value threw a fatal TypeError while rendering a component instance that has an override set for that key.
- The editor canvas twin `componentOverridableTransformer` dereferenced `originValue.$$type` without a null check, and `ComponentOverridable` typed `origin_value` as non-nullable while the zod schema declares it `.nullable()`.

`Overridable_Prop_Type` now also normalizes an empty `origin_value` to null, so documents saved by editors that wrote `{}` instead of null stop failing validation with `invalid_value` and get repaired on the next save.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

*

## Description

*

## Test instructions

*

## Quality assurance

Fixes #

## 1. Problem & Context

Legacy editor versions wrote empty objects `{}` instead of `null` for valueless overridable props, rendering documents unsavable on re-save. This normalizes empty origins to `null` across validation, sanitization, and render paths so the behavior is consistent regardless of save state.

## 2. What Changed (Where)

| File | Change |
|------|--------|
| `overridable-prop-type.php` | Extract `normalize_origin_value()` helper; apply in both `validate()` and `sanitize()` |
| `overridable-transformer.php` | Call `normalize_origin_value()` during render; add nullable type hint to `is_origin_value_override()` |
| `component-overridable-transformer.ts` | Mirror normalization logic; handle null origin values in override resolution |
| `types.ts` | Change `origin_value` from `TransformablePropValue<string>` to nullable variant |
| Tests (PHP/TS) | Add 9 test cases covering empty/null origins across save and render paths |

## 3. How It Works

Empty object → normalized to `null` at three gates: (1) validation rejects before save, (2) sanitization converts on save, (3) render path normalizes legacy stored `{}` before override resolution. TypeScript mirrors PHP normalization so frontend render behavior matches backend. Override detection checks `$type === 'override'` only after normalization, preventing null-reference crashes.

## 4. Risks

Type narrowing in TS (`isOriginValueOverride` type guard) assumes `originValue?.$type` is safe on null—correct here but fragile pattern. PHP version handles nullable comparisons correctly. Test coverage is comprehensive for the fix; backward compatibility confirmed by accepting both `{}` and `null` as equivalent "no value" signals.


max-zu:
All three test-coverage notes from the review are addressed on the branch (the bot reviewed the commit before the test commits landed), plus coverage beyond what was asked:

**1. Empty-array normalization** — `Test_Overridable_Prop_Type::test_validate__passes_with_empty_origin_value` and `test_sanitize__normalizes_empty_origin_value_to_null`, with `Video_Src_Prop_Type` variants, since `source` on Background Video is the reproduction case and has its own `id`/`url` validation rules.

**2. Null `origin_value` render path (PHP)** — `Test_Overridable_Transformer` covers both a matching override (returns the override value, the case that used to fatal) and no matching override (returns `null`).

**3. Null `origin_value` with active override (TS)** — `component-overridable-transformer.test.ts` covers the same two cases on the canvas side.

**Added on top:**

- `Test_Overridable_Props_Parsing` — integration coverage for the real save chain (`Props_Parser` → `Union_Prop_Type` → `Overridable_Prop_Type` → `Video_Src_Prop_Type`) with the legacy `origin_value: []` payload, asserting it validates and persists as `null`, and that a populated origin value is preserved.
- `component-overridable-prop-type.test.ts` — locks the client contract that caused the missing expose checkbox: `origin_value: null` is accepted by the zod schema, `{}` is rejected.

Run locally: `--filter Overridable` → 116 tests, 303 assertions, OK. Jest `component-overridable*` → 9/9.

### On the normalization-scope note

A related review comment observed that `normalize_origin_value()` runs only in validate/sanitize, so a legacy `{}` that has not been re-saved still reaches the transformer as `[]`. Verified behaviour:

| `origin_value` | matching override | result |
|---|---|---|
| `[]` (legacy) | no | `[]` |
| `null` | no | `null` |
| `[]` (legacy) | yes | override value |
| `null` | yes | override value |

So the divergence is limited to "legacy value not yet re-saved **and** no override matches". This is pre-existing behaviour rather than a regression from this PR — `[]` passed through unchanged before the fix too. Happy to close it here by promoting `normalize_origin_value()` to a shared helper and reusing it in `Overridable_Transformer`, so "empty means no value" is defined once and does not depend on whether the document has been re-saved. Say the word if you'd prefer that in this PR rather than a follow-up.

max-zu:
Closed the normalization-scope gap in this PR rather than deferring it.

`Overridable_Prop_Type::normalize_origin_value()` is now a shared public helper, reused by `Overridable_Transformer` (PHP) and by `componentOverridableTransformer` (TS). "Empty means no value" is therefore defined once and no longer depends on whether a document has been re-saved since upgrading, which was the last remaining way a legacy payload could resolve to an empty value instead of `null`.

Coverage added for the render path in both twins:

- `Test_Overridable_Transformer` — empty origin value with no matching override resolves to `null`; with a matching override it returns the override value.
- `component-overridable-transformer.test.ts` — the same two cases on the canvas side.

Keeping the PHP and TS twins aligned is the point here: the original bug came from exactly that pair drifting apart, so the normalizer is deliberately mirrored with a comment on each side pointing at the other.

Local runs on the branch: PHPUnit `--filter Overridable` → 118 tests, 305 assertions, OK. Jest `editor-components` → 30 suites, 236 tests, all passing. ESLint clean on the touched files.

modules/components/transformers/overridable-transformer.php
```diff
@@ -32,7 +35,7 @@ public function transform( $value, Props_Resolver_Context $context ) {
 		return $result;
 	}
 
-	private function is_origin_value_override( array $origin_value ): bool {
+	private function is_origin_value_override( ?array $origin_value ): bool {
 		return isset( $origin_value['$$type'] ) && Override_Prop_Type::get_key() === $origin_value['$$type'];
```
Ntnelbaba, line 39: Will this work if origin_value is null?

modules/components/prop-types/overridable-prop-type.php
```diff
@@ -43,7 +43,20 @@ protected function validate_value( $value ): bool {
 			return false;
 		}
 
-		return $origin_prop_type->validate( $value['origin_value'] );
+		return $origin_prop_type->validate( self::normalize_origin_value( $value['origin_value'] ) );
+	}
+
+	/**
+	 * Exposing a prop that has no value must persist `origin_value: null`. Editor versions that
```
IshayMaya, line 50: what raised this issue? this is not a new use case and was handled before

packages/packages/core/editor-components/src/component-overridable-transformer.ts
```diff
@@ -7,24 +7,36 @@ export const componentOverridableTransformer = createTransformer(
...
 		}
 
-		return value.origin_value;
+		return originValue;
 	}
 );
 
+/**
+ * Mirrors `Overridable_Prop_Type::normalize_origin_value()` on the PHP side. Rendering resolves
+ * props without validating them, so an empty origin value written by an older editor reaches this
```
IshayMaya, line 27: what does "older editor" mean?

packages/packages/core/editor-components/src/prop-types/__tests__/component-overridable-prop-type.test.ts
```diff
@@ -0,0 +1,31 @@
+import { componentOverridablePropTypeUtil } from '../component-overridable-prop-type';
```
IshayMaya, line 1: @max-zu please run the custom core against this pr to make sure everything passes (most of the component playwright tests are there, and i think we have tests on the overrides use cases)

modules/components/transformers/overridable-transformer.php
max-zu: Yes. `transform_overridable_override()` is only called after `is_origin_value_override()`, which returns `false` for `null` `(isset($origin_value['$$type']))`. In that case we take the else branch and return the matching instance override - that’s the path that used to `TypeError` on `array $origin_value`. Nested-override merge only runs when origin is actually `$$type: override`. Covered by `test_overridable_transformer_returns_override_value_when_origin_value_is_null`

modules/components/prop-types/overridable-prop-type.php
max-zu: Agreed - `null` as “no origin value” is the ED-22351 contract, not a new case. What wasn’t handled:

1. Consumers still assumed array / non-null (`Overridable_Transformer`, canvas twin). That fatal only showed up once Core started persisting null (ED-23521) while an instance override existed.
2. Pro’s expose write path still used `propType.default ?? {}`. Empty object fails origin validation (invalid_value, unsavable doc) and the client zod schema, so the expose checkbox disappears. Sanitize already kept `null`. it never rewrote `{} → null`.
3. Render skips validate/sanitize, so stored `{}` ( empty object ) reached the transformer as-is. The helper is shared so “empty means no value” doesn’t depend on a re-save.

packages/packages/core/editor-components/src/component-overridable-transformer.ts
max-zu: poor wording 😅. it means any editor still on the Pro write path propType.default ?? {} (current Pro until we ship the parked Pro change https://github.com/elementor/elementor-pro/pull/7525 ). That payload is JSON {} / PHP []. We kept the Pro write as null out of this PR because Core-old + Pro-new fatals on the transformer. Until that Pro change ships, Core has to treat empty origin as “no value” on render, not only on save.

packages/packages/core/editor-components/src/prop-types/__tests__/component-overridable-prop-type.test.ts
max-zu: custom core passed

IshayMaya (approved):

PR #37098, opened 2026-08-30
Internal: Atomic SVG Improvements [ED-25402]

ronenboxer:
## 1. Problem & Context

Add Font Awesome 7 support (ED-25402) to atomic widgets with proper SVG rendering, icon aliasing, and configurable JSON asset loading. Replaces hardcoded FA5 path handling with dynamic FA7 tuple unpacking and multi-path support.

## 2. What Changed (Where)

- **PHP Backend**: New `Font_Awesome_7_Icon_Resolver` class loads/caches FA7 JSON, resolves icons by name+alias, exposes config via localized settings filters
- **Icon Transformer (PHP)**: Refactored to branch on library type—FA7 builds SVG from tuple paths; managed icons use existing data manager
- **Frontend (TypeScript)**: Icon transformer validates config, fetches from dynamic URL, indexes aliases, handles multi-path normalization and escaping
- **Build Tooling**: New `generate-font-awesome-7` package converts @fortawesome npm packages to Elementor's FA5-shaped JSON envelope
- **Tests & Types**: Added comprehensive unit tests for tuple validation, aliasing, path traversal protection; updated global type definitions

## 3. How It Works

FA7 resolution splits into two paths: if library starts with `fa-`, extract JSON filename, fetch from `window.elementorCommon.config.fontAwesome.v7.jsonBaseUrl`, index icons+aliases by name, extract tuple `[width, height, aliases[], unicode, path(s)]`, normalize paths (string or array), build SVG with `<path>` elements per path, merge inline styles (`overflow: visible` for FA7 vs `unset` for managed). Non-FA7 libraries fall back to `Font_Icon_Svg_Data_Manager`. Path traversal blocked by whitelisting allowed filenames (`solid|regular|brands`). Caching prevents re-parsing large JSON files.

## 4. Risks

File access validation relies on `in_array(..., true)` against hardcoded whitelist—if FA7 package structure changes, generator must update. JSON file size capped at 5MB; oversized/corrupt files silently fail (acceptable since fallback returns empty). TypeScript escapes SVG paths but only basic XML chars—if Font Awesome changes path format dramatically, regex matching breaks. No version pinning between FA7 packages and generated artifacts; mismatch silently skips icons.


modules/atomic-widgets/props-resolver/font-awesome-7-icon-resolver.php
```diff
@@ -0,0 +1,157 @@
+<?php
+
+namespace Elementor\Modules\AtomicWidgets\PropsResolver;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+class Font_Awesome_7_Icon_Resolver {
```
Ntnelbaba, line 9: I think it is better to support the paths by filters

packages/packages/core/editor-canvas/src/transformers/shared/__tests__/icon-transformer.test.ts
```diff
@@ -1,7 +1,8 @@
 import { iconTransformer, resetFontAwesomeIconsCache } from '../icon-transformer';
 
-const STAR_PATH = 'M0 0h100v100H0z';
-const STAR_WIDTH = 512;
+const FA7_STAR_PATH =
```
Ntnelbaba: Weird, I was expecting to use the source directly and not provide a star icon inline

Ntnelbaba (changes requested):
I think this is questionable

modules/atomic-widgets/props-resolver/font-awesome-7-icon-resolver.php
```diff
@@ -0,0 +1,175 @@
...
+
+		if ( ! is_string( $base_path ) || '' === $base_path ) {
+			return self::JSON_BASE_PATH;
+		}
+
+		return trailingslashit( $base_path );
+	}
+
+	private static function get_json_file_name( string $library ): ?string {
+		if ( ! preg_match( '/^fa-(solid|regular|brands)$/', $library, $matches ) ) {
```
Ntnelbaba: The compare should come from the static value in the class

packages/packages/core/editor-canvas/src/transformers/shared/__tests__/icon-transformer.test.ts
ronenboxer: Fixed

packages/packages/core/editor-canvas/src/transformers/shared/icon-transformer.ts
```diff
@@ -7,14 +7,19 @@ type IconValue = {
 	library?: unknown;
 };
 
-type FontAwesomeIconJson = [ number, number, unknown, unknown, string ];
+type FontAwesomeIconJson = [ number, number, unknown[], unknown, string | string[] ];
```
Ntnelbaba, line 10: Ohh
Will it break changes?

packages/packages/core/editor-canvas/src/transformers/shared/icon-transformer.ts
```diff
@@ -7,14 +7,19 @@ type IconValue = {
...
 
 const FONT_AWESOME_JSON = {
 	width: 0,
 	height: 1,
+	aliases: 2,
+	unicode: 3,
 	path: 4,
 } as const;
 
+const FONT_AWESOME_7_JSON_FILES = [ 'solid', 'regular', 'brands' ] as const;
```
Ntnelbaba: Should be written at the PHP side and loaded using window

packages/packages/core/editor-canvas/src/transformers/shared/icon-transformer.ts
```diff
@@ -7,14 +7,19 @@ type IconValue = {
...
+const FONT_AWESOME_7_JSON_BASE_PATH = 'lib/font-awesome-7/json/';
```
Ntnelbaba: Should be written at the PHP side and loaded using window (filter make it get changed)

packages/packages/core/editor-canvas/src/transformers/shared/icon-transformer.ts
```diff
@@ -52,11 +62,9 @@ function getFontAwesomeIconName( iconValue: string ): string | null {
 }
 
 function getFontAwesomeJsonFileName( library: string ): string | null {
-	if ( ! library.startsWith( 'fa-' ) ) {
-		return null;
-	}
+	const match = library.match( /^fa-(solid|regular|brands)$/ );
 
-	return library.replace( /^fa-/, '' );
```
Ntnelbaba, line 59: I think this is better

packages/tsconfig.json
```diff
@@ -36,6 +36,7 @@
     },
   },
   "exclude": [
-    "**/dist/**"
+    "**/dist/**",
+    "packages/tools/generate-font-awesome-7"
```
Ntnelbaba: Why this is new here?

packages/packages/tools/generate-font-awesome-7/src/generate.ts
Ntnelbaba, line 1: Other tools using TS, why not to use it as well?

Ntnelbaba (dismissed):

packages/tsconfig.json
ronenboxer: to have the thing regenerated when npm run install/watch

packages/packages/core/editor-canvas/src/transformers/shared/icon-transformer.ts
ronenboxer: no breaking change; the type was corrected to the real pack shape

packages/packages/tools/generate-font-awesome-7/src/generate.ts
```diff
@@ -0,0 +1,163 @@
+import { copyFileSync, existsSync, mkdirSync, readFileSync, writeFileSync } from 'node:fs';
+import { createRequire } from 'node:module';
+import { dirname, join } from 'node:path';
+import { fileURLToPath, pathToFileURL } from 'node:url';
+
+import * as freeBrandsSvgIcons from '@fortawesome/free-brands-svg-icons';
```
Ntnelbaba: We are writing files to directory, why do we need to import all?

packages/types/global.d.ts
```diff
@@ -77,6 +77,12 @@ declare global {
 			config?: {
 				allowedHTMLWrapperTags?: string[];
 				experimentalFeatures?: Record< string, boolean >;
+				atomic?: {
```
Ntnelbaba: No need to this atomic wrapper
Suggesting to switch it to font-awesome and another key named v7 (for future usage)

Ntnelbaba (dismissed):

Ntnelbaba (approved):

packages/types/global.d.ts
jose-acosta-py: Confirmed and fixed in 57dc837f. You were right, and it is the same defect class this PR already removed once — my own comment on that line claimed the check refused the whole request while its placement made it per ticket.

Hoisted, not converted to a per-ticket outcome, because of what it IS: a property of the request. `assertCodeCoversEveryAccount` runs before any ticket is touched, over the distinct accounts from one `SELECT DISTINCT ... WHERE id = ANY($1)`.

Two things the fix decides that the finding did not name:

Accounts outside the caller scope are skipped. Validating the code against every account in the batch would turn a per-ticket `out_of_scope` refusal into a whole-request 400, and would answer whether a code exists in an account the caller has no access to. Those tickets are still refused one by one.

No second copy of the check is left in `closeOne`. Ticket accounts are immutable — no UPDATE in this package writes `account_id`, verified — so an in-loop re-check would be unreachable code claiming to guard something.

Three tests, each verified by mutation: the mixed-account batch with A ordered first (asserting nothing was written, not just the error — a test that only checked the error passes against the broken version too), the out-of-scope mirror, and a failed account read. Restoring the in-loop check fails the first and third.

Also: both reader ports now arrive as one value. Taken as two constructor parameters the command crossed the repo seven-argument limit, which was revive noticing they are one collaborator.

PR #37139, opened 2026-09-01
Fix: Skip no-op clearRemovedClasses write on style edits [ED-25455]

Omerisra6:
## Summary

- Skip `clearRemovedClasses` settings writes when no style IDs were removed, avoiding a no-op `updateElementSettings` → forced subtree re-render on Style-tab prop edits.
- Fixes editor jank on large atomic pages when editing style props (e.g. gap) without removing classes.

## Test plan

- [ ] Open a large atomic page; change gap on the page root container — editor should stay more responsive than before.
- [ ] Remove a local style class — classes prop still syncs correctly.
- [ ] `cd packages && npx jest packages/libs/editor-elements/src/styles/__tests__/update-element-style.test.ts`

## Jira

ED-25455


Omerisra6:
Superseded by split PRs: #37151 (style write fix) and #37150 (render optimization).

PR #37145, opened 2026-09-01
Tweak: Wrap compound atomic elements in flexbox on canvas drop [ED-25233]

avivu:
## Summary
Compound atomic elements (Tabs, Accordion, Loop) now auto-wrap in a flexbox container when dropped onto the canvas, matching the behavior of atom elements (Heading, Image, etc.).

https://github.com/user-attachments/assets/0f063e8a-42f8-43ab-a096-9f8a3c77bf0f

## Description

Previously, dragging Tabs or Accordion from the elements panel onto the canvas placed them as top-level elements. Users had no way to drop them directly inside an existing container — the only workaround was to first add a container, then drag the element inside it.

This change adds `e-tabs`, `e-accordion`, and `e-collection-loop` to the compound atomic types list that triggers the existing auto-wrap behavior:
- `preview.js` — includes compound types in the `shouldWrap` check
- `base-container.js` — includes compound types in the `isAtomic` check, ensuring they receive a V4 flexbox wrapper rather than a legacy container

Dropping a compound element directly into an **existing** container continues to place it as a direct child without an additional wrapper — same as atoms.

Existing pages with top-level compound elements are unaffected.

## Test Instructions

1. Enable the Atomic Editor experiment (Elementor > Settings > Atomic Editor)
2. Open any page in the editor
3. Drag **Accordion** from the elements panel onto the empty canvas — it should land inside an auto-created flexbox (same as dragging a Heading)
4. Drag **Tabs** onto the canvas — same behavior
5. Add a Flexbox container to the page, then drag Accordion **into** it — it should be placed directly inside without an extra wrapper
6. Verify undo (Ctrl+Z) works correctly for all cases

## PR Checklist
- [ ] My code follows the code style of this project
- [ ] My change requires a change to the documentation
- [ ] I have updated the documentation accordingly
- [ ] I have read the CONTRIBUTING document
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests passed

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

## 1. Problem & Context

Compound atomic elements (Tabs, Accordion, List, Collection Loop) weren't auto-wrapped in flexbox on canvas drop, unlike simple atoms. ED-25233 fixes this by marking them as compound so the editor treats them consistently with other atomic widgets.

## 2. What Changed (Where)

• Five atomic element constructors: added `$this->meta( 'is_compound', true )` declaration
• New utility function `isCompoundAtomicType()` checks element meta for compound flag
• Canvas drop logic updated to wrap compound atomics in flexbox containers
• Regression test added to enforce `is_compound` meta on compound element types

## 3. How It Works

When an element is dropped on canvas, the editor calls `getWrappingContainer()` which checks `isAtomic = elementor.helpers.isAtomicWidget( model ) || isCompoundAtomicType( model.elType )`. If true, it auto-wraps in a V4 container. Compound atomics now pass the `isCompoundAtomicType()` check by declaring `is_compound: true` in PHP meta, read via the config.elements registry.

## 4. Risks

Minimal. The meta flag is additive—only affects elements explicitly marked compound. Test suite validates the meta is present on all expected compound types, catching accidental removals.


assets/dev/js/editor/utils/element-types.js
```diff
@@ -7,6 +7,15 @@ const getAllElementTypes = () => {
 	return Object.keys( elementor.getConfig().elements );
 };
 
+/**
+ * Compound atomic element types that should be auto-wrapped in a flexbox container
+ * when dropped onto the canvas, consistent with atom elements (Heading, Image, etc.).
+ *
+ * @type {string[]}
+ */
+const COMPOUND_ATOMIC_TYPES = [ 'e-tabs', 'e-accordion', 'e-collection-loop', 'e-list', 'e-form' ];
```
Ntnelbaba: Why should this list be added here? The elements in the config already cover them, no?
If so, why not include them in the config?

avivu: Good call. Added `is_compound: true` meta to the 5 PHP element classes and replaced the hardcoded array with `isCompoundAtomicType(elType)` that reads from `elementor.getConfig().elements` — same pattern as `is_container`. New compound elements (e.g. Carousel) now only need the PHP flag, no JS changes required.

Ntnelbaba (dismissed):

Ntnelbaba (dismissed):

Ntnelbaba (approved):

PR #37150, opened 2026-09-01
Fix: Stop canvas render cascade on large pages [ED-25476]

Omerisra6:
## Summary

- Only dispatch `ELEMENT_STYLE_CHANGE_EVENT` when an element's `styles` reference actually changed after render (stops N style events per subtree re-render).
- Coalesce `document-elements` provider subscribe callbacks to one per animation frame.
- Skip child re-renders when a templated element's resolved settings are unchanged (`_domUpdateWasSkipped`), stopping the deep `rerenderExistingChildren` cascade on General-tab settings edits.

Profiling on a large page showed General-tab edits caused hundreds of nested renders and style events; these changes target that render/event storm independently of style-mutation writes.

## Test plan

- [ ] Open a large atomic page; change General-tab settings on root container — editor should stay responsive (no multi-second freeze).
- [ ] Change gap on page root container (Style tab) — CSS still updates.
- [ ] `cd packages && npx jest --testPathPattern="create-nested-templated-element-type|document-elements-styles-provider"`

## Jira

ED-25476

## 1. Problem & Context

Large pages trigger cascading canvas re-renders on style changes, causing performance degradation. This fix prevents redundant child re-renders when parent DOM updates are skipped and introduces debounced style change notifications (ED-25476).

## 2. What Changed (Where)

- **create-nested-templated-element-type.ts**: Replaced `_shouldReuseChildren()` logic with `_notifyStylesChanged()` that deduplicates style events using reference equality check; simplified `_renderChildren()` to skip parent render when `_domUpdateWasSkipped` is true.
- **create-templated-element-type.ts**: Applied same `_renderChildren()` simplification; removed `rerenderExistingChildren` call.
- **document-elements-styles-provider.ts**: Changed `subscribe` to batch style change callbacks via `requestAnimationFrame`, preventing multiple notifications per event burst.
- **types.ts**: Added `styles` field to `ElementModel` type.
- **Tests**: Added comprehensive coverage for `_renderChildren` skip logic and RAF-debounced subscription behavior.

## 3. How It Works

When `_domUpdateWasSkipped` is true (indicating template reuse), `_renderChildren()` skips parent render entirely. Style notifications now check object identity against `_lastRenderedStyles`; if unchanged, no event fires. The provider's `subscribe` defers callbacks to next frame and coalesces multiple rapid events into one, eliminating the per-descendant style event cascade when a single ancestor rerenders.

## 4. Risks

Relying on reference equality for style object identity could miss updates if styles are cloned rather than mutated in place. Unsubscribe cleanup cancels pending frames, but race conditions are unlikely given single-threaded event loop. Test coverage appears solid for the critical paths.


packages/packages/core/editor-canvas/src/legacy/create-templated-element-type.ts
```diff
@@ -122,19 +121,13 @@ export function createTemplatedElementView( {
 		}
 
 		async _renderChildren() {
-			if ( this._shouldReuseChildren() ) {
-				rerenderExistingChildren( this );
-			} else {
+			if ( ! this._domUpdateWasSkipped ) {
```
IshayMaya, line 124: doesn't it break the overrides?

packages/packages/core/editor-canvas/src/legacy/create-nested-templated-element-type.ts
```diff
@@ -164,6 +170,20 @@ export function createNestedTemplatedElementView( {
...
+		_notifyStylesChanged() {
+			const styles = this.model.get( 'styles' );
+
+			if ( styles === this._lastRenderedStyles ) {
+				return;
+			}
+
+			this._lastRenderedStyles = styles;
+
 			window.dispatchEvent( new CustomEvent( ELEMENT_STYLE_CHANGE_EVENT ) );
```
IshayMaya, line 187: do we currently just dispatch it on every render?
also - why is it only on nested?

packages/packages/core/editor-canvas/src/legacy/create-nested-templated-element-type.ts
```diff
@@ -164,6 +170,20 @@ export function createNestedTemplatedElementView( {
...
 			this.model.trigger( 'render:complete' );
+			this._notifyStylesChanged();
+		},
+
+		// Rendering a parent re-renders its whole subtree, so notifying unconditionally would
+		// emit one style event per descendant for a change that touched a single element.
+		_notifyStylesChanged() {
+			const styles = this.model.get( 'styles' );
+
+			if ( styles === this._lastRenderedStyles ) {
```
IshayMaya, line 181: make sure we always update the styles object ref when we update

Omerisra6, line 181: Check that we deeply compare the values

packages/packages/core/editor-canvas/src/legacy/create-nested-templated-element-type.ts
Omerisra6: +1 if this fix is only for nested elements maybe we can remove it

Omerisra6: Yes — previously we dispatched `ELEMENT_STYLE_CHANGE_EVENT` unconditionally on every `_afterRender`, only on the nested view (the templated widget view never did it).

Removed in 7be3d8915c. Verified on a large repro page: hundreds of render-dispatches fire per action for a change that wrote zero styles (`mutateElementStyles` never fired), triggering ~200 ms rAF handlers in the styles provider for no reason. `mutateElementStyles` already dispatches the event on every real style write, and `styleRerenderEvents` covers create/duplicate/import/paste command ends — the render-time dispatch was pure noise.

packages/packages/core/editor-canvas/src/legacy/create-nested-templated-element-type.ts
Omerisra6: Moot now — 7be3d8915c drops the render-time dispatch entirely, so there's no dedup left to worry about. `mutateElementStyles` is the sole write path and already allocates a new object via `structuredClone` + `Object.fromEntries` + `container.model.set('styles', ...)`, but we're no longer relying on that here.

packages/packages/core/editor-canvas/src/legacy/create-templated-element-type.ts
Omerisra6: No — checked. Override values live in the component instance's own `component-instance` prop (its `overrides` field) and therefore are part of the instance's `settings`. So the flow is:

1. Editing any override changes the instance's settings → `settingsHash` differs → `_domUpdateWasSkipped` stays `false` → children re-render with the new `renderContext.overrides`.
2. Even when the skip fires, `afterSettingsResolve` still runs before the skip check ([create-component-type.ts](https://github.com/elementor/elementor/blob/fix/canvas-render-optimization/packages/packages/core/editor-components/src/create-component-type.ts#L208-L218)), so `#componentRenderContext` stays current for any child that later re-renders on its own model change.
3. `_shouldReuseChildren` had no external overrides in `packages/` (grepped) — nothing outside these two files touched that hook.

packages/packages/core/editor-canvas/src/legacy/create-nested-templated-element-type.ts
```diff
@@ -273,20 +293,14 @@ export function createNestedTemplatedElementView( {
 		},
 
 		async _renderChildren() {
-			if ( this._shouldReuseChildren() ) {
-				rerenderExistingChildren( this );
-			} else {
+			if ( ! this._domUpdateWasSkipped ) {
```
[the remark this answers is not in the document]
Omerisra6: Added in 7be3d8915c — `_renderChildren` skip tests for both the nested view ([create-nested-templated-element-type.test.ts](https://github.com/elementor/elementor/blob/fix/canvas-render-optimization/packages/packages/core/editor-canvas/src/legacy/__tests__/create-nested-templated-element-type.test.ts)) and the templated view ([create-templated-element-type.test.ts](https://github.com/elementor/elementor/blob/fix/canvas-render-optimization/packages/packages/core/editor-canvas/src/legacy/__tests__/create-templated-element-type.test.ts#L124-L169)). Each asserts `parentRenderChildren` (nested) / `super._renderChildren` (templated) is called when `_domUpdateWasSkipped` is `false` and NOT called when it's `true`.

packages/packages/core/editor-canvas/src/legacy/create-nested-templated-element-type.ts
Omerisra6: Correction — reverted in 602b191433. Removing the render-time dispatch broke initial CSS materialization (only some styles loaded on page load).

Answers:

**"Do we dispatch on every render?"** — yes, but with a per-view-instance dedup: `_lastRenderedStyles === styles` short-circuits the second and later renders on the same view. First render on a view always fires (initial value is `undefined`), then any subsequent render with the same `styles` object reference is skipped.

**Why the render-time dispatch is load-bearing** — elements loaded from `_elementor_data` on page load don't go through `mutateElementStyles` (no command runs), so the first `_afterRender` per view is the only mount signal the styles provider gets. Without it, async-mounted elements never trigger a CSS regen and end up with classes but no rules. The provider's rAF batcher (this PR) coalesces the whole first-load burst into a single CSS regen per frame, so cost is negligible.

**Why only on nested?** — nested is the atomic container view (the one that actually holds v4 style trees). The templated view is used by widgets that don't participate in this style-mount signal path. Happy to also add it to the templated view if you'd rather have it symmetric — but it's currently a no-op for those.

packages/packages/core/editor-canvas/src/legacy/create-nested-templated-element-type.ts
Omerisra6: Correction — the render-time dispatch is restored in 602b191433, so reference-equality dedup is active again.

It's safe: `mutateElementStyles` always allocates a new styles object before calling `container.model.set('styles', ...)` — `structuredClone` + `Object.fromEntries(styles.map(...))` in every write path. So a real style mutation always produces a new object reference, and the `styles === _lastRenderedStyles` short-circuit only fires on renders where nothing actually changed.

PR #37151, opened 2026-09-01
Fix: Skip no-op clearRemovedClasses write on style edits [ED-25466]

Omerisra6:
## Summary

- Skip `clearRemovedClasses` settings writes when no style IDs were removed, avoiding a no-op `updateElementSettings` → forced subtree re-render on Style-tab prop edits.
- Fixes editor jank on large atomic pages when editing style props (e.g. gap) without removing classes.

## Test plan

- [ ] Open a large atomic page; change gap on the page root container — editor should stay more responsive than before.
- [ ] Remove a local style class — classes prop still syncs correctly.
- [ ] `cd packages && npx jest packages/libs/editor-elements/src/styles/__tests__/update-element-style.test.ts`

## Jira

ED-25455

## 1. Problem & Context
Prevents unnecessary writes to element settings when no classes are actually removed during style edits (ED-25466).

## 2. What Changed (Where)
- `mutate-element-styles.ts`: Added guard clause to skip processing if `removedIds` is empty.
- `update-element-style.test.ts`: Added test case for no-op scenario and updated existing tests.

## 3. How It Works
`clearRemovedClasses` now returns early if no IDs were removed, avoiding subsequent `structuredClone` and `updateElementSettings` calls.


packages/packages/libs/editor-elements/src/styles/__tests__/update-element-style.test.ts
```diff
@@ -271,16 +271,7 @@ describe( 'updateElementStyle', () => {
...
-			id: 'test-element-id',
-			props: {
-				classes: {
-					$$type: 'classes',
-					value: [ existingStyle.id ],
-				},
-			},
-			withHistory: false,
-		} );
+		expect( updateElementSettings ).not.toHaveBeenCalled();
```
IshayMaya, line 320: what does it actual test in that case?

Omerisra6: To ensure we won't call it on every change (the bug)

IshayMaya (approved):

PR #37164, opened 2026-09-02
[1/6] Internal: Gate V4-only MCP abilities by atomic experiment [ED-25071]

Omerisra6:
Split 1 of 6 from #37094.

Adds atomic-experiment gating so V4-only MCP abilities become unavailable when the V4 experiment is off, letting the V3 path stand on its own.

Owned files (12): the abstract-ability gate helper plus every V4-only ability whose only diff vs main is that gate.

- Standalone, base = main. Merge order does not matter across the split.
- Union of all 6 split PRs vs the parent branch tip is byte-identical.

See split plan: /Users/omeri/.claude/plans/https-github-com-elementor-elementor-pul-tidy-rainbow.md

Companion PRs: [2/6], [3/6], [4/6], [5/6], [6/6].

### DIFF
--- a/modules/mcp/abilities/abstract-ability.php
+++ b/modules/mcp/abilities/abstract-ability.php
@@ -100,10 +100,6 @@ abstract class Abstract_Ability {
 		return true;
 	}
 
-	public function is_available_for_current_mode(): bool {
-		return true;
-	}
-
 	private function is_resource(): bool {
 		return self::KIND_RESOURCE === ( $this->mcp_meta()['type'] ?? null );
 	}
--- a/modules/mcp/abilities/abstract-ability.php
+++ b/modules/mcp/abilities/abstract-ability.php
@@ -100,6 +100,10 @@ abstract class Abstract_Ability {
 		return true;
 	}
 
+	public function is_available_for_current_mode(): bool {
+		return true;
+	}
+
 	private function is_resource(): bool {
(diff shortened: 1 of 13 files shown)


modules/mcp/abilities/abstract-ability.php
```diff
@@ -100,6 +100,10 @@ public function is_exposed_on_server(): bool {
 		return true;
 	}
 
+	public function is_available_for_current_mode(): bool {
```
Omerisra6: lets just gate the registration instead of adding a new abstraction

Omerisra6:
Merge gate for ED-25071 split (post review round). This PR merges only after every listed check goes green.

Required checks: Lint PHP + PHPUnit matrix + PR Linter + Jira Ticket Check.

Merge order across the split:
1. [3/6] #37166
2. [4/6] #37167
3. [1/6] #37164
4. [2/6] #37165
5. [6/6] #37169 (rebase on main after the above; PHPUnit here is only green with [2/6] + [4/6] already on main)

Do not merge out of order.

Ref [ED-25071]

[ED-25071]: https://elementor.atlassian.net/browse/ED-25071?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #37168, opened 2026-09-02
[5/6] Internal: V4 global classes on V3 inner elements [ED-25071]

Omerisra6:
Split 5 of 6 from #37094.

## Bug this fixes

On `main`, attaching a V4 global class to a V3 widget writes the class name onto the HTML but the class's CSS rule never ships with the page — the widget renders unstyled.

- `Class_Applier` already routes V3 targets to `V3_Node_Bridge::apply_classes()`, which persists the label into `settings['_css_classes']`. Render emits `<... class="... my-label">` correctly.
- `Global_Classes_Relations::extract_class_ids_from_post()` scans the document to decide which global-class rules `Atomic_Global_Styles` should enqueue. It only reads the V4 atomic prop shape (`Atomic_Elements_Utils::collect_class_ids_from_element_data`), so V3 nodes' `_css_classes` string is invisible to it.
- Result: the label is on the element but the corresponding rule is never included in the enqueued stylesheet.

## Fix

Adds `V3_Elements_Utils::collect_class_labels_from_v3_element()` (reads `_css_classes`) and wires it into `extract_class_ids_from_post()` alongside the V4 scan, then resolves labels → class ids via `Global_Classes_Repository::all_labels()`. Rules for classes attached to V3 wrappers now ship with the page.

## Also included (scaffolding for [6/6])

- `Class_Applier` accepts a targeted-shape input `{ config_id: { wrapper: [...], <alias>: [...] } }` in addition to the flat legacy shape. Backwards-compatible.
- `V3_Node_Bridge::apply_classes_to_target()` replaces the wrapper-only `apply_classes()` and looks up the destination setting via `V3_Widget_Map_Loader::resolve_class_setting()`.
- Wrapper target is functional on its own. Inner-element aliases are dormant until [6/6] lands per-widget maps that declare `class_setting` on an alias.

## Dependency on [6/6]

`v3-node-bridge.php` imports `V3_Widget_Map_Loader` and `V3_Auto_Mapper` (added in [6/6]). This PR does not compile against `main` without [6/6]. Recommended merge order: [6/6] then [5/6], or bundle them.

## Files (6)

- `modules/global-classes/global-classes-relations.php`
- `modules/global-classes/utils/v3-elements-utils.php` (new)
- `modules/mcp/abilities/appliers/class-applier.php`
- `modules/mcp/abilities/appliers/v3-node-bridge.php`
- `tests/phpunit/.../test-class-applier-v3-targets.php` (new)
- `tests/phpunit/.../test-v3-node-bridge-classes.php` (new)

Companion PRs: [1/6], [2/6], [3/6], [4/6], [6/6].

### DIFF
--- a/modules/global-classes/utils/v3-elements-utils.php
+++ b/modules/global-classes/utils/v3-elements-utils.php
@@ -1,0 +1,52 @@
+<?php
+
+namespace Elementor\Modules\GlobalClasses\Utils;
+
+use Elementor\Modules\AtomicWidgets\Utils\Utils as Atomic_Utils;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * V3-element counterpart of Atomic_Elements_Utils. V3 widgets do not carry an atomic
+ * classes prop; instead, they store a space-separated list of CSS class names in the
+ * `_css_classes` setting. Those names ARE the labels of V4 global classes when the MCP
+ * class applier writes to a V3 wrapper — so a document scan must read them and resolve
+ * them back to global-class ids so the corresponding rule files ship with the page.
+ */
+class V3_Elements_Utils {
+
+	const V3_CSS_CLASSES_SETTING = '_css_classes';
+
+	/**
+	 * Returns the raw class-name tokens found in the V3 `_css_classes` setting of the
+	 * given element data. Returns an empty array when the element is atomic (V4) or when
+	 * the setting is absent / empty.
+	 *
+	 * @param array<string, mixed> $element_data Single element data node.
+	 * @return string[]
+	 */
+	public static function collect_class_labels_from_v3_element( array $element_data ): array {
+		$element_type = Atomic_Elements_Utils::get_element_type( $element_data );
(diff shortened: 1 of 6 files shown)


modules/global-classes/utils/v3-elements-utils.php
Omerisra6, line 1: What does it solve, and how? the inner elements dont have the support in the data model, how do we bridge it?

Omerisra6:
Absorbed into [#37169](https://github.com/elementor/elementor/pull/37169). The six files (bridge utility, class applier, node bridge, and their tests) are cherry-picked onto `ED-25071-split/F` (see [37169's updated body](https://github.com/elementor/elementor/pull/37169) for scope + the answer to the `v3-elements-utils` question). Closing without merging.

Ref [ED-25071]

[ED-25071]: https://elementor.atlassian.net/browse/ED-25071?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #37185, opened 2026-09-06
Internal: Exclude components from import/export until full support [ED-25468]

IshayMaya:
## Summary

- Gate components out of website template import/export until full round-trip support lands, using a local `IS_IMPORT_EXPORT_SUPPORTED` constant (not an experiment — kit export serializes experiment state and would silently disable the guard on destination sites).
- On export: exclude `elementor_component` posts from the zip.
- On import: strip dangling `e-component` widgets from imported document elements (covers legacy kits exported before this change); keep `Remap_Component_Instance_Ids` behind the flag for when full support is enabled.

## Test plan

- [ ] Export a site/page that uses nested components and confirm the zip contains no `elementor_component` posts.
- [ ] Import that zip on a fresh site and confirm no component posts are created; `e-component` widgets are stripped from imported pages.
- [ ] Import a kit exported before this change (with `e-component` widgets still in the JSON) and confirm dangling instances are stripped without panel errors.
- [ ] Run DB-less unit tests: `tests/phpunit/run-unit.sh tests/phpunit/elementor/modules/components/utils/test-strip-component-instances.php`

Fixes ED-25468


app/modules/import-export-customization/processes/import.php
```diff
@@ -785,7 +787,9 @@ private function save_elements_of_imported_posts() {
 			$document = Plugin::$instance->documents->get( $new_id );
 
 			if ( isset( $data['elements'] ) ) {
-				$data['elements'] = Remap_Component_Instance_Ids::apply( $data['elements'], $imported_data_replacements['post_ids'] ?? [] );
+				$data['elements'] = Components_Module::is_import_export_supported()
```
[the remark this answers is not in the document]
Omerisra6: Check that

Omerisra6: Why do we need that flag? When will it be false?

app/modules/import-export-customization/runners/export/elementor-content.php
```diff
@@ -32,6 +34,10 @@ public function export( array $data ) {
 			$excluded_post_types[] = 'post';
 		}
 
+		if ( ! Components_Module::is_import_export_supported() ) {
+			$excluded_post_types[] = Component_Document::TYPE;
```
Omerisra6: How would component instances work in that case?

modules/components/utils/detach-component-instances.php
```diff
@@ -0,0 +1,131 @@
...
+		$component = $this->repository->get( $component_id, false );
+
+		if ( ! $component ) {
+			return null;
+		}
+
+		$this->stack[] = $component_id;
+
+		try {
+			$overrides = $element['settings']['component_instance']['value']['overrides']['value'] ?? [];
```
Omerisra6: can we somehow not relay on this hardcoded path? maybe a module helper?

modules/components/utils/detach-component-instances.php
```diff
@@ -0,0 +1,131 @@
...
+
+			return null;
+		}
+
+		array_pop( $this->stack );
+
+		return $resolved;
+	}
+
+	private function is_component_instance( array $element ): bool {
```
Omerisra6: I think we have a helper for that

Omerisra6 (approved):

app/modules/import-export-customization/runners/export/elementor-content.php
```diff
@@ -32,6 +33,8 @@ public function export( array $data ) {
 			$excluded_post_types[] = 'post';
 		}
 
+		$excluded_post_types = array_merge( $excluded_post_types, Components_Module::excluded_import_export_post_types() );
+
 		$elementor_post_types = ImportExportUtils::get_elementor_post_types( $excluded_post_types );
 
 		$elementor_post_types = apply_filters( 'elementor/import-export-customization/elementor-content/post-types/customization', $elementor_post_types, $customization );
```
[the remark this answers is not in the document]
IshayMaya: The only registered callback on this filter (in Elementor Pro) is subtractive (`array_filter` on `page`), and this mirrors the pre-existing `'post'` exclusion a few lines above, which is also applied before the filter. Since the filter is a documented extension point, moving our exclusion after it would silently override integrators. Given this is a stopgap and the full fix will remove the gating entirely, I'd rather not reshuffle the ordering here. Happy to revisit if a real callback re-adds the type.

app/modules/import-export-customization/runners/import/elementor-content.php
```diff
@@ -57,6 +58,8 @@ public function import( array $data, array $imported_data ) {
 			$excluded_post_types[] = 'post';
 		}
 
+		$excluded_post_types = array_merge( $excluded_post_types, Components_Module::excluded_import_export_post_types() );
+
 		$post_types = ImportExportUtils::get_elementor_post_types( $excluded_post_types );
 
 		$post_types = apply_filters( 'elementor/import-export-customization/elementor-content/post-types/customization', $post_types, $customization );
```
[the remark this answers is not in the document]
IshayMaya: The only registered callback on this filter (in Elementor Pro) is subtractive (`array_filter` on `page`), and this mirrors the pre-existing `'post'` exclusion a few lines above, which is also applied before the filter. Since the filter is a documented extension point, moving our exclusion after it would silently override integrators. Given this is a stopgap and the full fix will remove the gating entirely, I'd rather not reshuffle the ordering here. Happy to revisit if a real callback re-adds the type.

PR #37196, opened 2026-09-06
Fix: Preserve global-class styles across component document switches [ED-25509]

annab1:
## Summary

When entering or exiting edit mode of a component (including a nested component), classes applied on the main page were losing their styling — flashing off during the switch, and in some paths staying unstyled until the editor was reloaded. This PR fixes two runtime-state defects and one missing repopulation path.

## Root causes

1. **Stale current-document read on `attach-preview`.** `loadCurrentDocumentClasses` was reading the current document from the V2 documents store via `getCurrentDocument()`. V2's `activeId` is only synced from V1 on `editor/documents/open` (see `syncActiveDocument` in `editor-documents/src/sync/sync-store.ts`), but `editor/documents/open` fires *after* `editor/documents/attach-preview`. So at attach-preview time V2 still returned the previous document — the hook fetched classes for the wrong post, wiped state, and repopulated with those wrong classes. Result: classes on the doc we just switched *to* were dropped.
2. **Item wipe caused a visible flash.** `loadCurrentDocumentClasses` did a two-step dispatch — first a "reset order/labels baseline" `load({ items: {}, ... })`, then a final `load({ items: currentDocItems, ... })`. Because `slice.actions.load` replaced `state.data` outright, `state.data.items` was empty in between the two dispatches for the duration of the styles ajax round-trip — every class rendered with an empty `placeholderDefinition`.
3. **Ancestor local styles were not re-registered.** `embeddedDocumentsManager.reset()` runs on every `attach-preview` and wipes the embedded-documents styles provider. Only components referenced from the *current* doc's `elements` were re-seeded, so main-page (and outer component) local styles disappeared while editing a nested component.

## Changes

- `editor-global-classes/load-document-classes.ts` — use `getV1CurrentDocument()` (fresh at attach-preview) instead of V2's `getCurrentDocument()` (stale until `open`). Same pattern `editor-components/init.ts` already uses.
- `editor-global-classes/store.ts` — `slice.actions.load` now merges `state.data.items` and `state.classLabels` with the payload instead of replacing them. Order and `initialData` still come from the payload (order is the authoritative live list; `initialData` is the undo baseline). Fresh values in the payload win over stale entries. Bounded memory (`MAX_CLASSES = 1000` in `globalClassesStylesProvider`).
- `editor-components/init.ts` + new `store/actions/load-ancestor-documents-assets.ts` — on every `attach-preview`, seed the initial page doc plus every component in the current path (except the current one) into the embedded-documents manager so their local styles land in the embedded-documents styles provider. Uses `getComponentDocumentData` which shares its ajax cache with the V1 documents manager, so no extra network round-trips.

## Test plan

- Unit: `editor-global-classes` (110/110) and `editor-components` (240/240) suites pass locally.
- New unit tests:
  - `store.test.ts` — `load` preserves already-loaded definitions when the payload has empty items (direct regression for the flash); fresh payload wins over stale entries.
  - `load-document-classes.test.ts` — `loadCurrentDocumentClasses` reads from `getV1CurrentDocument`.
  - `load-ancestor-documents-assets.test.ts` — no-op when no component is being edited; seeds initial page for top-level component edit; seeds initial + outer for nested; skips docs that fail to load.
- Manual acceptance:
  - Enter edit mode of a top-level component from a page that uses global classes → main page classes stay applied, no flash.
  - Enter edit mode of a nested component (component inside another component) → main page and outer component classes/local styles stay applied.
  - Exit nested → outer → main page: no flash, no dropped classes at any level.

## Related

- Jira: ED-25509

## 1. Problem & Context
ED-25509: Prevents style "flashing" and loss of global-class definitions when switching between component documents by preserving existing style data during state updates.

## 2. What Changed (Where)
| File | Change |
| :--- | :--- |
| `init.ts` | Triggers ancestor document asset loading on init. |
| `load-ancestor-documents-assets.ts` | New: Fetches and seeds ancestor documents into the manager. |
| `load-document-classes.ts` | Switches to `getV1CurrentDocument` for fresh document ID retrieval. |
| `store.ts` | Updates `load` reducer to merge items/labels instead of replacing them. |
| `**/__tests__/**` | Added unit tests for ancestor loading and style preservation. |

## 3. How It Works
`init()` now calls `loadAncestorDocumentsAssets`, which identifies the document hierarchy via `selectPath` and seeds the `embeddedDocumentsManager`. In the global classes store, the `load` action now merges new style definitions into the existing `data.items` and `classLabels` maps, ensuring styles from nested components persist while the current document's styles are being fetched.

### DIFF
--- a/packages/packages/core/editor-components/src/init.ts
+++ b/packages/packages/core/editor-components/src/init.ts
@@ -24,6 +24,7 @@ import { LoadTemplateComponents } from './components/load-template-components';
 import { COMPONENT_WIDGET_TYPE, createComponentType } from './create-component-type';
 import { PopulateStore } from './populate-store';
 import { initCircularNestingPrevention } from './prevent-circular-nesting';
+import { loadAncestorDocumentsAssets } from './store/actions/load-ancestor-documents-assets';
 import { loadComponentsAssets } from './store/actions/load-components-assets';
 import { removeComponentStyles } from './store/actions/remove-component-styles';
 import { slice } from './store/store';
@@ -66,6 +67,7 @@ export function init() {
 		removeComponentStyles( id );
 
 		void loadComponentsAssets( ( config?.elements as V1ElementData[] ) ?? [] );
+		void loadAncestorDocumentsAssets();
 	} );
 
 	embeddedDocumentsManager.onDocumentLoad( ( _documentId, data ) => {
--- a/packages/packages/core/editor-components/src/store/actions/load-ancestor-documents-assets.ts
+++ b/packages/packages/core/editor-components/src/store/actions/load-ancestor-documents-assets.ts
@@ -1,0 +1,42 @@
+import { getV1DocumentsManager } from '@elementor/editor-documents';
+import { embeddedDocumentsManager } from '@elementor/editor-embedded-documents-manager';
+import { __getState as getState } from '@elementor/store';
+
+import { type ComponentsSlice, selectCurrentComponentId, selectPath } from '../../store/store';
+import { getComponentDocumentData } from '../../utils/component-document-data';
+
+export async function loadAncestorDocumentsAssets() {
+	const ancestorIds = getAncestorDocumentIds();
+
+	if ( ancestorIds.length === 0 ) {
+		return;
+	}
+
+	await Promise.all(
+		ancestorIds.map( async ( id ) => {
+			const document = await getComponentDocumentData( id );
+
+			if ( document ) {
+				embeddedDocumentsManager.setDocument( id, document );
+			}
+		} )
+	);
+}
+
+function getAncestorDocumentIds(): number[] {
+	const state = getState() as ComponentsSlice;
+	const currentComponentId = selectCurrentComponentId( state );
+
+	if ( currentComponentId === null ) {
+		return [];
+	}
+
+	const path = selectPath( state );
+	const pathAncestors = path
+		.filter( ( item ) => item.componentId !== currentComponentId )
+		.map( ( item ) => item.componentId );
+
+	const initialId = getV1DocumentsManager().getInitialId();
+
+	return [ ...new Set( [ initialId, ...pathAncestors ] ) ];
+}
--- a/packages/packages/core/editor-global-classes/src/__tests__/store.test.ts
+++ b/packages/packages/core/editor-global-classes/src/__tests__/store.test.ts
@@ -159,6 +159,67 @@ describe( 'store', () => {
 		} );
 	} );
 
+	describe( 'load', () => {
+		it( 'preserves already-loaded class definitions when the payload has empty items (no flash on doc switch)', () => {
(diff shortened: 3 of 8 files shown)


packages/packages/core/editor-components/src/store/actions/load-ancestor-documents-assets.ts
```diff
@@ -0,0 +1,42 @@
+import { getV1DocumentsManager } from '@elementor/editor-documents';
+import { embeddedDocumentsManager } from '@elementor/editor-embedded-documents-manager';
+import { __getState as getState } from '@elementor/store';
+
+import { type ComponentsSlice, selectCurrentComponentId, selectPath } from '../../store/store';
+import { getComponentDocumentData } from '../../utils/component-document-data';
+
+export async function loadAncestorDocumentsAssets() {
```
IshayMaya, line 8: nested components styles flicker when entering edit mode, maybe we have a race condition with `embeddedDocumentsManager.reset` that runs on `attach-preview`

packages/packages/core/editor-components/src/init.ts
```diff
@@ -66,6 +67,7 @@ export function init() {
 		removeComponentStyles( id );
 
 		void loadComponentsAssets( ( config?.elements as V1ElementData[] ) ?? [] );
+		void loadAncestorDocumentsAssets();
```
IshayMaya, line 70: i'm not sure why the line above `loadComponentsAssets` was changed as well from `await` to `void` but not waiting for the styles to load means that component load initially without styles (look bad on the editor)

packages/packages/core/editor-global-classes/src/__tests__/store.test.ts
```diff
@@ -159,6 +159,67 @@ describe( 'store', () => {
 		} );
 	} );
 
+	describe( 'load', () => {
```
IshayMaya, line 162: maybe we should also add playwright tests? the the entire issue is regression from the styles split that wasn't caught event thought we already handled it in the past

packages/packages/core/editor-global-classes/src/store.ts
```diff
@@ -74,8 +74,17 @@ export const slice = createSlice( {
...
 			state.initialData.frontend = frontend;
 			state.initialData.preview = preview;
-			state.data = preview;
-			state.classLabels = classLabels;
+
+			// Merge (instead of replace) so definitions previously loaded for other documents
+			// (e.g. embedded components rendered on the current page) don't disappear between
+			// the "reset order/labels" step and the "load current document items" step in
+			// loadCurrentDocumentClasses. Fresh values in the payload always win.
+			state.data = {
```
IshayMaya, line 82: merge of items doesn’t cover the initialData
i'm not sure if its related to this but when testing it i got into a state where after a few moves on the component tree the styles stopped loading until i refreshed

PR #37203, opened 2026-09-07
Fix: Remove nested style-change dispatch and refresh on detach [ED-25476]

Omerisra6:
## Summary
- Remove `ELEMENT_STYLE_CHANGE_EVENT` dispatch from nested templated `_afterRender` (eliminates redundant per-descendant style rebuilds on parent re-render).
- After detach, undo, and redo, refresh canvas CSS once via `doAfterRender` on the replaced subtree root — `document/elements/create` runs before nested preview paint, so regenerated local style IDs need a post-render refresh.

## Test plan
- [ ] Pro Custom Core against this branch (detach + carousel shards)
- [ ] Detach component instance with locally styled heading — background color preserved on canvas and frontend
- [ ] Undo/redo detach — styles and selection still correct
- [ ] Edit nested container style — no regression in CSS regeneration latency vs main

## Jira
ED-25476

## 1. Problem & Context
Fixes ED-25476 where CSS provider missed local class IDs because `document/elements/create` fired before nested templates fully painted. Prevents redundant style dispatches during parent re-renders.

## 2. What Changed (Where)
- `create-nested-templated-element-type.ts`: Implemented `_notifyStylesChanged` with a reference-equality guard and sentinel value.
- `types.ts`: Added optional `styles` property to `ElementModel`.

## 3. How It Works
`_notifyStylesChanged` is called after render completion to trigger `ELEMENT_STYLE_CHANGE_EVENT`. It uses a `Symbol` sentinel and reference check against `_lastRenderedStyles` to ensure the event only fires when styles actually change or on initial mount.


packages/packages/core/editor-canvas/src/legacy/create-nested-templated-element-type.ts
```diff
@@ -108,6 +114,7 @@ export function createNestedTemplatedElementView( {
 	return AtomicElementBaseView.extend( {
 		_abortController: null as AbortController | null,
 		_lastResolvedSettingsHash: null as string | null,
+		_lastRenderedStyles: undefined as ElementModel[ 'styles' ],
```
IshayMaya: update original types

packages/packages/core/editor-canvas/src/legacy/create-nested-templated-element-type.ts
```diff
@@ -164,6 +171,20 @@ export function createNestedTemplatedElementView( {
 			} );
 
 			this.model.trigger( 'render:complete' );
+			this._notifyStylesChanged();
+		},
+
+		// Rendering a parent re-renders its whole subtree, so notifying unconditionally would
+		// emit one style event per descendant for a change that touched a single element.
+		_notifyStylesChanged() {
```
IshayMaya: why do we even need that? i don't see the use case for it

Omerisra6: I removed it, will revert if anything breaks

IshayMaya (dismissed):

IshayMaya (dismissed):

IshayMaya (approved):

PR #37204, opened 2026-09-07
Fix: Coalesce document elements styles provider subscribe with RAF [ED-25476]

Omerisra6:
## Summary
- Coalesce `documentElementsStylesProvider.subscribe` callbacks into one CSS regeneration per animation frame.
- Cancel pending rAF on unsubscribe to avoid stale CSS on document switch.
- Add Jest coverage for rAF batching and Playwright behavioral guards in `raf-styles-batching.test.ts`.

## Test plan
- [ ] `cd packages && npx jest --testPathPattern="document-elements-styles-provider"`
- [ ] `npm run test:playwright -- tests/playwright/sanity/modules/v4-tests/styles-editing/raf-styles-batching.test.ts`
- [ ] Change style-tab prop on atomic widget — CSS updates on canvas
- [ ] Rapid consecutive style edits — final CSS matches last edit

## Jira
ED-25476

Split from #37150 (PR A of 3).

Made with [Cursor](https://cursor.com)

## 1. Problem & Context
Prevents performance degradation caused by excessive style re-renders during rapid updates. Implements requestAnimationFrame (rAF) batching to coalesce multiple style events into a single render cycle. Jira: ED-25476.

## 2. What Changed (Where)
* `document-elements-styles-provider.ts`: Replaced direct event subscription with rAF-based coalescing logic.
* `document-elements-styles-provider.test.ts`: Added unit tests for batching and unsubscription.
* `raf-styles-batching.test.ts`: New Playwright sanity tests for E2E verification of batched CSS updates.

## 3. How It Works
The `subscribe` method now wraps the callback in a `requestAnimationFrame` loop. When a style event fires, it schedules a frame if one isn't already pending; subsequent events are ignored until the frame executes and resets the lock. Cleanup is handled via `cancelAnimationFrame`.

_Generated by LinearB AI and added by gitStream._
<sub>AI-generated content may contain inaccuracies. Please verify before using.
💡 **Tip:** You can customize your AI Description using **Guidelines** [Learn how](https://docs.gitstream.cm/automation-actions/#describe-changes)</sub>

PR #37205, opened 2026-09-07
Fix: Skip child re-render when parent DOM update was skipped [ED-25476]

Omerisra6:
## Summary

**Pass 1 (~5.1s) — Spurious re-render from LinkControl:**
- `LinkControl` was calling `setValue(null)` on *every* `commandEndEvent('document/elements/set-settings')` while restriction was already active. This echoed a no-op `set-settings { link }` → full 306-element render cascade.
- Fix: split the debounce into two functions. The `commandEndEvent` listener uses a transition-only guard (`restrictionJustAppeared = newRestriction.shouldRestrict && !prev.shouldRestrict`). `setValue(null)` only fires once, on the edge where restriction first becomes active.

**Pass 2 (~5.3s) — Full DOM remount on attribute-only HTML change:**
- When `_cssid` changed, Twig produced HTML with only one outer attribute different. The existing `$el.html()` / `_attachTwigContent()` calls wiped all child DOM, causing 305 descendant Twig re-renders of unchanged content.
- Fix: new `wrapper-shell.ts` with `parseShell(html)` and `syncAttributes(el, next)`. Before attaching Twig HTML, compare the parsed shell (tag + inner) against `_lastShell`. If only outer attributes changed, patch them in place and set `_domUpdateWasSkipped = true` — children reuse existing DOM via the existing `rerenderExistingChildren` path.

**Cleanup:** removed all debug instrumentation (`fetch` logs, timing fields, `_renderSequence`, window flags) from `create-dom-renderer.ts`, `set-settings.js`, and both element-type files.

**Result:** ~10.5s → <500ms wall clock for a single CSS-ID change on a 306-element page.

## Test plan
- [ ] `cd packages && npx jest --testPathPattern="create-nested-templated-element-type|create-templated-element-type|link-control|wrapper-shell"`
- [ ] `npm run test:playwright -- tests/playwright/sanity/modules/v4-tests/render-cascade/`
- [ ] `npm run test:playwright -- tests/playwright/sanity/modules/v4-tests/twig-containers-editor.test.ts`
- [ ] Change CSS ID on a flexbox with many descendants — verify fast update, no full remount
- [ ] Add/remove ancestor link — verify LinkControl restriction clears exactly once

## Jira
ED-25476

Split from #37150 (PR C of 3). Depends on PR A (#37204) and PR B (#37203) landing first.

## 1. Problem & Context
Fixes an issue where parent DOM updates triggered unnecessary child re-renders even when the parent's structural HTML remained unchanged (ED-25476). The goal is to prevent expensive leaf-node remounts during simple attribute updates.

## 2. What Changed (Where)

| File | Change |
| :--- | :--- |
| `create-nested-templated-element-type.ts` | Implements DOM skip logic and `_renderChildren` conditional check. |
| `create-templated-element-type.ts` | Adds DOM skip logic to base templated views. |
| `wrapper-shell.ts` | New utility to parse HTML shells and sync attributes without replacing DOM. |
| `create-dom-renderer.ts` | Wraps `environment.render` to avoid context binding issues. |
| `link-control.tsx` | Refines link restriction checks with debouncing and transition-only logic. |
| `tests/**/*` | Adds unit, integration, and Playwright sanity tests for render cascade. |

## 3. How It Works
`render()` now uses `parseShell` to compare new HTML against `_lastShell`. If the tag and inner content match, it calls `syncAttributes` to update the existing element and sets `_domUpdateWasSkipped = true`. Consequently, `_renderChildren` skips the parent render call, preserving the child DOM tree.

## 4. Risks
The `parseShell` utility uses a lightweight regex/string-based parser rather than a full DOM parser; highly complex or non-standard HTML structures in templates could potentially lead to incorrect "skip" decisions.

### DIFF
--- a/packages/packages/core/editor-canvas/src/legacy/create-nested-templated-element-type.ts
+++ b/packages/packages/core/editor-canvas/src/legacy/create-nested-templated-element-type.ts
@@ -8,7 +8,6 @@ import { canBeTemplated, type TemplatedElementConfig } from './create-templated-
 import {
 	createAfterRender,
 	createBeforeRender,
-	rerenderExistingChildren,
 	setupTwigRenderer,
 	waitForChildrenToComplete,
 } from './twig-rendering-utils';
@@ -273,18 +272,12 @@ export function createNestedTemplatedElementView( {
 		},
 
 		async _renderChildren() {
-			if ( this._shouldReuseChildren() ) {
-				rerenderExistingChildren( this );
-			} else {
+			if ( ! this._domUpdateWasSkipped ) {
 				parentRenderChildren.call( this );
 			}
 
 			await waitForChildrenToComplete( this );
 			this._removeChildrenPlaceholder();
-		},
-
-		_shouldReuseChildren() {
-			return this._domUpdateWasSkipped && this.children?.length > 0;
 		},
 
 		_removeChildrenPlaceholder() {
--- a/packages/packages/core/editor-canvas/src/legacy/__tests__/create-nested-templated-element-type.test.ts
+++ b/packages/packages/core/editor-canvas/src/legacy/__tests__/create-nested-templated-element-type.test.ts
@@ -1,0 +1,85 @@
+import { createNestedTemplatedElementView } from '../create-nested-templated-element-type';
+import { type LegacyWindow } from '../types';
+
+type CapturedPrototype = {
+	_renderChildren: () => Promise< void >;
+	_removeChildrenPlaceholder: () => void;
+	_domUpdateWasSkipped: boolean;
+	children?: { each: ( cb: ( child: unknown ) => void ) => void };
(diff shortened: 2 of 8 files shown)


packages/packages/core/editor-canvas/src/legacy/create-nested-templated-element-type.ts
```diff
@@ -273,20 +272,14 @@ export function createNestedTemplatedElementView( {
 		},
 
 		async _renderChildren() {
-			if ( this._shouldReuseChildren() ) {
-				rerenderExistingChildren( this );
-			} else {
+			if ( ! this._domUpdateWasSkipped ) {
```
IshayMaya, line 275: what if this is just a container (div) that was rendered without any changes in its settings, but one of its children should be updated?

PR #37207, opened 2026-09-07
Internal: Remap atomic prop IDs on kit import [ED-25469]

IshayMaya:
## Summary
- Add `Atomic_Prop_Remap` to walk atomic widget settings during kit import and rewrite ID references using `post_ids`, `term_ids`, and `user_ids` replacement maps.
- Register built-in handlers for `query`, `query-array`, `query-filter`, `query-filter-array`, `link`, and `dynamic` prop types.
- Invoke the remap step in both import runners after component instance remap and before legacy dynamic-content replacement; persist unresolved references in import session metadata.

## Test plan
- [x] `tests/phpunit/run-unit.sh tests/phpunit/elementor/modules/atomic-widgets/utils/test-atomic-prop-remap.php`
- [ ] Import a kit with atomic links using Internal URL dynamics and verify destination post IDs remap to destination-site pages.
- [ ] Import a kit with collection loop manual selection / term filters and verify loop content matches destination-site posts/terms.

Fixes ED-25469

## 1. Problem & Context
Ensures Atomic Widget properties (post, term, and user IDs) are correctly remapped during kit imports [ED-25469] to prevent broken references between environments.

## 2. What Changed (Where)
| File | Change |
| :--- | :--- |
| `app/modules/import-export-customization/processes/import.php` | Integrated `Atomic_Prop_Remap` and warning collection in import loop. |
| `app/modules/import-export/processes/import.php` | Integrated `Atomic_Prop_Remap` and warning collection in import loop. |
| `modules/atomic-widgets/module.php` | Registered builtin remap handlers during module initialization. |
| `modules/atomic-widgets/utils/atomic-prop-remap-handlers.php` | Implemented specific logic for remapping queries, links, and dynamic tags. |
| `modules/atomic-widgets/utils/atomic-prop-remap-registry.php` | Created a registry to map `$type` keys to their respective handlers. |
| `modules/atomic-widgets/utils/atomic-prop-remap.php` | Implemented the core recursive walking and ID replacement engine. |
| `tests/phpunit/.../test-atomic-prop-remap.php` | Added comprehensive unit tests for various remapping scenarios. |

## 3. How It Works
The import process now triggers `Atomic_Prop_Remap::apply()`, which recursively walks the element tree. When it encounters an atomic property (identified by `$type`), it delegates to a registered handler in `Atomic_Prop_Remap_Handlers` to swap source IDs with destination IDs based on the provided replacement maps.

## 4. Risks
Partial imports or missing map entries could leave stale IDs; mitigated by capturing and surfacing warnings via `consume_warnings()`.

[ED-25469]: https://elementor.atlassian.net/browse/ED-25469?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/atomic-widgets/utils/atomic-prop-remap-handlers.php
```diff
@@ -0,0 +1,210 @@
...
+		Atomic_Prop_Remap::FILTER_KEY_CURRENT_POST => null,
+	];
+
+	/**
+	 * @var callable|null
+	 */
+	private static $tag_resolver = null;
+
+	public static function register(): void {
+		Atomic_Prop_Remap_Registry::register( Query_Prop_Type::get_key(), [ self::class, 'remap_query' ] );
```
Omerisra6, line 42:
```suggestion
		Atomic_Prop_Remap_Registry::register( Query_Prop_Type::get_key(), $this->remap_query() );
```

and make functions private

modules/atomic-widgets/utils/atomic-prop-remap-handlers.php
```diff
@@ -0,0 +1,210 @@
...
+		Atomic_Prop_Remap_Registry::register( Query_Filter_Array_Prop_Type::get_key(), [ self::class, 'remap_descend_value' ] );
+		Atomic_Prop_Remap_Registry::register( Link_Prop_Type::get_key(), [ self::class, 'remap_link' ] );
+		Atomic_Prop_Remap_Registry::register( Dynamic_Prop_Type::get_key(), [ self::class, 'remap_dynamic' ] );
+	}
+
+	public static function set_tag_resolver( ?callable $resolver ): void {
+		self::$tag_resolver = $resolver;
+	}
+
+	public static function remap_query( array $atomic, array $replacements, callable $descend, array $context = [] ): array {
```
Omerisra6, line 54: What if the structure changes? How do you handle different versions between sites?

modules/atomic-widgets/utils/atomic-prop-remap-handlers.php
```diff
@@ -0,0 +1,210 @@
...
+			return $value;
+		}
+
+		$config = [
+			'id' => '',
+			'name' => $name,
+			'settings' => $value['settings'] ?? [],
+		];
+
+		if ( method_exists( $tag, 'on_import_replace_dynamic_content' ) ) {
```
Omerisra6, line 171: why?

modules/atomic-widgets/utils/atomic-prop-remap.php
Omerisra6, line 1: check if we can reuse some logic for `walk` and other stuff

tests/phpunit/elementor/modules/atomic-widgets/utils/test-atomic-prop-remap.php
```diff
@@ -0,0 +1,287 @@
...
+
+use Elementor\Modules\AtomicWidgets\Utils\Atomic_Prop_Remap;
+use Elementor\Modules\AtomicWidgets\Utils\Atomic_Prop_Remap_Registry;
+use PHPUnit\Framework\TestCase;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+class Test_Atomic_Prop_Remap extends TestCase {
```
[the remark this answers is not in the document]
IshayMaya: Added: new tests `test_apply__remaps_dynamic_query_term_id_setting` and `test_apply__remaps_dynamic_query_author_id_setting` cover `term_ids` and `user_ids` maps through `KIND_BY_SETTING_KEY` and `lookup_mapped_id` (ab0c001).

modules/atomic-widgets/utils/atomic-prop-remap.php
```diff
@@ -0,0 +1,243 @@
...
+
+		$remapped = array_map(
+			function ( $element ) use ( $replacements ) {
+				return self::remap_element( $element, $replacements );
+			},
+			$elements
+		);
+
+		if ( function_exists( 'apply_filters' ) ) {
+			self::$warnings = apply_filters( 'elementor/atomic_widgets/import_warnings', self::$warnings, $replacements );
```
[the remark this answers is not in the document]
IshayMaya: Fixed: hook renamed to `elementor/atomic-widgets/import_warnings` to match the sibling `elementor/atomic-widgets/prop-remap/register` (ab0c001).

modules/atomic-widgets/utils/atomic-prop-remap-handlers.php
IshayMaya: Applied the spirit of the suggestion: all handler methods are now `private` and `register()` wires them via a small `self::handler( 'remap_query' )` helper using `Closure::fromCallable`. Kept static (matches the rest of the class and PHP 7.4 constraint prevents first-class callable syntax), but callers no longer touch the internals (ab0c001).

modules/atomic-widgets/utils/atomic-prop-remap-handlers.php
IshayMaya: Not an actual issue in practice: the handler degrades gracefully — if `value` is missing, not an array, or lacks an `id` key we return `$atomic` untouched instead of throwing. Cross-version parity is out of scope for the remap: import already assumes the destination understands the same `$$type` shape as the source (it's stored verbatim in kit JSON). If `Query_Prop_Type`'s shape ever changes we'd rev the handler alongside the prop type; there's no meaningful "version-aware" behavior we can add here without a general prop-type versioning story.

modules/atomic-widgets/utils/atomic-prop-remap-handlers.php
IshayMaya: The condition is intentional, not a lint artifact — it mirrors the legacy dispatch in `includes/base/element-base.php::on_import_update_dynamic_content`, which does the same `method_exists( $tag, 'on_import_replace_dynamic_content' )` check first. Some legacy dynamic tags (e.g. `Internal_URL`, `popup`, `template`) still implement the older `on_import_replace_dynamic_content` signature (post-id map only) instead of the newer `on_import_update_dynamic_content` (full replacements + controls). Removing the check would silently skip remap on those tags.

modules/atomic-widgets/utils/atomic-prop-remap.php
IshayMaya: Looked into it and I don't think we can share meaningfully. The closest existing walker is `Props_Resolver` (`modules/atomic-widgets/props-resolver/`), including `Import_Export_Props_Resolver`. It's schema-driven: it walks a `Prop_Type` schema and delegates each node to a registered `Transformer_Base` to produce a *resolved* value (atomic-shape → plain shape). Our remap walker runs earlier in the import pipeline on raw kit JSON (no schema handy, and widget schemas aren't guaranteed to be registered for every source widget) and needs to mutate values in place while preserving the atomic wrapper. It also carries a `context` for kind inference (`terms`/`authors`/`current_post`/…), tracks element_id + path for warnings, and dispatches by `$$type` rather than by schema position. Shoehorning it into `Transformer_Base` would drop those and mean we'd have to build a synthetic schema for every widget. Happy to revisit if a common "walk atomic tree by $$type" primitive materializes, but I don't see a low-risk reuse today.

modules/atomic-widgets/utils/atomic-prop-remap-handlers.php
```diff
@@ -39,19 +40,23 @@ class Atomic_Prop_Remap_Handlers {
...
 	}
 
 	public static function set_tag_resolver( ?callable $resolver ): void {
 		self::$tag_resolver = $resolver;
 	}
 
-	public static function remap_query( array $atomic, array $replacements, callable $descend, array $context = [] ): array {
+	private static function handler( string $method ): Closure {
+		return Closure::fromCallable( [ self::class, $method ] );
+	}
```
Omerisra6: You don't need that, just use:

`() => $this->callMyMethod`

modules/atomic-widgets/utils/atomic-prop-remap-handlers.php
Omerisra6: So its not supported. OK.

modules/atomic-widgets/utils/atomic-prop-remap-handlers.php
IshayMaya: Fair — dropped the `Closure::fromCallable` helper. Converted the class to instance methods behind a private singleton and `register()` now wires each handler with an arrow function: `fn( ...$args ) => $instance->remap_query( ...$args )`. Handlers remain `private` so callers still can't touch internals (40ec9ef).

Omerisra6 (approved):

PR #37219, opened 2026-09-08
Internal: Expose role manager restrictions to the v4 editor [ED-25430]

Ntnelbaba:
## Summary

Stack 1/5 for ED-25430 (content-only role access in the Atomic editor). This PR only adds the permission signal; no behaviour changes yet.

The "Access to edit content only" restriction (token `design`, stored in `elementor_role-manager`) already reaches the v1 editor as `elementor.config.user.restrictions`. Nothing under `packages/packages/**` could read it, so the v4 panel had no way to know the current user is content-only.

This adds `restrictions` to the `GET elementor/v1/user-data/current-user` payload and a `useUserRestrictions()` hook in `@elementor/editor-current-user`, alongside the existing `useCurrentUserCapabilities()`.

## Changes

- `includes/user-data.php` — return `restrictions` from `Role_Manager::get_user_restrictions_array()`.
- `packages/packages/libs/editor-current-user` — `restrictions` on `User` / `UserModel`, new `useUserRestrictions()` exposing `isRestricted( token )` and `hasContentOnlyAccess`, plus the `DESIGN_RESTRICTION` constant.

## Notes for reviewers

- **Why REST and not `window.elementor.config`.** Reading the v1 config directly needs no backend change, but couples v4 packages to v1 editor bootstrap, is unavailable to any v4 surface hosted outside the classic editor, and can't be exercised through the React Query mock the other `editor-current-user` tests already use. The REST field costs two lines.
- **`restrictions` is a deny list, not a capability.** It is deliberately kept separate from `capabilities` rather than synthesised into a fake WP capability, which would misrepresent the data model and risk colliding with the real capability system used by global classes.
- **No admin bypass**, matching v1. The Role Manager admin screen skips the `administrator` role (`core/role-manager/role-manager.php` L112-115), so an admin can never carry the `design` restriction. Adding a `manage_options` bypass would be dead code that diverges from `elementor.userCan()`.
- **`array_values` wrapping is deliberate.** `get_user_restrictions_array()` ends in `array_unique()`, which preserves keys. A user in multiple restricted roles can produce non-sequential keys, which `json_encode` would emit as a JSON object and break `restrictions.includes()` on the client. The same latent issue exists in the v1 config path but is out of scope here.
- **Fail-open while loading.** `data` is undefined until the query resolves, so `hasContentOnlyAccess` is briefly `false`. `ensureCurrentUser()` prefetches after `editor/documents/attach-preview`, so the panel normally reads a warm cache, and failing open matches `useUserStylesCapability`'s all-true default. The server-side enforcement in stack 4/5 is what actually guarantees the restriction.

## Testing

- `npm run test:packages -- packages/libs/editor-current-user` — 9 passed, including 3 new cases.
- `npm run lint -w elementor-packages` — eslint and tsc clean.
- New PHPUnit coverage in `tests/phpunit/elementor/includes/test-user-data.php` for a restricted and an unrestricted role.

## Jira

[ED-25430](https://elementor.atlassian.net/browse/ED-25430)

<div><a href="https://cursor.com/agents/bc-72354742-39d3-4852-8f9f-bab3d4184b56?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-72354742-39d3-4852-8f9f-bab3d4184b56&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

[ED-25430]: https://elementor.atlassian.net/browse/ED-25430?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## 1. Problem & Context
Exposes Role Manager restrictions to the v4 editor (ED-25430) to enable client-side UI enforcement of user permissions.

## 2. What Changed (Where)

| File | Change |
| :--- | :--- |
| `includes/user-data.php` | Added `restrictions` to the user data REST response. |
| `tests/phpunit/.../test-user-data.php` | Added tests for restrictions, including JSON array serialization check. |
| `libs/editor-current-user/src/api.ts` | Updated `UserModel` and API client to fetch `restrictions`. |
| `libs/editor-current-user/src/types.ts` | Added `restrictions` field to `User` type. |
| `libs/editor-current-user/src/use-user-restrictions.ts` | New hook to check specific restrictions. |
| `libs/editor-current-user/src/use-has-content-only-access.ts` | New hook specifically for `DESIGN_RESTRICTION`. |
| `libs/editor-current-user/src/index.ts` | Exported new restriction hooks. |
| `libs/editor-current-user/src/__tests__/...` | Added unit tests for restriction hooks. |

## 3. How It Works
PHP fetches restrictions via `Role_Manager` and uses `array_values()` to ensure a sequential array for JSON encoding. The frontend consumes this via a new `useUserRestrictions` hook, which provides an `isRestricted` helper for the UI.


packages/packages/libs/editor-current-user/src/use-user-restrictions.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const useUserRestrictions = () => {
+	const { data } = useCurrentUser();
+
+	const isRestricted = ( restriction: string ) => Boolean( data?.restrictions?.includes( restriction ) );
+
+	return {
+		isRestricted,
+		restrictions: data?.restrictions,
+		hasContentOnlyAccess: isRestricted( DESIGN_RESTRICTION ),
```
max-zu: `hasContentOnlyAccess` feels out of place in a generic `useUserRestrictions` hook. Right now the hook does two unrelated things: it exposes a generic mechanism (`isRestricted` + raw `restrictions`), and it also encodes knowledge about one specific restriction and what it means for the product. That's a mix of abstraction levels — every new restriction-derived flag will have to be added here, so the hook keeps growing and every consumer imports it even if it only cares about one flag.

What about keeping this hook purely generic, and letting feature code compose it? E.g. `useUserRestrictions().isRestricted( DESIGN_RESTRICTION )`, or a thin dedicated hook next to the feature that owns it:

```ts
export const useHasContentOnlyAccess = () => useUserRestrictions().isRestricted( DESIGN_RESTRICTION );
```

That way `useUserRestrictions` stays the low-level primitive and domain meaning lives with the domain.

max-zu (dismissed):

max-zu (approved):

max-zu (approved):

max-zu (approved):

PR #37228, opened 2026-09-08
Fix: Ignore non-200 remote responses in atomic SVG src transformer [ED-25530]

Ntnelbaba:
## PR Checklist

## PR Type

- [x] Bugfix

## Summary

* Fix: A missing or unreachable SVG file no longer renders unrelated markup taken from the server's error page

## Description

`Svg_Src_Transformer::fetch_svg_content()` falls back to `wp_safe_remote_get()` and read `$response['body']` after only an `is_wp_error()` check. A 404/403/500 is not a `WP_Error`, so the error page HTML was treated as SVG source and passed to `process_svg()`. Because `WP_HTML_Tag_Processor::next_tag( 'svg' )` scans the whole document rather than just the root element, any inline `<svg>` on that error page (theme logo, social icons) was sanitized and rendered as the widget's icon. The result was silently wrong output instead of an empty icon.

Remote responses are now accepted only with HTTP 200, using `wp_remote_retrieve_response_code()` / `wp_remote_retrieve_body()` and `WP_Http::OK`.

### On the loopback request for deleted local attachments

The first revision of this PR also made same-site URLs skip the remote fallback entirely, since `resolve_local_path()` had already proven the file missing via `file_exists()`. **That was reverted** — it is incorrect. `resolve_local_path()` conflates "the file is gone" with "the URL could not be mapped to a path", and its naive `site_url()` → `ABSPATH` mapping legitimately fails to resolve files that do exist: a moved `WP_CONTENT_DIR`, symlinked plugins, or a `WP_HOME`/`WP_SITEURL` mismatch. In those setups the remote fetch is a working fallback, and skipping it turned existing SVGs into empty icons.

`Test_Atomic_Svg` demonstrated this: three snapshot tests failed because the default SVG (`ELEMENTOR_ASSETS_URL`) does not resolve under `ABSPATH` in the test environment and was then denied its fallback.

So the wasted loopback request for a genuinely deleted attachment still happens. Avoiding it safely needs a way to distinguish an unresolvable URL from a resolved-but-missing file, which is a larger change than this fix warrants. Worth a follow-up if it shows up in profiling.

Remote `Content-Type` is also still unvalidated; the sanitizer limits the blast radius there, and it was left out to keep this change minimal.

### Test changes

* Added `Test_Svg_Src_Transformer` covering a 404 for a same-site URL, a 404 for an external URL whose body contains an `<svg>`, and a successful 200 fetch.
* `Test_Atomic_Svg`'s `pre_http_request` mocks returned only a `body` key. That is not a valid HTTP response shape, so `wp_remote_retrieve_response_code()` had nothing to read. They now return a full response array with `WP_Http::OK`, via a small `mock_http_response()` helper.

## Test instructions

* Point an SVG prop at an external URL that returns 404 with an HTML body containing an inline `<svg>`. The widget should render an empty icon; before this change it rendered the `<svg>` scraped from that error page.
* Point an SVG prop at an external URL returning a real SVG with 200. It should still render inline with `fill="currentColor"`.
* Delete the file behind an SVG attachment while keeping the attachment record. The widget should render an empty icon rather than whatever the 404 page contains.
* `vendor/bin/phpunit --filter 'Test_Svg_Src_Transformer|Test_Atomic_Svg'`

## Quality assurance

Verified locally against MySQL + the WordPress test library: `Test_Svg_Src_Transformer` and `Test_Atomic_Svg` pass (15 tests, 23 assertions), as do the neighbouring `Test_Icon_Transformer`, `Test_Svg_Src_To_Icon_Migration` and `Test_Atomic_Accordion` suites (44 tests, 230 assertions). PHPCS is clean on the changed files.

<div><a href="https://cursor.com/agents/bc-a3cbfdbd-3e3d-4c4b-aaf9-127fda76aa34?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-a3cbfdbd-3e3d-4c4b-aaf9-127fda76aa34&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/modules/atomic-widgets/props-resolver/transformers/svg-src-transformer.php
+++ b/modules/atomic-widgets/props-resolver/transformers/svg-src-transformer.php
@@ -14,6 +14,8 @@ if ( ! defined( 'ABSPATH' ) ) {
 class Svg_Src_Transformer extends Transformer_Base {
 	const SVG_INLINE_STYLES = 'width: 100%; height: 100%; overflow: unset;';
 
+	private const HTTP_OK = 200;
+
 	public function transform( $value, Props_Resolver_Context $context ) {
 		$id = isset( $value['id'] ) ? (int) $value['id'] : null;
 		$url = $value['url'] ?? null;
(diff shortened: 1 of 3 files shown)


modules/atomic-widgets/props-resolver/transformers/svg-src-transformer.php
```diff
@@ -14,6 +14,8 @@
 class Svg_Src_Transformer extends Transformer_Base {
 	const SVG_INLINE_STYLES = 'width: 100%; height: 100%; overflow: unset;';
 
+	private const HTTP_OK = 200;
```
Ntnelbaba: Isn't there a WP const (or PHP)?

davseve (approved):

PR #37238, opened 2026-09-09
fix: alpine.js never starts if init() runs after DOMContentLoaded

### LINKED ISSUE #36443: Atomic Tabs (e-tabs) are never initialized in the editor canvas - Alpine components don't mount, tab content panels stay hidden (4.1.4)
### Description

In the editor canvas, Atomic Tabs elements (`e-tabs`) are never initialized by Alpine. The tab triggers render, but all `e-tab-content` panels keep the `hidden` attribute and inline `display: none`, so the canvas shows empty boxes where tab content should be. The same document renders its structure correctly (Structure panel shows all elements; ~300 elements present in canvas DOM), so this is an interactivity/initialization failure, not a data problem.

Diagnostic findings (measured via DevTools on the canvas iframe):

1. `window.elementorV2.alpinejs` is loaded (`{ Alpine, init, refreshTree }`), and `elementor-v2-alpinejs-js` script tag is present. No `window.Alpine` global exists (expected, the package is namespaced).
2. On a fresh editor load, every `e-tabs` element has **no `_x_dataStack`** (Alpine never initialized it), `aria-selected` is not set on any trigger, and all `e-tab-content` panels keep `hidden`.
3. Calling `window.elementorV2.alpinejs.init()` manually from the console **sometimes fixes it**: the `eTabs{id}` components resolve, the default trigger gets `aria-selected="true"`. This proves the component registrations exist and work.
4. However, this only succeeds if `init()` has not been "consumed" earlier in the session. In sessions where Alpine apparently started early (against an empty canvas), later manual `init()` calls are no-ops and the e-tabs elements can no longer be initialized at all. This looks like a one-shot `Alpine.start()` race: if the start happens before the canvas (Vue) renders the atomic elements, they are never picked up.
5. `refreshTree(element)` initializes individual late-rendered `e-tabs` instances correctly (again proving registrations are fine), but nothing in the editor calls it for canvas re-renders.
6. Even when initialization succeeds, `e-tab-content` panels keep their SSR inline `display: none`, so the active panel may still not be visible in the canvas.

Additional context / isolation: the site's theme originally enqueued its own Alpine.js into the canvas, which hijacked the `x-data="eTabs…"` elements and produced `Alpine Expression Error: eTabsXXX is not defined` floods. We dequeued the theme's Alpine for `is_preview_mode()` — the errors disappeared, the canvas now contains only `elementor-v2-alpinejs` — but Elementor's own initialization still never runs on the e-tabs elements. So this is reproducible without any third-party Alpine instance in the canvas.

We also tried injecting a bootstrap script via `wp_footer` in preview mode; the `<script>` tag appears in the canvas DOM but never executes (canvas DOM appears to be reconstructed via innerHTML, which does not execute inline scripts), so there is no user-land workaround.

### Steps to reproduce

1. Create a Theme Builder Header (or any V4 document) containing Atomic Tabs (`e-tabs`) with several triggers and content panels (in our case: two `e-tabs` instances nested inside mega-menu content of an `e-n-menu`).
2. Open the document in the editor (Atomic Editor enabled, Elementor 4.1.4).
3. Observe the canvas.

### Expected behavior

Atomic Tabs initialize in the canvas: the default trigger is selected and its content panel is visible/editable, as on a simple test page.

Expected behavior:
Atomic Tabs initialize in the canvas: the default trigger is selected and its content panel is visible/editable, as on a simple test page.

### Elementor System Info

```txt
Elementor 4.1.4, Elementor Pro (4.1.x), WordPress 7.0, PHP 8.x, WPML active. Full System Info below.

== Server Environment ==
	Software: nginx/1.18.0 (Linux, Ubuntu 22.04)
	MySQL: MariaDB 10.6.23
	PHP Version: 8.1.2
	PHP Memory Limit: 256M | WP Memory limit: 40M
	PHP Max Input Vars: 1000 | Max Post Size: 12M

== WordPress Environment ==
	Version: 7.0
	Site URL: https://www.imagoprinter.com
	Multisite: No | Language: pl_PL | Debug: Inactive
	Permalink Structure: /blog/%category%/%postname%/

== Theme ==
	Imago 1.0 (custom, by CroCoder) — no child theme.
	Note: the theme enqueues its own Alpine.js instance (handle "imago-alpine",
	scripts-alpine.min.js) on the frontend and, by default, into the editor
	preview. For this report it was dequeued in is_preview_mode() via a snippet
	— the e-tabs initialization failure persists with only
	elementor-v2-alpinejs present in the canvas.

== Key Plugins ==
	Elementor 4.1.4 | Elementor Pro 4.1.2 | Angie 1.1.10
	WPML Multilingual CMS 4.9.5 + String Translation 3.5.3 + ACFML 2.2.4
	ACF PRO 6.8.4 | Rank Math 1.0.272 | Code Snippets 3.9.6
	PixelYourSite 11.2.0.7 | CF7 6.1.6 (+addons) | UpdraftPlus 1.26.5
	(no caching/optimization plugin active)

== Relevant Experiments ==
	Editor V4: Active | Atomic widgets: Active | Nested elements: Active
	Global Classes: Active | Variables: Active | Components: Active
	Menu: Active | Pro Interactions: Default Active
	Editor MCP for atomic widgets: Default Active
	Optimized Markup: Default Inactive

== Relevant Log Entries ==
	JS 2026-06-03 [x25/x75] alpinejs.min.js (Elementor package):
	  "Undefined variable: headerInit" / "Undefined variable: megaDropdown"
	  → Elementor's Alpine instance evaluating the THEME's x-data components
	  (mirror image of the reported conflict; both instances walk the same tree)
	JS 2026-06-25 editor-loader-v2.min.js: Cannot read properties of
	  undefined (reading 'start')
	JS 2026-06-26 [x10] editor.min.js: "Element type already registered"
	JS 2026-07-03 editor.min.js: this.model.isValidChild is not a function
	PHP 2026-06-22 [x26] elementor-pro .../site-editor/data/endpoints/
	  templates.php::150 — Undefined array key "condition_type"
	PHP 2026-06-23 [x115] elementor/modules/interactions/
	  interactions-frontend-handler.php::61 — foreach() argument must be
	  of type array|object, null given
```

### Agreement

- [ ] I confirm I have read and followed all the guidelines and instructions outlined in the Editor V4 Bug Report form.
- [x] I agree that my issue may be closed without further action if it doesn't meet all the requirements outlined in the Editor V4 Bug Report form.

ZebaAfiaShama:
## PR Checklist

## PR Type
- [x] Bugfix

## Summary
* Fix: Alpine.js never starts if init() runs after DOMContentLoaded

## Description
`@elementor/alpinejs`'s `init()` only ever attached a one-shot `DOMContentLoaded` listener to call `Alpine.start()`. That event fires exactly once per document; if `init()` runs after it has already fired (e.g. the editor canvas iframe finishes its own document load before this async chunk executes), the listener never runs and Alpine never starts.

This left Atomic Tabs (and any other Alpine-driven atomic widget) permanently uninitialized in the editor canvas: no `_x_dataStack`, no `aria-selected`, tab content panels stuck on their SSR `hidden` state. Manually calling `window.elementorV2.alpinejs.init()` from the console only "fixed" it on documents where `DOMContentLoaded` hadn't fired yet - matching the reported behavior of the workaround being session/timing dependent.

The fix checks `document.readyState` first: if the document has already finished loading, start Alpine immediately instead of waiting for an event that will never fire again.

## Test instructions
1. Create a Theme Builder Header (or any V4 document) containing Atomic Tabs (`e-tabs`) with several triggers and content panels.
2. Open the document in the editor (Atomic Editor enabled).
3. Confirm the default tab trigger is selected (`aria-selected="true"`) and its content panel is visible on load, without needing any manual console workaround.
4. Unit tests: `npx jest --config=./packages/jest.config.js packages/packages/core/alpinejs`

## Quality assurance

Fixes #36443

## 1. Problem & Context

Alpine.js initialization was unconditionally listening for `DOMContentLoaded`, which never fires if the init module loads after that event. Affects async-loaded chunks in iframes or deferred scripts.

## 2. What Changed (Where)

- `init.ts`: Added `document.readyState` check to immediately start Alpine if DOM is already loaded, otherwise register listener with `{ once: true }`
- `init.test.ts`: New test suite covering three states (complete/interactive/loading) and idempotency

## 3. How It Works

On `init()` call: if `readyState === 'loading'`, attach `DOMContentLoaded` listener; otherwise call `Alpine.start()` immediately. The `{ once: true }` flag prevents double-start if event fires multiple times (browser quirk edge case).

## 4. Risks

None. Defensive fix with good test coverage. Edge case where browser fires `DOMContentLoaded` twice is already handled.

### DIFF
--- a/packages/packages/core/alpinejs/src/__tests__/init.test.ts
+++ b/packages/packages/core/alpinejs/src/__tests__/init.test.ts
@@ -1,0 +1,73 @@
+import { Alpine } from '@alpinejs/csp';
+import { init } from '../init';
+
+jest.mock( '@alpinejs/csp', () => ( {
+	Alpine: {
+		start: jest.fn(),
+	},
+} ) );
+
+function setReadyState( readyState: DocumentReadyState ) {
+	Object.defineProperty( document, 'readyState', {
+		value: readyState,
+		configurable: true,
+	} );
+}
+
+describe( 'init', () => {
+	afterEach( () => {
+		jest.clearAllMocks();
+	} );
+
+	it( 'starts Alpine immediately when the document has already finished loading', () => {
+		// Arrange.
+		setReadyState( 'complete' );
+
+		// Act.
+		init();
+
+		// Assert.
+		expect( Alpine.start ).toHaveBeenCalledTimes( 1 );
+	} );
+
+	it( 'starts Alpine immediately when the document is interactive', () => {
+		// Arrange.
+		setReadyState( 'interactive' );
+
+		// Act.
+		init();
+
+		// Assert.
+		expect( Alpine.start ).toHaveBeenCalledTimes( 1 );
+	} );
+
+	it( 'waits for `DOMContentLoaded` when the document is still loading', () => {
+		// Arrange.
+		setReadyState( 'loading' );
+
+		// Act.
+		init();
+
+		// Assert.
+		expect( Alpine.start ).not.toHaveBeenCalled();
+
+		// Act.
+		document.dispatchEvent( new Event( 'DOMContentLoaded' ) );
+
+		// Assert.
+		expect( Alpine.start ).toHaveBeenCalledTimes( 1 );
+	} );
+
+	it( 'never starts Alpine twice if `DOMContentLoaded` fires again', () => {
+		// Arrange.
+		setReadyState( 'loading' );
+		init();
+
+		// Act.
+		document.dispatchEvent( new Event( 'DOMContentLoaded' ) );
+		document.dispatchEvent( new Event( 'DOMContentLoaded' ) );
+
+		// Assert.
+		expect( Alpine.start ).toHaveBeenCalledTimes( 1 );
+	} );
+} );
--- a/packages/packages/core/alpinejs/src/init.ts
+++ b/packages/packages/core/alpinejs/src/init.ts
@@ -1,11 +1,20 @@
 import { Alpine } from '@alpinejs/csp';
 
 export function init() {
-	document.addEventListener(
-		'DOMContentLoaded',
-		() => {
-			Alpine.start();
-		},
-		{ once: true }
-	);
+	// `DOMContentLoaded` only ever fires once per document. If this module loads after that
+	// point (e.g. the editor canvas iframe finishes loading before this async chunk executes),
+	// the listener below would never run and Alpine would never start.
+	if ( document.readyState === 'loading' ) {
+		document.addEventListener(
+			'DOMContentLoaded',
+			() => {
+				Alpine.start();
+			},
+			{ once: true }
+		);
+
+		return;
+	}
+
+	Alpine.start();
 }


Ntnelbaba:
Hey @ZebaAfiaShama 👋🏽 
Netanel from Elementor team here
Did you do something special that this case happened?
How can we validate if this fix is valid?
We want to understand the use case and the way to reproduce 
Thanks!

ZebaAfiaShama:
Hello @Ntnelbaba 

thanks for the questions - I went and actually reproduced this end-to-end instead of just reasoning about it, here's what I found.

**Nothing special needed to trigger it.** This is a plain timing/race bug in `init.ts`, independent of the original reporter's setup (their theme's own Alpine instance was a red herring - they isolated it out and the failure persisted with only `elementor-v2-alpinejs` present).

**Root cause:** 

```
`init()` only ever did

```js
document.addEventListener('DOMContentLoaded', () => Alpine.start(), { once: true });

```

DOMContentLoaded fires exactly once per document. If this module executes after that event has already fired - which happens whenever the editor canvas iframe finishes its own document load before this async chunk gets to run - the listener is attached too late and never fires. Alpine.start() is simply never called, so nothing on the canvas ever initializes (no _x_dataStack, no aria-selected, panels stuck on their SSR hidden state).

How I reproduced it (verified, not just inferred):
1. Fresh Elementor 4.2.4 install, Atomic Elements experiment enabled.
2. Created a page with an Atomic Tabs widget (3 tabs, default active = Tab 1), published it.
3. Reloaded the editor on that page - a completely ordinary load, no network throttling, no special conditions.
4. Inspected the canvas iframe directly: the e-tabs element had no _x_dataStack, every tab trigger had aria-selected="false", and all three e-tab-content panels still had hidden + inline display: none from SSR - including Tab 1, which should have been the default active panel.

So this isn't a rare edge case - it reproduced on a plain, unthrottled first load every time I tried it, which suggests it's closer to the common case than a corner case.

How I validated the fix:
1. Patched init() to check document.readyState before deciding whether to wait for DOMContentLoaded or start immediately (patched the actual shipped bundle in the install, not just the source).
2. Reloaded the same editor/page. This time: _x_dataStack present, Tab 1's trigger aria-selected="true", and Tab 1's panel had its inline display cleared and was actually rendering on screen (confirmed visually, not just via attributes).
3. Separately, the included Jest unit tests (packages/packages/core/alpinejs/src/__tests__/init.test.ts) isolate the exact race deterministically: reverting init.ts to the version on main makes 3 of the 4 tests fail with the exact "listener never fires again" signature. Run via npx jest --config=./packages/jest.config.js packages/packages/core/alpinejs.

Happy to share the test page/screenshots or add more coverage (e.g. a Playwright repro) if that would help move this along.

PR #37239, opened 2026-09-09
Internal: Add generic responsive settings prop type [ED-25538]

max-zu:
## PR Checklist

## PR Type
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

* Internal: Add a generic responsive settings prop type so V4 settings can be authored per breakpoint [ED-25538]

## Description

V4 style props already cascade per breakpoint; settings props do not. Carousel (and later Loop Carousel) need per-viewport behavioural values without a new schema `kind`.

This adds `Responsive_Prop_Type` (`extends Object_Prop_Type`) and editor primitives (`responsivePropTypeUtil`, `resolveResponsiveValue`). There is **no** dedicated `responsive-number` control: `SettingsField` detects `key === 'responsive'` (including unions) and wraps the existing control in `ResponsiveBinding`, so a `Number_Control` can sit on a `Responsive_Prop_Type` schema. The active breakpoint comes from the canvas switcher (`useActiveBreakpoint`).

A fifth prop-type kind was rejected because it would break `Overridable_Schema_Extender` and touch 16+ editor sites. A `settings_variants[]` rewrite was also rejected as too global for this case — the breakpoint map stays in `settings`.

Unblocks Pro ED-25381. Story: https://elementor.atlassian.net/browse/ED-25538

## Test instructions

* PHPUnit: `Test_Responsive_Prop_Type` (sparse validation, disabled-breakpoint survival, fallback chain, resolver map, overridable wrapping)
* Jest: `resolveResponsiveValue` + `SettingsField` responsive binding (desktop/tablet writes, inherited placeholder, non-responsive props stay unwrapped)

## Quality assurance

- [x] Docs have been added / updated (for bug fixes / features)

Fixes ED-25538

[ED-25538]: https://elementor.atlassian.net/browse/ED-25538?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## 1. Problem & Context
Implements a generic responsive property type [ED-2538] to standardize how settings are handled across different breakpoints in the editor and backend.

## 2. What Changed (Where)

| File | Change |
| :--- | :--- |
| `responsive-prop-type.php` | New PHP prop type for responsive value mapping and fallback logic. |
| `test-responsive-prop-type.php` | Unit tests for PHP responsive prop validation and fallback chains. |
| `settings-field.tsx` | Integrated `ResponsiveBinding` to handle breakpoint-specific values and placeholders. |
| `is-responsive-prop-type.ts` | Utility to detect if a prop type is responsive or a responsive union. |
| `responsive.ts` / `resolve-responsive-value.ts` | TS definitions for responsive shapes and value resolution logic. |
| `settings-field.test.tsx` / `resolve-responsive-value.test.ts` | Comprehensive frontend tests for responsive binding and value inheritance. |

## 3. How It Works
The `Responsive_Prop_Type` (PHP) and `responsivePropTypeUtil` (TS) create an object where keys are breakpoints. When resolving a value, the system follows a `fallback_chain` from the active breakpoint upward to `desktop`, returning the first defined value encountered.

## 4. Risks
Reflection is used in PHP to set `initial_value` and `default` on cloned objects, which may break if internal class properties are renamed.


modules/atomic-widgets/prop-types/responsive-prop-type.php
```diff
@@ -0,0 +1,127 @@
...
+			$result = sprintf( 'var(--%s-%s, %s)', $css_var_prefix, $key, $result );
+		}
+
+		return $result;
+	}
+
+	private static function duplicate_inner_prop_type( Transformable_Prop_Type $inner ): Transformable_Prop_Type {
+		$class = get_class( $inner );
+
+		return $class::make();
```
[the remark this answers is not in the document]
max-zu: Fixed. `duplicate_inner_prop_type()` now clones the inner instance (so `float()`, meta, and settings survive) and then clears the copied default so only desktop keeps it. Added a PHPUnit case for float sanitize + meta on desktop and tablet entries.

packages/packages/libs/editor-controls/src/controls/__tests__/responsive-number-control.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+	useActiveBreakpoint: jest.fn( () => 'tablet' ),
+	useBreakpoints: jest.fn( () => [ { id: 'desktop' }, { id: 'tablet' }, { id: 'mobile' } ] ),
+} ) );
+
+const numberPropType = createMockPropType( { kind: 'plain', key: 'number' } );
+
+const responsivePropType = createMockPropType( {
+	kind: 'object',
+	key: 'responsive',
+	shape: {
```
IshayMaya: why do we need to wrap the number prop type with responsive, isn't it always a number?

packages/packages/libs/editor-controls/src/controls/responsive-number-control.tsx
```diff
@@ -0,0 +1,56 @@
+import * as React from 'react';
+import { resolveResponsiveValue, responsivePropTypeUtil, type ResponsivePropValue } from '@elementor/editor-props';
+import { useActiveBreakpoint, useBreakpoints } from '@elementor/editor-responsive';
+
+import { PropKeyProvider, PropProvider, useBoundProp } from '../bound-prop-context';
+import { createControl } from '../create-control';
+import { NumberControl } from './number-control';
+
+export const ResponsiveNumberControl = createControl(
```
IshayMaya: i'm not sure we need to register a control for each responsive one we ae going to have, maybe we can remove this (and the php part) and just add a layer to the `settings-field.tsx` that will be rendered conditionally for responsive prop type:
```
const DEFAULT_BREAKPOINT = 'desktop';

const ResponsiveBinding = ( { children }: { children: React.ReactNode } ) => {
	const { value, setValue, propType, disabled } = useBoundProp( responsivePropTypeUtil );
	const breakpoint = useActiveBreakpoint() ?? DEFAULT_BREAKPOINT;
	const activeBreakpoints = useBreakpoints().map( ( { id } ) => id );

	const inherited = responsiveFallbackChain( breakpoint )
		.filter( ( key ) => key !== breakpoint && activeBreakpoints.includes( key ) )
		.map( ( key ) => value?.[ key ] )
		.find( ( entry ) => entry !== null && entry !== undefined );

	return (
		<PropProvider
			propType={ propType }
			value={ value }
			setValue={ setValue }
			placeholder={ { [ breakpoint ]: inherited } }
			isDisabled={ () => disabled }
		>
			<PropKeyProvider bind={ breakpoint }>{ children }</PropKeyProvider>
		</PropProvider>
	);
};
```

modules/atomic-widgets/prop-types/responsive-prop-type.php
```diff
@@ -0,0 +1,132 @@
+<?php
```
IshayMaya: maybe we can model breakpoint as variant meta rather than part of the value shape?

styles have something similar, in the opposite direction: Style_Variant::build() puts breakpoint in meta beside props (not inside the prop value). That's why every style control is responsive without any control, prop type, or registry entry knowing breakpoints exist 

what if we could do the same for settings?

modules/atomic-widgets/prop-types/responsive-prop-type.php
```diff
@@ -0,0 +1,132 @@
...
+
+	protected function define_shape(): array {
+		return [];
+	}
+
+	public static function create_from( Transformable_Prop_Type $inner ): self {
+		$dependencies = $inner->get_dependencies();
+		$prop_meta = $inner->get_meta() ?? [];
+
+		$inner->set_dependencies( [] );
```
IshayMaya: i think this mutates the caller's object before any cloning happens
Number_Prop_Type::make()->... passed in from a widget schema loses its dependencies as a side effect

packages/packages/libs/editor-controls/src/controls/__tests__/responsive-number-control.test.tsx
max-zu: Agreed — wrapping was leftover from the first approach. The inner type is still just `number` (or whatever the schema already is). We dropped `Responsive_Prop_Type` / `ResponsiveNumberControl`. Opt-in is `Responsive_Settings::enable()` on the existing prop type, and `SettingsField` scopes writes by breakpoint.

packages/packages/libs/editor-controls/src/controls/responsive-number-control.tsx
max-zu: Took this further than a `ResponsiveBinding` wrapper: there is no responsive control and no `$$type: responsive` envelope. Controls stay breakpoint-agnostic. `SettingsField` writes desktop to `settings` and other breakpoints to `settings_variants` (same idea as styles).

modules/atomic-widgets/prop-types/responsive-prop-type.php
max-zu: This is the model we shipped. Breakpoint lives in variant meta (`settings_variants[]` with `{ meta: { breakpoint }, props }`), not inside the prop value. Desktop stays in `settings`. Parser/controls/registry do not know about breakpoints.

modules/atomic-widgets/prop-types/responsive-prop-type.php
max-zu: The mutation concern is gone with `Responsive_Prop_Type`. There is no wrapper that clones or mutates the inner prop type anymore — the schema stays `Number_Prop_Type` (etc.) plus `meta( Responsive_Settings::enable() )`.

packages/packages/libs/editor-controls/src/controls/responsive-number-control.tsx
max-zu: Implemented as suggested: `SettingsField` now wraps children in `ResponsiveBinding` when `isResponsivePropType` matches (`key === 'responsive'`, including unions). There is no dedicated `responsive-number` control — `Number_Control` binds to `Responsive_Prop_Type` and the breakpoint comes from the canvas `useActiveBreakpoint` switcher.

modules/atomic-widgets/prop-types/responsive-prop-type.php
max-zu: We rolled this back. After core-team review we are not putting the breakpoint in meta / `settings_variants`; per-breakpoint values live in `settings` again as `Responsive_Prop_Type` (`$$type: responsive`).

modules/atomic-widgets/prop-types/responsive-prop-type.php
```diff
@@ -0,0 +1,132 @@
...
+
+		foreach ( array_reverse( $chain ) as $key ) {
+			$result = sprintf( 'var(--%s-%s, %s)', $css_var_prefix, $key, $result );
+		}
+
+		return $result;
+	}
+
+	private static function duplicate_inner_prop_type( Transformable_Prop_Type $inner ): Transformable_Prop_Type {
+		$entry = clone $inner;
```
ronenboxer, line 134: Cloning the inner prop preserves its `required` setting for every breakpoint. Because `Object_Prop_Type::validate_value()` validates every shape entry and `to_json_schema()` marks required shape fields, wrapping a required prop makes sparse responsive maps invalid unless every breakpoint is present. Could we clear `required` on each cloned breakpoint entry and keep requiredness on the responsive wrapper instead? A test using `Number_Prop_Type::make()->required()` with only desktop/tablet values would cover this.

ronenboxer, line 134: The clone also retains the inner `initial_value` on every breakpoint entry. That means resetting a tablet/mobile field can resolve to an explicit initial value rather than removing the breakpoint value and restoring responsive inheritance. Could we clear initial values from cloned entries, at least for non-desktop breakpoints, and add a test asserting that non-desktop shape entries have no initial value?

max-zu: Fixed in `duplicate_inner_prop_type()`: cloned breakpoint entries are `optional()`, and requiredness is lifted onto the responsive wrapper. Sparse `{desktop, tablet}` now validates, and JSON schema no longer lists every breakpoint as required. Covered in `test_validate__accepts_a_sparse_value_when_inner_is_required`.

max-zu: Cloned entries now have `initial_value` cleared; only the desktop shape field (and the wrapper) keep the inner initial. Resetting tablet/mobile therefore writes `null` and restores inheritance. Covered in `test_create_from__clears_initial_value_on_non_desktop_entries`.

ronenboxer (approved):

IshayMaya (approved):

PR #37242, opened 2026-09-09
Internal: Add Agent Ready llms.txt, cache, and robots surfaces [APT-14]

bainternet:


modules/agents/classes/post-noindex.php
```diff
@@ -0,0 +1,40 @@
...
+
+		if ( is_string( $rm ) && false !== strpos( $rm, 'noindex' ) ) {
+			return true;
+		}
+
+		if ( '1' === (string) get_post_meta( $post_id, '_aioseo_noindex', true ) ) {
+			return true;
+		}
+
+		if ( 'yes' === get_post_meta( $post_id, '_seopress_robots_index', true ) ) {
```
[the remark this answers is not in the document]
Ntnelbaba: @bainternet is it valid?

modules/agents/module.php
```diff
@@ -15,106 +17,203 @@
...
 
+	public static function is_active(): bool {
+		return Plugin::$instance->experiments->is_feature_active( self::EXPERIMENT_NAME );
+	}
+
 	public function __construct() {
 		parent::__construct();
 
-		add_action( 'template_redirect', [ $this, 'maybe_serve_llms_txt' ], 1 );
+		$sanitizer              = new Prompt_Injection_Sanitizer();
```
Ntnelbaba, line 69: I wonder if the unit should be after checking the feature is active

modules/agents/module.php
```diff
@@ -15,106 +17,203 @@
...
+	// Kit tabs
+	// -------------------------------------------------------------------------
+
 	/**
-	 * @param \Elementor\Core\Kits\Documents\Kit $kit
+	 * @param Kit $kit
 	 */
 	public function register_kit_tabs( $kit ) {
-		if ( ! Plugin::$instance->experiments->is_feature_active( self::EXPERIMENT_NAME ) ) {
+		if ( ! self::is_active() ) {
```
Ntnelbaba, line 110: If the constructor will handle this, you will not need this

modules/agents/prompt-injection-sanitizer.php
```diff
@@ -0,0 +1,85 @@
...
+	 * Sanitize a string of text that will be embedded in an llms.txt file.
+	 *
+	 * The text is passed through each pattern in sequence. Only the suspicious
+	 * command-like fragments are removed or replaced; the surrounding readable
+	 * content is preserved.
+	 *
+	 * @param string $text Raw excerpt or description.
+	 * @return string Sanitized text safe to embed in the llms file.
+	 */
+	public function sanitize( string $text ): string {
```
Ntnelbaba, line 67: I talked with @ayalaharo , this can be problematic to filter out the content
Can we add a plain warning if some grep was found? And some indication that this grep occurred

modules/agents/classes/post-noindex.php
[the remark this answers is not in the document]
nirbhayel: Yes, this suggestion seems valid

modules/agents/module.php
nirbhayel: moved it behind the gate

modules/agents/module.php
nirbhayel: moved and tested

modules/agents/prompt-injection-sanitizer.php
nirbhayel: added

modules/agents/content-generator.php
```diff
@@ -0,0 +1,685 @@
...
+	 * Generate the full llms-full.txt content (links + inlined page/post content).
+	 *
+	 * WooCommerce products are linked but their content is not inlined.
+	 * Total output is capped at LLMS_FULL_HARD_CAP bytes; pages and posts are
+	 * prioritised over WooCommerce catalog data.
+	 *
+	 * @param array{intro?: string, optional?: string} $overrides
+	 * @return string
+	 */
+	public function generate_llms_full_txt( array $overrides = [] ): string {
```
DinetsV, line 119: Is it possible to receive a request timeout here?

modules/agents/module.php
```diff
@@ -15,106 +17,200 @@
...
+		return Plugin::$instance->experiments->is_feature_active( self::EXPERIMENT_NAME );
+	}
+
 	public function __construct() {
 		parent::__construct();
 
-		add_action( 'template_redirect', [ $this, 'maybe_serve_llms_txt' ], 1 );
-		add_action( 'elementor/kit/register_tabs', [ $this, 'register_kit_tabs' ] );
-		add_action( 'elementor/document/after_save', [ $this, 'maybe_invalidate_llms_txt_cache' ] );
+		add_action( 'plugins_loaded', [ $this, 'check_seo_plugin_conflict' ], 20 );
```
DinetsV, line 69: Do we need to check this `if ( ! self::is_active() )`? 
Maybe better to add an action after the is_active check?

nirbhayel: No because we need to check for conflict even if the feature is inactive.

modules/agents/prompt-injection-sanitizer.php
```diff
@@ -0,0 +1,111 @@
...
+		foreach ( self::PATTERNS as [ $pattern, $replacement ] ) {
+			$sanitized = preg_replace( $pattern, $replacement, $text );
+
+			// preg_replace returns null on error; keep original if that happens.
+			if ( null !== $sanitized ) {
+				if ( $sanitized !== $text ) {
+					$matched = true;
+				}
+
+				$text = $sanitized;
```
Ntnelbaba, line 85: I was more in favor of not removing and writing at the beginning of the page a warning that it might contain malicious instructions

modules/agents/content-generator.php
Ntnelbaba, line 119: **MEDIUM — first `llms-full.txt` request cost (DinetsV thread still relevant):** `on_global_change()` now bumps `INLINE_CACHE_GEN_OPTION` instead of scanning all postmeta (good). The remaining risk is the **first HTTP** build of `llms-full.txt`: this method can call `get_inline_content()` for many posts synchronously and still hit max execution time on large sites. Worth documenting, capping work per request, or warming async — not a stack-pin issue once downstream rebases.

modules/agents/prompt-injection-sanitizer.php
Ntnelbaba, line 85: **MEDIUM — product / open thread:** Prefer a **top-of-output warning** when patterns match, without stripping. Current code still removes/replaces matches and appends a notice; tests assert neutralization. Please confirm intended UX (warn-only vs strip) and align tests + llms-full/markdown paths.

modules/agents/content-generator.php
```diff
@@ -0,0 +1,685 @@
...
+			}
+
+			$output      .= "\n" . $section_text;
+			$output_size  = strlen( $output );
+		}
+
+		// Append the user-editable optional section to the output directly,
+		// since $output was already built from $lines before the content loop.
+		if ( ! empty( $overrides['optional'] ) ) {
+			$output .= "\n## Optional\n\n" . trim( $overrides['optional'] ) . "\n";
```
Ntnelbaba, line 216: **LOW — sanitization parity:** `generate_llms_txt()` runs optional through `append_optional_section()` (sanitizer). Here `generate_llms_full_txt()` appends `trim( $overrides['optional'] )` without `$this->sanitizer->sanitize()`. Saves go through `prepare_override_text()`, but consider reusing the same helper for parity if anything else can feed overrides.

Ntnelbaba (commented):
## Combined APT-14 stack note (this is PR 1/3)

**Stack order:** merge/rebase **#37242 → #37243 → #37244**. Do not merge #37243/#37244 until their bases match this tip (`6d863071`).

**This PR alone → main:** sound; `mergeable` but **behind** main — rebase/update before merge.

Most earlier threads look addressed on the discovery tip (cache gen bump, sanitizer on overrides, frontmatter tests, SEOPress, etc.). Remaining for this PR below (non-blocking for isolated ship).

Cloud agent: https://cursor.com/agents/bc-932f4115-5a93-52d8-94c7-ca7571cf9508

modules/agents/prompt-injection-sanitizer.php
nirbhayel: After discussing with @bainternet we don't want to add warning on every page so we remove it.

modules/agents/content-generator.php
nirbhayel: @Ntnelbaba @DinetsV  For now we have decided to add a max limit filter with a default to control the number of posts that will be fetched per post type reducing the risk of hitting max execution time.

modules/agents/prompt-injection-sanitizer.php
nirbhayel: I've discussed this with @bainternet and we've decided to strip the content with warning in the section itself instead of putting warning on top of each page with flagged content inside.

Ntnelbaba (dismissed):

modules/agents/content-generator.php
nirbhayel: Good point! Fixed

PR #37269, opened 2026-09-14
Fix: Restore adding elements into empty Atomic Form messages [ED-25493]

davseve:
## Summary
- Empty Atomic Form success/error messages can receive children again after the default paragraph is deleted (plus / panel / drop).
- In the editor, those messages stay visible when selected in the navigator or canvas, when a nested child is selected, or when empty so the drop target has a hit area. Frontend and unselected Normal stay hidden.
- Jira: https://elementor.atlassian.net/browse/ED-25493

## Test plan
- [ ] Add an Atomic Form. On Normal, select the success message in the navigator and confirm it appears on the canvas.
- [ ] Delete the default paragraph inside the success message. Confirm the empty plus/drop area is visible.
- [ ] Add a paragraph (and a heading) back via the plus button / widgets panel.
- [ ] Repeat independently for the error message.
- [ ] Confirm adding still works from the navigator.
- [ ] Confirm the live site still hides both messages until submit (or until the form States control is Success/Error).

## Video demo
**Broken:** Once the default paragraph was deleted, an empty Atomic form success or error message rejected every widget, so the slot could never be filled again.
**Where:** WP Admin → Pages → Add New → Edit with Elementor → drag **Atomic form** onto the canvas → Navigator
**Steps:** open the Navigator → expand **Atomic form** → select **Success message** and confirm it becomes visible on the canvas → delete the paragraph inside it → confirm the empty plus / drop area is visible → add a **Paragraph** via the plus button → drag a **Heading** in from the widgets panel → repeat the same flow for **Error message** → deselect and confirm both messages hide again
**Pass:** both empty messages show a drop area and accept the new Paragraph and Heading; each message is visible only while it or a child is selected
**Fail:** the empty message stays hidden, shows no drop target, or rejects the dropped widget

## 1. Problem & Context
ED-25493: Restricted child types in `Form_Message` prevented adding new elements to empty message containers because the whitelist rejected `widget` types, breaking the `addElement` flow.

## 2. What Changed (Where)
* `form-message.php`: Removed restricted child types to allow any element.
* `module.php`: Added CSS to ensure message slots remain visible/editable in editor mode.
* `test-atomic-form.php`: Added regression tests for child type configuration.

## 3. How It Works
Updating `define_allowed_child_types` to return an empty array bypasses the whitelist check, permitting the injection of various atomic elements. Complementary CSS ensures the `display: none` base style doesn't hide the target area during editor interaction.


modules/atomic-widgets/elements/atomic-form/form-message/form-message.php
```diff
@@ -57,7 +57,10 @@ protected static function define_props_schema(): array {
 	}
 
 	protected function define_allowed_child_types() {
-		return [ Atomic_Paragraph::get_element_type() ];
+		// Empty: same as accordion item content. Panel atoms are created as `elType: widget`
+		// (widgetType e-paragraph / e-heading). A whitelist of those type strings rejects
+		// `widget`, and addElement then calls children.last() which is undefined when empty.
+		return [];
```
davseve, line 60: Empty: same as accordion item content. Panel atoms are created as `elType: widget`
(widgetType e-paragraph / e-heading). A whitelist of those type strings rejects
`widget`, and addElement then calls children.last() which is undefined when empty.

Ntnelbaba (approved):

PR #37301, opened 2026-09-16
Internal: Prioritize V4 atomic elements in widget panel search results [ED-25570]

avivu:
## Summary
- When the `e_atomic_elements` experiment is active, V4 atomic elements (e.g. `e-heading`, `e-button`, `e-svg`) are sorted to the top of widget panel search results
- When the experiment is off, search results are unchanged
- When the search input is cleared, the comparator is removed and natural order is restored

## How it works
Sets a `comparator` on the panel elements collection during search: widgets whose `widgetType` starts with `e-` sort before all others. The comparator is cleared on empty search.

## Jira
https://elementor.atlassian.net/browse/ED-25570

## Test plan
- [ ] Enable `e_atomic_elements` experiment
- [ ] Search "heading" — V4 Heading should appear above V3 Heading widget
- [ ] Search "button" — V4 Button should appear above V3 Button widget
- [ ] Search "text" — V4 Heading and Paragraph should appear first
- [ ] Clear search — panel returns to default category order
- [ ] Disable `e_atomic_elements` experiment — search order is unchanged from before

## 1. Problem & Context
Prioritize V4 atomic elements over legacy V3 widgets in search results (ED-2570) to improve discoverability of the new atomic system.

## 2. What Changed (Where)

| File | Change |
| :--- | :--- |
| `atomic-background-video.php`, `div-block.php`, `flexbox.php`, `grid.php` | Expanded `get_keywords` for better search matching. |
| `elements.js` (LayoutView) | Maps the `atomic` property to the view model. |
| `elements.js` (CollectionView) | Implements `_syncAtomicComparator` to sort atomic items first. |
| `elements-panel-atomic-search.test.ts` | Added Playwright test to verify V4 priority in DOM order. |

## 3. How It Works
When a user types in the search bar, `onFilterChanged` triggers `_syncAtomicComparator`. If the atomic experimental feature is enabled, a custom `viewComparator` is assigned to the collection view, ensuring items with `atomic: true` are sorted to the top regardless of original index.

[ED-2570]: https://elementor.atlassian.net/browse/ED-2570?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


assets/dev/js/editor/regions/panel/pages/elements/views/elements.js
```diff
@@ -43,6 +43,24 @@ PanelElementsElementsView = Marionette.CollectionView.extend( {
 	onFilterChanged() {
 		const filterValue = elementor.channels.panelElements.request( 'filter:value' );
 
+		if ( filterValue && elementorCommon.config.experimentalFeatures?.e_atomic_elements ) {
```
Ntnelbaba: I'm not sure that this is the right location, maybe filter in advanced?
I think it can cause performances issues to sort every time the filter change

avivu: Good catch on the performance concern. Updated to only call `sort()` once — when the comparator is first set (filter becomes active). On subsequent keystrokes the comparator is already in place, so Backbone maintains order automatically without re-sorting.

On location: `onFilterChanged` is the natural hook since the sort needs to be tied to search state — we want atomic-first only during search, not in the default panel view. Moving it elsewhere would require the same event wiring anyway.

avivu: (Adding note: this also applies to large collections from 3rd party plugins — the sort runs exactly once per search session regardless of collection size, so the overhead is bounded even with 1000+ widgets.)

Ntnelbaba (dismissed):

Ntnelbaba (dismissed):

Ntnelbaba (approved):

PR #37311, opened 2026-09-16
Fix: Global class disappears from Class Manager after document switch and publish [ED-25545]

IshayMaya:
## Summary
- On frontend publish, merge the published order into preview order instead of overwriting it, so preview-only classes stay in the Class Manager until they are published or deleted.
- Filter lazy-load REST responses by the context registry and stop seeding missing frontend definitions in the editor store baseline, so `calculateChanges` correctly treats unpublished classes as added on publish.
- Add PHPUnit, Jest, and Playwright coverage for the document-switch + publish regression.

## Coordination
- Overlaps with store load/merge behavior in https://github.com/elementor/elementor/pull/37196 — please coordinate with that author before merge so the fix is not reintroduced.

## Test plan
- [x] `vendor/bin/phpunit --filter 'Global_Classes'` (105 tests)
- [x] `npm run test:packages -- --testPathPattern=editor-global-classes`
- [x] PHPCS on touched PHP files (0 errors)
- [ ] Manual: create class on page A (draft), publish page B, confirm class still in Class Manager, publish page A, confirm frontend label/CSS
- [ ] `tests/playwright/sanity/modules/v4-tests/global-classes/global-classes-unpublished.test.ts` (with v4 env)

## 1. Problem & Context
Fixes ED-2545 where publishing a document would overwrite the preview order, causing unpublished global classes to disappear from the Class Manager.

## 2. What Changed (Where)
* `global-classes-repository.php`: Added `propagate_order_to_preview` to merge published and unpublished class orders.
* `global-classes-rest-api.php`: Filtered `get_by_ids` calls to only request IDs present in the current context.
* `store.ts`: Added safety check to prevent missing frontend definitions from polluting the frontend baseline.
* `test-global-classes-rest-api.php` & `save-global-classes.test.ts` & `store.test.ts`: Added unit tests for order preservation and context filtering.
* `global-classes-unpublished.test.ts`: Added Playwright E2E test for unpublished class persistence.

## 3. How It Works
When publishing, the system now calculates the difference between existing preview classes and the new published set. It injects unpublished IDs back into the updated order list before saving to the preview context, ensuring they aren't wiped during synchronization.


modules/global-classes/global-classes-repository.php
```diff
@@ -176,6 +174,21 @@ public function apply_changes( array $touched_items, array $changes, array $orde
 		}
 	}
 
+	private function propagate_order_to_preview( array $published_order, array $deleted_ids ): void {
```
IshayMaya, line 177: Runs only on frontend publish (not preview saves). Adds one kit meta read plus an in-memory merge over the preview order (hard-capped at MAX_ITEMS / 1000). Unpublished ids are checked with an array_flip map (O(1) per id), so this stays negligible compared to the post/meta writes already done in apply_changes on the same request — not a meaningful performance regression.

Omerisra6 (approved):

PR #37324, opened 2026-09-17
Fix: Clear link value only when restriction becomes active [ED-25476]

Omerisra6:
## Summary
- `LinkControl` only clears the link value when restriction transitions from inactive → active, not on every `document/elements/set-settings` command end
- Prevents a spurious no-op `set-settings { link }` that was cascading a full page re-render (~5.1s on a 306-element page)

## Test plan
- [ ] `cd packages && npx jest --testPathPattern="link-control"`
- [ ] Add/remove an ancestor link — verify restriction clears the child link exactly once
- [ ] Change an unrelated setting (e.g. CSS ID) with an active ancestor link — verify no extra `set-settings { link }` echo

## Jira
ED-25476

Split from #37205 (Pass 1 / LinkControl only). Sibling: attribute-only DOM patch PR.

## 1. Problem & Context
Prevents redundant `setValue(null)` calls during `document/elements/set-settings` events when a link restriction is already active. Fixes ED-25476 by ensuring the link value is cleared only upon the transition to a restricted state.

## 2. What Changed (Where)
* `packages/packages/libs/editor-controls/src/controls/link-control.tsx`: Refactored restriction check into a parameterized `syncLinkRestriction` function.
* `packages/packages/libs/editor-controls/src/controls/__tests__/link-control.test.tsx`: Added regression test for non-triggering `setValue` when already restricted.

## 3. How It Works
`syncLinkRestriction` now accepts a `ClearActiveLinkWhen` flag. When triggered by `document/elements/set-settings`, it uses the `newly-restricted` mode, which compares the current restriction state against the previous state to detect the exact moment a restriction becomes active before clearing the value.

[ED-25476]: https://elementor.atlassian.net/browse/ED-25476?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/packages/libs/editor-controls/src/controls/link-control.tsx
```diff
@@ -75,14 +77,20 @@ export const LinkControl = createControl( ( props: Props ) => {
 		}
 
 		setLinkInLinkRestriction( ( prev ) => ( isSameRestriction( prev, newRestriction ) ? prev : newRestriction ) );
-	}, 300 );
+	};
+
+	const debouncedCheckRestriction = useDebouncedCallback( () => checkRestriction(), 300 );
+	const debouncedCheckRestrictionOnCommandEnd = useDebouncedCallback(
```
IshayMaya: its the same returned value from the useDebouncedCallback, why not just passing the param from where we need and have 1 function

packages/packages/libs/editor-controls/src/controls/link-control.tsx
```diff
@@ -63,10 +63,12 @@ export const LinkControl = createControl( ( props: Props ) => {
 
 	const shouldDisableAddingLink = ! isActive && linkInLinkRestriction.shouldRestrict;
 
-	const debouncedCheckRestriction = useDebouncedCallback( () => {
+	const checkRestriction = ( { transitionOnly = false }: { transitionOnly?: boolean } = {} ) => {
 		const newRestriction = getLinkInLinkRestriction( elementId, value ?? linkPlaceholder );
+		const restrictionJustAppeared = newRestriction.shouldRestrict && ! linkInLinkRestriction.shouldRestrict;
```
IshayMaya: what is the meaning of "just appeared" and "transition" here?

packages/packages/libs/editor-controls/src/controls/link-control.tsx
Omerisra6: Good call — collapsed into a single `debouncedSyncLinkRestriction` that takes the policy as an argument. Command end passes `"newly-restricted"`; mount + `inline-link-changed` use the default `"restricted"`. Pushed in `1c12fddfda`.

packages/packages/libs/editor-controls/src/controls/link-control.tsx
Omerisra6: Fair — the words were about the restriction state going from `off` → `on`, not a UI transition. Renamed to `becameRestricted`, and the flag is now a small union `ClearActiveLinkWhen = "restricted" | "newly-restricted"` that names the *when-to-clear* rule instead:

- `"restricted"` (mount, inline-link-changed): clear the active link whenever restriction is on.
- `"newly-restricted"` (set-settings command end): clear only on the inactive → active edge, so an already-active restriction does not re-fire `setValue(null)` and echo a no-op `set-settings` back into the render cascade.

Pushed in `1c12fddfda`.

IshayMaya (approved):

Issue #7661, opened 2019-03-29
.map file undefiened in Safari Browser Mac OS


sagarpatel124:
Hello Elementor,

Please check .map undefined issue. 

Screenshot : https://ibb.co/wLM1gBz 

[Error] Failed to load resource: the server responded with a status of 404 () (frontend-modules.js.map, line 0)
[Error] Failed to load resource: the server responded with a status of 404 () (ResizeSensor.min.js.map, line 0)
[Error] Failed to load resource: the server responded with a status of 404 () (common-modules.js.map, line 0)
[Error] Failed to load resource: the server responded with a status of 404 () (swiper.min.js.map, line 0)

Let us know, If you need us to share URL with you. 

Wordpress : 5.1.1 Elementor : 2.5.10 Jupiter X Theme

creakipik:
Hello,
Same for me.

hubertq:
No news about it ? EP 2.5.4 is out and the issue is not fixed ;(

backpackerstoday:
Still not fixed in the latest E and E Pro versions... 
Please fix that Elementor Team. The Widgets in Safari looks so blurry and unsharp.

hubertq:
Fixed for me since yesterday

lior-sh:
Close, fixed.

PR #37348, opened 2026-09-18
Fix: Allow clearing and multi-digit editing in V4 number control [ED-25510]

max-zu:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

## Summary

* Fix: Number fields in the V4 editing panel can now be cleared and typed into digit-by-digit, and are clamped to their allowed range when editing finishes instead of on every keystroke.

## Description

The shared V4 `NumberControl` applied `min`/`max` validation and prop defaults in ways that broke normal text-field editing. Reported on [ED-25510](https://elementor.atlassian.net/browse/ED-25510), discovered while implementing [ED-25377](https://elementor.atlassian.net/browse/ED-25377).

The two reported symptoms looked different because they had **different causes**, and there were two more defects in the same function that the card did not mention. All four are in `number-control.tsx`:

1. **Per-keystroke clamping rewrote the input.** `Math.min( Math.max( … ) )` ran on `onInput`. Because `max` was never passed to the native input, a max violation was clamped silently and instantly — this is why Items per page showed `200 → 100` as a visible rewrite.
2. **A native-validity gate blocked below-min values entirely.** `setValue( …, { validation: () => isInputValid } )` fed `event.target.validity.valid` into `useBoundProp`'s `validate()`. Since `min` *was* on the native input, typing `2` in a `min: 100` field set `rangeUnderflow`, which flipped the `isValid` latch, which made the resolved value `null`. So min violations blanked the field while max violations rewrote it.
3. **No local draft state, so `null` could never present as empty.** `setValue( null )` succeeded, but the next render resolved back through `fallbackValue = hasBaseValue ? null : propType.default` — hence `600` reappearing. Note this only happens on the **base breakpoint**; on tablet/mobile the inherited value renders as a placeholder instead.
4. **`restoreValue` was the only blur handler**, so nothing ever committed a clamped final value.

### The fix

`NumberControl` now keeps a local draft string while the field is focused. While typing, the input shows exactly what was typed and out-of-range values are **held** rather than committed; the canvas keeps the last valid value. On blur the draft is parsed, clamped, and committed. Parsing and range logic moved into a pure, separately tested `utils/number-value.ts`.

The `{ validation: () => isInputValid }` argument is gone — the control now owns range enforcement, so native HTML constraint validity no longer gates commits. That is the single change that fixes defect 2.

`components/number-input.tsx` is deliberately **untouched**: its blur-time remount (the `key` bump when `! validity.valid`) is load-bearing here, and it is shared with `SizeInput`, `UnstableSizeInput` and `TextFieldInnerSelection`.

### :warning: Two behaviour changes outside the card's text

Both follow directly from removing the validity gate, which was also catching `stepMismatch`. Reviewers should be aware of them:

* **Style → Transform → Scale now accepts `1.15`.** It passes `step={ 0.1 }` with no `min`, so the step base was `0` and any off-step decimal was silently rejected. This was never reported but was broken.
* **Carousel → Autoplay Delay now accepts `5050`.** It declares `step: 100`, so non-multiples were rejected. `step` remains on the input as a spinner increment; it was never a backend validation rule.

### Known follow-up, already split out

After blur, a cleared **optional** field still *displays* its prop default, because `useBoundProp` resolves unset values to `propType.default` for every V4 control. Making an unset prop render as a placeholder instead touches every control and was split to its own card by product decision. While the field is focused it now stays empty, which is what makes digit-by-digit editing possible.

### Null-safety note for reviewers

PHP declares `min`/`max`/`step` as `?int` and `Atomic_Control_Base::jsonSerialize()` passes them through unfiltered, so any field declared without `set_max()` sends `'max' => null`. A JavaScript destructuring default fires on `undefined` only, **never on `null`**, so the props are typed `number | null` and coalesced explicitly with `??` in the body. Getting this wrong makes `clamp( 5, 1, null )` return `0`. Two tests cover the PHP-shaped payload directly.

## Test instructions

**Verify on the desktop/base breakpoint** — per defect 3 the bug is breakpoint-dependent and tablet/mobile already behaved differently.

1. Insert a V4 Carousel (Pro). Open **Settings → Transition Speed** (shows `600`).
   * Select all → Delete. The field stays empty while focused.
   * Type `1`, `0`, `0`. It reads `1`, then `10`, then `100` — never rewriting early, never blanking.
   * Blur. Canvas transition speed is `100`.
2. Same field: type `5000`, blur → commits `3000` (clamped to max).
3. Same field: type `2`, blur → commits `100` (clamped to min).
4. **Autoplay Delay** (`step: 100`): type `5050`, blur → commits `5050`. Rejected outright before this PR.
5. Insert a Collection Loop. **Query → Items per page** (shows `3`): type `200` → reads `200`, not `100`; blur → commits `100`.
6. Insert a V4 Video. Clear **Start time**, blur → stays empty (`default: null`, so no default to fall back to).
7. **Style → Transform → Scale**: type `1.15` → commits `1.15`. Rejected before this PR.
8. **Style → Position → Z-index**: type `0` → shows `0`, not empty. Type `-5` → commits `-5`.
9. **Layout → Flex order**: negatives still typeable.
10. Pro **atomic form → Textarea → Rows** (declares `min` but no `max`): type `5`, blur → commits `5`, not `0`. This is the null-bound path.
11. Switch to tablet on Transition Speed and clear the override → the desktop value appears as a grey placeholder, unchanged from before.

## Quality assurance

**Automated verification:** `editor-controls` at 91 suites / 815 passed / 10 skipped (baseline on `main` was 90 / 766 / 10). `npm run lint` exit 0 — 0 eslint errors, 182 pre-existing warnings, `tsc` exit 0. No PHP changes, so no PHPUnit impact.

Two branches are covered by mutation testing rather than by coverage alone: deleting the `required` early-return, or the optional empty-on-blur commit, each makes a specific named test fail.

**Status:** the manual walkthrough above has not been run against a playground build yet. Marked ready for review so the Playwright suite runs, since it is gated on draft status. A downloadable build artifact for manual testing is linked in the CI comment below.

Fixes [ED-25510](https://elementor.atlassian.net/browse/ED-25510)

[ED-25510]: https://elementor.atlassian.net/browse/ED-25510?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ED-25377]: https://elementor.atlassian.net/browse/ED-25377?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ED-25510]: https://elementor.atlassian.net/browse/ED-25510?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/packages/libs/editor-controls/src/controls/number-control.tsx
```diff
@@ -26,57 +28,100 @@ const renderSuffix = ( propType: PropType ) => {
...
+			setDraft( null );
+			restoreValue();
+
+			if ( raw === null ) {
+				return;
+			}
+
+			if ( isEmptyDraft( raw ) ) {
+				if ( ! propType.settings.required ) {
+					setValue( null );
```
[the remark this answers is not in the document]
max-zu: Good catch — addressed in 44a37d0182. Optional clears now set a `committedNullOnInputRef` flag in `handleInput`, and `handleBlur` skips the redundant `setValue(null)` when that flag is set (same pattern as the in-range blur guard). Tests updated to assert blur does not re-commit after input already cleared.

packages/packages/libs/editor-controls/src/controls/__tests__/number-control.test.tsx
```diff
@@ -112,4 +112,371 @@ describe( 'NumberControl', () => {
...
+
+		// Act.
+		fireEvent.input( input, { target: { value: '1.5' } } );
+
+		// Assert.
+		expect( input ).toHaveDisplayValue( '1.5' );
+		expect( setValue ).toHaveBeenCalledWith( { $$type: 'number', value: 1 } );
+	} );
+
+	it( 'should follow the bound value when it changes externally and the field is not being edited', () => {
```
Ntnelbaba, line 346: **Medium — no coverage for external updates while focused**

There is a rerender test when **not** editing, but none for bound value changing while `draft !== null`. Until blur, the input keeps the stale draft and can commit it over a newer external value (undo, responsive inherit, another control).

Worth a test (or a short comment) if holding the draft is intentional — same class of gap as pre-PR, but easier to regress now that draft exists.

packages/packages/libs/editor-controls/src/controls/__tests__/number-control.test.tsx
```diff
@@ -112,4 +112,371 @@ describe( 'NumberControl', () => {
 		// Assert.
 		expect( input ).toHaveAttribute( 'placeholder', '123' );
 	} );
+
+	it( 'should not clamp or commit while typing through an intermediate below-min value', () => {
```
Ntnelbaba, line 117: **Medium — AC traceability**

Carousel Transition Speed (`100..3000`) and Collection Loop items-per-page (`max: 100`) are covered by equivalent min/max/draft tests, not named ED-25510 scenarios. Logic looks correct; one test each with those bounds (or a short comment mapping names → ticket scenarios) would make regressions obvious for future readers.

packages/packages/libs/editor-controls/src/controls/number-control.tsx
```diff
@@ -26,57 +28,109 @@ const renderSuffix = ( propType: PropType ) => {
...
+			}
 		};
 
+		const handleBlur = () => {
+			const raw = draft;
+			const committedNullOnInput = committedNullOnInputRef.current;
+
+			committedNullOnInputRef.current = false;
+			setDraft( null );
+			restoreValue();
```
Ntnelbaba, line 91: **Low — easy-to-misread order**

Calling `restoreValue()` before clamp/null handling is correct (`restoreValue` only clears the validation latch in `use-bound-prop`; it does not revert the model). A one-line comment here would prevent a future “move restoreValue after commit” regression.

packages/packages/libs/editor-controls/src/utils/number-value.ts
```diff
@@ -0,0 +1,17 @@
+export const isEmptyDraft = ( raw: string ) => raw.trim() === '';
+
+export const parseNumberDraft = ( raw: string, shouldForceInt = false ): number | null => {
+	const trimmed = raw.trim();
+
+	if ( trimmed === '' ) {
+		return null;
+	}
+
+	const parsed = shouldForceInt ? parseInt( trimmed, 10 ) : Number( trimmed );
```
Ntnelbaba, line 10: **Low — trailing-dot commit**

`Number('1.')` → `1` can commit on input while the draft still shows `1.` until the next keystroke (utils test documents it). Acceptable for this fix; decimal-typing polish can stay a follow-up.

packages/packages/libs/editor-controls/src/controls/number-control.tsx
```diff
@@ -1,11 +1,13 @@
...
 import { numberPropTypeUtil, type PropType } from '@elementor/editor-props';
 import { InputAdornment, Typography } from '@elementor/ui';
 
 import { useBoundProp } from '../bound-prop-context';
 import { NumberInput } from '../components/number-input';
 import ControlActions from '../control-actions/control-actions';
 import { createControl } from '../create-control';
+import { clamp, isEmptyDraft, isInRange, parseNumberDraft } from '../utils/number-value';
 
 const isEmptyOrNaN = ( value?: string | number | null ) =>
```
Ntnelbaba, line 13: **Low — naming only**

`isEmptyOrNaN` (bound prop / placeholder) vs `isEmptyDraft` (raw input) serve different layers — both still needed. Optional one-line note that this is not for draft parsing.

Ntnelbaba (commented):
## Review — APPROVE with nits

Draft/blur split, null-safe `??` bounds, blur clamp, required-empty via `restoreValue()` without wiping the model, and dropping the native-validity gate all match [ED-25510](https://elementor.atlassian.net/browse/ED-25510). AC is met in logic and unit tests.

**Not** REQUEST_CHANGES — no merge blockers. Inline nits below.

**Intentional stepMismatch change:** agree this is safe. Old `{ validation: () => isInputValid }` blocked legitimate values (Scale `1.15`, Autoplay Delay `5050`) that PHP never enforced. Control now owns min/max; `step` stays for spinner UX only.

**Documented follow-up (not a blocker):** optional fields with a numeric default can still *display* the default after blur while the model is `null`.

**Branch:** behind `main` — rebase before merge.

Ntnelbaba (dismissed):

packages/packages/libs/editor-controls/src/controls/__tests__/number-control.test.tsx
max-zu: Holding the draft until blur is intentional. Added `should keep an in-progress draft when the bound value changes externally` in d1a48108bb: an external bound-value change while the field is focused keeps the typed draft, and blur commits that draft (clamped), not the newer value.

packages/packages/libs/editor-controls/src/controls/__tests__/number-control.test.tsx
max-zu: Mapped the existing bounds onto the ticket scenarios in d1a48108bb: Carousel Transition Speed `100..3000` on the below-min typing test, and Collection Loop items-per-page `max: 100` on the above-max typing test.

packages/packages/libs/editor-controls/src/controls/number-control.tsx
max-zu: Added a comment above `restoreValue()` in d1a48108bb. It only clears the validation latch in `use-bound-prop` (`setIsValid(true)`) and does not revert the model.

packages/packages/libs/editor-controls/src/controls/number-control.tsx
max-zu: Added a note on `isEmptyOrNaN` in d1a48108bb: it is for the bound value and placeholder only. Draft parsing stays on `isEmptyDraft` / `parseNumberDraft`.

packages/packages/libs/editor-controls/src/utils/number-value.ts
max-zu: Leaving this as a follow-up, as noted. `parseNumberDraft('1.')` is already covered in `number-value.test.ts`, and committing `1` while the draft still shows `1.` is acceptable for this fix. No code change.

Ntnelbaba (commented):
## Follow-up review (paired with Pro #7661)

Re-checked the delta since this morning’s review (`44a37d01` → `50e3edbd`). Author addressed the nits: ED-25510-scoped test comments, `isEmptyOrNaN` / `restoreValue()` clarifications, and the intentional **draft-hold on external update** test. **Follow-up OK** — still APPROVE with nits posture; no REQUEST_CHANGES.

**Paired with** [elementor-pro#7661](https://github.com/elementor/elementor-pro/pull/7661) (carousel canvas). Soft coupling; prefer merging **this Core PR first**, then Pro, for Transition Speed / Autoplay Delay QA.

packages/packages/libs/editor-controls/src/controls/__tests__/number-control.test.tsx
```diff
@@ -112,4 +112,405 @@ describe( 'NumberControl', () => {
...
+		// Act.
+		fireEvent.input( input, { target: { value: '2' } } );
+
+		// Assert.
+		expect( input ).toHaveDisplayValue( '2' );
+		expect( setValue ).not.toHaveBeenCalled();
+
+		// Act — external update while the draft is still held.
+		rerender( <NumberControl min={ 100 } max={ 3000 } />, {
+			value: { number: { $$type: 'number', value: 900 } },
```
Ntnelbaba, line 383: **Low** — External-update `rerender` passes `value: { number: { $$type: 'number', value: 900 } }` while the initial render uses `value: { $$type: 'number', value: 600 }`. Confirm `renderControl`’s second argument shape actually drives the bound prop change; if the nested `number` key is wrong, the test could pass without exercising an external update. (Same shape as the existing “not being edited” test — worth a quick harness check.)

Ntnelbaba (approved):

